PR #50, opened 2018-03-13
(refactor): Test server rebuilds with one less image

tahb:
* Having Spring as a separate docker-compose service was creating a third docker image at rebuild time. One for web, test and spring. Given Web and Spring are the main things we want to run we can cut our build time by a third by instead calling the test service with a non-spring command as and when we need it, which is just bin/dtests I believe.
* I was finding repeated issues in dev with a spring image not being reliably cleaned up on bin/drebuild. This caused gems still not to be present after a lengthy rebuild.

### DIFF
--- a/bin/dclean-slate
+++ b/bin/dclean-slate
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# Sometimes you need to remove all state including in persistent database volumes and all connected images.
+# This process takes much more time to rebuild from but aims to guarantee a fresh state.
+
+echo "Removing all traces of the test server…"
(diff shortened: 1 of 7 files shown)


tahb:
@despo I have noted we are wasting a lot of time when using bin/rebuild because it always removes the volumes - with all of our imported and scraped data - to guarantee a reliable state. I've added another commit here that splits out `bin/drebuilds` responsibility of guaranteeing that reliable state into a new file called `bin/dclean-slate` which should only need to be used if something weird happens with state between 2 branches.

PR #54, opened 2018-03-13
Feature/vacancies scrape scheduled task

Stretch96:


terraform.tf
```diff
@@ -57,7 +57,10 @@ module "ecs" {
   ecs_service_task_port                 = "${var.ecs_service_task_port}"
   ecs_service_task_definition_file_path = "${var.ecs_service_task_definition_file_path}"
   ecs_import_schools_task_definition_file_path = "${var.ecs_import_schools_task_definition_file_path}"
+  ecs_vacancies_scrape_task_definition_file_path = "${var.ecs_vacancies_scrape_task_definition_file_path}"
   import_schools_entrypoint              = "${var.import_schools_entrypoint}"
+  vacancies_scrape_entrypoint            = "${var.vacancies_scrape_entrypoint}"
+  vacancies_scrape_schedule_expression   = "${var.vacancies_scrape_schedule_expression}"
```
tahb: This feels like a useful value to include in our config file `staging.tfvars.example` so that it continues to be as straight forward as possible to tweak this value.

terraform/modules/ecs/ecs.tf
```diff
@@ -37,6 +37,56 @@ resource "aws_ecs_service" "web" {
 ECS task definitions
 ======*/
 
+/* scheduled task role */
+resource "aws_iam_role" "scheduled_task_role" {
+  name = "${var.project_name}-${var.environment}-scheduled-task-role"
+
+  assume_role_policy = <<EOF
```
tahb: Would it be possible to remain consistent with the rest of our policy definitions and move this to a file in `./terraform/policies` and load it into here via `File` eg: 

```  
policy = "${file("./terraform/policies/ecs-autoscale-role-policy.json")}"
```

terraform/modules/ecs/ecs.tf
```diff
@@ -92,6 +142,33 @@ data "template_file" "import_schools_task" {
 
 }
 
+/* vacancies_scrape task definition*/
+data "template_file" "vacancies_scrape_task" {
+  template = "${file(var.ecs_vacancies_scrape_task_definition_file_path)}"
+
+  vars {
```
tahb: This is now the third place we are repeating our application variables. I wonder if there is anything we can do at this point to control this duplication and making adding or changing variables less painful.

terraform/modules/ecs/ecs.tf
```diff
@@ -37,6 +37,56 @@ resource "aws_ecs_service" "web" {
...
+  ]
+}
+EOF
+}
+
+/* scheduled task policy */
+resource "aws_iam_role_policy" "scheduled_task_policy" {
+  name        = "${var.project_name}-${var.environment}-scheduled-task-policy"
+  role        = "${aws_iam_role.scheduled_task_role.id}"
+  policy = <<EOF
```
tahb: Same comment regarding the location of our policy definitions.

terraform/modules/ecs/ecs.tf
```diff
@@ -114,6 +191,34 @@ resource "aws_ecs_task_definition" "import_schools" {
...
+
+resource "aws_cloudwatch_event_rule" "vacancies_scrape_task_cron" {
+  name                = "${var.ecs_service_task_name}_vacancies_scrape_cron"
+  description         = "Run vacancies_scrape_task at specified cron time"
+  schedule_expression = "${var.vacancies_scrape_schedule_expression}"
+}
+
+resource "aws_cloudwatch_event_target" "vacancies_scrape_task_cron_event" {
+  target_id          = "${var.ecs_service_task_name}_vacancies_scrape"
+  rule          = "${aws_cloudwatch_event_rule.vacancies_scrape_task_cron.name}"
```
tahb: A non-blocking spot of the value alignment here.

variables.tf
```diff
@@ -105,6 +105,16 @@ variable "ecs_import_schools_task_definition_file_path" {
...
 }
 
+variable "ecs_vacancies_scrape_task_definition_file_path" {
+  description = "Task definition to scrape vacancies"
+  default = "./vacancies_scrape_task_definition.json"
+}
+
+variable "vacancies_scrape_schedule_expression" {
+  description = "vacancies_scrape schedule expression"
+  default = "rate(60 minutes)"
```
tahb: It would be really good to point users to the documentation behind the accepted values here in the description.

If I wanted to change this value to every day I'd like to be confident in a small change without having to understand the ins and outs of our Terraform code to learn that this value eventually gets received by CloudWatch, and it's that API I need to look at.

Stretch96: Yep I'll stick that in the description

terraform.tf
Stretch96: This is in the staging.tfvars.example file now

PR #56, opened 2018-03-14
(fix) makes published highlight box full width

gazaston:
https://trello.com/c/3zub6dzS/24-vacancy-published-confirmation-page

Before:
screen_shot_2018-02-22_at_15 01 20

After: 
screen shot 2018-03-14 at 11 07 09


gazaston:
[vacancy published confirmation page](https://trello.com/c/3zub6dzS/24-vacancy-published-confirmation-page)

PR #61, opened 2018-03-15
[151] Remove html tags from scraped content

tahb:
* Based on https://github.com/dxw/teacher-vacancy-service/pull/59 for the test suite
* Nokogiri was doing a good job of removing tags from what I could see after running the scraper so I have just added some tests to support this behaviour

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@
 # Ignore application configuration
 /config/application.yml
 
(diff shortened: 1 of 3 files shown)


tahb:
[Remove HTML tags from job vacancy description](https://trello.com/c/0TwS10Jx/151-remove-html-tags-from-job-vacancy-description)

PR #64, opened 2018-03-19
Fix formatting of terraform files

rjw1:
The result of running `terraform fmt`. We should be consistent in our in
our formatting and since terraform provides a tool we should use it.

There is discussion to be had about whether we want the lines between sections
which caused the alignment of values to be variable.


rjw1:
If they are intentionally logical groupings its fine for the alignment to vary between them.

PR #66, opened 2018-03-19
[183] Search results display when paginated

tahb:
* Kaminari `total_count` returns the full list of records without filtering.
* Moved helper method and found it was missing tests so added them.
* I deleted vacancies at random to try and break the pagination, it held up okay.

Before:
screen shot 2018-03-19 at 13 32 28

After: 
screen shot 2018-03-19 at 13 33 05

### DIFF
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -22,6 +22,14 @@ module VacanciesHelper
             class: "sortby--#{order}#{active_class || ''}"
   end
 
+  def pluralize_vacancy_count(count, i18n_id, plural_i18n_id = nil)
+    if count == 1
+      I18n.t(i18n_id, count: count)
+    else
+      I18n.t(plural_i18n_id || (i18n_id + '_plural'), count: count)
+    end
+  end
+
   private
 
   def vacancy_params_whitelist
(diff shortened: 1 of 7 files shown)


app/helpers/vacancies_helper.rb
```diff
@@ -22,6 +22,14 @@ def link_to_sort_by(title, column:, order:, sort:)
...
   end
 
+  def pluralize_vacancy_count(count, i18n_id, plural_i18n_id = nil)
+    if count == 1
+      I18n.t(i18n_id, count: count)
+    else
+      I18n.t(plural_i18n_id || (i18n_id + '_plural'), count: count)
+    end
+  end
+
```
despo: This count is specific to vacancies so it would make more sense to be in the VacanciesPresenter, with a beteer name, as the purpose of the presenter is to remove unecessary noise from the views e.g. `pluralize_vacancy_count` could be `vacancies.total_count`

tahb: As mentioned I cannot use the built in `total_count` method, it returns everything regardless of filters. 

The source of this bug is that the pagination is half handled inside the presenter, half handled by the controller. The controller is in currently responsible for the reading of data and the pagination. The presenter only every gets those paginated vacancies, it doesn't have access to the total value.

PR #69, opened 2018-03-20
(Feture) Add Dotenv

Stretch96:
* In production, secrets are retrieved and added to .env

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,6 +17,8 @@ require 'sprockets/railtie'
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+Dotenv::Railtie.load
+
 module TeacherVacancyService
   class Application < Rails::Application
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -17,6 +17,8 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+Dotenv::Railtie.load
```
tahb: From their docs I believe this loading is redundant however I think it will be helpful to have it and be explicit in what it's purpose is. Perhaps an additional comment along the lines of:

> For secret management in live environments only. Development ENV still managed through `docker-compose.env`

What do you think?

despo:
@tahb @Stretch96 can we close this or is it being addressed?

tahb:
Closing pending further conversations as I don't believe this is being working on.

PR #70, opened 2018-03-20
Feature/secrets

Stretch96:
Moves environment variables to AWS parameter store and bootstraps the containers, pulling the parameters and storing them in .env

**DO NOT MERGE** until https://github.com/dxw/teacher-vacancy-service/pull/69 is merged 👍


terraform/modules/container_bootstrap/output.tf
```diff
@@ -0,0 +1,7 @@
+output "entrypoint" {
+  value = "apt-get update && apt-get -y install python-dev python2.7 curl && curl -O https://bootstrap.pypa.io/get-pip.py && python2.7 get-pip.py && rm get-pip.py && pip install awscli && aws s3 cp s3://${aws_s3_bucket.container_bootstrap_bucket.id}/container_bootstrap.sh . && /bin/bash container_bootstrap.sh -b ${aws_s3_bucket.container_bootstrap_bucket.id} -p ${var.parameter_store_path} -r ${var.region} -u ${var.dotenv_user} &&"
```
tahb: Installing additional packages into our running containers makes me uncomfortable as we would be deliberately mutating our immutable images and therefore remove our parity between dev, test and live. This is a key reason to use containers in the first place so I would like to be clear on alternatives before taking this route. Perhaps mounting a .env/ to a volume within each container, have we explored that?

tahb:
Closing pending further conversations as I don't believe this is being working on.

PR #71, opened 2018-03-20
[163] Lockdown vacancy CRUD endpoints

tahb:
* Switch from using School URN in the URL to our internal UUID as the reliability of the URN is in question. We are not in a position of knowledge with the URN so we don't know it if is always unique to that school or even where that data is mastered is to check. This results in calls to `@school.id` in our routing as with just `@school` it defaults to URN.
* Add the new basic auth credentials for hiring staff 
* Nest the schools controller within a new Hiring Staff scope as they are the users who should be able to access these features going forwards
* There is a bit of deliberate duplication in the basic_auth_spec for staging and production. The idea is that eventually when we have a production we change the setup code to no longer expect a global basic auth.


tahb:
@despo I think this one is best reviewed by yourself given you're working closely with this code too.

tahb:
[Lockdown CRUD endpoints](https://trello.com/c/EqUq35kh/163-lockdown-crud-endpoints)

spec/features/users_cannot_manage_vacancies_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    fake_env = double.as_null_object
+    allow(Figaro).to receive(:env).and_return(fake_env)
+    allow(fake_env).to receive(:http_user).and_return(nil)
+    allow(fake_env).to receive(:http_pass).and_return(nil)
+  end
+
+  scenario 'An unauthenticated user tries to create a vacancy' do
+    school = create(:school)
+
+    visit "/schools/#{school.id}/vacancies/new"
```
despo: can we use url helpers across all specs

spec/features/users_cannot_manage_vacancies_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    expect(page).to have_content('HTTP Basic: Access denied.')
+  end
+
+  scenario 'An unauthenticated user tries to update a vacancy' do
+    school = create(:school)
+    vacancy = create(:vacancy, school: school)
+
+    visit "/schools/#{school.id}/vacancies/#{vacancy.slug}"
```
despo: can we use url helpers across all specs

.rubocop.yml
```diff
@@ -52,11 +52,15 @@ Metrics/MethodLength:
     Exclude:
       - 'spec/**/*'
 
+Metrics/AbcSize:
+    Max: 20
+    Exclude:
+      - 'spec/**/*'
+
```
despo: 👍

spec/requests/basic_auth_spec.rb
```diff
@@ -35,60 +28,134 @@
...
+
+      it_behaves_like 'basic auth is required', '/schools', :hiring_staff_http_user, :hiring_staff_http_pass
+
+      it 'posting to the create vacancy endpoint still requires the hiring staff basic auth' do
+        school = create(:school)
+
+        path = school_vacancies_path(school.id)
+        post path, params: { vacancy: { foo: :bar } }, env: { 'HTTP_AUTHORIZATION': encoded_credentials }
+
+        expect(response.status).to eq(401)
```
despo: can we change it to `expect(response).to have_http_status(:unauthorized)` as it's easier to read

tahb:
@depso I've made those changes now 👍

PR #76, opened 2018-03-22
[202] Hiring staff can only find their school

tahb:
* This moves the publishing journey from clicking a link in search results (which have just been removed) to the schools show page.
* The notion of `DEFAULT_SCHOOL_URN` exists in order for the product team to log in to given school (this could be made configurable if the need arises) and to give our test suite some stability until we switch to Open ID.
* Hiring staff are now always asked for authentication even in development and test. With the switch to a better authenticate mechanism we will need to effectively set this up in development and test too to make sure we are working with some parity to live environments.
* When we add in the second school that should drive the next iteration of refactoring

1:
screen shot 2018-03-22 at 16 58 03

2:
screen shot 2018-03-22 at 16 58 12

3:
screen shot 2018-03-22 at 16 58 20


app/controllers/hiring_staff/base_controller.rb
```diff
@@ -1,32 +1,43 @@
...
 
+  def current_school
+    @current_school ||= School.find_by! urn: session[:urn]
+  rescue ActiveRecord::RecordNotFound
+    not_found
+  end
+
   def authenticate_hiring_staff?
-    !(Rails.env.development? || Rails.env.test?)
+    true
```
despo: Is there any value in having `authenticate_hiring_staff?` anymore then or was this removed accidentaly?

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -1,32 +1,43 @@
...
+        true
+      elsif name == http_user && password == http_pass
+        session[:urn] = DEFAULT_SCHOOL_URN # Enables the product team to access publishing and the test suite a default
+        true
+      end
     end
   end
 
+  def current_school
+    @current_school ||= School.find_by! urn: session[:urn]
```
despo: I have been using `id`, which is our identifier, and not `urn`. This will still work for now but we should decide what to use as our identifying key for the admin facing and the user facing controllers

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -1,32 +1,43 @@
 class HiringStaff::BaseController < ApplicationController
   before_action :authenticate
+  DEFAULT_SCHOOL_URN = '140181'.freeze
```
despo: I'm not sure I understand this..  Why do we have hardcoded urns in the app?

app/controllers/hiring_staff/base_controller.rb
tahb: Oh good spot. This can go :)

app/controllers/hiring_staff/base_controller.rb
tahb: I can switch to our school ID that's a better idea.

tahb: I've remembered why I didn't use the school ID: they are always going to be different in each environment so they aren't something we can pin down reliably tie to a log in. The only identifying data that can works must come from Get Information About Schools.

app/controllers/hiring_staff/base_controller.rb
tahb: @despo I've had a crack: https://github.com/dxw/teacher-vacancy-service/pull/76/commits/080a22663ddd3f649a75a85589feb63b5a7b13e2

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -7,7 +7,7 @@ def authenticate
         session[:urn] = '110627'
         true
       elsif name == http_user && password == http_pass
-        session[:urn] = School.first.urn # Enables the product team to access publishing and the test suite a default
+        session[:urn] = ENV.fetch('DEFAULT_SCHOOL_URN') { School.first.urn }
```
despo: 👍

PR #83, opened 2018-03-27
Feature/216 add a field to provide a link to application form pack

despo:


despo:
[Add a field to provide a link to application form/pack/whatever to enable interested applicants to take next step](https://trello.com/c/pmaUCR4m/216-add-a-field-to-provide-a-link-to-application-form-pack-whatever-to-enable-interested-applicants-to-take-next-step)

config/locales/en.yml
```diff
@@ -96,3 +98,7 @@ en:
     save_and_continue: 'Save and continue'
     find: 'Find'
     delete: 'Delete'
+
+  errors:
```
tahb: This should probably live in in the activerecord.yml file.

app/views/hiring_staff/vacancies/application_details/new.html.haml
```diff
@@ -16,6 +16,12 @@
                 as: :email,
                 required: true
 
+      = f.input :application_link,
```
tahb: I feel like a hint would work well here as I can imagine `Application URL` will not be clearly understood by all.

despo: fixed, it was mapped to the wrong locale as I copy pasted the gds fields below and edited 🤔

PR #86, opened 2018-03-29
(Feature) minimum salary is not less than lowest pay scale salary

despo:
- update factories to so that by generated pay scale values are always less than minimum salary
- fix broken specs by updating specified salary values in Factory


despo:
[Ensure minimum salary is not less than minimum payscale salary](https://trello.com/c/Xly6WEm8/11-ensure-minimum-salary-is-not-less-than-minimum-payscale-salary)

PR #87, opened 2018-03-29
(feature) empty state for vacancy list on school show page

gazaston:
- when a school has no vacancies, the table is removed and a list of information required to create a vacancy is displayed

before:
before

after: 
after


gazaston:
[Empty state for vacancy list on school show page](https://trello.com/c/asVoTMyA/1-empty-state-for-vacancy-list-on-school-show-page)

app/views/hiring_staff/schools/show.html.haml
```diff
@@ -5,14 +5,10 @@
...
   .column-full
-    %table.vacancies
-      %th= t('vacancies.job_title')
-      %th= t('vacancies.publish_on')
-      %th= t('vacancies.expires_on')
-      %th= t('vacancies.status')
-      %th= t('vacancies.actions')
-      - @school.vacancies.each do |vacancy|
-        = render partial: 'vacancy', locals: { vacancy: vacancy, school: @school }
+    - if @school.vacancies.count > 0
```
despo: can you change this to `.any?` please

spec/features/hiring_staff_can_view_vacancies_spec.rb
```diff
@@ -4,6 +4,15 @@
   include_context 'when authenticated as a member of hiring staff',
                   stub_basic_auth_env: true
 
+  scenario 'A school should see advisory text when there are no vacancies', elasticsearch: true do
+    school = FactoryGirl.create(:school)
+    visit school_path(school.id)
+
+    expect(page).to have_content(I18n.t('schools.vacancies.index', school: school.name))
+    expect(page).not_to have_css('table.vacancies')
+    expect(page).to have_content(I18n.t('schools.no_vacancies.heading'))
```
despo: For expectations, it would be better if use the actual expected text, as lack of verifying content was what caused me to override the application_link.

gazaston: You think that's preferable to having the text in one place only?

despo: Since this is specifically about the content you see when you get to that page yes, I don't class this it's duplication.

app/views/hiring_staff/schools/show.html.haml
```diff
@@ -5,18 +5,14 @@
...
+      = render partial: 'vacancies'
+    - else
+      = render partial: 'no_vacancies'
 
   .column-full
     .form-group.mt1
-      = link_to 'Publish a vacancy', new_school_vacancy_path(school_id: @school.id), class: 'button'
```
despo: 👍

PR #88, opened 2018-03-29
Chore/remove redundant routes and amend school retrieval

despo:
- retrieve schools by id and not urn


despo:
[Default school retrieval should be by id and not by urn](https://trello.com/c/LYAEaCyU/13-default-school-retrieval-should-be-by-id-and-not-by-urn)

PR #90, opened 2018-03-30
Feature/6 hiring staff can edit vacancies after they are published

despo:


despo:
[Hiring staff can edit vacancies after they are published](https://trello.com/c/fr6Emb5D/6-hiring-staff-can-edit-vacancies-after-they-are-published)

app/views/hiring_staff/vacancies/edit.html.haml
```diff
@@ -0,0 +1,92 @@
+%h1.heading-large
+  = t('vacancies.edit_heading', school: @vacancy.school.name)
+
```
gazaston: Do we need some way for the user to cancel and return to the `schools#show` page?

app/views/hiring_staff/schools/_vacancy.html.haml
```diff
@@ -3,5 +3,6 @@
   %td= format_date(vacancy.publish_on)
   %td= format_date(vacancy.expires_on)
   %td= vacancy.status.titleize
+  %td= link_to 'Edit', edit_school_vacancy_path(school, vacancy.id)
```
gazaston: We'll need to add a `colspan=2` attribute to the 'Actions' table heading cell — the border cuts off prematurely.

screen shot 2018-04-03 at 10 37 32

app/views/hiring_staff/vacancies/edit.html.haml
despo: @gazaston yes. Do you have any suggestions?

gazaston: Probably just a simple link above the `h1`?

despo: @gazaston done! After this PR is in we should maybe address navigation more generally. E.g. review page -> school page, vacancy#show -> school page etc.

gazaston: I'll merge this in as-is, but agree that we need to tackle this nav pattern. I'll address that immediately.

PR #91, opened 2018-04-02
Feature/14 basic audit to identify and rollback malicious changes

despo:
add public activity and track changes

- on create and update
- add simple user model for mapping to changes
- expose /activities on hiring school vacancies controller
- includes tweaks to the default public_activity table creation to support uuids

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,8 @@ gem 'rubocop' # Not in Test group due to: https://github.com/chapmanu/imposter/i
 
 gem 'activerecord-session_store'
 
+gem 'public_activity'
+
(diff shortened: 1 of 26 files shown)


despo:
[Basic audit to identify and rollback malicious changes](https://trello.com/c/dNyHVuMk/14-basic-audit-to-identify-and-rollback-malicious-changes)

despo:
@tahb there is no User dependency on the gem, I just had to associate each action with a record. I'm not using tracked as we don't necessarily want any action to be logged at the moment.

I'll drop the user and see how to map to the session.

tahb:
Rebased

PR #92, opened 2018-04-02
chore/18 a vacancy must pass all validations before it can be published

despo:


despo:
[A vacancy must pass all validation rules before it can be published](https://trello.com/c/NtzXEkR0/18-a-vacancy-must-pass-all-validation-rules-before-it-can-be-published)

PR #94, opened 2018-04-03
(feature) adds breadcrumbs for hiring staff views

gazaston:
displays breadcrumbs for hiring staff to return to their school page when they view/edit/review a vacancy

before:
before

after:
after


gazaston:
[Breadcrumb navigation](https://trello.com/c/0vuLw69o/21-breadcrumb-navigation)

PR #95, opened 2018-04-03
Feature/8 implement basic validation to ensure data from hiring staff is good enough

despo:


despo:
[Implement basic validation to ensure data from hiring staff is 'good enough'](https://trello.com/c/7dflbiQ1/8-implement-basic-validation-to-ensure-data-from-hiring-staff-is-good-enough)

app/models/concerns/vacancy_job_specification_validations.rb
```diff
@@ -2,12 +2,36 @@ module VacancyJobSpecificationValidations
   extend ActiveSupport::Concern
   include ApplicationHelper
 
+  MAX_INTEGER = 2147483647
```
tahb: Does this need a freeze to make sure it's not accidentally mutated?

despo: @tahb no. Rubocop ` Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.`

PR #96, opened 2018-04-03
(feature) prevent access to hiring_staff/vacancies#show if

despo:
a vacancy has not been published and redirect to review page

- utilise factories to create draft and published vacancies
- simplify path helper parameters, no need to specify school_id and vacancy_id

(the current vacancy#show page assumes a vacancy's fields have been succesfuly validated and does not check for empty values, which currently causes some draft vacancies to break the page)

### DIFF
--- a/app/controllers/hiring_staff/vacancies_controller.rb
+++ b/app/controllers/hiring_staff/vacancies_controller.rb
@@ -1,18 +1,21 @@
 class HiringStaff::VacanciesController < HiringStaff::Vacancies::ApplicationController
   def show
     vacancy = school.vacancies.active.find(id)
+    unless vacancy.published?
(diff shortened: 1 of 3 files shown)


despo:
[A school should only be able to preview vacancies set to published (not draft)](https://trello.com/c/paltoPvd/22-a-school-should-only-be-able-to-preview-vacancies-set-to-published-not-draft)

gazaston:
I18n error here. Also I wonder if this semantically should be an `alert` rather than a `notice`?

screen shot 2018-04-04 at 16 19 31

despo:
@gazaston fixed and changed to alert

screen shot 2018-04-04 at 18 34 20

PR #98, opened 2018-04-04
(chore) correct naming of pay scale options

despo:
- moves data creation in migrations rather than seeds
- add data update as migration


despo:
[Correct naming of pay scale options](https://trello.com/c/GSRskwVf/12-correct-naming-of-pay-scale-options)

db/migrate/20180404100617_add_pay_scale_data.rb
```diff
@@ -0,0 +1,64 @@
+class AddPayScaleData < ActiveRecord::Migration[5.1]
+  def change
```
tahb: As is, if you rollback this migration will it leave the table untouched?

I have previously utilised both `up` and `down` to ensure the migration works in both directions.

db/migrate/20180404102429_update_pay_scale_data.rb
```diff
@@ -0,0 +1,61 @@
+class UpdatePayScaleData < ActiveRecord::Migration[5.1]
```
tahb: This looks to add labels to existing pay scales but that job seems to have been done by the previous migration `AddPayScaleData`, do we need this one?

despo: @tahb  It updates the existing labels.

I transferred the existing rake task in add_payscale_data and then did the update as I didnt want to just drop the first step.

db/migrate/20180404100617_add_pay_scale_data.rb
despo: I'm not sure what the value of that in the specific migration is. So you think I should change this to up and add a down that removes all payscale data?

tahb:
My initial thinking was to try and make these things rerunnable and revertable incase they go wrong when we deploy, however a rollback would be much more complicated as new vacancies could become associated with this data and deleting them would leave a damaging black hole.

This situation seems like these tasks are more appropriately suited to being one off rake tasks as they should only ever be run once. The migration approach also presents us problems should we ever want to remigrate things instead of schema:load as if we change the name of the PayScale class to PayRange the migration process dies as it is referencing it directly.

This post is an excellent description of this situation: https://robots.thoughtbot.com/data-migrations-in-rails

I'm reluctant to merge because of this, even through I've just merged https://github.com/dxw/teacher-vacancy-service/pull/101 which sets a precedent.

I'd be more comfortable with the rake task approach, do you know of a reason that might be problematic?

despo:
@tahb  payscale data are part of the application so it makes more sense to add them in a migration. A rake task implies some sort of reusability as it can be run again - when it shouldn't - and it also adds another step in the project setup process. An irreversible change in the records will happen in both cases. 
if we rename everything will be impacted anyway and class names will have to be changed across the app (tasks or migrations..)

Interesting article.

despo:
Closing in favor of #112

PR #102, opened 2018-04-04
(chore) styled error pages

gazaston:
Uses custom, dynamic error pages rather than static files from the `public` folder. 

Based on [this article](https://mattbrictson.com/dynamic-rails-error-pages).

unknown


gazaston:
[Styled error pages](https://trello.com/c/GDRV3X3F/20-styled-error-pages)

app/views/errors/internal_server_error.html.haml
```diff
@@ -0,0 +1,2 @@
+%h1.heading-xlarge Something went wrong at our end.
+%p You can email #{mail_to 'email@example.com'} if the problem persists.
```
despo: @gazaston can we retrieve the mail from the locale file?

despo:
@gazaston approved 👍 it need to be rebased.

PR #103, opened 2018-04-04
(Chore) change payscale salary db type to integer

despo:


despo:
[Change pay_scale salary db type to integer](https://trello.com/c/qEBCFA5k/27-change-payscale-salary-db-type-to-integer)

tahb:
Rebased

PR #109, opened 2018-04-06
(fix) change app title

gazaston:
before:
before

after:
after


gazaston:
[Change app name to School Jobs](https://trello.com/c/CpGDpqvK/44-change-app-name-to-school-jobs)

PR #113, opened 2018-04-10
(fix) reword salary range hint to not include (£)

deberny:
Partially addresses this trello card [34-using-in-salary-range-does-not-work](https://trello.com/c/Wza31XO5/)

The purpose for rewording the salary range hint is to help the user avoid errors.
There is a long discussion on the card and asking the user not to put the '£' is a first small step in improving the form.


deberny:
[Using £ in salary range does not work](https://trello.com/c/Wza31XO5/34-using-in-salary-range-does-not-work)

PR #114, opened 2018-04-11
(fix): Containers in all env will always be greater or equal to 2

tahb:
.. or as much as 10
* Should one container fail for any reason there will *always* be another to take over while a new one is spun up inline with desired_count.


tahb:
3 it is 👍

PR #115, opened 2018-04-11
(chore): Add user mapping

tahb:
* The first onboarding process has just completed and so we'd like to make sure this user is associated with the right school. Afterwards I will be able to remove the OVERRIDE_URN environment variable.


tahb:
Replaced by a new duplicate pull request: https://github.com/dxw/teacher-vacancy-service/pull/116

PR #118, opened 2018-04-12
(chore) add rake task to create ES indices

despo:
`rake elastic search:vacancies:index`


despo:
[elastic search requires seed to be run on boot](https://trello.com/c/jQFxaBqK/39-elastic-search-requires-seed-to-be-run-on-boot)

PR #119, opened 2018-04-12
(feature) configure task to update pay_scale

despo:


despo:
[in vacancy creation, correct dropdown labelling in 'Pay scale (optional)' to read 'Main Pay Range' rather than 'Minimum Pay Range'](https://trello.com/c/u6QNU34w/33-in-vacancy-creation-correct-dropdown-labelling-in-pay-scale-optional-to-read-main-pay-range-rather-than-minimum-pay-range)

tahb:
Rebased

PR #120, opened 2018-04-12
(feature): Automate the deployment of terraform changes

tahb:
* The process is currently too manual and could be automated.
* Because of the above it is also prone to errors and I have found myself being in the wrong Terraform workspace with the wrong tfvars, which (depending on what you change) can be damaging.
* Could have done a Terraform `plan` but since running apply without `-auto-approve` always provides a prompt to check planned changes it does the same job of giving the user a final confirmation step incase the terraform change might be dangerous.

### DIFF
--- a/bin/terraform-deploy
+++ b/bin/terraform-deploy
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -e
+
+if [ $# -eq 0 ]
+then
+  echo "No terraform workspace supplied, aborting."
+  exit 1
+fi
+
+TERRAFORM_WORKSPACE=$1
+GIT_BRANCH=""
+
+case $TERRAFORM_WORKSPACE in
+  "production") GIT_BRANCH="master"
+    ;;
+  "staging") GIT_BRANCH="develop"
+    ;;
+  "edge") GIT_BRANCH="edge"
+    ;;
+  "testing") GIT_BRANCH="testing"
+    ;;
+  *)
+    echo "Could not find an known environment with the name: $TERRAFORM_WORKSPACE"
+    exit 1
+    ;;
+esac
+
+echo "Deploying…"
+echo "* Terraform workspace: $TERRAFORM_WORKSPACE"
+echo "* Git branch: $GIT_BRANCH"
+
+read -r -p "Continue (y/n)?" CONT
+if [ "$CONT" = "y" ]; then
+  echo
+else
+  exit 1
+fi
+
+echo "-> Checking out terraform workspace $TERRAFORM_WORKSPACE …"
+terraform workspace select "$TERRAFORM_WORKSPACE"
+
+echo "-> Fetching latest git changes …"
+git fetch
+git pull origin $GIT_BRANCH
+
+echo "-> Checking out branch $GIT_BRANCH …"
+git checkout $GIT_BRANCH
+
+echo "-> Applying Terraform command on the $TERRAFORM_WORKSPACE workspace …"
+terraform apply -var-file="workspace-variables/$TERRAFORM_WORKSPACE.tfvars"


bin/terraform-deploy
```diff
@@ -0,0 +1,50 @@
...
+if [ $# -eq 0 ]
+then
+  echo "No terraform workspace supplied, aborting."
+  exit 1
+fi
+
+TERRAFORM_WORKSPACE="$@"
+GIT_BRANCH=""
+
+case $@ in
```
rjw1: You set `TERRAFORM_WORKSPACE` but then dont use it.

bin/terraform-deploy
```diff
@@ -0,0 +1,50 @@
...
+
+echo "-> Fetching latest git changes …"
+git fetch
+git pull origin $GIT_BRANCH
+
+echo "-> Checking out branch $GIT_BRANCH …"
+git checkout $GIT_BRANCH
+
+echo "-> Applying Terraform command on the $@ workspace …"
+terraform apply -var-file="workspace-variables/$@.tfvars"
```
rjw1: `shellcheck` isnt happy with a lot of this
```
bob@roti:~/git/dxw/teacher-vacancy-service ☃ >shellcheck bin/terraform-deploy

In bin/terraform-deploy line 10:
TERRAFORM_WORKSPACE="$@"
^-- SC2034: TERRAFORM_WORKSPACE appears unused. Verify it or export it.
                    ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In bin/terraform-deploy line 23:
    echo "Could not find an known environment with the name: $@"
                                                             ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In bin/terraform-deploy line 29:
echo "* Terraform workspace: $@"
                             ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In bin/terraform-deploy line 32:
read -p "Continue (y/n)?" CONT
^-- SC2162: read without -r will mangle backslashes.


In bin/terraform-deploy line 39:
echo "-> Checking out terraform workspace $@ …"
                                          ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In bin/terraform-deploy line 40:
terraform workspace select $@
                           ^-- SC2068: Double quote array expansions to avoid re-splitting elements.


In bin/terraform-deploy line 49:
echo "-> Applying Terraform command on the $@ workspace …"
                                           ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In bin/terraform-deploy line 50:
terraform apply -var-file="workspace-variables/$@.tfvars"
                                               ^-- SC2145: Argument mixes string and array. Use * or separate argument.
```

bin/terraform-deploy
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -e
+
+if [ $# -eq 0 ]
+then
+  echo "No terraform workspace supplied, aborting."
+  exit 1
+fi
+
+TERRAFORM_WORKSPACE="$@"
```
rjw1: I'd probably use $1 instead here.
getopts might be an option too

tahb:
Thanks @rjw1, shellcheck is happy now and I made the other 2 suggestions.

bin/terraform-deploy
```diff
@@ -0,0 +1,50 @@
...
+if [ $# -eq 0 ]
+then
+  echo "No terraform workspace supplied, aborting."
+  exit 1
+fi
+
+TERRAFORM_WORKSPACE=$1
+GIT_BRANCH=""
+
+case $@ in
```
rjw1: This should probably be `TERRAFORM_WORKSPACE` as well.

PR #122, opened 2018-04-13
(feature) remove vacancy#headline

despo:


despo:
[remove the headline field](https://trello.com/c/WW5fB1W9/59-remove-the-headline-field)

PR #123, opened 2018-04-13
(chore) update seeds 

despo:
- extend db:seed to run `data:seed:pay_scale` and `data:update:pay_scale`
- remove pay scale and leadership data creation from seeds
- use real school details
- map to known urns


despo:
[Fix seed data](https://trello.com/c/PSMsQBZN/52-fix-seed-data)

PR #124, opened 2018-04-13
(feature) reverse date fields on application details page

despo:
screen shot 2018-04-13 at 15 49 37


despo:
[Reverse date fields on application_details so that Date role will be listed appears above Application deadline and set hinttext dynamically to future dates.](https://trello.com/c/fT58TO3b/51-reverse-date-fields-on-applicationdetails-so-that-date-role-will-be-listed-appears-above-application-deadline-and-set-hinttext-d)

PR #126, opened 2018-04-16
(feature): 503 and 502 will be served with a static offline page

tahb:
* Whenever CloudFront receives a 503 service unavailable or 502 bad gateway from our application or load balancer, it will respond with a different behaviour. Instead of passing up a basic 503 error page it will look to a specified bucket to supply the web page instead. This will enable us to selectively put the site offline by scaling the number of instances or containers down to 0 or go offline automatically if our whole service becomes unavailable.
* https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages.html

screen shot 2018-04-16 at 17 25 03

### DIFF
--- a/terraform.tf
+++ b/terraform.tf
@@ -47,24 +47,24 @@ module "core" {
 module "ecs" {
   source = "./terraform/modules/ecs"
 
-  environment                                    = "${terraform.workspace}"
-  project_name                                   = "${var.project_name}"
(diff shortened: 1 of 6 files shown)


tahb:
[Static holding page for application offline](https://trello.com/c/9PfoZQry/19-static-holding-page-for-application-offline)

PR #128, opened 2018-04-17
(Feature) change salary fields to strings

despo:
- implemented `SalaryValidator` - checks for
  - blank?
  - format `/^\d+\.{0,1}\d{2}{0,1}$/` should match XX or X*.XX
  - minimum_allowed
  - maximum_allowed

- extended `VacancyForm` to delegate `=` attributes so validation conditions can be simplified
- extract min and max allowed salaries to constants

### DIFF
--- a/app/form_models/vacancy_form.rb
+++ b/app/form_models/vacancy_form.rb
@@ -3,7 +3,7 @@ class VacancyForm
 
   attr_accessor :vacancy
 
-  delegate *Vacancy.attribute_names.map { |attr| [attr, "#{attr}="] }.flatten, to: :vacancy
+  delegate *Vacancy.attribute_names.map { |attr| [attr, "#{attr}=", "#{attr}?"] }.flatten, to: :vacancy
   delegate :save, to: :vacancy
 
   def initialize(params = {})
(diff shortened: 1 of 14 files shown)


despo:
[Using £ in salary range does not work](https://trello.com/c/Wza31XO5/34-using-in-salary-range-does-not-work)

tahb:
Rebased

tahb:
Thoughts on using this for the first time.

As this salary is the first field I get to that asks for money, I tried entering similar content to the ones seen on Schools North East which include scales:

screen shot 2018-04-20 at 14 10 34

I think the user experience here needs more work to be better than the previous iteration. I think it begins when the user is not given clear instructions on what's expected from them and then the subsequent error doesn't provide enough to correct let the user correct the situation first time.

Pound symbols don't seem to be accepted either which was the primary problem with using the number input type, if I recall correctly?:

screen shot 2018-04-20 at 14 11 16

If in the form the Pay scales were brought to the attention of the user before this field I _might_ have been able to make the mental leap to assume that this next field shouldn't include pay scale.

I've had a tinker with how we could address these:
screen shot 2018-04-20 at 14 34 05

tahb:
Closing in favour of https://github.com/dxw/teacher-vacancy-service/pull/148

PR #130, opened 2018-04-17
(feature) adds a text banner for public beta phase

gazaston:
screen shot 2018-04-17 at 17 18 17

### DIFF
--- a/app/views/shared/_beta_banner.html.haml
+++ b/app/views/shared/_beta_banner.html.haml
@@ -1,0 +1,6 @@
+.flash.notice
+  .small This service is still in development, so for now only lists jobs in select schools in Cambridgeshire and the North East. When it launches, the full service will list jobs in schools throughout England.
+  .small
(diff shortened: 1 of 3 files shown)


app/views/shared/_beta_banner.html.haml
```diff
@@ -0,0 +1,6 @@
+.flash.notice
+  .small This service is still in development, so for now only lists jobs in select schools in Cambridgeshire and the North East. When it launches, the full service will list jobs in schools throughout England.
+  .small
+    Interested in listing jobs at your school? Email
+    =mail_to 'isobel.croot@digital.education.gov.uk'
```
despo: Can we extract this to an environment variable and reuse so it's not stored in the codebase.

tahb:
I think this has been merged as this PR https://github.com/dxw/teacher-vacancy-service/pull/131 has already been merged in. I think we can close.

PR #134, opened 2018-04-18
[83]  Error pages are shown when errors are caused by SSO

tahb:
* When a request comes in from another source that raises an error, eg. SSO, then currently the application does not provide a 500 page and the request results in an unstyled browser 500.
* Replicate by signing in and then visiting https://tvs.edge.dxw.net/auth/azureactivedirectory/callback directly

Before:
screen shot 2018-04-18 at 11 33 10

After:
screen shot 2018-04-18 at 11 44 12

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,4 +1,7 @@
 class ErrorsController < ApplicationController
+  skip_before_action :verify_authenticity_token,
+                     only: %i[not_found unprocessable_entity internal_server_error]
+
   def not_found
     respond_to do |format|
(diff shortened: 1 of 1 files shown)


tahb:
[Errors caused by sign-in are rendered correctly](https://trello.com/c/w5oNHjBi/83-errors-caused-by-sign-in-are-rendered-correctly)

tahb:
[Errors caused by sign-in are rendered correctly](https://trello.com/c/w5oNHjBi/83-errors-caused-by-sign-in-are-rendered-correctly)

PR #136, opened 2018-04-18
[81] Production is publicly available

tahb:
.. to job seekers, not hiring staff.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,7 +26,7 @@ class ApplicationController < ActionController::Base
   end
 
   def authenticate?
-    !(Rails.env.development? || Rails.env.test?)
+    Rails.env.staging?
   end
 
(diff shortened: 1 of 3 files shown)


tahb:
[Remove basic auth for production and go-live](https://trello.com/c/pdJNJ7KW/81-remove-basic-auth-for-production-and-go-live)

app/controllers/application_controller.rb
```diff
@@ -25,7 +25,7 @@ def check_staging_auth
   end
 
   def authenticate?
-    !(Rails.env.development? || Rails.env.test?)
+    !(Rails.env.production? || Rails.env.development? || Rails.env.test?)
```
despo: is there any value in having this then? And if it's just for staging can we change it to `Rails.env.staging?`

tahb:
@despo Done 👍

PR #142, opened 2018-04-19
[57] Create vacancy button above the fold

tahb:
* Includes minor tweaks to the language used in respect to required and optional that help facilitate the break up of the long list.
* Double sign posting the start of the create journey with a hyper link that explains what the user should do, if they don't scroll all the way down there's a strong chance this may be remembered and used instead of the primary, large green call to action.

Before:
screen shot 2018-04-19 at 13 39 09

After:
screen shot 2018-04-19 at 13 50 15

After(mobile)
screen shot 2018-04-19 at 13 15 52


tahb:
[create a vacancy button - move to top of page](https://trello.com/c/x3vO6Xvm/57-create-a-vacancy-button-move-to-top-of-page)

PR #144, opened 2018-04-19
(feature) add unique index to school’s urn

despo:
Before running this in development you have to update any records with duplicate entries manually or recreate the database and seed


despo:
[Add unique index to school urn](https://trello.com/c/CBr6dHKX/65-add-unique-index-to-school-urn)

PR #150, opened 2018-04-23
[91] Hiring staff can view their own vacancies

tahb:


tahb:
[Hiring staff viewing a hiring_stagg/vacancy falls over](https://trello.com/c/3lezkqHf/91-hiring-staff-viewing-a-hiringstagg-vacancy-falls-over)

PR #151, opened 2018-04-24
[94] Hiring can see the no content copy after deleting all their vacancies

tahb:
Before:
screen shot 2018-04-24 at 10 40 12

After:
screen shot 2018-04-24 at 10 49 20


tahb:
[Hiring staff that delete all of their vacancies see an empty table](https://trello.com/c/NvwbhtMk/94-hiring-staff-that-delete-all-of-their-vacancies-see-an-empty-table)

PR #152, opened 2018-04-24
[68] Vacancy validation is more permissive

tahb:


tahb:
[Vacancy validations](https://trello.com/c/8PN6OC0g/68-vacancy-validations)

PR #156, opened 2018-04-24
Fix/96 scraped listings dont map to school if in doubt

tahb:
* Take the scraped vacancies from 8 to 17 on a fresh database
* Stop importing 20k schools that are marked as 'Closed' - I found this when debugging the wrong school allocation with the new logging in this PR. This now speeds up import and increases the chances we match the right school.
* When there are schools that have 3 words the same but differ by 1 eg. `Eltham Sixth Form Centre` then this is a dangerously incorrect match. This change makes the search match on the same number of words.


tahb:
[Scraped vacancies match the wrong school](https://trello.com/c/byfqHGF0/96-scraped-vacancies-match-the-wrong-school)

PR #160, opened 2018-04-24
(chore) use ES’s `import`

despo:


lib/tasks/elasticsearch.rake
```diff
@@ -3,7 +3,7 @@ namespace :elasticsearch do
   namespace :vacancies do
     task index: :environment do
       Rails.logger.debug("Indexing vacancies in #{Rails.env}")
-      Vacancy.__elasticsearch__.create_index! force: true
+      Vacancy.import
```
tahb: @despo Do you have a link to where you learnt about this? After using it to test the pagination without success I would like to see if I can learn the difference which makes me think that would be good to have in the commit for future people.

tahb: I found their documentation but the way it is described there is only for first time use. I was hoping for more information to help me understand what problem this change resolves so that I can check if it does it or not. I've never needed this rake task so my knowledge of the problem is very low.

Currently it looks like we might want need to clear the existing indexes first before it will have any impact perhaps with a flush.. but as I said, I don't know what the problem is.
```
Vacancy.__elasticsearch__.client.indicies.flush
Vacancy.import
```

Locally I've tried to work out what the problem is..

- Running `Vacancy.__elasticsearch__.create_index! force: true` removes all my vacancies from the front end.
- Running `Vacancy.import` adds them back.

- Running `Vacancy.import` alone returns only `0` and doesn't appear to change anything

lib/tasks/elasticsearch.rake
```diff
@@ -3,7 +3,8 @@ namespace :elasticsearch do
   namespace :vacancies do
     task index: :environment do
       Rails.logger.debug("Indexing vacancies in #{Rails.env}")
-      Vacancy.__elasticsearch__.create_index! force: true
+      Vacancy.__elasticsearch__.client.indicies.flush
```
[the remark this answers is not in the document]
despo: 🤔

PR #161, opened 2018-04-26
Chore/testlogin cannot sign into production

tahb:


tahb:
[Azure test user cannot be used on production](https://trello.com/c/gaUtoTpY/99-azure-test-user-cannot-be-used-on-production)

PR #164, opened 2018-04-27
(Feature) 84 payscale as a range

despo:


despo:
[payscale as a range](https://trello.com/c/C26tu7HB/84-payscale-as-a-range)

app/views/hiring_staff/vacancies/job_specification/edit.html.haml
```diff
@@ -31,6 +31,25 @@
                 wrapper_html: {id: 'subject'},
                 collection: Subject.order(:name),
                 required: false
+      %fieldset.form-group#pay_scale_range
+        %label.form-label.form-label-bold= t('vacancies.pay_scale_range')
```
tahb: The form header no longer says this is an "(optional)" field.

app/views/hiring_staff/vacancies/job_specification/edit.html.haml
```diff
@@ -31,6 +31,25 @@
...
+                  collection: pay_scale_options,
+                  label_method: :label,
+                  label: false,
+                  required: false,
+                  wrapper: false
+
+        to
+
+        = f.input :max_pay_scale_id,
+                  collection: pay_scale_options.reverse,
```
tahb: When you click this drop down you get the following content which fills the entire height of my screen:

screen shot 2018-05-02 at 12 07 39


This AC isn't totally clear in regards to the second field but it is the only field referencing ordering and doesn't talk about reversing:
> - order content of drop down as main, then upper, then lead (ascending ££ value and from most to least commonly used)

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -28,7 +28,8 @@
     let!(:leaderships) { create_list(:leadership, 3) }
     let(:vacancy) do
       VacancyPresenter.new(build(:vacancy, :complete,
-                                 pay_scale: pay_scales.sample,
+                                 min_pay_scale: pay_scales.sample,
```
tahb: Has this change been driven through tests as with this as the only change to the integration spec it alone does not produce a failing spec which I believe has led to a feature that looks tested but doesn't work when the user tries to do it.

I think we can improve this test file by removing the vacancy object as I assume it is the part that is to blame for the false confidence. Feature tests that are concerned with testing the full flow of creating a particular record in the database should probably not build objects of the same. Perhaps we should pass a hash of explicit parameters to our form helpers that help us fill in the pages and test for them exactly at the end of the process?

When I add values, they don't appear:
screen shot 2018-05-02 at 12 11 16

When I edit the vacancy they are blank:
screen shot 2018-05-02 at 12 12 01

app/views/hiring_staff/vacancies/job_specification/edit.html.haml
despo: From the trello card:
`both dropdowns are optional ([blank] to [UPS 5], [MPS3] to [blank]`

I had asked @Ellayanor about this as at first I thought the order might need to be different since the leadership scales were not mentioned, who confirmed that this meant reverse order.

Regarding the height, I don't get this behavior on my local so maybe it has to do with your local browser settings. But, it's also a QA issue
selected_max_payscale

Ellayanor: hey team - I think this one is on me - I tried to give an example to make it clearer but think I made it worse, sorry. 

There are two aspects: 
1. order content of drop down as main, then upper, then lead (ascending ££ value and from most to least commonly used). The order should be the same for both dropdowns. 

2. it should be accepted if a user **only** makes a selection from either one of the dropdowns, leaving the other one blank.

app/views/hiring_staff/vacancies/job_specification/new.html.haml
```diff
@@ -33,6 +33,25 @@
...
+                  collection: pay_scale_options,
+                  label_method: :label,
+                  label: false,
+                  required: false,
+                  wrapper: false
+
+        to
+
+        = f.input :max_pay_scale_id,
+                  collection: pay_scale_options,
```
despo: removed reversing order

app/views/hiring_staff/vacancies/job_specification/edit.html.haml
```diff
@@ -31,6 +31,25 @@
...
+                  collection: pay_scale_options,
+                  label_method: :label,
+                  label: false,
+                  required: false,
+                  wrapper: false
+
+        to
+
+        = f.input :max_pay_scale_id,
+                  collection: pay_scale_options,
```
despo: removed reversing order

tahb:
All works 👍 

We could go with it but I have a suggestion with the copy used when only one range is provided. Here's what it looks like now:

screen shot 2018-05-08 at 17 02 48
screen shot 2018-05-08 at 17 02 16

We could capitalise the first word.

The words `from` and `up to` make perfect sense for numbers eg `above £30,000` but their use with a range gave me pause. Would `Above main pay scale 1` and `Below upper pay scale 2` read any better?
 
- [ ]

PR #165, opened 2018-04-30
[106] Track user applications

tahb:
* Noticed that the test suite could make any outbound request it wanted when making sure Google Analytics would not be called in test.
* I was unsuccessful in TDDing this feature after an hour of digging and investigation that was aided by [Puffing Billy](https://github.com/oesmith/puffing-billy). It seems that this page will always send the same `get` request to GA with or without this feature. The body of the request is always empty and there are not other data to set assertions on, screenshot below.
screen shot 2018-04-30 at 12 37 29
* Refactored Capybara and DatabaseCleaner configuration as it was noisy during this investigation


tahb:
[Track when a user clicks apply](https://trello.com/c/UCbJ0o71/106-track-when-a-user-clicks-apply)

PR #171, opened 2018-04-30
[88, 33] Vacancies are always sorted when pagination, fragile test fixed

tahb:
Failing test seed that can be used to reproduce this error without this change: rspec spec --seed=65486

* I think because elasticsearch-model defines the Kaminar page method and calls self.search, this doesn’t work as expected when self is an array of records instead of a query object. Changing that seems to fix the issue.
* I looked around to see the ordering other people use and found that this ordering was typical: https://github.com/elastic/elasticsearch-rails/issues/16#issuecomment-35596319

—

method definition: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb#L29

troublesome line: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb#L36


tahb:
[bug: sort by posting date doesn't work - job just posted shows as last result (3rd page)](https://trello.com/c/nUnbsNEO/88-bug-sort-by-posting-date-doesnt-work-job-just-posted-shows-as-last-result-3rd-page)

despo:
@tahb this looks good but I we shouldn't need the elasticsearch tag on any features that dont get to the root_path since it doesnt use Elastic Search.

despo:
@tahb I am little bit confused here - isn't this the work I have already done last week on my open pull request? I realise I switched to `will_paginate`, but other than that what is the benefit of this and why was all the work redone?
https://github.com/dxw/teacher-vacancy-service/pull/158/files#diff-e0a51d3baef42a3146156050a1c8b0b6R15

despo:
can you also remove kaminari and rails from application.rb

https://github.com/dxw/teacher-vacancy-service/pull/158/files#diff-b1fe55db50c712fef0673345e5b9c0d9

tahb:
@despo It's now a lot smaller which feels much better, back with you.

despo:
@tahb requiring the gem before `elasticsearch-model` in the Gemfile should address that based on what I came across when looking this up.

tahb:
I know I've replied on Slack already but for the record of https://github.com/dxw/teacher-vacancy-service/pull/171#issuecomment-385614639.

I began with the intent of fixing the intermittent test failure we've had for the last few weeks but in doing so fixed the pagination bug so named my pull request in a way that described the new state of the system. 

I wanted to improve my understanding of what was really going wrong with the test failures because my understanding was that our only working strategy was to swap from Kaminari to WillPaginate in the hope it might fix it. As it turned out we both arrived at the same controller change in the end but hopefully understand how we arrived at this place with a test to prevent it happening again.

tahb:
https://github.com/dxw/teacher-vacancy-service/pull/171#issuecomment-385651439 @despo tests that start passing when our gems are ordered differently makes me feel very uneasy.

That would seem to point to the load order of Kaminari being important and given that it is coupled with more Kaminari code within Elasticsearch-model I worry we don't understand this issue well enough to bolt a commit onto the end of this PR with confidence.

PR #173, opened 2018-05-01
(feature) job count text is more contextual

gazaston:
Previously, the initial view of the vacancies#index page would say ‘there are n vacancies that match your search’ even though as far as the user is concerned, they hadn’t ‘searched’ yet.

- the `VacanciesController` has a new private method `searched?` that checks for the presence of the `:commit` key which appears in the query params when the user applies filters.
- this method is passed into new instances of the `VacanciesPresenter` where it’s used to conditionally return the correct text via the `total_count` method

before (initial):
before

after (initial):
after initial

after (with search):
after with search


gazaston:
[service still shows '27 results match your search' when you first land on the site](https://trello.com/c/xLWbjClV/98-service-still-shows-27-results-match-your-search-when-you-first-land-on-the-site)

app/controllers/vacancies_controller.rb
```diff
@@ -73,4 +73,8 @@ def sort_column
   def sort_order
     params[:sort_order]
   end
+
+  def searched?
+    params[:commit]
```
tahb: Worried that this could return true by accident in future if another form is added to the page. I can't foresee what form that might be at the moment but the trouble it would present in the form of a silent bug that would skip past our tests isn't worth a small tweak now:

2 ideas:
- assert that the `commit` value is for our search form eg `params[:commit] && params[:commit].eql?(I18n.t('buttons.apply_filters'))`
- add a new hidden field to the form explicitly for this purpose

PR #175, opened 2018-05-01
Hiring staff that are not yet associated with a school should be told that

despo:


despo:
The error occurs when refreshing `http://localhost:3000/auth/azureactivedirectory`

despo:
[Handle bad log in error](https://trello.com/c/AMDccmNJ/54-handle-bad-log-in-error)

despo:
[Hiring staff that are not yet associated with a school should be told that](https://trello.com/c/KT5OHcL6/101-hiring-staff-that-are-not-yet-associated-with-a-school-should-be-told-that)

despo:
[Bad login error message](https://trello.com/c/RUUoY1mP/50-bad-login-error-message)

app/controllers/hiring_staff/sessions_controller.rb
```diff
@@ -9,13 +9,11 @@ def new
...
     permission = Permission.new(identifier: oid)
 
-    if permission.valid?
-      session.update(session_id: oid)
-      session.update(urn: permission.school_urn)
-      redirect_to school_path(current_school.id)
-    else
-      redirect_to root_path, notice: I18n.t('errors.sign_in.unauthorised')
-    end
+    return redirect_to root_path, notice: I18n.t('errors.sign_in.unauthorised') unless permission.valid?
```
tahb: Reads nicer having the unhappy case return early leaving the rest of the code to tell the story of the happy path 👍 .

tahb:
When implementing the DfE Sign in work I was looking to undo this change but then realised it had not been merged. It is likely that this will no longer be required very shortly.

despo:
@tahb I've pushed a new branch with the Azure error fix but I think we should leave it as it is Azure specific.  This PR is just the updated error message now.

config/locales/en.yml
```diff
@@ -155,7 +155,7 @@ en:
     jobs:
       unable_to_publish: We were unable to publish this job. Make sure that you have filled all mandatory fields
     sign_in:
-      unauthorised: You are not authorised to act on behalf of this school
+      unauthorised: Your school has not been onboarded yet. Please try again when it has.
```
tahb: @despo This doesn't tell the user _how_ to resolve their issue and become 'on-boarded' which risks a frustrating experience. I think we could make this better by giving an instruction to contact us, perhaps:

> Your school has not been invited yet, you can register your interest by emailing isobel.croot@digital.education.gov.uk. You will be able to try again after receiving your invitation.

tahb: We were just talking about this @Ellayanor, do you have anything further to add?

Ellayanor: I think you're right @tahb - worth checking with Isobel whether the generic email contact is imminent, but otherwise I like what you've suggested as a starting point, we definitely need to give users some idea of how to proceed. Michael needs input into this before we deploy it.

Ellayanor: https://trello.com/c/BbGZrTGU

tahb: I've pointed Michael at this thread.

despo:
@Ellayanor @tahb can we close this PR and move the card back? The effort to update the label is a lot less than having to keep the PR up to date and also notice further comments on it.  This requirement should have either been the card was added to the backlog, or went to staging and came back as an improvement.

Ellayanor:
@despo that makes sense to me - the original card never went through product in order to make sure requirements were clear (which is what our new process tries to fix) so hopefully this won't happen in future.

despo:
@Ellayanor I've linked the card mentioned above to the existing  development card. Closing.

Please add any further updates to the [trello card](https://trello.com/c/KT5OHcL6/101-hiring-staff-that-are-not-yet-associated-with-a-school-should-be-told-that)

PR #182, opened 2018-05-02
(chore) job spec number field JS consistency

gazaston:
The inline JS for removing commas in IE8 number fields wasn’t being applied consistently.

- abstracted inline JS into a function
- added a teaspoon spec
- added consistently across new and edit views

### DIFF
--- a/app/views/hiring_staff/vacancies/job_specification/edit.html.haml
+++ b/app/views/hiring_staff/vacancies/job_specification/edit.html.haml
@@ -43,7 +43,7 @@
                     required: true,
                     wrapper: false,
                     label: false,
-                    input_html: { class: 'form-control-1-8', step: '1', min: '1000', max: '1000000', onchange: "this.value = this.value.replace(/,/g, '')" }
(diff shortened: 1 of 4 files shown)


gazaston:
[Job specification form: 'new' needs same number field JS as 'edit'.](https://trello.com/c/vZ3OX9la/122-job-specification-form-new-needs-same-number-field-js-as-edit)

app/views/hiring_staff/vacancies/job_specification/edit.html.haml
```diff
@@ -54,7 +54,7 @@
                     wrapper: false,
                     required: true,
                     label: false,
-                    input_html: { class: 'form-control-1-8', step: '1', min: '1000', max: '1000000', onchange: "this.value = this.value.replace(/,/g, '')" }
+                    input_html: { class: 'form-control-1-8', step: '100', min: '1000', max: '1000000', onchange: "this.value = removeCommaFromNumber(this.value)" }
```
despo: shouldn't the step here be 1?

gazaston: Good spot.

PR #185, opened 2018-05-02
Fix/add sns policy for cloudwatch events

Stretch96:


tahb:
[The team need to know when a deployment fails](https://trello.com/c/6k59yzWw/123-the-team-need-to-know-when-a-deployment-fails)

PR #186, opened 2018-05-02
(Fix) Don't send opsgenie alerts for pipeline fails

Stretch96:
* If the 'NewStateValue' in the message is 'ALARM', the message gets
sent to Opsgenie. This fix changes it to 'SOFT_ALARM' so that it will
still send to Slack.


tahb:
[The team need to know when a deployment fails](https://trello.com/c/6k59yzWw/123-the-team-need-to-know-when-a-deployment-fails)

PR #189, opened 2018-05-02
Chore/117 change the way flexible working is displayed to jobseekers

gazaston:
- if no flexible working, show nothing in the job post
- if flexible working, show an invitation to email for further information in the job post
- add some extra hint text to inform hiring staff that users will be invited to email for info

form before:
form before
form after:
form after
job post before
job post before
job post after
job post after
job post no flexible working
job post no flexible working


gazaston:
[change the way flexible working is displayed to jobseekers](https://trello.com/c/cZU0CiNl/117-change-the-way-flexible-working-is-displayed-to-jobseekers)

app/presenters/vacancy_presenter.rb
```diff
@@ -53,7 +54,12 @@ def publish_today?
   end
 
   def flexible_working
-    @flexible_working = model.flexible_working ? 'Yes' : 'No'
+    if model.flexible_working?
+      mailto = mail_to(model.contact_email, model.school.name)
+      @flexible_working = safe_join([I18n.t('vacancies.flexible_working_info', mailto: mailto).html_safe])
+    else
+      'No'
```
despo: I would have expected this to be `false` since we will never utilise this text

spec/presenters/vacancy_presenter_spec.rb
```diff
@@ -80,16 +80,16 @@
...
   describe '#flexible_working' do
-    context 'it converts flexible working' do
-      it 'to Yes' do
-        vacancy = VacancyPresenter.new(build(:vacancy, :complete))
-        expect(vacancy.flexible_working).to eq('Yes')
-      end
+    it 'shows nothing if flexible working is not available' do
+      school = create(:school)
+      vacancy = VacancyPresenter.new(build(:vacancy, school: school, flexible_working: false))
+      expect(vacancy.flexible_working).to eq('No')
```
despo: the same not about returning false applies here. The spec would also make more sense as it currently doesn't desribe what it tests.

app/presenters/vacancy_presenter.rb
gazaston: The 'No'? It's shown in the review page, isn't it?

despo: So we are reusing this in the review page. A bit confusing but ok 👍

gazaston: I only changed the presenter method, not where it's called ¯\_(ツ)_/¯

PR #192, opened 2018-05-02
Production deploy #5

tahb:
- to verify the new url works with sign-in we need to temporarily add access for a dev (me) to sign in


tahb:
I can test this with the root account, no need for this.

PR #197, opened 2018-05-02
Feature/add logspout service

Stretch96:


tahb:
[The team need to have the ability to set up notifications for events](https://trello.com/c/Vdgdq4ZG/124-the-team-need-to-have-the-ability-to-set-up-notifications-for-events)

logspout_container_definition.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "name": "${task_name}",
+    "image": "gliderlabs/logspout:latest",
```
tahb: The last time I used logspout their upgrade from v2 to v3 broke things for us because we used `:latest`. What do you think about locking this to the current image version we've tested it with? Looks like that would be `:v3.2.4` in this case: https://hub.docker.com/r/gliderlabs/logspout/tags/

terraform/modules/ecs/ecs.tf
```diff
@@ -34,6 +34,23 @@ resource "aws_ecs_service" "web" {
...
+resource "aws_ecs_service" "logspout" {
+  name            = "logspout-${var.environment}"
+  cluster         = "${aws_ecs_cluster.cluster.id}"
+  task_definition = "${aws_ecs_task_definition.logspout.arn}"
+  desired_count   = "${var.ecs_logspout_task_count}"
+
+  deployment_minimum_healthy_percent = 50
+
+  placement_constraints {
+    type = "distinctInstance"
```
tahb: I see the commit message says 'Places one container per instance' however does this also ensure that no matter the `desired_count` we only ever have 1 container running, never more? Thinking about the potential to misconfigure this and end up with multiple containers duplicating all logs to Papertrail.

Stretch96: If the `desired_count` is higher than the current number of instances running, the service will fail, resulting in no containers being launched.
If the `desired_count` is less than the current number of instances running, the containers will be launched, but there will be instances without the container running.
`desired_count` needs to be stepped up and down along with the same metric that autoscales the instances

tahb:
Because the amount of traffic going to Papertrail was very high and it was hard to alert on certain events as they went over multiple lines I've added in [Lograge](https://github.com/roidrage/lograge) to make this change easier to use.

It also gives us the ability to remove /check requests from the logs which will dramatically reduce the amount of storage we need to page for.

PR #198, opened 2018-05-03
(chore) remove ref no from job posting

gazaston:
before:
before

after:
after


gazaston:
[remove vacancy reference numbers from user facing pages](https://trello.com/c/u2suYgyh/115-remove-vacancy-reference-numbers-from-user-facing-pages)

despo:
@gazaston do we utilise this number anywhere now? If not, do we need a card to remove it completely from the database as well?

PR #200, opened 2018-05-03
(chore): Papertrail has a reliably way to hook into an event

tahb:
.. and send notifications as appropriate.
* There are no SQL logs in production so we couldn’t look for an SQL statement of INSERT for example.

screen shot 2018-05-03 at 15 29 33


tahb:
[The team need to have the ability to set up notifications for events](https://trello.com/c/Vdgdq4ZG/124-the-team-need-to-have-the-ability-to-set-up-notifications-for-events)

PR #207, opened 2018-05-07
(feature) rename routes `/vacancies` to `/jobs`

despo:


despo:
[Consistent use of language — "Vacancies" should become "Jobs" site-wide](https://trello.com/c/BYQz1H3G/102-consistent-use-of-language-vacancies-should-become-jobs-site-wide)

gazaston:
Do you think we should change the instances of 'Vacancy' in `en.yml` as part of this PR?

despo:
@gazaston I did begin working on that but there were a lot of places where I was unsure if it should be `job`, or `job vacancy` so I think it will be simpler to do it as part of another card.

tahb:
This will break the Google Analytic funnels so we must remember to update them as part of this on every view (environment).

despo:
@tahb are there some instructions on how to do this? Can @Ellayanor do it or did you set it up?

tahb:
@despo I set them up but don't have any instructions to share. It likely involves adding more regex in order to support both old an new paths.

screen shot 2018-05-08 at 11 37 11

PR #213, opened 2018-05-09
(chore) change ‘vacancy’ to ‘job’ in UI

gazaston:


gazaston:
[Change "Vacancies" to "Jobs" in the UI, site-wide](https://trello.com/c/8k3PD8NH/143-change-vacancies-to-jobs-in-the-ui-site-wide)

config/locales/en.yml
```diff
@@ -24,7 +24,7 @@ en:
     flexible_working_info: 'Yes. Email %{mailto} for more information.'
     starts_on: 'Expected start date'
     ends_on: 'End date'
-    expired: 'This vacancy has expired'
+    expired: 'This job has expired'
```
despo: @gazaston  This is where I got stuck before - should this not be `This job vacancy has expired?` It does not read correctly as it is.

gazaston: Perhaps 'This job posting has expired' instead?

despo: Actually `job post` as `job vacancy` is even more confusing.

PR #215, opened 2018-05-09
(feature): Development uses HTTPS

tahb:
screen shot 2018-05-09 at 15 12 34

* Required for DfE Sign in
* This resource was very useful in creating this: https://medium.freecodecamp.org/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec
* All developers will need to follow the instructions to trust a CA pem file that will need to be distributed separately
* We’ve decided not to include the above file in source control because even though it’s password protected it still presents a risk of a hostile actor brute forcing the password to create a certificate and targeting a developer via a phishing attack
* To mitigate the inconvenience of not having this close to the code there are instructions on how to replicate the process and create new files
* This change will require SSO and other integrations to communicate in dev with HTTPS too
* There is no auto redirect from http://localhost to https://localhost - this might be irritating for developers. Appreciate any ideas to help with this.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,8 @@
 .env
 .dockerignore
 config/application.yml
+
+# Terraform
 *.tfstate
 *.tfstate.backup
 .terraform/environment
@@ -10,3 +12,7 @@ terraform.tfstate
 ./*.tfstate
 .terraform/
 .terraform.tfstate.lock.info
+
+# Certificates
+config/localhost/https/*.key
+config/localhost/https/*.crt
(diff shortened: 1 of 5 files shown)


tahb:
[Support https in dev for dfe sign-in integration](https://trello.com/c/zE6CWd19/144-support-https-in-dev-for-dfe-sign-in-integration)

tahb:
I would like to change approach on this so could we park please!

tahb:
Ready to go.

PR #221, opened 2018-05-10
 Production deploy #10

despo:


despo:
@fionamurray that change was deployed in a previous release so it shouldn't prevent this from going in as it doesn't impact it.

tahb:
I've made a PR addressing Fiona's feedback https://github.com/dxw/teacher-vacancy-service/pull/226

PR #224, opened 2018-05-10
[150] Rake tasks can log in production

tahb:
* Followed this guide to get rake tasks to log http://macbury.ninja/2014/8/better-loging-in-rake-tasks
* When `RAILS_ENV` is production or staging rake tasks will not log to STDOUT and therefore don't get collected by logspout and piped to Papertrail.
* This will allow the Papertrail alerts to ping when we scrape a vacancy

screen shot 2018-05-10 at 15 08 46

### DIFF
--- a/lib/tasks/logging.rake
+++ b/lib/tasks/logging.rake
@@ -1,0 +1,9 @@
+desc 'switch rails logger to stdout'
+task verbose: [:environment] do
+  Rails.logger = Logger.new(STDOUT)
+end
+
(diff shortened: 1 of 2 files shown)


tahb:
[Rake task output should log to Papertrail](https://trello.com/c/ZrBFesQe/150-rake-task-output-should-log-to-papertrail)

despo:
@tahb it needs rebasing

PR #225, opened 2018-05-10
Chore/140 add microcopy for benefits

gazaston:
before:
before

after:
after


gazaston:
[Add microcopy for 'Benefits'](https://trello.com/c/mCCvdpEU/140-add-microcopy-for-benefits)

PR #233, opened 2018-05-14
(chore) lastest vacancy updates

despo:


despo:
These are actually for the 10th of May, my branch name is wrong as I based it on the trello card check item, which must be out of date.

PR #245, opened 2018-05-16
[171] Track empty search results

tahb:
* When no results are returned from ES then we send an info event to Rollbar with the parameters given that caused it.
* VacancyFilters and VacancySort classes were not being tested when run alone and have been moved from models to services. More info on commit.
* When a search is carried we notify as close to ElasticSearch as possible to allow reuse should we allow search from more places than the existing controller action.
* Fix a dev issue with the test-server stating up with errors
* VacancySort default parameters are defined in the class, not on every constructor. This cleans the controller up and makes it easier to test other code that is dependent upon it.


tahb:
[Understand why users are getting no search results](https://trello.com/c/yAtByC2A/171-understand-why-users-are-getting-no-search-results)

PR #252, opened 2018-05-17
(fix) better spacing for job list sort links

gazaston:
#### before:
before

#### after:
after


gazaston:
[Sort links need spacing from filter section](https://trello.com/c/xZ0bdGOC/46-sort-links-need-spacing-from-filter-section)

PR #254, opened 2018-05-17
(fix) school jobs table no longer breaks on mobile

gazaston:
- scrolls horizontally instead of poking out of the layout

#### before
before

#### after

after


gazaston:
[Vacancies table breaks layout at small widths](https://trello.com/c/xcdLHOY3/45-vacancies-table-breaks-layout-at-small-widths)

PR #269, opened 2018-05-24
[212] Handle the origin IP not our CloudFront IP in logs and integrations

tahb:
* We have tried to trust Rack to return the right IP by using `request.remote_ip` but it appears that this is not enough. It is not correct in Papertrail and in Rollbar who also do similar logic to acquire the correct IP from Rack.
* On a deeper look we need to configure a list of allowed proxies so that Rack can return us the ‘correct’ IP. Configuring AWS IP ranges on boot is a future proof way of allowing any CloudFront IP as we cannot whitelist them or reasonably hope to keep in sync by defining our own static list of the same: https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L164
* I’ve designed this for best-effort. If any Net::HTTP error is raised on boot in staging or production it will default back to logging the CloudFront IP and log a warning.
* Following advice for Net::HTTP exception handling here http://tammersaleh.com/posts/rescuing-net-http-exceptions/ instead of resucing from StandardError for the same reasons as mentioned.
* This curl was useful in diagnosing the problem that was appearing in staging but not in dev: curl -XGET -H 'X-Forwarded-For: 213.123.63.147, 54.182.244.82' 'https://localhost:3000/jobs’
* Trust AWS IP range: https://forums.aws.amazon.com/ann.jspa?annID=2051

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,3 +1,4 @@
+require 'lib/aws_ip_ranges'
 Rails.application.configure do
   # Settings specified here will take precedence over those in
   # config/application.rb.
@@ -105,4 +106,6 @@ Rails.application.configure do
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.action_dispatch.trusted_proxies = cloudfront_ips.map { |proxy| IPAddr.new(proxy) }
 end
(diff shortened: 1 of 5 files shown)


tahb:
[Second attempt to log the IP address of our users not our dynamic CloudFront IP](https://trello.com/c/rjStSKyk/212-second-attempt-to-log-the-ip-address-of-our-users-not-our-dynamic-cloudfront-ip)

PR #273, opened 2018-05-25
(feature)34 string salary fields

despo:


despo:
[Make error messaging consistent with GOV UK patterns](https://trello.com/c/rouiBElA/159-make-error-messaging-consistent-with-gov-uk-patterns)

despo:
This is on [edge](http://tvs.edge.dxw.net/) fo review

gazaston:
This is heroic work :)

A couple of things — the gov.uk elements documentation states that when errors are present on a page:

> you should reflect that there's been an error in the <title> by prefixing the existing title with "Error: ". That will make sure screen readers are alerted to there being an error as soon as possible.

> The error summary must appear at the top of the page, so it is visible when the page is refreshed and is immediately read out by assistive technology. 

^ we currently anchor to the first error, when we should just show the top of the page so the error summary is immediately visible

Also, I think the error message that shows a comma (below) might be misleading — because we specify no commas in our formatting message:

screen shot 2018-05-29 at 15 55 30

despo:
@gazaston happy to remove the scrolling to the specific field - it does only happens for single errors though so if there are multiple error messages you are taken to the top of the error box

Ok, I will remove the formatter from the error message.

Should I also apply the changes in the other errors PR changing the displayed summary error message?

gazaston:
> Should I also apply the changes in the other errors PR changing the displayed summary error message?

Makes sense to me :)

PR #283, opened 2018-05-29
(chore) jobs listing updates 25/05

despo:


despo:
[25 May scraped listing amends](https://trello.com/c/IPipNcph/217-25-may-scraped-listing-amends)

PR #287, opened 2018-05-31
fix/222-apply-link-mentioned-twice

gazaston:
before:
before

after:
after


gazaston:
[Apply link mentioned twice in school#show page under the job post required fields](https://trello.com/c/dixl5XWl/222-apply-link-mentioned-twice-in-schoolshow-page-under-the-job-post-required-fields)

PR #288, opened 2018-05-31
(fix) use strict GOV.UK palette for flashes

gazaston:
- removes sass color functions, applies set tints from GOV.UK palette
- uses GOV.UK `$black` for borders and text in flashes 

before:
before

after:
after


gazaston:
[Flash message colours](https://trello.com/c/KK3DRG4i/180-flash-message-colours)

PR #290, opened 2018-05-31
(fix) apply .wordwrap class to school URL

gazaston:
before:
before

after:
after


gazaston:
[Apply word-wrap to URLs in school details](https://trello.com/c/dCnhijir/182-apply-word-wrap-to-urls-in-school-details)

PR #292, opened 2018-05-31
(fix) correct markup of school info table

gazaston:
This table was marked up as a definition list, using code based on [an example from the GOV UK prototype kit which contained invalid markup](https://govuk-prototype-kit.herokuapp.com/docs/examples/check-your-answers-page) (as detailed in the DAC report, see attachment).

I've altered the markup to match [another example within the prototype kit marked up as a `<table>`](https://govuk-prototype-kit.herokuapp.com/docs/examples/pass-data/vehicle-check-answers), which doesn't require any change to `check_your_answers.scss`. 

I've added a `.strong` class to `_utilities.scss` for emboldening the first column.

before:
before

after:
after

### DIFF
--- a/app/assets/stylesheets/base/_utilities.scss
+++ b/app/assets/stylesheets/base/_utilities.scss
@@ -22,3 +22,7 @@
 .overflow-scroll {
   overflow: scroll;
 }
+
+.strong {
+  font-weight: bold;
+}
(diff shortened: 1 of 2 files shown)


despo:
[School info table on dashboard page incorrectly marked up](https://trello.com/c/wGiMaChr/185-school-info-table-on-dashboard-page-incorrectly-marked-up)

PR #296, opened 2018-06-04
(fix) add breadcrumb to job show page

gazaston:
before:
before

after:
after


gazaston:
[More than one way is not available to locate a web page within a set of web pages](https://trello.com/c/5AJY5TCU/189-more-than-one-way-is-not-available-to-locate-a-web-page-within-a-set-of-web-pages)

PR #298, opened 2018-06-04
(chore) remove school id from session redirect url

despo:


despo:
@gazaston could you have a look at this please? Not creating any issues but after logging in to staging it redirects users to `staging.*/school.<school_id>`

PR #301, opened 2018-06-05
(fix) adds specific page titles globally

gazaston:


gazaston:
[Add descriptive page titles globally](https://trello.com/c/WXs4x79O/206-add-descriptive-page-titles-globally)

PR #306, opened 2018-06-06
(fix) add legend and aria-label to pay scale range

gazaston:
Two inputs cannot share a single ambiguous <label> — replaced `<label>` with a descriptive`<legend>` (see [‘When to use a fieldset and legend’](https://accessibility.blog.gov.uk/2016/07/22/using-the-fieldset-and-legend-elements/) on the GDS a11y blog)

Also added [`aria-label` attributes](https://webaim.org/techniques/forms/advanced#arialabel) to both `<select>`s for assistive technology.


gazaston:
[‘Pay scale range (optional)’ combo boxes are not labelled correctly](https://trello.com/c/htKde0ce/194-pay-scale-range-optional-combo-boxes-are-not-labelled-correctly)

PR #307, opened 2018-06-07
Fix/188 review page change links should not link directly to fields

gazaston:
review before:
review before

review after:
review after

edit before:
edit before

edit after:
edit after

job post before:
job post before

job post after:
job post after


gazaston:
[Review page 'change' links should not link directly to fields](https://trello.com/c/BfgXGWZ0/188-review-page-change-links-should-not-link-directly-to-fields)

PR #308, opened 2018-06-07
(fix) add aria-label to sort links

gazaston:
- this is a more elegant solution than that implemented in https://github.com/dxw/teacher-vacancy-service/pull/303


gazaston:
[Job list sort links need more descriptive text](https://trello.com/c/XNC66tG0/191-job-list-sort-links-need-more-descriptive-text)

PR #309, opened 2018-06-07
(fix) add aria attributes to salary range fields

gazaston:
- add a `<legend>` with an ID to the `<fieldset>`
- add `aria-label` to each field
- add `aria-labelledby` to each field to use the legend text as a label for assistive technology


gazaston:
[The ‘Salary range’ text fields are not labelled correctly.](https://trello.com/c/FC40GI6V/196-the-salary-range-text-fields-are-not-labelled-correctly)

PR #310, opened 2018-06-11
(fix) put aria-labels into i18n

gazaston:


gazaston:
[Review page 'change' links should not link directly to fields](https://trello.com/c/BfgXGWZ0/188-review-page-change-links-should-not-link-directly-to-fields)

PR #311, opened 2018-06-11
(fix) improve text and location of ‘apply’ button

gazaston:
- change button text to ‘Get more information’ (also add aria-label fo a11y purposes)
- move button to top of sidebar for more prominence
- add some explainer text above the button

before:
before

after:
after


gazaston:
['Apply for this job' button text revision](https://trello.com/c/3Bvqx25K/238-apply-for-this-job-button-text-revision)

PR #312, opened 2018-06-11
Fix/192 use descriptive text when displaying website links

gazaston:
before:
before

after:
after


gazaston:
[Use descriptive text when displaying website links](https://trello.com/c/NVbRogyj/192-use-descriptive-text-when-displaying-website-links)

PR #313, opened 2018-06-11
Fix/199 flexible working optional check box is not labelled correctly

gazaston:


gazaston:
[‘Flexible working (optional)’ check box is not labelled correctly](https://trello.com/c/EaAEu12U/199-flexible-working-optional-check-box-is-not-labelled-correctly)

PR #314, opened 2018-06-11
(fix) prefix page title on pages with errors

gazaston:
as per guidance on [GOV.UK Elements](https://govuk-elements.herokuapp.com/errors/):

> Additionally, you should reflect that there's been an error in the <title> by prefixing the existing title with "Error: ". That will make sure screen readers are alerted to there being an error as soon as possible.


gazaston:
[Page titles should indicate that there are errors on a page](https://trello.com/c/BBOFIst9/203-page-titles-should-indicate-that-there-are-errors-on-a-page)

PR #315, opened 2018-06-11
(fix) adds min. char. count to hint text

gazaston:
- as per Digital Accessibilty Centre recommendation

before:
before

after:
after


gazaston:
[Insufficient hint text for minimum character counts & values](https://trello.com/c/oXnqwLLB/200-insufficient-hint-text-for-minimum-character-counts-values)

PR #316, opened 2018-06-11
(fix) add a title attribute to google maps iframe

gazaston:
- uses [a bit of jquery](https://css-tricks.com/snippets/jquery/run-javascript-only-after-entire-page-has-loaded/) to wait until the page is fully loaded before targeting the gmaps `<iframe>` and setting the `title` attribute
- if there's a better way to do this I'd love to know


gazaston:
[Google Maps iframe needs title attribute](https://trello.com/c/2MdlqlVF/187-google-maps-iframe-needs-title-attribute)

PR #317, opened 2018-06-12
(fix) .overflow-scroll only applied on mobile

gazaston:
Removes visible scroll bars around table.

before:
before

 after:
after


gazaston:
[Apply overflow: scroll only on mobile](https://trello.com/c/65WEssqG/177-apply-overflow-scroll-only-on-mobile)

PR #319, opened 2018-06-12
WIP: Fix/167 display public job listing url on school job post page

gazaston:


gazaston:
[Display public job listing URL on school/job post page to enable schools to easily share/access](https://trello.com/c/BR8QYQLM/167-display-public-job-listing-url-on-school-job-post-page-to-enable-schools-to-easily-share-access)

app/views/hiring_staff/schools/_vacancy.html.haml
```diff
@@ -1,5 +1,5 @@
 %tr.vacancy[vacancy]
-  %td= link_to vacancy.job_title, school_job_path(vacancy.id), class: 'view-vacancy-link'
+  %td= link_to vacancy.job_title, job_path(vacancy), class: 'view-vacancy-link'
```
despo: This changes existing functionality and removes access to the admin job page.
The dashboard should be extended to cater for displaying (or copying to clipboard maybe?) the public job listing url - and only for published jobs as we should never link to non existing links.

gazaston:
**This branch is a WIP.**

- only shows public share url if viewed via admin url
- if user has js enabled, they get the option to click to copy url (using [clipboard.js](https://clipboardjs.com/)
- if user has no js, the url is rendered in full for manual copying

**Todos:**

- [ ] Make `hostname` dynamic in `config/environments/[env]`

app/views/vacancies/_show.html.haml
```diff
@@ -7,6 +7,8 @@
       .banner-warning
         = t('jobs.expired')
 
+  = render partial: 'share_url' if controller_path == 'hiring_staff/vacancies'
```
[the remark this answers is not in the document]
gazaston: I defer to your judgement here 👍

app/views/hiring_staff/vacancies/_share_url.html.haml
```diff
@@ -0,0 +1,7 @@
+.column-full
+  .share-url.display-none
+    %h2.heading-small.mt0 Share this job
+    #{link_to t('jobs.view_public_link'), job_path(@vacancy), id: 'share-url', 'aria-label': t('jobs.view_public_link')}, or #{link_to t('jobs.copy_public_url'), '#', class: 'copy-to-clipboard', 'data-clipboard-text': @vacancy.share_url, 'aria-label': t('jobs.copy_public_url')}
+  .share-url-no-js
+    %h2.heading-small.mt0 Share this job
```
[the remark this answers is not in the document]
gazaston: What do you think the accessibility considerations are for that though? Feels quite unsemantic?

docker-compose.env.sample
```diff
@@ -4,3 +4,4 @@ DATABASE_URL=postgres://postgres@db:5432/tvs_development?template=template0&pool
 REDIS_URL=redis://redis:6379
 AAD_CLIENT_ID=
 AAD_TENANT=
+DOMAIN=localhost:3000
```
despo: can we use `domain = URI(ENV['DOMAIN'])` in development and test since we do set the `DOMAIN` variable?

spec/features/hiring_staff_can_share_their_vacancy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  scenario 'A school can visit their page as the job seeker would' do
+    vacancy = create(:vacancy, school: school)
+
+    visit school_path
+
+    click_on(vacancy.job_title)
+
+    click_on(I18n.t('jobs.view_public_link'))
+
```
despo: Can we remove the empty lines between the actions? I think setup - space - actions - space - expectation reads quite well so we dont really need the new lines.

app/views/hiring_staff/vacancies/_show.html.haml
```diff
@@ -0,0 +1,124 @@
...
+.vacancy.grid-row
+  - if @vacancy.expired?
+    .column-full
+      .banner-warning
+        = t('jobs.expired')
+
+  .column-full
+    .share-url.display-none
+      %h2.heading-small.mt0 Share this job
+      #{link_to t('jobs.view_public_link'), @vacancy.share_url, id: 'share-url', 'aria-label': t('jobs.view_public_link')}, or #{link_to t('jobs.copy_public_url'), '#', class: 'copy-to-clipboard', 'data-clipboard-text': @vacancy.share_url, 'aria-label': t('jobs.copy_public_url')}
```
despo: how does a link_to work on an `aria-label`? Does it just generate the full url?

app/views/hiring_staff/vacancies/_show.html.haml
```diff
@@ -0,0 +1,124 @@
...
+
+    %p= @vacancy.school.description
+
+    %dl.school--details
+      %dt= t('schools.address')
+      %dd= @vacancy.school.full_address
+
+      - if @vacancy.school.phase.present?
+        %dt= t('schools.phase')
+        %dd= @vacancy.school.phase&.titleize
```
despo: Since we already check if phase is present I think we can drop the `&` since there will always be something to apply titleize on.

app/views/hiring_staff/vacancies/_show.html.haml
```diff
@@ -0,0 +1,124 @@
...
+
+      - if @vacancy.school.school_type.present?
+        %dt= t('schools.type')
+        %dd= @vacancy.school.school_type.label
+
+      - if @vacancy.school.url.present?
+        %dt= t('schools.website')
+        %dd
+          = link_to("#{@vacancy.school.name} website (opens in a new window)", @vacancy.school.url, class: 'wordwrap', target: '_blank')
```
despo: Can't we move all this to an `aria-label` instead of having it on the page visible to all users? It seems quite long winded.

spec/features/hiring_staff_can_share_their_vacancy_spec.rb
gazaston: Done.

app/views/hiring_staff/vacancies/_show.html.haml
gazaston: Done.

app/views/hiring_staff/vacancies/_show.html.haml
gazaston: DAC advise semantic text for URLs, which I agree is a good idea for all users.

gazaston:
- Schools can use and share the correct URL when they publish their vacancy
- Uses [clipboard.js](https://clipboardjs.com/) to enable click-to-copy 
- Users with JS turned off get the full URL to copy and paste

### After:
after

### After (without JS)
after without javascript

app/views/hiring_staff/vacancies/_show.html.haml
gazaston: It's just using the translations?

docker-compose.env.sample
tahb: Not sure I understand the purpose of changing to `domain = URI(ENV['DOMAIN'])` however you are right in that the value it not needed in dev or test given the config in staging.rb and production.rb, so perhaps it's not useful and in fact should be removed.

PR #321, opened 2018-06-14
(fix) fix aria labels for pay scale and salary

gazaston:
- remove `aria-labelledby` attributes — these were overriding the individual `aria-label` attributes which in this instance allow a more accurate description
- improve `aria-label` text for pay scale
- put `aria-label` text for salary into i18n


gazaston:
[remove aria-labelledby attributes from pay scale and salary inputs](https://trello.com/c/WQKg29ur/250-remove-aria-labelledby-attributes-from-pay-scale-and-salary-inputs)

despo:
Merging this as it resolves some issues that have already made it to development.

PR #322, opened 2018-06-19
Fix/autoscaling

Stretch96:


tahb:
[Optimise auto-scaling of containers and instances](https://trello.com/c/NODIabq5/152-optimise-auto-scaling-of-containers-and-instances)

terraform/modules/core/core.tf
```diff
@@ -360,7 +416,7 @@ resource "aws_appautoscaling_policy" "down" {
 
     step_adjustment {
       metric_interval_upper_bound = 0
-      scaling_adjustment          = -1
+      scaling_adjustment          = -2
```
tahb: I agree in going up by 2 containers at a time.

I wonder if it would be wise to scale down more slowly than we scale up. There's no real cost to us in taking a more cautious approach but there would be a benefit of having resources already in place if the high load is intermittent.

This also relates to the `period` field above on the scale down alarm which is currently 60 seconds: https://github.com/dxw/teacher-vacancy-service/pull/322/files#diff-8a7db2d745b3f970e5c7652791121c2dR343. 

What do you think about scaling down by only 1 container if CPU is normal for a solid 5 minutes?

Stretch96: Yep, I'll change that to `-1` 👍

despo:
@tahb are you happy with this? Can you review please?

tahb:
I still need Chris to find sometime to make some testing steps for this as I have very low confidence that it achieves the goal at the mo.

Stretch96:
I've managed to test this with small instances

To scale up -

At the initial state there are 3 containers and 2 small instances
The small instances have 2GB memory, 1vCPU (1024 cpu shares)
Each of the containers reserve 512Mb Memory, 256 cpu shares
(The amount of memory and cpu shares should be the same percentage of the total memory and cpu shares. Eg. 512Mb is 25% of the total memory available, so we should set cpu shares at 25% of the total, 256)

Run stress on each of the containers:
```
apt-get update && apt-get -y install stress && stress -c 4
```

The auto scaling for the containers will trigger when the CPUUtilisation for the containers has been above 85% for 5 minutes.
When there are 6 running containers, run stress on those containers as well.

The reserved CPU will now be at ~75% for the cluster.
After 2 minutes above 70%, 2 new instances will be launched and containers placed on them.

To scale down -

Stop stress on all the containers

Either wait half an hour for the containers to slowly scale down
or
run `terraform apply -var-file=workspace-variables/the-env.tfvars` to immediately reduce the amount of containers to 3

After 2 minutes, an instance will be terminated. Then the other after another 2 minutes.

tahb:
Scaling up looks good 👍 I shared a screen recording in our dfe_beta channel. Scaling down hasn't happened well after it should have based on the testing steps so waiting on some help from Chris with it.

tahb:
2 hours after having stopped stressing the containers, we are still at 4 instances with only 3 containers running. I'm cautious of deploying this and all environments silently (in the background) running max instance and incurring twice the costs.

tahb:
Fixed by manually increasing the scale down threshold from 10 to 25 as current usage was ~20%.

PR #323, opened 2018-06-19
Feature/elasticache redis

Stretch96:


tahb:
[Setup REDIS on all environments](https://trello.com/c/QgWzLHV2/244-setup-redis-on-all-environments)

terraform/modules/elasticache-redis/outputs.tf
```diff
@@ -0,0 +1,15 @@
+output "cache_security_group_id" {
```
tahb: Nitpick. This file is named `outputs.tf` however throughout we have been using `output.tf`.

terraform/modules/elasticache-redis/main.tf
```diff
@@ -0,0 +1,41 @@
+# Pulled from https://github.com/turnerlabs/terraform-aws-elasticache-redis
+resource "aws_security_group" "redis" {
+  vpc_id = "${var.vpc_id}"
+
+  tags {
+    Name          = "${var.tag_name}"
```
tahb: Can we name things consistently with the rest of the resources so we can identify them. Instead of `tvs2-staging` it would be good to have the resource in the name so: `tvs2-staging-redis-sg`

terraform/modules/elasticache-redis/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  maintenance_window   = "${var.maintenance_window}"
+  node_type            = "${var.instance_type}"
+  num_cache_nodes      = "1"
+  parameter_group_name = "${var.parameter_group_name}"
+  port                 = "6379"
+  subnet_group_name    = "${aws_elasticache_subnet_group.default.name}"
+  security_group_ids   = ["${aws_security_group.redis.id}"]
+
+  tags {
+    Name          = "${var.tag_name}"
```
tahb: Can we name things consistently with the rest of the resources so we can identify them. Instead of `tvs2-staging` it would be good to have the resource in the name so: `tvs2-staging-redis`

terraform/modules/core/output.tf
```diff
@@ -17,3 +17,11 @@ output "default_security_group_id" {
 output "ecs_autoscaling_group_name" {
   value = "${aws_autoscaling_group.ecs-autoscaling-group.name}"
 }
+
+output "private_subnet_ids" {
+  value = ["${aws_subnet.private_subnet.*.id}"]
+}
+
+output "public_subnet_ids" {
```
tahb: We add this as an output but I don't see that we're using it so looks like it can safely be trimmed before it goes in.

terraform/modules/elasticache-redis/outputs.tf
```diff
@@ -0,0 +1,15 @@
...
+output "hostname" {
+  value = "${aws_elasticache_cluster.redis.cache_nodes.0.address}"
+}
+
+output "port" {
+  value = "${aws_elasticache_cluster.redis.cache_nodes.0.port}"
+}
+
+output "endpoint" {
+  value = "${join(":", list(aws_elasticache_cluster.redis.cache_nodes.0.address, aws_elasticache_cluster.redis.cache_nodes.0.port))}"
```
tahb: Cool! Does this return a string? `"hostname:5678"`

despo:
@tahb no, since it has already been verified on testing. I already switched to `REDIS_URL` and was going to verify on staging after merging.

Stretch96:
@tahb The module has been copied as is from https://github.com/turnerlabs/terraform-aws-elasticache-redis

We have found it better to use external modules, but as this project doesn't yet have 'Terrafile' module installation set up, I just copied it into the modules directory. In an ideal situation, this module would be installed to the vendor directory. I don't feel we'll gain anything by changing this module to follow our naming conventions, and it should just be kept up to date with the upstream module.

terraform/modules/elasticache-redis/outputs.tf
Stretch96: Yep 👍

terraform/modules/core/output.tf
Stretch96: From what I've seen with other 3rd party modules, outputs should be added when modules are created, as we may need to use them in future.

tahb:
The new Redis ENV should go into the `workspace-variables/workspace.tfvars.example` for reference.

terraform.tf
```diff
@@ -95,6 +95,7 @@ module "ecs" {
   aws_elasticsearch_region = "${var.region}"
   aws_elasticsearch_key    = "${module.es.es_user_access_key_id}"
   aws_elasticsearch_secret = "${module.es.es_user_access_key_secret}"
+  redis_url                = "${module.elasticache_redis.endpoint}"
```
tahb: The `endpoint` is just the hostname so for it to be a URL (which our Redis library needs and is correctly named as the ENV variable) we need to at least attach a protocol when we assign this data to the variable named URL.

I tested successfully with the app using:
``` 
redis_url                = "redis://${module.elasticache_redis.endpoint}"
```

tahb:
Once I changed the REDIS_URL (see comment) this works and receives connections.

terraform/modules/core/output.tf
tahb: The principle of YAGNI (you ain't going to need it) is a useful one to bear in mind in these types of situation as it helps keep the code base simpler and easier to comprehend. Ignored enough times things can quickly get out of control and fill up with code we never even used.

However as this is an external module that we've copied across for connivence (which I think was a good balance to strike) we can leave it on the basis that we don't 'own it' even though it's in our codebase, someone who misses this fact in future might change things anyway thinking they're helping and tidying things up! 

Worth always thinking about.

terraform.tf
Stretch96: This has been updated now 👍

tahb:
Rebased

PR #335, opened 2018-07-05
[279] Add DfE Sign In with Azure

tahb:
Built on: https://github.com/dxw/teacher-vacancy-service/pull/331

* This should continue to sign in all users through Azure unless in TESTS where the new ENV has been set on only for testing purposes. The DfE Sign In journey should not be accessible to our users.
* This isn't designed to be temporarily turned on in production for testing because we don't yet have the journey where users can select azure or sign in. Users would get stuck and there's no way to limit it to when only we are looking at it.


tahb:
[Deploy support for DfE Sign In](https://trello.com/c/1JuidkL0/279-deploy-support-for-dfe-sign-in)

PR #337, opened 2018-07-09
(chore) default radius to 1 if none set via params

despo:


despo:
[Handle no radius via url parameters gracefully - default to minimum allowed value](https://trello.com/c/tDprHxfp/285-handle-no-radius-via-url-parameters-gracefully-default-to-minimum-allowed-value)

tahb:
Related to: https://rollbar.com/dxw/teacher-vacancies/items/107/

PR #338, opened 2018-07-09
Production deployment #28

despo:
- bug fix (if no location radius is specified then default value to 1)
- fix issue with being unable to run individual specs


tahb:
There are no feature changes here that affect users in the main so I'm happy to merge.

PR #341, opened 2018-07-11
feature/switch to tva

despo:
pending terraform configuration for required environment variables + correcting DFE Sign in extracted identifier


app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -21,4 +22,8 @@ def create
   private def oid
     auth_hash['uid']
   end
+
+  private def identifier
+    auth_hash['uid']
```
despo: this should be updated to retrieve the email, unless we opt for using different mappings for each service DFE Sign in vs Azure.

tahb:
Here's a PR for the blocker; https://github.com/dxw/teacher-vacancy-service/pull/343

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
tahb: @despo  This PR returns the information from DfE Sign In and includes an example payload: https://github.com/dxw/teacher-vacancy-service/pull/344

PR #347, opened 2018-07-17
[280] Sign In decision point

tahb:
* Support the use of both SSO services without interrupting existing users who are signing in through Azure. This should allow DfE Sign in to be enabled without being used. The hypothesis is that people who might manage to end up will have got the answer to the question wrong, adding a link to the canonical source of schools to help users answer the question should help with this. 
* To establish the effectiveness of this journey we log the attempt which we should be to tie to events i.e the invitation to a new region or if many people are trying to use 'Any other area'.

screen shot 2018-07-17 at 14 03 37

* This also fixes the `My Jobs` navigation link that used to route to `sessions/new`, note the bottom left of this screenshot: 

screen shot 2018-07-17 at 13 17 40

* Removed Central Bedfordshire as Milton Keynes is the next district 
* Choosing other will halt the process and set new expectations

screen shot 2018-07-17 at 14 05 19


tahb:
[Hiring staff sign in using the authentication service they were registered with](https://trello.com/c/YkwfJyil/280-hiring-staff-sign-in-using-the-authentication-service-they-were-registered-with)

tahb:
[Hiring staff sign in using the authentication service they were registered with](https://trello.com/c/YkwfJyil/280-hiring-staff-sign-in-using-the-authentication-service-they-were-registered-with)

app/controllers/hiring_staff/identifications_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def choice
+    params.require('identifications').permit('name')['name']
+  end
+
+  private def halt_other_regions
+    return unless choice.eql?('Other')
+    flash[:notice] = "Other areas have not been invited yet, \
+      please register your interest by emailing us at #{I18n.t('help.email')}."
```
gazaston: Should make this a `mailto` link.

tahb:
I spoke with Matt about the language we use for the school's area whether that be: region, district, local authority. We have invited Cambridge as a district (as defined by GIAS) but NE as a regional office. Moving forward we will be working with `district` as the smallest unit of schools to invite at one time. We need to be careful how many options we could end up presenting to users as it could become unclear quite quickly. Matt advises that until we reach 2500 users we have chosen the highest mass districts, which should enable us to get much closer to December without too many problems.

As we are using a mixture of region and district it's hard to point users to the right place in GIAS so I've picked `area` in the hope that they will be able to read the list then pick the relevant answer.

spec/features/hiring_staff_can_only_see_their_school_spec.rb
```diff
@@ -12,7 +12,7 @@
     end
   end
 
-  context 'when the log in is NOT connected to a school' do
+  context 'when the session in is NOT connected to a known school' do
```
gazaston: typo ('when the session in is')?

app/controllers/hiring_staff/identifications_controller.rb
tahb: I've added this in using `.html_safe`

tahb:
screen shot 2018-07-17 at 17 43 50

app/controllers/hiring_staff/identifications_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    OpenStruct.new(
+      name: 'Milton Keynes',
+      to_radio: ['Milton Keynes', 'Milton Keynes']
+    )
+  ].freeze
+
+  AZURE_SIGN_IN_OPTIONS = [
+    OpenStruct.new(name: 'Cambridgeshire', to_radio: ['Cambridgeshire', 'Cambridgeshire']),
+    OpenStruct.new(name: 'The North East', to_radio: ['The North East', 'The North East'])
+  ].freeze
```
despo: Can we extract these to a helper?

app/controllers/hiring_staff/identifications_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def new
+    @identification_options =
+      AZURE_SIGN_IN_OPTIONS.map(&:to_radio) +
+      DFE_SIGN_IN_OPTIONS.map(&:to_radio) +
+      OTHER_SIGN_IN_OPTION.map(&:to_radio)
```
despo: these should also be moved to a helper.

app/controllers/hiring_staff/identifications_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    params.require('identifications').permit('name')['name']
+  end
+
+  private def halt_other_regions
+    return unless choice.eql?('Other')
+
+    flash[:notice] = safe_join(
+      [
+        'Other areas have not been invited yet, please register your interest by emailing us at ',
+        "<a href=mailto:#{I18n.t('help.email')}>#{I18n.t('help.email')}</a>".html_safe
```
despo: Can we move this to the locale please and a helper.

tahb:
@despo I've addressed this feedback and also added a new bit that stops showing DfE Sign In in production. As per our sprint goal that aims to start with usability testing with that first before going live.

app/controllers/hiring_staff/identifications_controller.rb
tahb: I'm not sure this should live in a helper since it's not technically rendered by any view, it's only added to `flash`.

PR #348, opened 2018-07-17
(chore) add page for currently unauthorised users

gazaston:
- adds a static page at `/pages/user-not-authorised`

after


gazaston:
[Hiring staff who aren't authorised shouldn't get stuck](https://trello.com/c/1ueIw68G/284-hiring-staff-who-arent-authorised-shouldnt-get-stuck)

PR #350, opened 2018-07-18
(feature) Add instructions to add environment variables

robbiepaul:
For now there are multiple places where we need to define environment variables. Having a list of these places is useful for reference.


robbiepaul:
[Add instructions for adding new environment variables](https://trello.com/c/qykTJJ9m/292-add-instructions-for-adding-new-environment-variables)

tahb:
Rebasing

PR #352, opened 2018-07-18
(feature) set up performance platform page

despo:
**Implements submission of data to the performance platform**

A single task `performance_platform:submit` configured to run every 24 hours submits data for both publication transactions and user satisfaction for the previous day.

An additional `performance_platform:submit_data_up_to_today` iterates through every day since April the 20th and submits to both endpoints as well. This task needs to be run manually.

It also implements a `TransactionAuditor` that logs successful and failed submissions to prevent duplicate entries. To achieve that it ensures uniquness of `task` and `date` and it always attempts to retrieve, or alternatively creates, and then updates the transaction_auditor entry.


spec/support/rake_tasks.rb
```diff
@@ -0,0 +1,24 @@
...
+    subject(:task) { tasks[task_name] }
+  end
+end
+
+RSpec.configure do |config|
+  config.define_derived_metadata(file_path: %r{/spec/tasks/}) do |metadata|
+    metadata[:type] = :task
+  end
+
+  config.include TaskExampleGroup, type: :task
```
robbiepaul: This is very cool 👍

terraform/modules/ecs/ecs.tf
```diff
@@ -417,6 +468,17 @@ resource "aws_ecs_task_definition" "reindex_vacancies_task" {
   task_role_arn            = "${aws_iam_role.ecs_execution_role.arn}"
 }
 
+resource "aws_ecs_task_definition" "performance_platform_submit_all_task" {
+  family                   = "${var.ecs_service_task_name}_performance_platform_submit_all_task"
+  container_definitions    = "${data.template_file.performance_platform_submit_all_container_definition.rendered}"
+  requires_compatibilities = ["EC2"]
+  network_mode             = "bridge"
+  cpu                      = "256"
```
robbiepaul: I know very little about Terraform but this seems like a high number, is it correct?

workspace-variables/workspace.tfvars.example
```diff
@@ -61,3 +61,4 @@ google_maps_api_key =
 google_analytics =
 rollbar_access_token =
 domain = # "tvs.dxw.net"
+pp_transactions_by_channel_token =
```
robbiepaul: Shall we add `pp_user_satisfaction_token` to the example tfvars as well?

terraform/modules/ecs/ecs.tf
despo: I just copy pasted one of the existing one-off task definitions, the are all identiacal and use cpu 256 and memory 512.
@Stretch96 can you advise on this?

workspace-variables/workspace.tfvars.example
despo: yes, that sounds like a good idea 👍

workspace-variables/workspace.tfvars.example
```diff
@@ -61,3 +61,5 @@ google_maps_api_key =
...
+pp_user_satisfaction_token =
```
despo: @robbiepaul added

robbiepaul:
Approved, not sure if you want to wait for Chris' response before merging?

terraform/modules/ecs/ecs.tf
tahb: 256 CPU and 512 Memory is what we provision a container running Puma based only on a sensible default and our initial provision testing. Our instances running on non-prod environments are running t2.small instances which get 2Gb of memory and 2Ghz of CPU.

For this we want to choose a number that gives enough resources for the rake task to succeed (some can be quite hungry) but not too much that it might starve the other Puma containers of resources and fail to start. The EC2 servers will now scale if this load is detected but it won't have enough time to respond to a rake task so it will likely fail with a provisioning error from ECS if you specify this as for example: 2Gb.

tahb: There's no reason this particular task couldn't be lower as 128 CPU however we _have_ had problems with the schools:import task running out of CPU. Once we found that 256 worked it seems easy to keep it as out default as there are no notable down sides.

lib/performance_platform.rb
```diff
@@ -0,0 +1,65 @@
+module PerformancePlatform
+  PP_DOMAIN = 'https://www.performance.service.gov.uk'.freeze
+
+  class Base
+    attr_reader :headers
+
+    def initialize(token = nil)
+      token || abort('No token set. Note that this task should only be executed in production.')
```
tahb: If not running in production is important enough to write this error perhaps it also justifies a guard clause that checks for `Rails.env.production?`

tahb:
Left some feedback

app/models/transaction_auditor.rb
```diff
@@ -0,0 +1,30 @@
...
+  class Logger
+    attr_reader :task, :date
+
+    def initialize(task, date)
+      @task = task
+      @date = date
+    end
+
+    def performed?
+      TransactionAuditor.exists?(task: task, date: date, success: true)
```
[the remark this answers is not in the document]
despo: it is indexed https://github.com/dxw/teacher-vacancy-service/pull/352/files#diff-2122010f5dbd941260cd367d7f17714aR11

lib/performance_platform.rb
despo: I don't want to guard this as I intend to use it in staging first, and then request for any submitted information to be reset so we can start using it in production.

As abort raises an exception any issues with the the token not being set should be identifiable via rollbar errors.

despo:
Don't merge please, waiting on current production deployment PR to be merged in first.

PR #357, opened 2018-07-19
(feature): Copy for 0 listings includes total listed count

tahb:
* This will provide the reader some context on the life of this system and the traction the service is getting.

screen shot 2018-07-19 at 14 06 02


tahb:
[Explain how many jobs _have_ been listed so far when there are 0 results](https://trello.com/c/IJcUYvjO/301-explain-how-many-jobs-have-been-listed-so-far-when-there-are-0-results)

PR #358, opened 2018-07-19
Add support for future pay scales

robbiepaul:
## User need
As a member of hiring staff
I need to only see current pay scales
So that I am up-to-date with the job market

## Description 
Adds `#current` scope to `PayScale` which is any records that have a start date earlier than today and an end date later than today.


robbiepaul:
[Add support for future payscales](https://trello.com/c/ytKBZRR6/290-add-support-for-future-payscales)

PR #360, opened 2018-07-19
(feature) Redirect unauthorised users to static page

robbiepaul:
We now have a static page for unauthorised users, we should redirect to that when the user doesn’t have the correct permissions.


robbiepaul:
[Hiring staff who aren't authorised shouldn't get stuck](https://trello.com/c/1ueIw68G/284-hiring-staff-who-arent-authorised-shouldnt-get-stuck)

PR #361, opened 2018-07-19
(fix) don’t process TVA responses with non 200 status code

despo:


app/services/teacher_vacancy_authorisation.rb
```diff
@@ -25,11 +25,11 @@ def school_urn
     private
 
     def parsed_response
-      JSON.parse(response.body)
+      response.code == 200  ? JSON.parse(response.body) : nil
```
tahb: Extra space but only a nitpick.

despo: yeah rubocop failed us here

tahb:
For the record this was the error we saw on staging: https://rollbar.com/dxw/teacher-vacancies/items/115/

PR #362, opened 2018-07-19
(fix) Provide Google ENV to ECS terraform module

tahb:
* This was missed during the review of  https://github.com/dxw/teacher-vacancy-service/pull/342 and has been added manually to a terraform deploy to support the feature that relies upon it until this is merged.


tahb:
[Add missing Terraform variables](https://trello.com/c/Obh2jiGf/303-add-missing-terraform-variables)

despo:
Part of the commit I removed. Does it break anything since the task will not be utilised?

PR #367, opened 2018-07-23
 Production deploy #31

despo:
- Regenerate url slug when a job listing is updated to resolve issue with URL not matching job title
- Add support for future payscales and rake task set start date for existing payscales
- Fix issue with authorisation service response code (check against string not integer)
- Adds Sign in decision point (in Staging 'Milton Keynes' will be presented to allow testing of DfE Sign In.


robbiepaul:
@despo do you know how we run the update pay scales task on staging?

At the moment, they don't have a start date so the options look like this:
screen shot 2018-07-23 at 14 21 26

despo:
@robbiepaul MK is the only one who should take you to DFE Sign in (which doesnt currently work on staging), however this is only on for staging so not a blocker for the production release.

PR #368, opened 2018-07-23
(fix): Hiring staff who are not in the right region don’t go to Azure

tahb:
* This fixes a bug introduced in a refactoring and makes sure it's tested https://github.com/dxw/teacher-vacancy-service/pull/347


tahb:
[Fix users who click the 'Other' sign in region don't go to Azure](https://trello.com/c/OVju66fU/311-fix-users-who-click-the-other-sign-in-region-dont-go-to-azure)

tahb:
Rebased and the old commit went away. I have no idea how.

PR #371, opened 2018-07-24
Automate browser testing

robbiepaul:
# Description

The is now ready for review. It allows us to test our happy path feature specs in five different browsers as part of our CI process. This will allow us to catch any browser incompatibilities as they happen.

# Things to consider

## Parallelisation 

I attempted to run the specs in parallel using the `parallel` gem however it threw a `ActiveRecord::RecordNotUnique` error. I guess because the tests attempted to use the same database and the same time and couldn't insert the same content.

By running the tests serially it takes about 15 minutes to run the tests in all the browsers.

## Internet Explorer 11 driver

The IE11 driver has many quirks, including:
- not being able to find checkboxes that are hidden (which is an issue for us because we're using GOV.UK styles that hide standard checkboxes)
- `page.html` has a strange issue where you can't see all the correct HTML on the page. I'm not sure what's going on here but the same test works in IE10 so I went with that instead.

## Edge driver

The Edge driver will fail to find an element if it's too far away from the element that is currently focused.

## Commit history

I decided to leave all the commits in the PR as it really helps describe the issues I ran in to trying to get this working, it might help people in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,7 @@ workspace-variables/*.tfvars
 # Certificates
 config/localhost/https/*.key
 config/localhost/https/*.crt
+
+# BrowserStack
+browserstack.err
+local.log
--- a/Gemfile
+++ b/Gemfile
@@ -79,6 +79,8 @@ group :development, :test do
   gem 'teaspoon-mocha'
   gem 'coffee-rails'
   gem 'timecop'
+  gem 'browserstack-local'
+  gem 'selenium-webdriver'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,6 +81,7 @@ GEM
     bindex (0.5.0)
     breasal (0.0.1)
     browser (2.5.3)
+    browserstack-local (1.3.0)
     builder (3.2.3)
     byebug (10.0.2)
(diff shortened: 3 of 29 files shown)


robbiepaul:
Aha, looks like it was the database cleaner strategy...

robbiepaul:
Added `browserstack: true` to most happy paths, all tests passing in BrowserStack.
```
Finished in 4 minutes 43.2 seconds (files took 6.23 seconds to load)
32 examples, 0 failures
```

robbiepaul:
[Automate cross browser testing](https://trello.com/c/xVhnZUWl/278-automate-cross-browser-testing)

robbiepaul:
Still not resolved. I messaged Chris to ask about possible connectivity issues between the build server and browserstack, just waiting for a response

tahb:
We haven't been able to make this work as intended on CI on either AWS CodeBuild or Codeship. The only value I can see that remains is having this as a tool that can be run locally from time to time. In reality I predict this will quickly be forgotten as the team changes and this code will linger on and deliver very minimal value.

If we get hurt by the absence of this again, we can revisit this closed pull request and salvage the parts that allow it to work locally as currently this includes a lot of Terraform that's not required.

PR #374, opened 2018-07-24
(chore) update copy for sign in region selection

despo:
and drop `Other` option


despo:
[Copy text changes for hiring staff authentication](https://trello.com/c/M1m4o1dR/313-copy-text-changes-for-hiring-staff-authentication)

despo:
updated-sign-in-page-copy

PR #377, opened 2018-07-24
(chore) update seeds to match auth service school

despo:
as we currently use the staging auth service for development, this matches the assigned permissions to all developers on that.


despo:
[Update seeds and set one of the school urns to `137138` to match school that devs are authorised with on staging access API.](https://trello.com/c/rqL7g44D/309-update-seeds-and-set-one-of-the-school-urns-to-137138-to-match-school-that-devs-are-authorised-with-on-staging-access-api)

PR #378, opened 2018-07-25
[310] Users that reset their DFE Sign-in password can recover from t…

tahb:
…he error

* DfE Sign-in will always redirect users to our redirect_uri that’s used for normal authentication requests: auth/dfe/callback.
* This is not currently configurable in their service and they advise us to handle multiple types of request on the same endpoint.
* This was a problem because OmniAuth was identifying this request as unauthorised because our service was not the origin, no nonce was created and stored (I think). This then is intercepted in Rack in the Omniauth wrapper of our callback endpoint so we have to look inside our omniauth strategy to handle this behaviour and stop it rendering a white page with “Not authorised”.
* This change is the quickest thing I can think of doing that will allow users to recover from this error when they reset their password. Giving us more time to identify a better solution.
* Both types of requests are GET so I cannot distinguish on that but as QUERY_STRING is populated with params like `code` as part of an OIDC request, I can check for their absence

### DIFF
--- a/app/views/errors/unauthorised.html.haml
+++ b/app/views/errors/unauthorised.html.haml
@@ -1,0 +1,4 @@
+- content_for :page_title_prefix, t('error_pages.unauthorised')
+
+%h1.heading-xlarge=t('error_pages.unauthorised')
+%p You can email #{mail_to t('help.email')} if the problem persists.
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,4 +1,42 @@
-class OmniAuth::Strategies::Dfe < OmniAuth::Strategies::OpenIDConnect; end
+class OmniAuth::Strategies::Dfe < OmniAuth::Strategies::OpenIDConnect
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  # rubocop:disable Metrics/LineLength
+  # rubocop:disable Style/GuardClause
+  # Please refer to this commit to read why this has been copied from:https://github.com/m0n9oose/omniauth_openid_connect/blob/master/lib/omniauth/strategies/openid_connect.rb
+  def callback_phase
+    logger.info('Override callback phase…')
+    error = request.params['error_reason'] || request.params['error']
+    if error
+      raise CallbackError.new(request.params['error'], request.params['error_description'] || request.params['error_reason'], request.params['error_uri'])
+    elsif request['QUERY_STRING'].blank?
+      return Rack::Response.new(['Oops there was a problem. Please visit this URL to continue: https://teaching-jobs.service.gov.uk/identifications/new'], 200)
+    elsif request.params['state'].to_s.empty? || request.params['state'] != stored_state
+      return Rack::Response.new(['401 Unauthorized'], 401).finish
+    elsif !request.params['code']
+      return fail!(:missing_code, OmniAuth::OpenIDConnect::MissingCodeError.new(request.params['error']))
+    else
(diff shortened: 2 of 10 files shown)


tahb:
[Investigate and improve how 401 responses from DFE Sign in are handled](https://trello.com/c/HHEV4etF/310-investigate-and-improve-how-401-responses-from-dfe-sign-in-are-handled)

tahb:
Don't merge this for the moment please, brainwave 💡

tahb:
Ready to go again.

app/views/errors/unauthorised.html.haml
```diff
@@ -0,0 +1,4 @@
+- content_for :page_title_prefix, t('error_pages.unauthorised')
+
+%h1.heading-xlarge=t('error_pages.unauthorised')
+%p You can email #{mail_to t('help.email')} if the problem persists.
```
despo: This can probably be extracted to the locale and reused between this template & the internal_server_error one

config/initializers/omniauth.rb
```diff
@@ -1,3 +1,48 @@
...
+        elsif request.params.blank?
+          redirects = { '/auth/dfe/callback' => '/dfe/sessions/new' }
+          req = Rack::Request.new(env)
+          return redirect(redirects[req.path]) if redirects.include?(req.path)
+        elsif request.params['state'].to_s.empty? || request.params['state'] != stored_state
+          return redirect('/401')
+        elsif !request.params['code']
+          return fail!(:missing_code, OmniAuth::OpenIDConnect::MissingCodeError.new(request.params['error']))
+        else
+          options.issuer = issuer if options.issuer.blank?
```
despo: can we use `||=` instead of an if statement

spec/requests/dfe_sign_in_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  before do
+    ensure_omniauth_has_not_been_mocked_in_another_test
+  end
+
+  context 'when DfE Sign-in respond with an OIDC payload for authentication purposes' do
+    context 'when that payload is unauthorised' do
+      it 'redirects the user to the not authorised page' do
+        params = {
+          'code' => 'a-long-secret',
+          'session_state' => '123.456'
```
despo: can we use the new hash format here?

spec/requests/dfe_sign_in_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        expect(response.status).to eq(302)
+        expect(response.body).to eq('Redirecting to /401...')
+      end
+    end
+  end
+
+  context 'when a user is linked back to our service with a redirect' do
+    it 'routes the request back to the new sign-in page' do
+      request = get '/auth/dfe/callback'
```
despo: can we use the `auth_dfe_callback_path` path helpers here - as it's easier to spot failures if anything changes.
I can't recall if request specs have access to path helpers. Same for line 16.

app/views/errors/unauthorised.html.haml
tahb: I was being consistent with the other error pages and did not want to take this opportunity to address them all incase those changes got locked into this more complex change.

config/initializers/omniauth.rb
tahb: This code has been directly copy and pasted from the omniauth-connect gem so I was reluctant to rewrite any of it for style reasons.

spec/requests/dfe_sign_in_spec.rb
tahb: It's hard coded in the routes as there is no resource to attach the routing to so I don't think this can be done.

tahb:
@despo I think this is ready for you to read again.

despo:
@tahb Approved 👍

PR #381, opened 2018-07-30
WIP (chore) render all Sign in options across env

despo:
- remove production check and relevant spec


tahb:
Last minute problem with this one so please don't merge, the use of 'Milton Keynes' now seems to be in doubt! We should have a new name shortly.

PR #387, opened 2018-08-02
(fix) ensure OmniAuth test mode is disabled …

despo:
after any features that enable it
- use rails route helpers in request spec
- remove redundant before(:each) block from request spec


tahb:
[Update to be deployed to testing only: Relabel 'Milton Keynes' option with 'Other' on the pre-sign in region selection page.](https://trello.com/c/1MyQUMp0/336-update-to-be-deployed-to-testing-only-relabel-milton-keynes-option-with-other-on-the-pre-sign-in-region-selection-page)

PR #388, opened 2018-08-02
[336] Enable generic sign-in option for testing

tahb:
* This should only need to life as long as we need to present sign-in options to users but it's still needed in usability testing sessions as the testers school will likely not be in the same regions we have rolled out to, at least to begin with.
* Refactor existing options using locale
* Switched the logic of the guard clause around sign-in redirection. Instead of Azure it will always try to go to DfE Sign-in unless one of the Azure areas. This makes support of 'Other' very easy and clear as well as aligning with the longer term strategy of moving away from Azure.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,11 @@
 inherit_from: .rubocop_todo.yml
+Rails/UnknownEnv:
+  Environments:
+    - production
+    - development
+    - test
+    - staging
+    
 AllCops:
     TargetRubyVersion: 2.4
     Exclude:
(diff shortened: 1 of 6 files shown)


tahb:
[Update to be deployed to testing only: Relabel 'Milton Keynes' option with 'Other' on the pre-sign in region selection page.](https://trello.com/c/1MyQUMp0/336-update-to-be-deployed-to-testing-only-relabel-milton-keynes-option-with-other-on-the-pre-sign-in-region-selection-page)

tahb:
No longer needed as Azure has been completely extracted from the product.

PR #399, opened 2018-08-06
Chore/340 validated changes to mobile layouts

gazaston:
This PR implements (some of) the changes made to try and improve the experience for job seekers on mobile:

- jobseekers' index page: adding the ' _n_ results' and 'refine search' anchor link UI at the top of the results page, and moving the filters to the very bottom
- job show page: apply listing layout changes as previously deployed to testing, except place the map below the about school info (rather than above)

Browser detection is performed using the [Browser gem.](https://github.com/fnando/browser) Different views are served to mobile devices using [Rails' ActionPack Variants.](https://richonrails.com/articles/action-pack-variants-in-rails-4-1)

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -1,6 +1,4 @@
 class VacanciesController < ApplicationController
-  # include VariantConcerns
-
   helper_method :location,
                 :keyword,
                 :minimum_salary,
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -1,4 +1,6 @@
 class VacanciesController < ApplicationController
+  # include VariantConcerns
+
   helper_method :location,
                 :keyword,
                 :minimum_salary,
--- a/app/views/vacancies/_show.html+phone.haml
+++ b/app/views/vacancies/_show.html+phone.haml
@@ -18,22 +18,7 @@
       = @vacancy.school.name
     %p= @vacancy.school.description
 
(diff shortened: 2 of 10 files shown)


gazaston:
[Validated changes to mobile layouts](https://trello.com/c/MTrSWcdL/340-validated-changes-to-mobile-layouts)

app/controllers/vacancies_controller.rb
```diff
@@ -1,4 +1,6 @@
 class VacanciesController < ApplicationController
+  # include VariantConcerns
```
robbiepaul: Did you mean to leave this in?

app/views/vacancies/_show.html+phone.haml
```diff
@@ -0,0 +1,123 @@
...
+      = @vacancy.job_title
+      %span.heading-secondary
+        = @vacancy.location
+
+    %h2.heading-medium
+      = t('schools.about')
+      = @vacancy.school.name
+    %p= @vacancy.school.description
+
+    %dl.school--details
```
robbiepaul: Do you think we should move the school details to a partial? It looks like it uses the same content as the normal `_show.html.haml`

app/controllers/vacancies_controller.rb
gazaston: Nope :/

app/views/vacancies/_show.html+phone.haml
gazaston: 👍

PR #401, opened 2018-08-06
All sign in options now use DfE Sign In

robbiepaul:
As part of the switch to using DfE Sign In for all our users we need to make sure existing users see the correct sign in screen.

~~In a future PR we will remove Azure sign in completely, but for now just pointing existing options to DfE sign in should suffice.~~

I've also removed all Azure AD code except in the Auditor class as I guess we want to keep that information going forward. 

EDIT: Please don't merge this until we're ready to send out the new invites for existing users.


spec/features/hiring_staff_can_sign_in_with_azure_spec.rb
```diff
@@ -1,216 +0,0 @@
...
-
-    scenario 'adds entries in the audit log' do
-      authentication = PublicActivity::Activity.first
-      expect(authentication.key).to eq('azure.authentication.success')
-
-      authorisation = PublicActivity::Activity.last
-      expect(authorisation.key).to eq('azure.authorisation.success')
-      expect(authorisation.trackable.urn).to eq(school.urn)
-    end
-  end
```
despo: If these are being removed then the relevant controller code should also be removed.

app/helpers/identifications_helper.rb
```diff
@@ -1,18 +1,13 @@
...
-  AZURE_SIGN_IN_OPTIONS = [
+    OpenStruct.new(name: 'milton_keynes', to_radio: ['milton_keynes', 'Milton Keynes']),
     OpenStruct.new(name: 'cambridgeshire', to_radio: ['cambridgeshire', 'Cambridgeshire']),
     OpenStruct.new(name: 'the_north_east', to_radio: ['the_north_east', 'The North East'])
   ].freeze
 
   def identification_options
     @identification_options ||= begin
-      opts = AZURE_SIGN_IN_OPTIONS.map(&:to_radio)
-      opts += DFE_SIGN_IN_OPTIONS.map(&:to_radio) unless Rails.env.production?
```
tahb: Without this guard we will be putting the Milton Keynes into production here and we have since changed plan. This was handy to help us test and prototype DSI but now that we have 2 real options for use with DSI it's no longer needed.

robbiepaul: Added a commit that removes Milton Keynes ✨

PR #402, opened 2018-08-06
[353] Sign-in copy doesn’t include Milton Keynes 

tahb:
* Remove "Milton Keynes" from the copy only as we change plan to first invite all Azure users across and scale up in those regions. Remember that MK is shown as an option in Staging only for the purposes of demonstration and testing DSI.
* This list has already fallen out of sync on it’s first iteration and as we look to add more districts and regions, we want to make this copy always align with the options available

Before
screen shot 2018-08-06 at 16 40 56

After
screen shot 2018-08-09 at 15 34 19


tahb:
[Remove Milton Keynes from sign-in step copy](https://trello.com/c/rGq1ytv7/353-remove-milton-keynes-from-sign-in-step-copy)

tahb:
[Remove Milton Keynes from sign-in step copy](https://trello.com/c/rGq1ytv7/353-remove-milton-keynes-from-sign-in-step-copy)

config/locales/en.yml
```diff
@@ -17,7 +17,7 @@ en:
     school_page_link: 'My jobs'
   sign_in:
     title: 'Sign in to Teaching Jobs'
-    intro: 'While the service is in development it lists teaching jobs in select areas, currently Cambridgeshire, the North East and Milton Keynes. In order for you to sign into the service we need to know which area your school is in. '
+    intro: 'While the service is in development it only lists teaching jobs in the following areas. In order for you to sign into the service we need to know which area your school is in. '
```
despo: The copy on the PR doesn't read very well if MK is only removed but no further amendments take place.

`While the service is in development it only lists jobs in the following areas. In order for you to sign into the service we need to know which area  your school is in.`

in the following areas... reads as if it's missing some continuation

Should this be split into two and and the first part be inset text (similar to here https://govuk-elements.herokuapp.com/typography/#typography-inset-text ) so it's highlited?

e.g.

> While the service is in development it only lists jobs in the areas listed below.

In order for you to sign into the service we need to know which area  your school is in.

PR #403, opened 2018-08-06
(feature) Redirect signed in users from the identifications page

robbiepaul:
It's possible for already signed in users to come back to our service via a bookmark or direct link. If they are signed in we should redirect them to the schools page instead of showing them the sign in page again needlessly.


robbiepaul:
[auto redirect from the identification/new page to the school page if there's an active session](https://trello.com/c/k2UHhMtX/337-auto-redirect-from-the-identification-new-page-to-the-school-page-if-theres-an-active-session)

PR #411, opened 2018-08-07
(feature) Allow hiring staff to select from multiple schools

robbiepaul:
# User need

As a DfE sign user who is invited to use Teaching Jobs I need to be able to manage all the schools (organisations) I represent, so that I can publish job listings for my school/s 

# Description

This PR allows for a few different sign in cases:

- A user with multiple DfE Sign In organisations and permission in TVA selects the school they want to manage and we sign them in to that school
- A user with one DfE Sign In organisation and permission in TVA signs in to the school they are associated with from DfE Sign In
- A user with no DfE Sign In organisations and permission in TA signs in the the school they are associated with in TVA

### DIFF
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -24,7 +24,7 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
   private
 
   def update_session(school_urn, permissions)
-    session.update(session_id: oid, urn: school_urn, tva_permissions: permissions.all_permissions)
+    session.update(session_id: oid, urn: school_urn, multiple_schools: permissions.many?)
     Auditor::Audit.new(current_school, 'dfe-sign-in.authorisation.success', current_session_id).log
   end
 
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -9,12 +9,11 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
 
   def create
     permissions = TeacherVacancyAuthorisation::Permissions.new
-    permissions.authorise(identifier)
+    permissions.authorise(identifier, selected_school_urn)
     Auditor::Audit.new(nil, 'dfe-sign-in.authentication.success', current_session_id).log_without_association
 
-    if permissions.all_permissions.any?
-      school_urn = selected_school_urn || permissions.school_urn
-      update_session(school_urn, permissions.all_permissions)
+    if permissions.authorised?
+      update_session(permissions.school_urn, permissions)
       redirect_to school_path
     else
       Auditor::Audit.new(nil, 'dfe-sign-in.authorisation.failure', current_session_id).log_without_association
@@ -24,8 +23,8 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
 
   private
 
-  def update_session(school_urn, all_permissions)
-    session.update(session_id: oid, urn: school_urn, tva_permissions: all_permissions)
+  def update_session(school_urn, permissions)
+    session.update(session_id: oid, urn: school_urn, tva_permissions: permissions.all_permissions)
     Auditor::Audit.new(current_school, 'dfe-sign-in.authorisation.success', current_session_id).log
   end
 
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -14,7 +14,7 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
 
     if permissions.all_permissions.any?
       school_urn = selected_school_urn || permissions.school_urn
-      update_session(school_urn)
+      update_session(school_urn, permissions.all_permissions)
       redirect_to school_path
     else
       Auditor::Audit.new(nil, 'dfe-sign-in.authorisation.failure', current_session_id).log_without_association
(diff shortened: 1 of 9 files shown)


robbiepaul:
[Hiring staff managing multiple schools](https://trello.com/c/rFA7OFY5/348-hiring-staff-managing-multiple-schools)

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -12,8 +12,9 @@ def create
     permissions.authorise(identifier)
     Auditor::Audit.new(nil, 'dfe-sign-in.authentication.success', current_session_id).log_without_association
 
-    if permissions.school_urn.present?
-      update_session(permissions.school_urn)
+    if permissions.all_permissions.any?
```
tahb: What would happen if the `selected_school_urn` was present but not included within `permissions.all_permissions`?

tahb:
On the card we have 

> a user can change organisations that they represent when signed in to the service 

Could we add the `Change organisation` link back in as done in https://trello.com/c/qwOoI6Et and have it route to an action that removes the school urn from the session and somehow punt it back to DfE Sign-in to make them choose their options again?

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
robbiepaul: As long as they had some permission in TVA it would sign them in using the selected school.

I think this makes sense as we're authorising users not schools? 

If we were to check the school urn was present in TVA we could get to a situation where a user has 10 schools in DfE Sign In and only 5 in TVA... how do we keep those in sync?

tahb: The current design of TVA provides us granular control in granting and revoking access between users and schools, we _are_ authorising users but the only thing we care about is their access to an individual school. 

Examples where losing this would cause confusion:
- Individual hiring staff should only have access to a subset of schools they are responsible for
- Hiring staff moves between organisations but continue to have full access to listings they should no longer be managing
- Hiring staff is added to another organisation in error (by another service perhaps) and has full access to listings they should not be managing
- Hiring staff are attached to multiple organisations, valid schools and other inappropriate organisations eg. Universities.

In general I think this approach would introduce a grey area of uncertainty and we'd be less confident in knowing what users had access to what school at any point in time.

You're right these is a synchronisation problem to solve. I was thinking that until DfE Sign-in can replace TVA we could get by with the quick and dirty solution of noting down the log output when an association could not be made during our invitations and manually creating it through the support console. If that cannot be achieved, we talk to DSI and GIAS to find out why that school urn is not yet in DSI. It may be that we have to wait a day or so before that association can be made.

robbiepaul:
@tahb ready for another look

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -23,8 +23,8 @@ def create
 
   private
 
-  def update_session(school_urn)
-    session.update(session_id: oid, urn: school_urn)
+  def update_session(school_urn, permissions)
+    session.update(session_id: oid, urn: school_urn, tva_permissions: permissions.all_permissions)
```
tahb: Given we are no longer rendering our own page for them to select options I wonder if a basic `yes` or `no` as the session value would now be enough to make this work and make it clearer to the reader.

app/services/teacher_vacancy_authorisation.rb
```diff
@@ -23,13 +24,20 @@ def all_permissions
...
+      return user_permissions_for_school.first['school_urn'] if @school_urn.present?
+      user_permissions.first['school_urn']
     end
 
     def many?
       user_permissions && user_permissions.count > 1
     end
 
+    def authorised?
+      return user_permissions.any? if @school_urn.blank?
```
tahb: I'm confident in our ability to associate our users during the invitation having looked at the first 60 so I think that if DfE Sign-in doesn't return an organisation that we can match on with a permission, we fail and return unauthorised. 

My previous comment described why I think it's risky to authorise a user in this scenario where a user doesn't have an organisation in DSI but has any other granted authorisation with us, we allow them in as the first organisation we find. I think this can safely go. Do you have any worries about losing it?

We may later look at displaying a friendly page in future that says "Hey we can see you're in this organisation but you've not been granted access by this organisation to use this service yet. Do x,y,z to make it happen".

app/views/hiring_staff/schools/show.html.haml
```diff
@@ -5,7 +5,7 @@
     %h1.heading-large
       = t('schools.jobs.index', school: @school.name)
     - if @multiple_schools
-      = link_to t('sign_in.organisation.change'), new_azure_path
+      = link_to t('sign_in.organisation.change'), auth_dfe_callback_path
```
tahb: Will the Azure users who have multiple schools lose the ability to choose organisations with our 'choose organisation' page? 

I'm not overly concerned if they do because none of them have logged in in at least 8 weeks and we're about to migrate them away. They can still use the 'Choose organisation' feature when signing in but will lose this ability to change organisation once in. They can always sign out and sign back in if we get any support requests in the few days inbetween now and the migration where they can't use it at all.

app/services/teacher_vacancy_authorisation.rb
robbiepaul: Roger, I guess I was just trying to break as few things as possible, but in hindsight it doesn't matter as we're reimporting everyone anyway. I'll make it so the organisation returned from DSI has to match one for the user's permission in TVA.

app/views/hiring_staff/schools/show.html.haml
robbiepaul: Yeah you're right there could be some time where Azure users can't switch organisation. When do you think we're gonna migrate over? Maybe we could delay merging this change until then?

tahb: In the next 2 days I hope!

spec/features/hiring_staff_can_sign_in_with_dfe_spec.rb
```diff
@@ -1,4 +1,38 @@
 require 'rails_helper'
+
+RSpec.shared_examples 'a successful sign in' do
```
tahb: Nice 👍

robbiepaul:
@tahb I've added a commit that requires the school urn returned from DSI to match one in our permissions app, if no urn is returned from DSI the user won't be authorised.

Also renamed that session key that shows the 'Change organisation' link and made it a boolean.

tahb:
Thanks for your work on this.

PR #420, opened 2018-08-08
Feature/zero downtime ecs instance redeploy

Stretch96:


tahb:
We've since made logspout daemon, and learnt that we don't want web to be a daemon as we'd like to have many of them as scale dictates.

PR #421, opened 2018-08-09
Production deploy #35

tahb:
* 10 gem updates after the install of Dependabot
* Changes to the jobs listing view on mobile
* Reporting to the performance platform
* Hiring staff who land on the sign-in step when are already logged in will be automatically redirected to their schools page and will no longer see this page
* Users of DfE Sign-in will see the DSI 'Choose organisation' page if they have more than one organisation
* Users who do not have a DSI organisation configured will not be authorisable we now use a combination of email and organisation_id from DSI to check our authorisation service for approval
* CloudFront caches all files under the /assets directory for efficiency improvements


despo:
...to avoid any confusion, until I get the rake task update in.

tahb:
Sure. Thanks for the review.

PR #422, opened 2018-08-09
Revert "(feature) Require school urn from DfE Sign In to log in succe…

tahb:
…ssfully"

This reverts commit 6d93fab152e2f87210f9d7f3eababa859a1201b9.

* As we have at the last minute found we cannot depend on the organisations being associated as reliably as we first thought we need to have a fall back option after all until a fix can be made.
* This is not ideal because if no org is returned by DSI it will choose the first school_urn from the permissions list we hold. This will work okay for single user schools but not for MATs. We are going to work around this by not inviting any new MATs until a fix is in.
* We were still showing the 'Choose organisation' link which wouldn't have worked so I replaced it with a temporary message explaining that while we can see they have multiple, we're adding support for them to switch soon. To do this I reused the `.share-url` class as it has the styles needed for this message. I chose not to refactor this style for reuse as I plan for this pull request to be reverted in the whole soon.

screen shot 2018-08-09 at 12 02 00

### DIFF
--- a/app/services/teacher_vacancy_authorisation.rb
+++ b/app/services/teacher_vacancy_authorisation.rb
@@ -34,6 +34,7 @@ module TeacherVacancyAuthorisation
     end
 
     def authorised?
+      return user_permissions.any? if @school_urn.blank?
       user_permissions_for_school.any?
     end
 
--- a/app/views/hiring_staff/schools/show.html.haml
+++ b/app/views/hiring_staff/schools/show.html.haml
@@ -2,11 +2,11 @@
 
 .school.grid-row
   .column-full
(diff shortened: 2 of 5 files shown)


tahb:
[When no DSI organisations allow single school users to sign in](https://trello.com/c/bQEPzsly/359-when-no-dsi-organisations-allow-single-school-users-to-sign-in)

fionamurray:
Should the copy text say "organisations" rather than schools?

tahb:
I think 'schools' is a sensible place to start as we reference it quite a lot in the copy on the service whilst we haven't yet used 'organisation'.

PR #447, opened 2018-08-16
Chore/368 migrate to gov uk design system

gazaston:
To futureproof the application, we have switched over from the [deprecated GOV UK Elements](https://govuk-elements.herokuapp.com/) to the new [GOV UK Frontend and Design System.](https://design-system.service.gov.uk/)


tahb:
I have added the GOV.UK Frontend in via Docker. It looks like it's ready to continue @gazaston.

despo:
@gazaston I have updated the gem to add the error classes in date fields
screen shot 2018-09-14 at 14 36 22 1

PR #449, opened 2018-08-17
(chore) enable heroku review apps

despo:
`app.json` is required to configure heroku review apps


despo:
[Heroku Review apps](https://trello.com/c/PXOz7xz0/375-heroku-review-apps)

PR #453, opened 2018-08-21
(feature) expose json activity feed

despo:


despo:
[Set-up a way to monitor and visualise the health (sign-in/publish/search) of the Teaching Jobs](https://trello.com/c/lsCGP3Fh/372-set-up-a-way-to-monitor-and-visualise-the-health-sign-in-publish-search-of-the-teaching-jobs)

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @trashed = Vacancy.trashed.count
+    @scraped = @published - (@audited_publications['vacancy.publish'] || 0)
+
+    @audit = Activity.order(created_at: :desc).page(page) if audit?
+  end
+
+  private
+
+  def page
+    params[:page] || 1
```
tahb: Unclear why we need to paginate this endpoint when it just returns a summary of information.

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,38 @@
+class StatsController < ApplicationController
+  include ActionController::HttpAuthentication::Token::ControllerMethods
+
+  before_action :authenticate, :set_access_control_headers
```
tahb: As mentioned on the Trello card: when we discussed the actions for this card we agreed that it didn't need to begin with authentication as it could solve the need of basic statistic counts without the extra wrapper of complexity and barrier to people accessing these numbers (especially within the department). Did something change since then?

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def index
+    @audited_publications = PublicActivity::Activity.where(key: 'vacancy.publish')
```
tahb: Could we display the 2 stats that are driving this change along with all counts with a small change to the following: `PublicActivity::Activity.group(:key).count`? This means as soon as we start tracking a future activity, they are automatically included without having to make an additional change to this page.

app/views/stats/index.json.jbuilder
```diff
@@ -0,0 +1,22 @@
...
+
+  json.sign_ins do
+    json.authenticated @signin_audit['dfe-sign-in.authentication.success']
+    json.authorised @signin_audit['dfe-sign-in.authorisation.success']
+    json.unauthorised @signin_audit['dfe-sign-in.authorisation.failure']
+  end
+
+  if @audit.present?
+    json.audit @audit do |audit|
+      json.partial! 'show.json.jbuilder', audit: audit
```
tahb: We said we would display only a summary of information (in JSON) as that's all that is needed by the people who need to consume it, unsure why we are now surfacing all information we hold instead. This does now begin to explain why authentication has been added in.

tahb:
Tests are failing for everything currently built on top of develop. Fix is in https://github.com/dxw/teacher-vacancy-service/pull/456

PR #454, opened 2018-08-21
[380] Rely on DSI organisations for authentication

tahb:
This reverts our previous revert. This was done because of a bug in DfE Sign-in that wasn't going to be fixed in time for us to invite more users. The bug was that we couldn't attach organisations to users through the API, they were wiped as soon as the user activated their account.

To get around this issue we decided to add this code and to only support single school users for the time being, as they wouldn't have the need to switch organisations - the feature DSI was currently missing. Now we've tested DSI organisation attachments in production and it's now working as expected, so we can now reliably depend on users having 1 or more DSI organisations.

They can have their organisation removed from within DSI, the user would see a not authorised screen within our service and would need to be added back to the organisation before they can sign-in again.

### DIFF
--- a/app/services/teacher_vacancy_authorisation.rb
+++ b/app/services/teacher_vacancy_authorisation.rb
@@ -34,7 +34,6 @@ module TeacherVacancyAuthorisation
     end
 
     def authorised?
-      return user_permissions.any? if @school_urn.blank?
       user_permissions_for_school.any?
     end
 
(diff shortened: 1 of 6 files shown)


tahb:
Tests are failing for everything currently built on top of develop. Fix is in https://github.com/dxw/teacher-vacancy-service/pull/456

PR #457, opened 2018-08-22
Feature/376 a more obvious link to the search landing page

gazaston:
Testing revealed that some jobseekers were having difficulty navigating back from a single job view to the list view.

We’ve replaced the breadcrumb (which was only ever one level deep) with a GOV UK ‘back’ link pattern.

**Before**
before

**After**
after


gazaston:
[A more obviously visible clickable link to the search landing page would improve the job-seeker experience](https://trello.com/c/06khqkiq/376-a-more-obviously-visible-clickable-link-to-the-search-landing-page-would-improve-the-job-seeker-experience)

PR #458, opened 2018-08-22
[354] Allow search engines to index public listing pages

tahb:
* Switch from `none` to the equivalant `noindex, nofollow` as it’s more explicit and clear. None might mean, no robots tag. Details here: https://developers.google.com/search/reference/robots_meta_tag
* Continue to stop indexing all pages - including new ones - by default without us having to remember this in future
* Allow search engines to index pages but not to follow links on our home page and any public listing
* Remove robots.txt to allow the headers to be reliably respected per the advice of GDS: https://www.gov.uk/service-manual/technology/get-a-domain-name#telling-search-engines-not-to-index-pages

Now:

```
HTTP/1.1 200 OK
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Referrer-Policy: strict-origin-when-cross-origin
X-Robots-Tag: noarchive, nofollow
Content-Type: text/html; charset=utf-8
ETag: W/"6a666fe6c90cefc9d81bddd8314e54d6"
Cache-Control: max-age=0, private, must-revalidate
Set-Cookie: _teachingjobs_session=f15f342fe672335b0229e23a368f52ab; path=/; expires=Wed, 22 Aug 2018 13:35:45 -0000; HttpOnly
X-Request-Id: 7870d19d-a305-4f12-9879-0b2ad64bbd42
X-Runtime: 1.303874
Transfer-Encoding: chunked
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -66,6 +66,6 @@ class ApplicationController < ActionController::Base
   end
 
   def set_headers
-    response.set_header('X-Robots-Tag', 'none')
+    response.set_header('X-Robots-Tag', 'noindex, nofollow')
   end
 end
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -85,4 +85,8 @@ class VacanciesController < ApplicationController
   def searched?
     params[:commit]&.eql?(I18n.t('buttons.apply_filters'))
   end
+
+  def set_headers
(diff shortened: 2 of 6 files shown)


tahb:
[Investigate how we allow Google Jobs to show our results in their search results](https://trello.com/c/yuwYtsKm/354-investigate-how-we-allow-google-jobs-to-show-our-results-in-their-search-results)

PR #459, opened 2018-08-22
Production deploy #38

tahb:
* Replace job listing breadcrumbs with a back link
* Remove temporary patch that prevents us inviting MAT hiring staff due to a DfE Sign-in bug. This bug has been fixed and testing with Suzanne and Hilary to ensure that MAT hiring staff can now use the DSI 'Choose organisation' screen as planned
* 2 minor updates to gems
* Unexpected changes to Schema from an update in a previous deployment have been committed


tahb:
I've gone through Trello with Suzanne and all code changes are now in accepted. There are 3 other cards but they each respond to another task or process from code.

PR #460, opened 2018-08-22
[379] Sign-in step no longer requires a choice of area

tahb:
* Hopefully this change will help users who've not yet been invited from trying to start a journey they cannot succeed in but make it clear to users who are applicable that they should proceed. The previous use of areas was used to differentiate users between DfE Sign-in and Azure, with Azure now gone the use of area names is no longer required based on current understanding

screen shot 2018-08-22 at 17 08 16


tahb:
[Update [remove] the area selection sign-in step page to reflect roll-out](https://trello.com/c/DaqmXn5T/379-update-the-area-selection-sign-in-step-page-to-reflect-roll-out)

tahb:
I've asked Michael today to have a look for us.

tahb:
I'm struggling to get hold of the copy we need to duplicate but the need to remove these redundant area choices is high going into the first week of term. Reducing the copy to as minimal as possible and linking to the start page should it not suffice.

screen shot 2018-08-31 at 17 05 44

PR #461, opened 2018-08-23
(chore) ‘apply filters’ becomes ‘(refine) search’

gazaston:
Some testing has revealed the button text "Apply filters" can be confusing.

We've changed it to 'Search' for the initial state, and 'Refine search' when a search has already been performed.

**Before**
before

**After (no search applied)**
after no search

**After (with search applied)**
after with search

### DIFF
--- a/spec/features/job_seekers_can_view_vacancies_spec.rb
+++ b/spec/features/job_seekers_can_view_vacancies_spec.rb
@@ -125,6 +125,21 @@ RSpec.feature 'Viewing vacancies' do
     expect(page).not_to have_content(I18n.t('jobs.no_jobs'))
   end
 
+  scenario 'The search button text changes from \'Search\' to \'Refine search\' when the filters are applied' do
+    Vacancy.__elasticsearch__.client.indices.flush
+    visit jobs_path
+
+    within '.filters-form' do
+      fill_in 'keyword', with: 'English'
+      expect(find('.button').value).to eq('Search')
+      click_on I18n.t('buttons.apply_filters')
+    end
+
+    within '.filters-form' do
+      expect(find('.button').value).to eq(I18n.t('buttons.apply_filters_if_criteria'))
+    end
+  end
+
   context 'when the vacancy is part_time' do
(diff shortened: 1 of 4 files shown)


spec/features/job_seekers_can_view_vacancies_spec.rb
```diff
@@ -125,6 +125,28 @@
     expect(page).not_to have_content(I18n.t('jobs.no_jobs'))
   end
 
+  scenario 'Should have button with text \'Search\' if default search criteria are applied' do
+    Vacancy.__elasticsearch__.client.indices.flush
+    visit jobs_path
+
+    # expect(page).to have_content(I18n.t('buttons.apply_filters'))
```
despo: can we remove this please?

spec/features/job_seekers_can_view_vacancies_spec.rb
```diff
@@ -125,6 +125,28 @@
...
+    expect(find('.button').value).to eq(I18n.t('buttons.apply_filters'))
+  end
```
despo: I wonder if we do need this scenario at all since the other scenario caters for both states. Maybe rewording the title eg `The search button text changes from 'Search' to 'Refine search' when the filters are applied`.

What do you think?

gazaston: Agreed 👍

spec/features/job_seekers_can_view_vacancies_spec.rb
```diff
@@ -125,6 +125,27 @@
...
+  scenario 'The search button text changes from \'Search\' to \'Refine search\' when the filters are applied' do
+    Vacancy.__elasticsearch__.client.indices.flush
+    visit jobs_path
+
+    expect(find('.button').value).to eq(I18n.t('buttons.apply_filters'))
+  end
+
+  scenario 'Should have button with text \'Refine search\' if search criteria are applied' do
+    visit jobs_path
+
```
despo: Can remove lines 133-137. Then the one scenario verifies the existence of both `Search` and `Refine Search` after applying the criteria.

PR #462, opened 2018-08-23
[149] Fix salary filtering bug

despo:
Updated checks to ensure that maximum salary is correctly utilised in filters. As the maximum salary can be nil then an ElasticSearch OR clause (look at example below) is used to check that it is less than the maximum specified **or** null.
 
```
    {
      bool: {
        should: [
          less_than(:maximum_salary, @maximum_salary.to_i),
          bool: {
            must_not: {
              exists: {
                field: 'maximum_salary'
              }
            }
          }
        ]
      }
    }
```


despo:
[Filtering bug: When I filter by both a minimum and a maximum salary, any job posts without a maximum salary but still within the range don't show up.](https://trello.com/c/ToLvkwfJ/149-filtering-bug-when-i-filter-by-both-a-minimum-and-a-maximum-salary-any-job-posts-without-a-maximum-salary-but-still-within-the-r)

despo:
@tahb I've amended the features and added checks for the expected number of listed jobs.

PR #463, opened 2018-08-24
[394] Log anonymised email address for users who weren’t authorised

tahb:
* This will be helpful during the early stages of inviting new users in gaining confidence in the process to know whether it is sensible to invite more users. We want to know that for all users that see the not-authorised page, it is because they have been invited to DSI by another service and are truly not allowed access yet, as opposed to a bug in our invitation process.
* We don't want or need the full personal email addresses that are coming back from DfE Sign-in to be sent to Papertrail as I think we can get away with checking every other character against the list we've invited before looking up in DfE Sign-in and our authorisation tool to debug.

screen shot 2018-08-24 at 12 03 31


tahb:
[Understand if hiring staff who aren't authorised are a symptom of a problem and a reason not to invite more users](https://trello.com/c/zgpevSXM/394-understand-if-hiring-staff-who-arent-authorised-are-a-symptom-of-a-problem-and-a-reason-not-to-invite-more-users)

PR #465, opened 2018-08-24
Fix - prevent editing publication date of published job

despo:
**`publish_on`: disable validation check for already published jobs and enable editing and validation for jobs with a `published` status and a future publication date**

## A job that has already been published
**`status: published` and `publish_on.past?`**

- Remove `change` button of publication date
edit-view-of-published-with-past-publish-on

- Display text instead of date field
edit-published-with-past-publish-on

## A job that is ready to be published
**`status: published` and `publish_on.future?`**

- The `change` button is still available
edit-view-of-published-with-future-publish-on

- The publication date can be edited
screen shot 2018-08-24 at 17 10 42

- Validation enforced - `publish_on` can not be set in the past
unable-to-set-past-publish-on-on-published

### DIFF
--- a/app/controllers/hiring_staff/vacancies/application_details_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_details_controller.rb
@@ -31,6 +31,7 @@ class HiringStaff::Vacancies::ApplicationDetailsController < HiringStaff::Vacanc
   def update
     vacancy = school.vacancies.published.find(vacancy_id)
     @application_details_form = ApplicationDetailsForm.new(application_details_form)
+    @application_details_form.status = vacancy.status
     @application_details_form.id = vacancy.id
 
     if @application_details_form.valid?
--- a/app/controllers/hiring_staff/vacancies/application_details_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_details_controller.rb
@@ -24,6 +24,8 @@ class HiringStaff::Vacancies::ApplicationDetailsController < HiringStaff::Vacanc
 
     @school = school
     @application_details_form = ApplicationDetailsForm.new(vacancy_attributes)
+    vacancy ||= school.vacancies.find(@application_details_form.id)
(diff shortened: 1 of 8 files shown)


despo:
[Attempting to edit a published job's application details, triggers a check to ensure that the publication date is not earlier than today. This means that if hiring schools attempt to specifically change any details rendered on that section will also prompt them to update the date the job was published.](https://trello.com/c/zfQ9EdoZ/383-attempting-to-edit-a-published-jobs-application-details-triggers-a-check-to-ensure-that-the-publication-date-is-not-earlier-than)

tahb:
There's no user need that I can find to help me understand the bug/problem we are fixing here. It looks like users are editing their published date after publication and that that's a problem but I'm not sure of the problem this raises?

If we don't make this change, what would happen?

Looking for a evidence before we potentially prohibit all hiring staff from extending their posts by say 2-4 weeks if their listings aren't been filled, or why we might not want to allow that? The current flexibility of the service seems valuable, and for us to be more restrictive again we should be very clear.

tahb:
Further to that I have some feedback for the design:

In the case where a listing has been published, I think we need to convey the reason why the 'change' button is whipped away from them.

Perhaps we can do this by changing the copy from:

> Date role will be listed

to:

> Date role was listed

And for accessibility, users will screen readers will have to work out that the change link is missing. Perhaps we should have a replacement statement that says 'Cannot be changed after publication'?

despo:
@tahb when a user wants to edit any of the application details, they are forced to change their publication date as that validation would always be trigerred when attempting to save the form. It does not prohibit extending the expiry date of a listing.

I can make the amendments to the text, maybe @gazaston can provide some feedback on the current content as well?

tahb:
Rebased

tahb:
On second thoughts using 'Date role was listed' would not make sense if the publish date was in the future and the hiring staff was making edits before that date. It could cause confusion so let's wait and see how this change plays out.

PR #466, opened 2018-08-27
(fix) A hiring school user can sign out…

despo:
… if they have not accepted the ToCs


despo:
[Sign out cannot be clicked if the user doesn't accept the terms and conditions page](https://trello.com/c/abtffgqB/389-sign-out-cannot-be-clicked-if-the-user-doesnt-accept-the-terms-and-conditions-page)

despo:
@tahb I've removed the commits

PR #476, opened 2018-08-30
Fix/aws ip ranges does not stop boot

tahb:
* Add timeout for AWS IP ranges to improve the chances our containers can boot successfully
* Handle the case where AWS give us a 403 and some XML instead of JSON

### DIFF
--- a/lib/modules/aws_ip_ranges.rb
+++ b/lib/modules/aws_ip_ranges.rb
@@ -6,8 +6,13 @@ module AWSIpRanges
     uri = URI(PATH)
 
     begin
-      response = Net::HTTP.get(uri)
-      return parse_json_for_ips(response)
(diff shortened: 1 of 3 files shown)


lib/modules/aws_ip_ranges.rb
```diff
@@ -6,8 +6,13 @@ def self.cloudfront_ips
     uri = URI(PATH)
 
     begin
-      response = Net::HTTP.get(uri)
-      return parse_json_for_ips(response)
+      http_connection = Net::HTTP.new(uri.host, uri.port)
+      http_connection.read_timeout = 5
```
despo: is 5 enough? Should we push up to 10 instead?

tahb: Yeah you're right, we don't have to be as short tempered with AWS on the read if a connection has been opened quickly.

PR #482, opened 2018-08-31
[chore] add pp variables in rake task definition

despo:


despo:
@tahb made some amendments but looks good now 👍

PR #486, opened 2018-09-03
[355] configure job queue

despo:
Configures active job and sidekiq and sets up a worker service 

The terraform configuration includes
- a worker ecs service
- ~load balancer to be used by worker~ (removed)
- a new container definition including the required variables
- exposed configuration for the worker port and command to be run


variables.tf
```diff
@@ -106,6 +106,11 @@ variable "ecs_service_task_port" {
   default     = 3000
 }
 
+variable "ecs_service_worker_task_port" {
+  description = "The port the worker is listening on (ALB will map these with ephemeral port numbers)"
+  default     = 3000
```
tahb: Is the default sidekiq port 4567, could this be conflicted with puma on 3000?

despo: The actual port is `8082`, this was my initial configuration but after going through the sidekiq code I tweaked it via the edge tfvars but missed adding the entry to the `workspace.tfcars.example`

`ecs_service_worker_task_port = 8082`

terraform/modules/core/core.tf
```diff
@@ -249,6 +249,72 @@ resource "aws_alb_target_group" "alb_target_group" {
...
+    target_group_arn = "${aws_alb_target_group.worker_target_group.arn}"
+    type             = "forward"
+  }
+
+  depends_on = ["aws_alb_target_group.worker_target_group"]
+}
+
+resource "aws_alb_listener" "worker_https" {
+  load_balancer_arn = "${aws_alb.worker.arn}"
+  port              = "443"
```
despo: Change to use redis port

terraform/modules/core/core.tf
```diff
@@ -249,6 +249,72 @@ resource "aws_alb_target_group" "alb_target_group" {
...
+  protocol          = "HTTPS"
```
despo: Set protocol to `TCP`

terraform/modules/core/core.tf
```diff
@@ -249,6 +249,72 @@ resource "aws_alb_target_group" "alb_target_group" {
   depends_on = ["aws_alb.alb_default"]
 }
 
+# worker load balancer
+
+resource "aws_alb" "worker" {
+  name            = "${var.project_name}-alb-worker-${var.environment}"
+  subnets         = ["${aws_subnet.public_subnet.*.id}"]
```
despo: Use a private subnet.

PR #487, opened 2018-09-03
(feature) users can clear search parameters

gazaston:
> As a jobseeker  
I need to be able to get back to the way the search page was initially  
So I can start from an unfiltered list of results

**This commit adds a 'Clear search' link if search parameters are currently applied.**

before (desktop)
before desktop

after (desktop)
after desktop

before (mobile)
before mobile

after (mobile)
after mobile

### DIFF
--- a/app/assets/stylesheets/base/_utilities.scss
+++ b/app/assets/stylesheets/base/_utilities.scss
@@ -36,3 +36,11 @@
 .display-none {
   display: none;
 }
+
+.inline {
+  display: inline;
+}
+
+.nobreak {
+  white-space: nowrap;
+}
--- a/app/views/vacancies/index.html+phone.haml
+++ b/app/views/vacancies/index.html+phone.haml
@@ -5,9 +5,12 @@
 .grid-row
 
   .column-two-thirds
-    %p.heading-medium.mt0.mb1
-      %p.heading-medium.mt0= @vacancies.total_count_message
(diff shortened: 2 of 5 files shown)


gazaston:
[Add a 'Clear search' feature](https://trello.com/c/hMpjScWq/391-add-a-clear-search-feature)

tahb:
Did you consider putting this function with the other search control we have in 'Search' or 'Refine search'? It took me a moment to find where it was.

screen shot 2018-09-04 at 16 45 57

A change to the font size to become consistent with the preceding copy might make it less jarring to my eye, however a large link would also be inconsistent with links else where on the site, the `date_posted` one directly beneath as an example. 

These tricky things make me think that it's positioned like a call to action to assist with a problem the proceeding copy introduces instead of a search function that can be used regardless. When no problem needs to be solved, or is there?

gazaston:
This feature is essentially a panic button — arising from a need identified when people were struggling to return to a default state. 

I believe it has more context when it's presented in proximity to the number of results returned, particularly in a mobile setting.

That said, on desktop it _could_ work under the 'Search job listings' in the sidebar. However, on mobile it would be lost as the sidebar is pushed to the bottom of the layout. My preference would be for the link to appear in the same(ish) place regardless of device.

tahb:
Rebased

PR #492, opened 2018-09-05
Revert "(chore) enable heroku review apps"

tahb:
* This reverts commit c44b2a428c191c01dd1c6b7de4f2d0204358be83.
* We want to revert this because it doesn't support the new work we've done with the govuk design system (due to packages required) and doesn't work with DfE Sign-in. 
* To use review apps (just for the job seekers part of the service) we'd need to duplicate deployment effort to deploy containers to Heroku. The buildspec.yml code to deploy to AWS ECS cannot be reused.


tahb:
[Heroku Review apps](https://trello.com/c/PXOz7xz0/375-heroku-review-apps)

PR #493, opened 2018-09-05
(feature) adds a mobile-specific feature test

gazaston:
[Using `Capybara.page.driver.header` we can manipulate the request header](https://github.com/teampoltergeist/poltergeist#manipulating-request-headers) and pass it a user agent string that will trigger the action pack variants for our mobile views.

This way, we can test for elements of UI that differ in mobile and desktop.

We’ve added a constant `USER_AGENTS` to `specs/rails_helper.rb` that extracts the user agent strings that come packaged up with the gem.


gazaston:
[Add a way to write specs that are specific to action pack variant views](https://trello.com/c/9iuKulQ8/418-add-a-way-to-write-specs-that-are-specific-to-action-pack-variant-views)

PR #494, opened 2018-09-05
(fix) searched? method is more robust

gazaston:
- caters for both possible button texts (‘Search’ or ‘Refine search’)


gazaston:
['Refine search' button prompts display of 'This is because schools are on summer holidays…' text](https://trello.com/c/6PyhzGUG/421-refine-search-button-prompts-display-of-this-is-because-schools-are-on-summer-holidays-text)

PR #501, opened 2018-09-10
(chore) ensure `Change` button is always rendered for contact email

despo:


tahb:
[Bug: When no contact email or url is set, the review page still generates a mail link (without an actual email) and a link for more information (without a url, which maps to the current page.](https://trello.com/c/mwfIX1jY/384-bug-when-no-contact-email-or-url-is-set-the-review-page-still-generates-a-mail-link-without-an-actual-email-and-a-link-for-more)

PR #503, opened 2018-09-11
[390] google indexing

despo:
- integrates Google Indexing API
- adds calls when publishing, updating and deleting a listed job

The default quota of the service is `200` requests per day. To change that we need to send a [quota increase request](https://docs.google.com/forms/d/e/1FAIpQLSc_mpLw3WnnCt3pVbUHYZZ6ZdOS-c0GIj-WZ_k54SG-jDqCXQ/viewform?entry.1781339011=despo.pentara@digital.education.gov.uk&entry.840738641=689616473831&entry.1191128052&entry.1776911313) and explain why we think 200 requests are not sufficient. This can be done from the [Google API Cloud console](https://console.cloud.google.com/apis/api).

The service key can be found in 1password.

**Note** As we don't manage slug updates, any changes to the job title will result in an update request to Google's indexing API,  which will ensure that the new url will be indexed but not that the old one has been removed.


despo:
[Use Google search indexing API to update or remove results based on our hiring staff actions](https://trello.com/c/juQVi2Ug/390-use-google-search-indexing-api-to-update-or-remove-results-based-on-our-hiring-staff-actions)

despo:
@esteedqueen CodeFactor fails to load the default rubocop configuration from  .rubocop.yml so behaves differently [eg rails_helper.rb should be ignored](https://github.com/dxw/teacher-vacancy-service/blob/develop/.rubocop.yml#L10) - we do however use it as it's free and it does make it easier to identify real issues as well.

As we run rubocop as [part of the main rake task](https://github.com/dxw/teacher-vacancy-service/blob/develop/lib/tasks/rubocop.rake#L11) any issues should cause the build to fail.

PR #504, opened 2018-09-12
(fix) deleting job confirmation appears twice

gazaston:
- `jquery_ujs` was included to support IE8; we’re no longer compelled to do so, therefore we can remove it


gazaston:
[Hiring staff: deleting a job — confirmation dialogue appears twice](https://trello.com/c/ovg4SatY/369-hiring-staff-deleting-a-job-confirmation-dialogue-appears-twice)

PR #507, opened 2018-09-17
[433] hiring staff can specify a listing as nqt eligible

esteedqueen:
### Hiring staff can specify a listing as nqt eligible (completed)
- added newly_qualified_teacher attribute to vacancies table
- added newly_qualified_teacher attribute to job specification forms (new & edit)
- added newly_qualified_teacher status to all job details pages
- modified existing test specs to test for newly_qualified_teacher spec

PR #508, opened 2018-09-17
[417] better filtering options for nqts

esteedqueen:
### Better filtering options for nqts (wip)
- added newly_qualified_teacher to filter options, controllers and views
- jobs can be filtered by newly_qualified_teacher value
- added newly_qualified_teacher to elasticsearch indexes and queries
- added test specs for job seekers filtering by NQTs

_Dependent on #507_
Should only be reviewed/merged after #507 is merged in.

### DIFF
--- a/app/views/vacancies/_filters.html.haml
+++ b/app/views/vacancies/_filters.html.haml
@@ -25,7 +25,6 @@
       = label_tag 'phase', t('jobs.filters.education_phase'), class: 'form-label'
       = select_tag 'phase', options_for_select(school_phase_options, phase), class: 'form-control form-control-block', include_blank: 'Any'
     .form-group
-      %span.form-label= t('jobs.filters.newly_qualified_teacher')
       .multiple-choice
         = check_box_tag :newly_qualified_teacher, 'true', nqt_suitable_checked?(newly_qualified_teacher)
         = label_tag 'newly_qualified_teacher', t('jobs.filters.nqt_suitable'), class: 'form-label'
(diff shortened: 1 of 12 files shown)


esteedqueen:
[Better filtering options for NQTs](https://trello.com/c/Sm9W9spw/417-better-filtering-options-for-nqts)

app/views/vacancies/_filters.html.haml
```diff
@@ -24,6 +24,9 @@
     .form-group
       = label_tag 'phase', t('jobs.filters.education_phase'), class: 'form-label'
       = select_tag 'phase', options_for_select(school_phase_options, phase), class: 'form-control form-control-block', include_blank: 'Any'
+    .form-group
+      = label_tag 'newly_qualified_teacher', t('jobs.filters.newly_qualified_teacher'), class: 'form-label'
+      = select_tag 'newly_qualified_teacher', options_for_select(nqts_options, newly_qualified_teacher), class: 'form-control form-control-block'
```
despo: This component should be a checkbox rather than a dropdown.
Sorry about this, I think it's not very clear on the trello card as both design ideas are attached but UR's final decision was to use a checkbox.

esteedqueen: Thanks @despo, will change this to a checkbox

app/views/vacancies/_filters.html.haml
```diff
@@ -24,6 +24,11 @@
     .form-group
       = label_tag 'phase', t('jobs.filters.education_phase'), class: 'form-label'
       = select_tag 'phase', options_for_select(school_phase_options, phase), class: 'form-control form-control-block', include_blank: 'Any'
+    .form-group
+      %span.form-label= t('jobs.filters.newly_qualified_teacher')
```
despo: I spoke to UR who confirmed that they don't want this label.
Sorry about the lack of clarity on this on the trello card.

esteedqueen: Ok done

PR #510, opened 2018-09-19
[436] Increase session timeout to 8 hours

esteedqueen:
- increased session timeout to 8 hours with integration test


esteedqueen:
[Increase session time out duration](https://trello.com/c/Rc53I39B/436-increase-session-time-out-duration)

PR #511, opened 2018-09-20
[feature] move Google index calls to background jobs

despo:
- run calls to Google indexing API via sidekiq
- ensure that service is only called when API credentials are specified.


despo:
[Move Google search indexing API calls to background jobs](https://trello.com/c/PmB6V8Dt/442-move-google-search-indexing-api-calls-to-background-jobs)

PR #515, opened 2018-09-21
[feature] Map JobPosting school region to GIAS region

despo:
Update mapped region field to ensure value is always set and avoid JobPosting validation warnings.

screen shot 2018-09-21 at 18 03 00


despo:
[Investigate improving mapped address region field in JobSchema to resolve Google search console issue.](https://trello.com/c/u5hpbUy4/439-investigate-improving-mapped-address-region-field-in-jobschema-to-resolve-google-search-console-issue)

PR #521, opened 2018-09-27
(feature) 444 pro rata salary

gazaston:
> As hiring staff, I can publish pro-rata salary for part time roles, so that I can accurately convey the pay for part time roles

> As a job-seeker, I can see the pro-rata salary for a part time roles, so that I have a clear understanding of the pay for that role

### list view
list view
### post view
post view
### post view mobile
post view mobile
### create job form
create job form
### edit job page
edit job page
### edit job form
edit job form
### job review page
job review page

### DIFF
--- a/app/views/vacancies/_show.html+phone.haml
+++ b/app/views/vacancies/_show.html+phone.haml
@@ -30,7 +30,6 @@
 
     %h2.mt1.govuk-heading-m= t('jobs.key_info')
 
-<<<<<<< HEAD
     %dl.app-check-your-answers.app-check-your-answers--short
       .app-check-your-answers__contents
         %dt.app-check-your-answers__question= t('jobs.salary')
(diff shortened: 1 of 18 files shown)


tahb:
Rebased

spec/features/hiring_staff_can_specify_a_part_time_role.rb
```diff
@@ -23,18 +23,65 @@
...
+  scenario 'a job page should not indicate a pro rata salary if the role is full time' do
+    school = create(:school)
+    vacancy = create(:vacancy, :published, school: school, working_pattern: :full_time)
+
+    visit job_path(vacancy.id)
+
+    within('.vacancy--metadata') { expect(page).not_to have_content('pro rata') }
+  end
+
+  context 'when the user is on mobile' do
```
tahb: This is complete coverage. It adds 5 tests to ensure pro rata is always shown and it should catch all regressions in the future which is valuable.

I would use these features to drive the development and then look to slim them down by moving some to unit tests, while keeping my confidence high that we'd have a strong change of knowing if it broke in future.

I'd think about updating the [happy path test](https://github.com/dxw/teacher-vacancy-service/blob/develop/spec/features/hiring_staff_can_publish_a_vacancy_spec.rb#L75) to also test for the presence of the correct `pro rota` value. Once here we can see how we're verifying all the other data we add is set as we expect https://github.com/dxw/teacher-vacancy-service/blob/develop/spec/support/vacancy_helpers.rb#L41 and add our new one on. You can see that there's a big reliance on using the `VacancyPresenter` for this purpose, which is another benefit of moving the logic out of the view.

For the rest I'd keep a feature test for what happens when the vacancy is `part_time` which keeps it inline with the spec file is named.

This leaves the mobile tests. Would it be an idea to have a deliberate test for this `_job_seekers_can_see_listings_on_mobile` which tests this and more to make sure that any changes made to show.html+phone.haml are caught.

app/views/vacancies/_vacancy.html.haml
```diff
@@ -5,7 +5,7 @@
     %dt= t('jobs.salary')
     %dd.double
       = vacancy.salary_range
-      = t('jobs.per_year')
+      = vacancy.working_pattern == 'Full time' ? t('jobs.per_year') : t('jobs.per_year_pro_rata')
```
tahb: For view logic like this I'd try and group it in the presenter for reuse, testability and smaller views.

app/views/vacancies/_show.html+phone.haml
```diff
@@ -32,12 +32,15 @@
...
-          %td.cya-answer= @vacancy.salary_range
         %tr
           %td.cya-question= t('jobs.working_pattern')
           %td.cya-answer= @vacancy.working_pattern
+        %tr
+          %td.cya-question= t('jobs.salary')
+          %td.cya-answer
+            = @vacancy.salary_range
+            - if @vacancy.part_time?
+              pro rata
```
tahb: All the other copy is managed by locale, not a blocker but perhaps this was missed?

app/views/vacancies/_show.html.haml
```diff
@@ -59,15 +59,18 @@
...
 
         %dt= t('jobs.salary')
-        %dd= @vacancy.salary_range
-
-        %dt= t('jobs.working_pattern')
-        %dd= @vacancy.working_pattern
+        %dd
+          = @vacancy.salary_range
+          - if @vacancy.part_time?
+            pro rata
```
tahb: All the other copy is managed by locale, not a blocker but perhaps this was missed?

PR #523, opened 2018-10-01
Fix/design system snags

gazaston:


gazaston:
[Migrate to GOV UK Design System](https://trello.com/c/cOVdtive/368-migrate-to-gov-uk-design-system)

PR #525, opened 2018-10-01
[434] Hiring staff sign in attempts log their DSI ID

tahb:
* This also corrects the fact that no logs using Rails.logger were going
to papertrail. This is because without being configured logs were going
to <environment>.log, which isn't sent to PT due to how logspout
collects and forwards logs on the hosted environments. This is desipite
the configuration we have for lograge which just concerns itself with
formatting and colouring.
* Group logging config together for clarity and remove prod config for log level which was being overwritten
screen shot 2018-10-01 at 17 43 30

### DIFF
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -9,6 +9,8 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
   end
 
   def create
+    Rails.logger.warn("Hiring staff signed in: #{oid}")
+
     permissions = TeacherVacancyAuthorisation::Permissions.new
     permissions.authorise(identifier, selected_school_urn)
     Auditor::Audit.new(nil, 'dfe-sign-in.authentication.success', current_session_id).log_without_association
@@ -18,7 +20,7 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
       redirect_to school_path
     else
       Auditor::Audit.new(nil, 'dfe-sign-in.authorisation.failure', current_session_id).log_without_association
-      log_debug_information
+      Rails.logger.warn("Hiring staff not authorised: #{oid} for school: #{selected_school_urn}")
       redirect_to page_path('user-not-authorised')
     end
   end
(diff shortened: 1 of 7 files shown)


app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -45,9 +47,4 @@ def identifier
   def selected_school_urn
     auth_hash.dig('extra', 'raw_info', 'organisation', 'urn')
   end
-
-  def log_debug_information
-    anonymised_email = identifier.gsub(/(.)./, '\1*')
```
tahb: Anonymising this email wouldn't have been useful in the cases we've had so far. We need to find the exact email address of a person looking to sign in to confirm they are the ones requesting support, favouring the DSI ID will keep it completely anonymous in our logs and the full ID value can only be used with access to DSI support console.

config/environments/production.rb
```diff
@@ -50,13 +50,6 @@
   config.force_ssl = true
   config.ssl_options = { redirect: { exclude: ->(request) { request.path =~ /check/ } } }
 
-  # Use the lowest log level to ensure availability of diagnostic information
-  # when problems arise.
-  config.log_level = :debug
```
despo: 👍

config/environments/production.rb
```diff
@@ -78,14 +71,13 @@
...
 
-  # Use default logging formatter so that PID and timestamp are not suppressed.
-  config.log_formatter = ::Logger::Formatter.new
-
-  # Use a different logger for distributed setups.
-  # require 'syslog/logger'
+  # Logging
+  config.log_level = :info
+  config.log_tags = [:request_id] # Prepend all log lines with the following tags.
+  config.logger = ActiveSupport::Logger.new(STDOUT)
```
despo: 👍

config/environments/staging.rb
```diff
@@ -78,6 +71,11 @@
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Logging
+  config.log_level = :debug
```
despo: Do we need `debug` on staging?

tahb:
[Log the DSI ID for users who sign into our service](https://trello.com/c/umEkh2S6/434-log-the-dsi-id-for-users-who-sign-into-our-service)

PR #528, opened 2018-10-02
[461] Docker tests can be run through Docker Compose

tahb:
* This was apart of docker-compose.yml used to start the services locally but it wasn’t copied across into the test docker-compose.yml, meaning the required deps in node_modules were indeed unavailable, as the 540 errors pointed to.


tahb:
[Fix: Development tests run with the new design system](https://trello.com/c/F4pyTui4/461-fix-development-tests-run-with-the-new-design-system)

PR #533, opened 2018-10-03
(458) Validates publish_on date on the vacancy review change if

esteedqueen:
- Added publish_on date validation on the vacancy review page if the date is in the past - regardless of vacancy status (draft or published) and whether the date value changed or not.

This addresses a bug experienced by a user during an unusual review page flow


esteedqueen:
[Production bug: New Error: NoMethodError: undefined method `publish_on_change?' for #<Vacancy:0x00000004822888> https://rollbar.com/dxw/teacher-vacancies/items/169/](https://trello.com/c/eq24urp9/458-production-bug-new-error-nomethoderror-undefined-method-publishonchange-for-vacancy0x00000004822888-https-rollbarcom-dxw-teacher)

app/controllers/hiring_staff/vacancies/publish_controller.rb
```diff
@@ -8,12 +8,20 @@ def create
       reset_session_vacancy!
       redirect_to school_job_summary_path(vacancy.id)
     else
+      skip_to_change_publish_on_date(vacancy.id) && return if vacancy.publish_on.past?
```
despo: When there are any errors when publishing we just want to display the generic error message and leave the user on the review page.

In the *How migh we session* we run last week, we did decided for some changes to be implemented so that more information is displayed to enable users to identify any old draft vacancy errors, but that work will be done as part of https://trello.com/c/x14tp4v4/487-display-detailed-error-messages-if-vacancy-publication-fails

app/controllers/hiring_staff/vacancies/application_details_controller.rb
```diff
@@ -31,7 +31,7 @@ def edit
 
   # rubocop:disable Metrics/AbcSize
   def update
-    vacancy = school.vacancies.published.find(vacancy_id)
+    vacancy = school.vacancies.find(vacancy_id)
```
despo: Is this change required? As only published vacancies need to be updated (I think `create` updates any drafts)

esteedqueen:
@despo I've updated the PR to only display cannot publish error and remain on the review page if publish_on is in the past.

app/models/concerns/vacancy_application_detail_validations.rb
```diff
@@ -12,17 +12,13 @@ module VacancyApplicationDetailValidations
   end
 
   def validity_of_publish_on
-    errors.add(:publish_on, publish_on_before_today_error) if publish_on_in_past? && publish_on_check?
+    errors.add(:publish_on, publish_on_before_today_error) if publish_on_in_past?
```
despo: This condition is used to prevent displaying errors messages for published vacancies with a past publication date, so now the message is incorrectly rendered on the edit form.

screen shot 2018-10-10 at 14 17 51

PR #545, opened 2018-10-08
(483) Add custom callback to refresh elasticsearch indices after update

esteedqueen:
- Added custom callback to refresh elasticsearch indices after create and update, and removed elasticsearch automatic callback

This is a fix for: automatic callback not handling changes in vacancy
association models so changes are not reflected in search after vacancy
has been updated

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -10,7 +10,6 @@ class Vacancy < ApplicationRecord
   include VacancyApplicationDetailValidations
 
   include Elasticsearch::Model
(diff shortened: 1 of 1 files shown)


app/models/vacancy.rb
```diff
@@ -73,6 +72,10 @@ class Vacancy < ApplicationRecord
 
   validates :slug, presence: true
 
+  after_commit on: %i[create update] do
+    __elasticsearch__.index_document
+  end
+
```
despo: 👍

PR #547, opened 2018-10-09
[feature] cache jobs views 

despo:


despo:
[Hiring staff know how many views their listings are receiving](https://trello.com/c/BNbjjJbi/432-hiring-staff-know-how-many-views-their-listings-are-receiving)

PR #551, opened 2018-10-10
[feature] always redirect to latest slugged vacancy path

despo:
## Trello card URL:
https://trello.com/c/Pnx3NERl/314-links-that-hiring-staff-share-should-not-break-if-the-job-title-is-updated

## Changes in this PR:
Utilise friendly_id history module to enable redirection to latest slugged url. Do not redirect when requests are JSON.

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -21,6 +21,7 @@ class VacanciesController < ApplicationController
 
   def show
     vacancy = Vacancy.listed.friendly.find(id)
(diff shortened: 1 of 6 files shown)


despo:
[Links that hiring staff share should not break if the job title is updated](https://trello.com/c/Pnx3NERl/314-links-that-hiring-staff-share-should-not-break-if-the-job-title-is-updated)

despo:
[Links that hiring staff share should not break if the job title is updated](https://trello.com/c/Pnx3NERl/314-links-that-hiring-staff-share-should-not-break-if-the-job-title-is-updated)

PR #552, opened 2018-10-10
(chore) Update the README to be more explicit

lozette:
## Trello card URL:

https://trello.com/c/zl3rUSbt/496-update-readme-to-include-rake-task-info

## Changes in this PR:

I have updated the README to be a bit more explicit. I have also added some additional steps I needed when setting up the application locally, and a "Troubleshooting" section detailing some problems I encountered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Before you can log in to the application locally you will need a __DfE Sign-in__
 
 ### Importing school data
 
-Populate your environment with real school data from the Edubase archive
+Populate your environment with real school data. This is taken from [GIAS](https://get-information-schools.service.gov.uk/)
 
 ```bash
 bin/drake data:schools:import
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,29 +3,52 @@
...
 bin/dstart
 ```
 
+## User accounts & data
+
+Before you can log in to the application locally you will need a __DfE Sign-in__ and an __invitation to join Teaching Jobs__. Talk to the team to get these set up.
+
+### Importing school data
+
+Populate your environment with real school data from the Edubase archive
```
despo: The service that we populate school information from is GIAS ([Get information about schools](http://get-information-schools.service.gov.uk/)) - edubase a legacy domain they use. It might be worth linking to this as this is also how we verify schools when issues come up e.g. https://get-information-schools.service.gov.uk/Establishments/Establishment/Details/140212

README.md
```diff
@@ -3,29 +3,52 @@
...
+
+```bash
+bin/drake data:schools:import
+```
+
+_db/seeds.rb contain sample school data so this is not required for development_
+
+### Indexing the vacancies
+
+Index the test vacancies in Elasticsearch
```
despo: this should index all vacancies, in both test and development.

PR #553, opened 2018-10-10
[fix] handle both GET and POST publish requests

despo:
## Trello card URL:
https://trello.com/c/7qMOkZoI/498-fix-bug-with-publish-button-not-working-when-js-is-disabled

## Changes in this PR:
Exposes and handles both POST and GET requests to publish vacancies, as `link_to` does not correctly convert POST requests when JS is disabled. This means that users with JS disabled are unable to publish.


despo:
[Fix bug with publish button not working when JS is disabled.](https://trello.com/c/7qMOkZoI/498-fix-bug-with-publish-button-not-working-when-js-is-disabled)

PR #554, opened 2018-10-10
(fix) revert to correct locale

despo:
## Changes in this PR:
Rerverts mistakenly changed locale value and relevant feature.


despo:
[Fix bug with publish button not working when JS is disabled.](https://trello.com/c/7qMOkZoI/498-fix-bug-with-publish-button-not-working-when-js-is-disabled)

PR #557, opened 2018-10-11
(487) Display detailed error messages on review page if vacancy publish fails

esteedqueen:
## Trello card URL:
https://trello.com/c/x14tp4v4/487-display-detailed-error-messages-if-vacancy-publication-fails

## Changes in this PR:

- Display detailed error messages on the review page if vacancy publish fails and each error message links to the relevant component on the page.

### Before
screen shot 2018-10-11 at 14 12 53

### After
screen shot 2018-10-11 at 14 13 29

### DIFF
--- a/app/controllers/hiring_staff/vacancies/application_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_controller.rb
@@ -49,6 +49,10 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
     school_job_review_path(school_id: school.id, job_id: vacancy.id)
   end
 
+  def review_path_with_errors(vacancy)
+    school_job_review_path(school_id: school.id, job_id: vacancy.id, anchor: 'errors', source: 'publish')
+  end
+
   def redirect_unless_vacancy_session_id
     redirect_to job_specification_school_job_path(school_id: school.id) unless session_vacancy_id
   end
--- a/app/controllers/hiring_staff/vacancies/application_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_controller.rb
@@ -50,7 +50,7 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
   end
 
   def review_path_with_errors(vacancy)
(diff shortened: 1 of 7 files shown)


esteedqueen:
[Display detailed error messages if vacancy publication fails.](https://trello.com/c/x14tp4v4/487-display-detailed-error-messages-if-vacancy-publication-fails)

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -49,6 +49,10 @@ def review_path(vacancy)
     school_job_review_path(school_id: school.id, job_id: vacancy.id)
   end
 
+  def review_path_with_errors(vacancy)
+    school_job_review_path(school_id: school.id, job_id: vacancy.id, anchor: 'errors', source: 'publish')
```
despo: even though school_id is also set for review path, it's actually something we missed removing when the routes were refactored a while ago, as it is no longer required (we use the one stored in the sesion and this could lead to confussion). Can you remove that please.

app/views/hiring_staff/vacancies/review.html.haml
```diff
@@ -1,5 +1,4 @@
-- content_for :page_title_prefix, "Review the job details — Publish a job for #{@school.name}"
-
+- content_for :page_title_prefix, "#{@vacancy.errors.present? ? 'Error: ' : ''}Review the job details — Publish a job for #{@school.name}"
 =render partial: 'school_vacancy_breadcrumb'
```
despo: Can the title and error be extracted to the locale.
I wonder if it the logic for this can also be moved in the VacancyPresenter as `review_page_title` so we can clean up the view.

app/views/hiring_staff/vacancies/review.html.haml
```diff
@@ -1,5 +1,4 @@
-- content_for :page_title_prefix, "Review the job details — Publish a job for #{@school.name}"
-
+- content_for :page_title_prefix, "#{@vacancy.review_page_title}"
```
despo: I think you shouldnt need to wrap the vacancy method in quotes and evaluate. Other than that looks great!

esteedqueen: Ah yeah, good catch, thanks! 👍

PR #563, opened 2018-10-16
[fix] only refresh analytics data for listed non expired vacancies

despo:
## Trello card URL:
https://trello.com/c/CuwMQb6z/505-bug-restrict-vacancies-updated-with-google-information-to-currently-published-and-not-expired

## Changes in this PR:
Only update cached view data from Google Analytics for published non expired vacancies. Google Analytics API requests are limited and we don't want to end up doing thousands of requests daily for no reason as this may have an impact on the allowed service quota and/or request costs.


despo:
[Bug: Restrict vacancies updated with Google information to currently published and not expired.](https://trello.com/c/CuwMQb6z/505-bug-restrict-vacancies-updated-with-google-information-to-currently-published-and-not-expired)

PR #564, opened 2018-10-16
[fix] ensure that only published non expired vacancies are included in sitemap

despo:
## Trello card URL:
https://trello.com/c/mwY3kTX3/514-sitemap-should-also-only-include-published-but-not-expired-vacancies

## Changes in this PR:
- Only lists published non expired vacancies in generated sitemap. We want the crawler to know and keep our published vacancies up-to-date, not every single job that was ever published!
- Fix to ensure secure paths are linked


despo:
[Sitemap should also only include published but not expired vacancies](https://trello.com/c/mwY3kTX3/514-sitemap-should-also-only-include-published-but-not-expired-vacancies)

PR #568, opened 2018-10-18
Production release #47

despo:
- [x] Add missing key to worker container configuration
- [x] Amend fuzziness level on search to improve retrieve results (resolves bug where `art` matches `Part time`)


tahb:
On final test it doesn't appear that https://github.com/dxw/teacher-vacancy-service/pull/565 is having the desired effect on Staging. Despo and I have concluded that the change doesn't have any negative impacts on experience of searching so it can go into production and this card can go back to the backlog for another pass.

screen shot 2018-10-18 at 12 29 30

PR #569, opened 2018-10-18
(fix) Remove async defer attributes from Google maps API link

lozette:
## Trello card URL:

https://trello.com/c/zS86O5L4/503-bug-copying-a-job-listing-link-doesnt-change-clipboard

## Changes in this PR:

The async defer attributes were causing the Google maps API to be included in the page multiple times. See https://stackoverflow.com/questions/33115375/prevent-google-maps-js-executing-multiple-times-caused-by-rails-turbolinks


lozette:
[Bug: copying a job listing link doesn't change clipboard](https://trello.com/c/zS86O5L4/503-bug-copying-a-job-listing-link-doesnt-change-clipboard)

PR #600, opened 2018-10-30
(feature) move API under /api and skip filters

despo:
## Trello card URL:
https://trello.com/c/yQzCQbTG/541-jobs-api-should-respond-with-all-vacancy-data-regardless-of-state

## Changes in this PR:
- adds new Api::VacanciesController to handles JSON requests 
- existing filters utilised to ensure only non expired published jobs are returned
- no additional filters can be set
- no sorting
- no pagination


despo:
[Jobs API should respond with all vacancy data regardless of state](https://trello.com/c/yQzCQbTG/541-jobs-api-should-respond-with-all-vacancy-data-regardless-of-state)

PR #610, opened 2018-11-02
[feature][556] version our api per the gds standard

despo:
## Trello card URL:
https://trello.com/c/NcIX7BCB/556-version-our-api-per-the-gds-standard

## Changes in this PR:
- Adds version to API endpoint e.g. `/api/v1`
- Adds info object with version and other information
- Change `vacancies` to `data` following the [JSON API](https://jsonapi.org/format/) specification where `data` is the document's primary data. As this is a significant change we should include it as part of versioning so we are not required to move to v2.

### Before
screen_shot_2018-11-02_at_14_17_08

### After
screen_shot_2018-11-02_at_14_19_37_1

### DIFF
--- a/app/views/api/vacancies/index.json.jbuilder
+++ b/app/views/api/vacancies/index.json.jbuilder
@@ -1,3 +1,19 @@
-json.vacancies @vacancies.decorated_collection do |vacancy|
+json.info do
+  json.title "GOV UK - #{I18n.t('app.title')}"
+  json.description I18n.t('app.description')
+  json.termsOfService terms_and_conditions_url(protocol: 'https', anchor: 'api')
+  json.contact do
+    json.name "#{I18n.t('app.title')} API Support"
+    json.email I18n.t('help.email')
+  end
+  json.license do
+    json.name 'Open Government License'
+    json.url 'https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/'
+  end
+  json.version '0.0.1'
+end
+json.openapi '3.0.0'
+
+json.data @vacancies.decorated_collection do |vacancy|
   json.partial! 'show.json.jbuilder', vacancy: vacancy
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,7 +16,9 @@ Rails.application.routes.draw do
   end
 
   namespace :api do
-    resources :jobs, only: %i[index show], controller: 'vacancies'
+    scope 'v:api_version', api_version: /[1]/ do
+      resources :jobs, only: %i[index show], controller: 'vacancies'
+    end
   end
 
   resources :stats, only: [:index]
--- a/spec/controllers/api/vacancies_controller_spec.rb
+++ b/spec/controllers/api/vacancies_controller_spec.rb
@@ -7,18 +7,18 @@ RSpec.describe Api::VacanciesController, type: :controller do
     request.accept = 'application/json'
   end
 
-  describe 'GET /jobs', elasticsearch: true do
+  describe 'GET /api/v1/jobs.json', elasticsearch: true do
     render_views
 
     context 'sets headers' do
       it 'robots are asked to index but not to follow' do
-        get :index
+        get :index, params: { api_version: 1 }
         expect(response.headers['X-Robots-Tag']).to eq('noarchive')
       end
     end
 
     it 'returns status :not_found if the request format is not JSON' do
-      get :index, format: :html
+      get :index, params: { api_version: 1 }, format: :html
 
       expect(response.status).to eq(Rack::Utils.status_code(:not_found))
     end
@@ -28,7 +28,7 @@ RSpec.describe Api::VacanciesController, type: :controller do
 
       Vacancy.__elasticsearch__.refresh_index!
 
-      get :index
+      get :index, params: { api_version: 1 }
 
       expect(response.status).to eq(Rack::Utils.status_code(:ok))
       expect(json[:vacancies].count).to eq(3)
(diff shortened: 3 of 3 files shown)


app/views/api/vacancies/index.json.jbuilder
```diff
@@ -1,3 +1,19 @@
-json.vacancies @vacancies.decorated_collection do |vacancy|
+json.info do
```
tahb: Reading through https://jsonapi.org/format/#document-top-level should this be `meta`?

app/views/api/vacancies/index.json.jbuilder
```diff
@@ -1,3 +1,19 @@
...
+  json.title "GOV UK - #{I18n.t('app.title')}"
+  json.description I18n.t('app.description')
+  json.termsOfService I18n.t('service.url')
+  json.contact do
+    json.name "#{I18n.t('app.title')} API Support"
+    json.url 'tbc'
```
tahb: I see there was service.url added to locale, was the intent to use that here?

spec/controllers/api/vacancies_controller_spec.rb
```diff
@@ -7,49 +7,59 @@
...
     end
 
     it 'returns status :not_found if the request format is not JSON' do
-      get :index, format: :html
+      get :index, params: { api_version: 1 }, format: :html
 
       expect(response.status).to eq(Rack::Utils.status_code(:not_found))
     end
 
+    it 'returns the API\'s info', wip: true do
```
tahb: `wip: true` left in

config/routes.rb
```diff
@@ -16,7 +16,9 @@
   end
 
   namespace :api do
-    resources :jobs, only: %i[index show], controller: 'vacancies'
+    scope 'v:api_version', api_version: /[1]/ do
```
tahb: If we needed to bump to version 2, is the plan that we would add a new block with a different controller name like this:

```
scope 'v:api_version', api_version: /[2]/ do
  resources :jobs, only: %i[index show], controller: 'v2/vacancies'
end
```

If so, might it be prudent to namespace the existing controller to v1?

app/views/api/vacancies/index.json.jbuilder
despo: @tahb actually I was unsure of what to use here as this is meant to be a contact url and we don't have a contact page. I think we can remove this field since there isnt one, and just leave the email address.

app/views/api/vacancies/index.json.jbuilder
```diff
@@ -1,3 +1,19 @@
-json.vacancies @vacancies.decorated_collection do |vacancy|
+json.info do
+  json.title "GOV UK - #{I18n.t('app.title')}"
+  json.description I18n.t('app.description')
+  json.termsOfService I18n.t('service.url')
```
tahb: Can we use a rails url method to return `https://teaching-jobs.service.gov.uk/pages/terms-and-conditions` here?

app/views/api/vacancies/index.json.jbuilder
tahb: What about `mailto:teachingjobs@digital.education.gov.uk`?

app/views/api/vacancies/index.json.jbuilder
despo: No. We are following the Open API standard [as specified in the GDS guide](https://www.gov.uk/guidance/gds-api-technical-and-data-standards#document-your-api).

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md

screen_shot_2018-11-06_at_13_28_46_1

app/views/api/vacancies/index.json.jbuilder
despo: It specifically indicates that it should be a url. However it's not a required field. 

screen shot 2018-11-06 at 13 33 53 1

config/routes.rb
despo: Yes. We would also have to edit v1. The following example should do it.

```
  namespace :api do
    scope 'v:api_version', api_version: /[1]/ do
      resources :jobs, only: %i[index show], controller: 'v1/vacancies'
    end

    scope 'v:api_version', api_version: /[2]/ do
      resources :jobs, only: %i[index show], controller: 'v2/vacancies'
    end

    get '*path', :to => redirect("/api/v2/%{path}")
  end
```

app/views/api/vacancies/index.json.jbuilder
tahb: 👍 thanks for linking

app/views/api/vacancies/index.json.jbuilder
tahb: So it does, URL rather than URI. I'll stop trying to be clever! If it's optional I'm in favour of removing it as it's unlikely it will ever be corrected from 'tbc' even if a contact page is eventually created.

config/routes.rb
tahb: The wildcard redirector worries me a bit. 

Whilst I can see the utility in helping users who haven't declared a version number and want to get started quickly with the current latest version, I can also see that by being relaxed it's omission we may accidentally break automated integrations with us. When we make breaking changes in v2 all requests set up initially with `api/jobs` as the path will suddenly stop working or behalf unexpectedly. I think this is one of the benefits of versioning the API. I would think about returning a 404 instead, and ensure all requests specify a version number explicitly.

Do you see this being an issue?

tahb: I'm also unclear about your thoughts on changing the controller's namespace now, rather than later when/if v2 comes?

despo: I don't think we should change it now. There is no plan to extend this API to be writable at any point so I don't see the value in making a change like that.

tahb: We are more likely to add or remove data from the representation of job listings. eg. flexible working. The controller name spacing can be done later when we need to do it 👍 

The wildcard redirector is still a worry for me. Thoughts?

despo: Regarding the redirection, I came across it when looking up how to best structure the routes for the API and it seems to be common practise. There is a possibility that it will break, however, the goal is to enable the users to be able to discover the latest version rather than to use it to retrieve data by default. I'm happy to take it out though.

despo: @tahb removed

tahb: Was there a particular article you found that you could link?

despo: https://stackoverflow.com/questions/389169/best-practices-for-api-versioning
https://stackoverflow.com/questions/9627546/api-versioning-for-rails-routes

tahb: Thanks for the links.

I'm unconvinced by their rationale and am struggling to see the valuable benefits.  

I feel like the safer route is to require the API user to explicitly choose a version which might take them a few more minutes to pick or discover however it should give long term assurance of a stable interface for integration.

A luxury of this route is that we can always add an alias in later if we really need to. We cannot however responsibly remove it once we put it in users hands.

despo: @tahb I have already removed that commit.

PR #622, opened 2018-11-12
(fix) Only show these warnings in Rails development env

lozette:
## Trello card URL:

https://trello.com/c/Clj34m9S/570-prod-stage-report-that-no-authorisation-env-is-set-bug

## Changes in this PR:

Wrap the missing `AUTHORISATION_SERVICE_URL` and AUTHORISATION_SERVICE_TOKEN` warnings in Rails.env.development? so that they don't appear in production.

We decided to only show these warnings in development. In production this warning will be handled by Rollbar.

I wanted to test this, perhaps in `spec/services/teacher_vacancy_authorisation_spec.rb` but it's hard to test initializers with rspec. I can keep plugging away at it if it's important?


lozette:
[Prod & stage report that no authorisation ENV is set, bug?](https://trello.com/c/Clj34m9S/570-prod-stage-report-that-no-authorisation-env-is-set-bug)

PR #623, opened 2018-11-12
(fix) Static pages should not have NOINDEX unless it is the user-not-…

lozette:
…authorised page

## Trello card URL:

https://trello.com/c/JfT23ZhJ/579-fix-google-search-coverage-error-for-static-pages-that-are-included-in-the-sitemap

## Changes in this PR:

Remove the NOINDEX header from static pages so that Google will index them.

However, I have left NOINDEX on the `user-not-authorised` page as I don't think we want Google to crawl that (@tahb can you confirm?)


tahb:
@lozette agreed, we don't want Google directing any users to the not authorised page as it is effectively an error page.

PR #624, opened 2018-11-13
[591] Redirect traffic to the old service hostname

tahb:
## Trello card URL:
https://trello.com/c/EMDwIScy

## Changes in this PR:
* I've tested this on Edge. you can test by curling tvs.edge.dxw.net and tvs.edge2.dxw.net. The first should be a 301 redirect and the second should be as normal.
* Both HTTP and HTTPS
* Permanent solution part of the core infrastructure rather than using another service for redirection. This change is designed to be permanent since teaching-jobs.service.gov.uk has been seen, used and linked to by users in public beta
* More reliable than using a redirection service hosted somewhere
* Can turn it on/off per environment via environment variable so it can the redirect will only be running in production where more than 1 domain exists

## Next steps:

### DIFF
--- a/terraform.tf
+++ b/terraform.tf
@@ -39,6 +39,9 @@ module "core" {
   asg_min_size     = "${var.asg_min_size}"
   asg_desired_size = "${var.asg_desired_size}"
 
+  domain = "${var.domain}"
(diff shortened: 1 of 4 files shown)


tahb:
[Redirect all old traffic to the new service domain](https://trello.com/c/EMDwIScy/591-redirect-all-old-traffic-to-the-new-service-domain)

PR #625, opened 2018-11-14
[594] API returns as many jobs by default as possible

tahb:
## Trello card URL:
https://trello.com/c/S84E2Fvl

## Changes in this PR:
* We cannot yet make assumptions about the users of the API, if this is someone who wants information on expired listings or someone else who wants to use the data to publish on their site. By folllowing the standard, we make all data that’s been made public past and present available by default. In future, we can add filter parameters to help users get to the subsets of data they want more efficiently.
* Does not couple itself to the other user search code as the needs of this endpoint are different from the needs of users searching through the front end. We may introduce “API search” in future but we shouldn’t set early assumptions that they will use the same code. This is already an issue as lots more code is being run than nescassry to fulfill the task at hand and is prone to any changes we do for job seeker users.
* We don’t want to accidentally call any extra code like Rollbar.info(‘0 results searched’) which could lead to misleading data.

## Next steps:

### DIFF
--- a/app/controllers/api/vacancies_controller.rb
+++ b/app/controllers/api/vacancies_controller.rb
@@ -3,8 +3,7 @@ class Api::VacanciesController < Api::ApplicationController
   before_action :verify_json_or_csv_request, only: %w[index]
 
   def index
-    filters = VacancyFilters.new({})
-    records = Vacancy.public_search(filters: filters, sort: {}).records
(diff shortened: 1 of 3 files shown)


tahb:
[Fix API bulk downloads include expired jobs](https://trello.com/c/S84E2Fvl/594-fix-api-bulk-downloads-include-expired-jobs)

PR #632, opened 2018-11-19
(fix) Testing with expired vacancies is less fragile

tahb:
## Changes in this PR:
Sometimes when this test rungs and `publish_on { Faker::Time.backward(14) }` results in a date in the range of 7-14 days instead of 0-7 days, it fails validation that states the expiry must be after publish.

### Before
screenshot 2018-11-19 at 16 57 55

## Next steps:

### DIFF
--- a/spec/factories/vacancies.rb
+++ b/spec/factories/vacancies.rb
@@ -79,8 +79,8 @@ FactoryBot.define do
     trait :expired do
       status { :published }
       sequence(:slug) { |n| "slug-#{n}" }
-      publish_on { Faker::Time.backward(14) }
-      expires_on { Faker::Time.backward(7) }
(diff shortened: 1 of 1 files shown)


spec/factories/vacancies.rb
```diff
@@ -79,7 +79,7 @@
     trait :expired do
       status { :published }
       sequence(:slug) { |n| "slug-#{n}" }
-      publish_on { Faker::Time.backward(14) }
+      publish_on { Faker::Time.between(Time.zone.today - 14.days, Time.zone.today - 7.days) }
       expires_on { Faker::Time.backward(7) }
```
despo: We should change this to 6 to ensure that the publication and expiry date don't overlap.

despo: @tahb I think this change should be applied to the expires_on, as now there is ac hance of the publication date being after the expiry date.

PR #654, opened 2018-12-10
[673] Refactor DfE Sign-in to handle password resets without a monk…

tahb:
…ey patch

* CCS came up with a more elgant approach and enables us to completely remove the monkey patch for the OpenIDConnect gem: https://github.com/Crown-Commercial-Service/crown-marketplace/commit/2d169633f7e6c7ba5c9ff3425c47b87e3651a40b
* I was also able to refactor several bits of configuration away
* I simplified the guard clauses on every ENV fetch to a single line at the top of the file. It clearly explains that Docker is the reason we need to handle the scenario where ENV is not available, as it isn’t using the `docker build` phase.
* To remind us of why we need to do this. At the end of a password reset journey using DfE Sign-in, that service redirects services back to our service. They use the same URL as the one used during the authentication phase, however this time it doesn’t have a valid payload. Omniauth promptly treats this request as unauthorised instead of linking the user to the sign in page. Since we cannot configure 2 callback URL’s within DSI, we are checking the payload in a before step, when it doesn’t exist at all we redirect that use to the sign in step.

## Trello card URL:
https://trello.com/c/0fBiG9GY/673-replace-dfe-sign-in-monkey-patch

## Next steps:

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,74 +1,46 @@
-module OmniAuth
-  module Strategies
-    class OpenIDConnect
-      # rubocop:disable Metrics/AbcSize
-      # rubocop:disable Metrics/CyclomaticComplexity
-      # rubocop:disable Metrics/PerceivedComplexity
-      # rubocop:disable Metrics/LineLength
-      # rubocop:disable Style/GuardClause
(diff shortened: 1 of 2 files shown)


tahb:
[Replace DfE Sign-in monkey patch](https://trello.com/c/0fBiG9GY/673-replace-dfe-sign-in-monkey-patch)

tahb:
Closing in favour of https://github.com/dxw/teacher-vacancy-service/pull/658 which has an improved understanding of the situation.

PR #656, opened 2018-12-11
[Feature] 1.1.1 Email alerts - subscription model

despo:
## Trello card URL:
https://trello.com/c/5oJ6NTzf/656-111-email-alerts-subscription-model

## Changes in this PR:
Adds subscription model and essential validations.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,0 +1,19 @@
+class Subscription < ApplicationRecord
+  enum status: %i[active trashed]
+  enum frequency: %i[daily]
+
+  validates :email, email_address: { presence: true }
+  validates :frequency, presence: true
+  validates :search_criteria, uniqueness: { scope: %i[email expires_on frequency] }
+
+  scope :ongoing, -> { active.where('expires_on >= current_date') }
+
+  before_save :set_reference
+
+  def set_reference
+    self.reference ||= loop do
+      reference ||= SecureRandom.hex(8)
+      break reference unless self.class.exists?(email: email, reference: reference)
+    end
+  end
+end
--- a/app/validators/concerns/validator_concerns.rb
+++ b/app/validators/concerns/validator_concerns.rb
@@ -1,0 +1,17 @@
+module ValidatorConcerns
+  extend ActiveSupport::Concern
+
+  def check_presence?
+    options.key?(:presence) && options[:presence]
+  end
+
+  private
+
+  def error_message(record, attribute, message)
+    record.errors[attribute] << message
+  end
+
+  def cant_be_blank_message
+    I18n.t('errors.messages.blank')
+  end
+end
--- a/app/validators/email_address_validator.rb
+++ b/app/validators/email_address_validator.rb
@@ -1,0 +1,23 @@
+class EmailAddressValidator < ActiveModel::EachValidator
+  EMAIL_FORMAT = /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
+
+  def validate_each(record, attribute, value)
+    return record.errors[attribute] << cant_be_blank_message if check_presence? && value.blank?
+    record.errors[attribute] << invalid_error_message(options) if value[EMAIL_FORMAT].nil?
+  end
+
+  private
+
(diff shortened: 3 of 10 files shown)


despo:
[1.1.1 Email alerts: Subscription model](https://trello.com/c/5oJ6NTzf/656-111-email-alerts-subscription-model)

db/migrate/20181210151520_add_index_to_subscriptions.rb
```diff
@@ -0,0 +1,7 @@
+class AddIndexToSubscriptions < ActiveRecord::Migration[5.2]
+  def change
+    add_index :subscriptions, [:email, :status, :frequency, :expires_on], unique: true,
```
despo: this needs to be removed as it will restict creating multiple alerts on the same day (as search_criteria is a json field and cannot be used as part of the index)

app/validators/email_address_validator.rb
```diff
@@ -0,0 +1,16 @@
+class EmailAddressValidator < ActiveModel::EachValidator
+  include ValidatorConcerns
+  EMAIL_FORMAT = /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
```
lozette: Not for now, but I'm wondering if we can re-use this email validator for the contact_email field in the job posting eventually. We have a ton of validators all over the place and it might be nice to reconcile them a bit.

app/models/subscription.rb
```diff
@@ -0,0 +1,19 @@
...
+  enum status: %i[active trashed]
+  enum frequency: %i[daily]
+
+  validates :email, email_address: { presence: true }
+  validates :frequency, presence: true
+  validates :search_criteria, uniqueness: { scope: %i[email expires_on frequency] }
+
+  scope :ongoing, -> { active.where('expires_on >= current_date') }
+
+  before_save :set_reference
```
tahb: Active record hooks make be wary as they can be called at unexpected times. I found some considerations [1] which I think are worth considering. 

- should this run with `after_commit` instead of `after_save` or `before_save`
- do we need to set this after save at all? could we set it on the initialisation, or are there other considerations to be made? 

https://www.justinweiss.com/articles/a-couple-callback-gotchas-and-a-rails-5-fix/

app/validators/concerns/validator_concerns.rb
```diff
@@ -0,0 +1,17 @@
+module ValidatorConcerns
```
tahb: Nice extraction and reuse.

spec/models/subscription_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  context 'reference' do
+    it 'generates a reference if one is not set' do
+      expect(SecureRandom).to receive(:hex).and_return('ABCDEF')
+      subscription = create(:subscription, frequency: :daily)
+
+      expect(subscription.reference).to eq('ABCDEF')
+    end
+
+    it 'does not generate a reference if one is set' do
```
tahb: 👍

app/models/subscription.rb
despo: I will take a better look, however the Subscription model is not indexed by ES, and the value should be generated within the user session (if the user does not set a value, which is why it cannot be done on initialisation) so I don't think this is necessary. 

What I think I missed here is enforcing a value in the reference field, as the subscription mockups rely on having that value set. That would enforce why this needs to happen in a before hook as well.

db/migrate/20181210145601_create_subscriptions.rb
```diff
@@ -0,0 +1,14 @@
+class CreateSubscriptions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :subscriptions, id: :uuid do |t|
+      t.string :email
+      t.integer :frequency
+      t.date :expires_on
+      t.integer :status, default: 0
+      t.jsonb :search_criteria
+      t.string :reference, null: false
```
despo: Latest amendment. Don't allow an empty reference at db level as we don't want to enforce model validation (as we will generate the value if the user does not set one).

PR #657, opened 2018-12-12
[Feature] 1.2.1 Email alerts - New subscription page

despo:
Pending merge of #656 

## Trello card URL:
https://trello.com/c/gykP6TKh/657-121-email-alerts-new-subscription-page

## Changes in this PR:
- Adds subscription page
- Formats search_criteria 
- Generates subscription reference
- Adds reusable email_address validator

## Screenshot

image

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,0 +1,23 @@
+class SubscriptionsController < ApplicationController
+  def new
+    subscription = Subscription.new(search_criteria: search_criteria.to_json)
+    @subscription = SubscriptionPresenter.new(subscription)
+  end
+
+  def create
+    subscription = Subscription.new(subscription_params)
+    @subscription = SubscriptionPresenter.new(subscription)
+
+    render 'new' unless @subscription.valid?
+  end
+
+  private
+
+  def subscription_params
+    params.require(:subscription).permit(:email, :reference, :search_criteria)
+  end
+
+  def search_criteria
+    params.require(:search_criteria).permit(*VacancyFilters::AVAILABLE_FILTERS)
+  end
+end
--- a/app/views/subscriptions/new.html.haml
+++ b/app/views/subscriptions/new.html.haml
@@ -1,0 +1,39 @@
+- content_for :page_title_prefix do
+  = t('subscriptions.new')
+
+- content_for :page_description do
+  = t('subscriptions.new')
+
+.govuk-grid-row
+  .govuk-grid-column-full
+    %h1.govuk-heading-xl
+      = t('subscriptions.new')
+.govuk-grid-row
+  .govuk-grid-column-full
+    %p.govuk-body-l
+      = t('subscriptions.intro')
+
+    %div.govuk-inset-text
+      %ul.govuk-list
+        - @subscription.filtered_search_criteria.each_pair do |filter, value|
+          %li
+            - if filter.present?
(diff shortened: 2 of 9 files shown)


despo:
[1.2.1 Email alerts: New subscription page](https://trello.com/c/gykP6TKh/657-121-email-alerts-new-subscription-page)

app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  def new
+    subscription = Subscription.new(search_criteria: search_criteria.to_json)
+    @subscription = SubscriptionPresenter.new(subscription)
+  end
+
+  def create
+    subscription = Subscription.new(subscription_params)
+    @subscription = SubscriptionPresenter.new(subscription)
+
+    render 'new' unless @subscription.valid?
```
tahb: If it is valid, where does the user go?

app/views/subscriptions/new.html.haml
```diff
@@ -0,0 +1,37 @@
...
+        - @subscription.filtered_search_criteria.each_pair do |filter, value|
+          %li
+            - if filter.present?
+              %span{ class: 'govuk-!-font-weight-bold'} #{filter.humanize}:
+            =value
+
+.govuk-grid-row
+  .govuk-grid-column-one-half
+    = simple_form_for @subscription do |f|
+      = f.input :email, as: :email, input_html: { class: 'govuk-input' }, required: true
```
tahb: Sometimes in this file we are putting spaces after `=` and sometimes not. Not sure it's worth the effort to add a rubocop for haml but I thought I'd flag in case you hadn't noticed.

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.feature 'A job seeker can subscribe to a job alert' do
+  context 'A job seeker' do
+    scenario 'can access the new subscription page when search criteria have been specified' do
+      expect { visit(new_subscription_path) }.to raise_error(ActionController::ParameterMissing)
```
tahb: We are going directly to the page here. I presume that's because another story will come along and add the start page for the user after they finish a search?

app/controllers/subscriptions_controller.rb
despo: Nowhere on this PR 😈 

https://trello.com/c/naFQHYhF/659-123-email-alerts-subscription-confirmation addreses that.

app/views/subscriptions/new.html.haml
despo: Good point. I noticed that I was doing this earlier 🙄 I'll add a card.

despo: https://trello.com/c/hNBToacJ/680-rubocop-investigate-add-rubocop-haml-cop

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
despo: The first page is still missing as it won't be implemented until this card https://trello.com/c/OFbNOs2j/660-124-email-alerts-a-visitor-can-subscribe-to-a-search

However, I thought that that page was left in the design as it might be needed later on, and not for implementing now.

app/views/subscriptions/new.html.haml
despo: I've also updated the file to include a space in all assignments.

PR #658, opened 2018-12-12
[673] Refactor omniauth_openid_connect monkey patch to make it manageable

tahb:
## Trello card URL:
https://trello.com/c/0fBiG9GY

## Changes in this PR:

This is a refactoring exercise to manage the use of our monkey patch, and bring us closer to a point where we can remove it.

Before going into this it wasn't clear that we were using it for 2 purposes:
1. stop password reset completions being incorrectly routed to the 401 page by DSI  (DfE Sign-in)
2. ensure that any authentication failures resulted in a styled and monitored 401 page

Trying to test this with feature specs was an absolute nightmare. It's easy to mock OmniAuth but it also wraps up the behaviour of the strategies, so I couldn't create a realistic simulation. The only feature test I did make, was actually a lie because of the Omniauth stubbing layer.
```
  context 'with invalid credentials' do
    before(:each) { OmniAuth.config.mock_auth[:dfe] = :invalid_credentials }

    scenario 'redirects the user to the not authorised page' do
      visit root_path
      click_on(I18n.t('nav.sign_in'))
      click_on(I18n.t('sign_in.link'))

      expect(page).to have_content(I18n.t('error_pages.unauthorised')) => this passes but isn't what actually happens as OmniAuth defers the callback phase to the strategy in omniauth_openid_connect!
      a_plain_401_rack_response = '401 Unauthorized'
      expect(page).not_to have_content(a_plain_401_rack_response)  => I could never make this appear in the first place via tests, only postman
    end
  end
```

My view is that it it is better to have a styled 401 page with GA tracking for monitoring, and to manage this code than it would be to remove it, serve a white 401 page and never know how much it was a problem for users.

Further details are on the commits. 

Next steps are to open a pull request on the omniauth_openid_connect gem to change their basic rack response with a redirect to /auth/failure.

### Before
screenshot 2018-12-12 at 15 18 55

### After
screenshot 2018-12-12 at 15 20 54

## Next steps:

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -11,10 +11,6 @@ module OmniAuth
         error = request.params['error_reason'] || request.params['error']
         if error
           raise CallbackError.new(request.params['error'], request.params['error_description'] || request.params['error_reason'], request.params['error_uri'])
-        elsif request.params.blank? && request.path == '/auth/dfe/callback'
-          response = Rack::Response.new
-          response.redirect('/dfe/sessions/new')
-          response.finish
         elsif request.params['state'].to_s.empty? || request.params['state'] != stored_state
           return redirect('/401')
         elsif !request.params['code']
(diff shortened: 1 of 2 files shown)


tahb:
[Replace DfE Sign-in monkey patch](https://trello.com/c/0fBiG9GY/673-replace-dfe-sign-in-monkey-patch)

PR #664, opened 2018-12-14
[676] Hiring staff using the new table can get back to the table in …

tahb:
…1 click

* No JS and mobile views with small aspects are automatically switched from tabs to a single page of stacked tables by the design system - no problem for us
* On Desktop if you clicked through the new available tabs in the jobs table (draft, expired, etc), then clicked to view or edit a job and used the browsers back function to go backwards, the url would change but the page content would not.
* Instead the browser would cycle back through all the history /school#draft, /school#published, /school#published before it got to /school and it would then render the page.
* This is presumably a bug within the design system itself.
* The table works great for everything else apart from this issue
* I manipulated the browser’s history for browsers that use HTML5 [1]. This means that when I detect a click off of the /school page to another part of our service, I add a new popstate into the browser history ready for use if the user comes back via the browser back function.
* Listening for onpopstate and checking for the special state that’s been added allows me to force the /school pages to reload.
* This is not perfect. Reasonably if the user was on the Draft tab before they went somewhere else, they should expect to come back to the draft tab, rather than the starting point of the /school page. I think this is preferable to the back button in the browser appearing completely broken.
* This behaviour is very similar to how the design system  functions on mobile. No hashes/anchors are used and since all the tables are on one page, if you go forward and press back, you always come back to the /school page.
* Make the reload smarter to fix this is probably viable. We could go through the browser history states and detect the last one with a hash and add that to the window.location on reload. This could work but would be very fragile as any new anchor link would break it or if the link history involved another service, weird things could start happening!

[1] https://developer.mozilla.org/en-US/docs/Web/API/History_API

Pull Request template

- Fill in or remove sections as necessary
- Delete this section before submitting your PR

## Trello card URL:
https://trello.com/c/pbTHPIT0

### Before
https://drive.google.com/file/d/1XMZ4_XeA2wqXbY4z8HnPe-X3U6qXvx2R/view

### After

With Javascript enabled:
ezgif com-crop

Without Javascript enabled: 
ezgif com-crop 1

## Next steps:


tahb:
[Bug: Back browser navigation not working correctly from tab content link (attached video)](https://trello.com/c/pbTHPIT0/676-bug-back-browser-navigation-not-working-correctly-from-tab-content-link-attached-video)

app/assets/javascripts/govuk_frontend_tab_patch_for_using_browser_back.js
```diff
@@ -0,0 +1,12 @@
+window.onpopstate = function(event) {
+  if (event.state == null) return
+  if (event.state.name == 'CustomHistoryManipulator') {
+    window.location = document.location;
+  }
+};
+
+$(document).on('click', '.govuk-link', function(e) {
```
tahb: This should be flexible enough so that any links added and remove to the school#show page don't have to remember to add a custom class for this to listen to. This seems okay since I'm only including that JS on this page.

despo:
@tahb https://github.com/dxw/teacher-vacancy-service/compare/feature/676-hiring-staff-table-can-be-navigted-back-to-2?expand=1

PR #667, opened 2018-12-17
(feature) Copy an existing job

lozette:
## Trello card URL:

https://trello.com/c/tEijeDTC/562-hiring-staff-ui-user-can-make-a-copy-of-an-existing-job

## Changes in this PR:

A hiring staff can make a copy of an existing published or pending job (_not_ a draft or expired job).

The Published and Pending job tables now feature a Duplicate link. 

Clicking this link calls `dup` on the original vacancy.

The copied vacancy is updated to have a `draft` status.

The copied vacancy's `publish_on` date is updated to today if the original vacancy's publication date is before today (because a new vacancy cannot have a publish_on date in the past). 

The copied vacancy's title is also updated to be "Copy of {original vacancy job title}". 

The hiring staff is then shown the Review page, from which they can edit or publish their new job. 

The copied vacancy is a completely new record with a new id.

### DIFF
--- a/app/controllers/hiring_staff/vacancies/copy_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/copy_controller.rb
@@ -1,8 +1,7 @@
 class HiringStaff::Vacancies::CopyController < HiringStaff::Vacancies::ApplicationController
   def create
     vacancy = Vacancy.find(vacancy_id)
-    vacancy_copy = vacancy.dup
-    update_fields(vacancy_copy)
+    vacancy_copy = CopyVacancy.new(vacancy: vacancy).copy
     if vacancy_copy.save
       Auditor::Audit.new(vacancy_copy, 'vacancy.copy', current_session_id).log
       redirect_to review_path(vacancy_copy)
@@ -12,13 +11,6 @@ class HiringStaff::Vacancies::CopyController < HiringStaff::Vacancies::Applicati
   end
 
   private
-
-  def update_fields(vacancy)
-    vacancy.job_title = "#{I18n.t('jobs.copy_of')} #{vacancy.job_title}"
-    vacancy.status = :draft
-    vacancy.publish_on = Time.zone.today if vacancy.publish_on <= Time.zone.today
-    vacancy
-  end
 
   def vacancy_id
     params.permit(:job_id)[:job_id]
--- a/app/controllers/hiring_staff/vacancies/copy_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/copy_controller.rb
@@ -4,6 +4,7 @@ class HiringStaff::Vacancies::CopyController < HiringStaff::Vacancies::Applicati
     vacancy_copy = vacancy.dup
     update_fields(vacancy_copy)
     if vacancy_copy.save
+      Auditor::Audit.new(vacancy_copy, 'vacancy.copy', current_session_id).log
       redirect_to review_path(vacancy_copy)
     else
       redirect_to school_path, notice: I18n.t('errors.jobs.unable_to_copy')
--- a/app/controllers/hiring_staff/vacancies/copy_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/copy_controller.rb
@@ -1,0 +1,25 @@
+class HiringStaff::Vacancies::CopyController < HiringStaff::Vacancies::ApplicationController
+  def create
+    vacancy = Vacancy.find(vacancy_id)
+    vacancy_copy = vacancy.dup
+    update_fields(vacancy_copy)
+    if vacancy_copy.save
+      redirect_to review_path(vacancy_copy)
+    else
+      redirect_to school_path, notice: I18n.t('errors.jobs.unable_to_copy')
+    end
+  end
+
+  private
+
+  def update_fields(vacancy)
+    vacancy.job_title = "#{I18n.t('jobs.copy_of')} #{vacancy.job_title}"
+    vacancy.status = :draft
+    vacancy.publish_on = Time.zone.today if vacancy.publish_on <= Time.zone.today
+    vacancy
+  end
+
+  def vacancy_id
+    params.permit(:job_id)[:job_id]
+  end
+end
--- a/app/views/hiring_staff/schools/_vacancy.html.haml
+++ b/app/views/hiring_staff/schools/_vacancy.html.haml
@@ -3,6 +3,5 @@
   %td.govuk-table__cell= format_date(vacancy.publish_on)
   %td.govuk-table__cell= format_date(vacancy.expires_on)
   %td.govuk-table__cell= vacancy.status.titleize
-  %td.govuk-table__cell= link_to 'Edit', edit_school_job_path(vacancy.id), class: 'govuk-link'
-  %td.govuk-table__cell= link_to 'Delete', school_job_path(id: vacancy.id), class: 'govuk-link', method: :delete, data: { confirm: "Are you sure you want to delete the '#{vacancy.job_title}' job?" }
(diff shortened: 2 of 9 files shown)


lozette:
[Hiring staff UI: user can make a copy of an existing job](https://trello.com/c/tEijeDTC/562-hiring-staff-ui-user-can-make-a-copy-of-an-existing-job)

lozette:
I have a couple of questions here:

1) Do we need to record the id of the vacancy a copied vacancy was copied from?
2) Do we need a new event logged, e.g. `vacancy.duplicated`? 
3) Is the text OK for the Duplicate button and the error message I show if a vacancy cannot be copied?

lozette:
This PR is dependent on the hiring staff UI changes, so if #668 goes through then this one should be placed on hold 😢

app/controllers/hiring_staff/vacancies/copy_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    if vacancy_copy.save
+      redirect_to review_path(vacancy_copy)
+    else
+      redirect_to school_path, notice: I18n.t('errors.jobs.unable_to_copy')
+    end
+  end
+
+  private
+
+  def update_fields(vacancy)
```
tahb: Wrapping this logic up and moving it to a small service object would get it out of the controller which would keep this file simpler and give us a way to add light unit tests for it. 

The existing condition of `if vacancy.publish_on <= Time.zone.today` is currently tested with 2 dedicated feature specs. As feature tests are expensive we should try to push as many as we can into unit tests whilst keeping our confidence high that we'd catch a regression, I think these are perfect candidates:

```
scenario 'the publish_on date is updated to today'
scenario 'the publish_on date is the same as the pending job it was copied from'
```

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    click_on I18n.t('jobs.duplicate_link')
+    click_on I18n.t('jobs.submit')
+
+    expect(page).to have_content(I18n.t('jobs.confirmation_page.submitted'))
+  end
+
+  context 'an error copying a job' do
+    before do
+      copy = build(:vacancy, school: school, status: 'draft')
+      allow_any_instance_of(Vacancy).to receive(:dup).and_return(copy)
```
tahb: Mocking and stubbing of internal classes and objects in feature tests where we are meant to test all the way through our stack seems a little odd to me. 

What can a real user of our tool do to achieve an error here?

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    create(:vacancy, school: school, status: 'published')
+    create(:vacancy,
+           school: school,
+           status: 'published',
+           expires_on: Time.zone.today + 4.days,
+           publish_on: Time.zone.tomorrow)
+
+    visit school_path
+
+    published_table = page.find('section#published-jobs table')
```
tahb: These tests are specific about where in the DOM they expect to find the content. I tend to find being specific in 1 test makes it much easier to accommodate changes to the DOM in future.

In this case we now know we're going to have to change this layout imminently from the design system to something with links, and there's a good chance that 4 of these tests will have to be reworked to get things to go green. 

If you extrapolate that approach out across the whole application, changing something in a common view (root_path) could take someone a very long time to work through and could be quite painful for them.

My advice would be to have 1 test for the happy path of this feature that _does_ go into specifics and trust that it will fail if something is unexpectedly changed. The edge cases can likely get by by asserting content is present on the page.

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  before(:each) do
+    stub_hiring_staff_auth(urn: school.urn)
+  end
+
+  scenario 'hiring staff can see a Duplicate link on published and pending jobs' do
+    create(:vacancy, school: school, status: 'published')
+    create(:vacancy,
+           school: school,
+           status: 'published',
```
tahb: [You can use FactoryBot traits for building objects with a particular and repeatable state](https://robots.thoughtbot.com/remove-duplication-with-factorygirls-traits), this enables us to change the definition of what a status means to the system centrally, without having to update every test. You may be able to lose both the `expires_on` and `publish_on` field as a result throughout this file.

app/controllers/hiring_staff/vacancies/copy_controller.rb
```diff
@@ -0,0 +1,25 @@
+class HiringStaff::Vacancies::CopyController < HiringStaff::Vacancies::ApplicationController
+  def create
+    vacancy = Vacancy.find(vacancy_id)
+    vacancy_copy = vacancy.dup
```
tahb: Good idea to use dup.

app/controllers/hiring_staff/vacancies/copy_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    update_fields(vacancy_copy)
+    if vacancy_copy.save
+      redirect_to review_path(vacancy_copy)
```
tahb: Before we hand over to the review and existing publish code, I think it would be sensible to start collecting audit data here as soon as its shipped.

`Auditor::Audit.new(vacancy_copy, 'vacancy.copy', current_session_id).log`

This will at least make this feature deployable while we work out how we're going to collect this going forward, and allow us to backfill data reliably.

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
lozette: These selectors are already used in another feature test: https://github.com/dxw/teacher-vacancy-service/blob/develop/spec/features/hiring_staff_can_see_their_vacancies_spec.rb

If they should be removed in this PR then the other feature test will probably need refactoring as well.

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
lozette: They technically can't get here but I was worried there was no test which exposes the error message otherwise! I guess I can put it in a unit test and check the flash.

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
lozette: I think I used non-FactoryBot here as most of the existing feature tests don't use it and I assumed it was the style 😳 🤷‍♀️   Probably worth logging a refactor ticket to get them all using the same structure.

app/views/hiring_staff/schools/_vacancy.html.haml
```diff
@@ -3,6 +3,5 @@
   %td.govuk-table__cell= format_date(vacancy.publish_on)
   %td.govuk-table__cell= format_date(vacancy.expires_on)
   %td.govuk-table__cell= vacancy.status.titleize
-  %td.govuk-table__cell= link_to 'Edit', edit_school_job_path(vacancy.id), class: 'govuk-link'
-  %td.govuk-table__cell= link_to 'Delete', school_job_path(id: vacancy.id), class: 'govuk-link', method: :delete, data: { confirm: "Are you sure you want to delete the '#{vacancy.job_title}' job?" }
-
+  %td.govuk-table__cell= link_to I18n.t('jobs.edit_link'), edit_school_job_path(vacancy.id), class: 'govuk-link'
```
tahb: Much less fragile 👏

PR #697, opened 2019-01-21
Production release #60

tahb:
- remove explanation of service rename from blue service banner
- allow whitespaces to be added either side of a salary field when creating a new job listing without causing an error
- minor automatic gem update to ipaddr 
- the temporary fix for the service supporting sign-ins from `www.` becomes permanent in the infrastructure
- fix worker containers running out of memory and being constantly restarted every few minutes
- fix the job that records searches made from intermittently running out of threads


tahb:
Around this time, or because of this pull request our staging environment has been raising a steady stream of `Redis::CannotConnectError`. After an investigation where we have gone as far as to replace and split the Redis instances we cannot identify why this is, perhaps a coincidence in timing. This code does not cause the same affect on the identical edge environment after 2 days.

PR #651, opened 2018-12-10
Hiring staff UI improvements

lozette:
## Trello card URL:

This PR covers a few cards:
https://trello.com/c/iWZh1bSZ/653-hiring-staff-ui-improvements (@gazaston)
https://trello.com/c/FJE5XW0Y/646-create-new-job-status-draft (@lozette)
https://trello.com/c/UwuesOnr/645-create-new-job-status-pending (@lozette)
https://trello.com/c/YFLoKl9S/647-create-new-job-status-expired (@lozette)

## Changes in this PR:

[Note that this PR does not include adding page view data to the expired and published jobs tabs, or republishing an expired job]

Improve the hiring staff view of their jobs, with tabs to split the jobs into Published | Draft | Pending | Expired. 

For non-JS users the tab contents will be stacked on top of each other.

Associated back-end work, creating new scopes for expired and pending vacancies. 

## Screenshots of UI changes:

screenshot 2018-12-10 at 12 39 30

screenshot 2018-12-10 at 12 39 39

screenshot 2018-12-10 at 12 39 44

Non-JS view

screenshot 2018-12-10 at 12 41 38

PR #698, opened 2019-01-22
Feature/686 new job statuses

pezholio:
## Trello card URL:

https://trello.com/c/JXmAScnE/686-hiring-staff-ui-replace-design-system-component-with-rails-pages-to-fix-bug

## Changes in this PR:

This is a reroll of #651, with all the changes from that PR (which were rolled back in #668), but a few tweaks, so that vacancies are loaded on their own separate pages, rather than loading them all and relying on JS to handle the tabs.

### Before

image

#### Desktop

image

#### Small screen

image


pezholio:
[Hiring staff UI: replace design system component with Rails pages to fix bug](https://trello.com/c/JXmAScnE/686-hiring-staff-ui-replace-design-system-component-with-rails-pages-to-fix-bug)

app/views/hiring_staff/schools/_vacancies_expired.html.haml
```diff
@@ -2,9 +2,9 @@
...
-            %th.govuk-table__header= t('jobs.expires_on')
-            %th.govuk-table__header{ colspan: 2}= t('jobs.actions')
+            %th.govuk-table__header= t('jobs.date_to_be_posted')
+            %th.govuk-table__header= t('jobs.expired_on')
+            %th.govuk-table__header= t('jobs.actions')
     %tbody.govuk-table__body
         - vacancies.each do |vacancy|
-            = render partial: 'vacancy_published', locals: { vacancy: vacancy, school: school }
\ No newline at end of file
+            = render partial: 'vacancy_expired', locals: { vacancy: vacancy, school: school }
```
tahb: 👍

app/controllers/hiring_staff/schools_controller.rb
```diff
@@ -32,4 +32,21 @@ def update
   def session_has_multiple_schools?
     session.key?(:multiple_schools) && session[:multiple_schools] == true
   end
+
+  def set_vacancies
```
tahb: I remember you mentioning that you thought the use of a switch statement here wasn't the most elegant. 

Now we've got this in a good place perhaps a way to refactor this would be to introduce a small and testable presenter object that would move this logic away from the controller.

```
@school_vacancy_presenter = SchoolVacancyPresent.new(type: params[:type])
@school_vacancy_presenter.draft
@school_vacancy_presenter.pending
@school_vacancy_presenter.expired
@school_vacancy_presenter.live
@school_vacancy_presenter.current_type
```

`SchoolVacancyPresenter` is a bit long, but we already have a `VacancyPresenter` for the job seekers!

app/views/layouts/application.html.haml
```diff
@@ -26,7 +26,7 @@
     = javascript_include_tag 'application'
     %meta{content: asset_path("govuk-opengraph-image.png"), property: "og:image"}/
     = csrf_meta_tags
-  %body.govuk-template__body.app-body-class
+  %body{class: "govuk-template__body app-body-class #{authenticated? ? 'hiring-staff': '' } #{controller_name + '_' + action_name}"}
```
tahb: We could move this logic into a tested helper method to keep the view simpler and to help us catch any regressions.

tahb:
Noticed that on the after screenshot there isn't enough padding between the table and it's header.

pezholio:
Fixed the spacing issue

image

app/views/hiring_staff/schools/_vacancies.html.haml
```diff
@@ -2,13 +2,8 @@
   %h2.govuk-tabs__title
     = t('aria_labels.content')
   %ul.govuk-tabs__list.tab-list
-    - [:published, :pending, :draft, :expired].each do |t|
+    - SchoolVacanciesPresenter.valid_types.each do |t|
```
tahb: Oh this is nice 👏

spec/presenters/school_vacancies_presenter_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+  let!(:live_vacancies) { create_list(:vacancy, 7, :published, school: school) }
+
+  describe 'initialize' do
+    it 'raises an error if the type is not recognised' do
+      expect { SchoolVacanciesPresenter.new(school, 'something') }.to raise_error(ArgumentError)
+    end
+  end
+
+  it 'returns the vacancy type' do
```
tahb: The convention of wrapping methods in a `describe` block is useful as it makes it easier for devs as a grouping method so we can find all the relevant tests to a method and we'll also be able to add 'contexts' and 'before' steps accurately and only affect the relevant tests. Without this, it is all too easy for methods to get added out of order or moved around in other refactoring.

If we start off the file with describe blocks even when we don't obviously need them right now we set the precedent that the next contributor can follow to ensure the file grows in a more manageable way. We remove the need for them to question "why was this done and should I refactor this first?" which I see very much as value.

app/views/hiring_staff/schools/_vacancies_draft.html.haml
```diff
@@ -1,9 +1,15 @@
...
+            %thead.govuk-table__head
+                %tr.govuk-table__row
+                    %th.govuk-table__header= t('jobs.job_title')
+                    %th.govuk-table__header= t('jobs.date_drafted')
+                    %th.govuk-table__header{ colspan: 2}= t('jobs.actions')
+            %tbody.govuk-table__body
+                - presenter.draft.each do |vacancy|
+                    = render partial: 'vacancy_draft', locals: { vacancy: vacancy, school: presenter.school }
+    - else
+        %p= t("jobs.no_draft_jobs")
```
tahb: Good spot!

PR #700, opened 2019-01-24
[715] Stop requests being sent to google to index our listing pages …

tahb:
…on non-production environments

## Trello card URL:
https://trello.com/c/HN9oqZY8

## Changes in this PR:
* When testing these methods they were all grouped under a `private` group right at the top, we’re unable to test these methods at the unit level with this as is.
* On reflection, these ‘application controller’ methods seem to be grouped in this file for reuse to the inherent controllers for which Rails makes them available. We can see this as the methods we are testing are called directly from controllers without issue. I’m not sure what private was protecting us from. here so i’ve removed it, I’d welcome any other guesses for why might have needed it!


spec/features/hiring_staff_can_delete_vacancies_spec.rb
```diff
@@ -48,9 +48,9 @@
 
   scenario 'Notifies the Google index service' do
     vacancy = create(:vacancy, school: school)
-    vacancy_url = job_url(vacancy, protocol: 'https')
 
-    expect(RemoveGoogleIndexQueueJob).to receive(:perform_later).with(vacancy_url)
+    expect_any_instance_of(HiringStaff::Vacancies::ApplicationController)
```
pezholio: Yep, I like this. Stub the controller method, rather than the job, especially given we're in test here

spec/controllers/hiring_staff/vacancies/application_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+RSpec.describe HiringStaff::Vacancies::ApplicationController, type: :controller do
+  describe '#update_google_index' do
+    let!(:vacancy) { create(:vacancy) }
+
+    context 'when in production' do
+      before(:each) do
+        allow(Rails).to receive(:env)
+          .and_return(ActiveSupport::StringInquirer.new('production'))
```
pezholio: This won't have any dangerous side effects will it? I'm just wondering if it'd be safer to have a controller method called something like `in_production?` and stub that instead, rather than stubbing Rails internals

tahb: I share the concern. 

My thinking is that in this scenario I feel reasonably confident that we're safe in regards to side effects since the method, and code that is executed, is a single task: put a job on a queue. 

The scenario where I see this being dangerous is if it was a big hairy method with lots of responsibilities including doing things like sending emails inline or perhaps a feature spec where a user is clicking around testing lots of parts of the application. In that case I couldn't honestly say there wouldn't be nasty unexpected consequences.

I suppose that the risk to this code is that _in the future_ the method may change to become complex rather than sticking to the principle single responsibility.

Also I chose to copy the approach from others tests for our basic auth that have to date served us well: https://github.com/dxw/teacher-vacancy-service/blob/e6eeb744335cbf9c42588c3b8d11b8b895f3f841/spec/controllers/application_controller_spec.rb

PR #701, opened 2019-01-24
[716] Add a new Redis instance to address connection errors

tahb:
- We've been struggling with a bug recently where Redis is seemingly randomly raising connection errors only on Staging
- After a lengthy investigation we've discovered that our configuration
is not following the current advice from Sidekiq to use different
instances Redis for different tasks of caching and queuing [2]. Before
this change we are using a single rdis instance for both caching
geocoding requests and asynchronous jobs.
- We believe that our problem could be explained by a problem with the
instance including latency issues and we want to remove having
everything on the same instance from the investigation.
- Splitting it out also allows us the AWS console to distinguish how many jobs are
in the redis instance compared to cache records, as at the moment it's
just an ever increasing graph as the cache grows
- This change will remove the existing redis instance in order to add
explicit names of 'redis_cache' and 'redis_queue' in the hope that they
will not be confused in the future. The impact of this is that we lose
the outcome of any failing jobs at the time of change, and our
geolocation cache has to be started again. This should be of minimal
impact.
- This change is an important consideration before we ship the work of
job alert nofitications which will be a critical feature using our
queue infrastructure. In the future this change will be much harder as
the impact will be considerably higher
- Both redis instance types can be independently configured/changed so
we can react to problems more easily in the future and scale the job
queue if we need to without impacting the cache redis. This will also
allow us to scale as part of our overall investigation of redis
connection issues if it is a network latency or AWS issue

[1] https://rollbar.com/dxw/teacher-vacancies/items/276/
[2] https://github.com/mperham/sidekiq/wiki/Using-Redis#multiple-redis-instances

Pull Request template

- Fill in or remove sections as necessary
- Delete this section before submitting your PR

## Next steps:


tahb:
[Caching and job queues should be using different Redis databases - fix for work being done to investigate redis errors on staging](https://trello.com/c/beMHMjap/716-caching-and-job-queues-should-be-using-different-redis-databases-fix-for-work-being-done-to-investigate-redis-errors-on-staging)

tahb:
I've tested this on the `testing` environment and made a change as a result. With removing the original redis instance in the same change, there are lots of redis connection errors that are thrown during the deployment of this change. Which ever deployment goes first Git or Terraform, the app will be unable to establish a connection until the new redis instance is in place.

To fix this, I'm leaving the original redis instance in place so that any containers still referencing the old REDIS_URL can succeed until they're updated with the new REDIS_CACHE_URL and REDIS_QUEUE_URL. Another benefit of this is that we can manually check the jobs left in this redis instance to make sure there is no unprocessed jobs there before we switch it off. 

We will need a separate change to clear up this old redis instance later.

pezholio:
Yep, that seems sensible 👍

PR #727, opened 2019-02-12
Send daily email alerts to subscribers

pezholio:
## Trello card URL:

https://trello.com/c/tu3w9BnL/668-43-email-alerts-senddailyemailalertjob-rake-task

## Changes in this PR:

This adds a mailer, job and Rake task to send daily job alerts to users. The job gets all ongoing subscriptions, checks if there are any posted vacancies in the last day for a subscription's criteria, and queues up a mailer if there is.

## Screenshots of UI changes:

image

## Next steps:


app/jobs/send_daily_alert_email_job.rb
```diff
@@ -0,0 +1,10 @@
+class SendDailyAlertEmailJob < ApplicationJob
+  def perform
+    Subscription.ongoing.each do |s|
+      vacancies = s.vacancies_for_range(Time.zone.yesterday, Time.zone.today)
+      next if vacancies.blank?
+
+      AlertMailer.daily_alert(s.id, vacancies.pluck(:id)).deliver_later
```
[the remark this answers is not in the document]
pezholio: So, after our chat about this, I've limited the number of vacancies to 500, and added a log message to tell us how many alerts are being returned, so we can keep an eye on the sorts of alerts that are being sent. I've confirmed that both Sidekiq and Notify can cope with 500 results being returned, so I think we can go with this for now 👍

app/jobs/send_daily_alert_email_job.rb
```diff
@@ -0,0 +1,10 @@
+class SendDailyAlertEmailJob < ApplicationJob
+  def perform
```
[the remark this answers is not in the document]
pezholio: I've had a stab at this now, I'm using ActiveJob callbacks (which are amazing) to create runs, and also check the age of a job.

tahb:
[Fix race condition where alert run does not exist when an alert is created](https://trello.com/c/4C5yFIYv/751-fix-race-condition-where-alert-run-does-not-exist-when-an-alert-is-created)

app/jobs/daily_alert_mailer_job.rb
```diff
@@ -0,0 +1,37 @@
+class DailyAlertMailerJob < ActionMailer::DeliveryJob
+  EXPIRES_IN = 4.hours
+
+  before_enqueue do
```
tahb: Difficult to spot the diff for the `after_queue` to `before_queue` fix, do we have test coverage for this case that will catch regressions?

pezholio: Good call. I rushed it a bit because it was breaking in testing. I'll see what I can do to catch the regressions

pezholio: I've added a test here that checks everything gets run in order https://github.com/dxw/teacher-vacancy-service/blob/cdeba62b9194fac725c9f8d773de45f40097593b/spec/jobs/daily_alert_mailer_job_spec.rb#L26-L33

tahb: As discussed, we discovered that this meant the `job_id` was never being set. We thought it would work if it went into the `after_enqueue` block.

pezholio: Done this now. I've also added another spec the checks the job ID gets checked after the job is run:

https://github.com/dxw/teacher-vacancy-service/blob/c569fa9d670beae2e1c6745407b602b42d9e95dd/spec/jobs/daily_alert_mailer_job_spec.rb#L38-L47

config/initializers/default_url_options.rb
```diff
@@ -0,0 +1,13 @@
+default_domains = {
```
tahb: This change is great 👍 Clear to understand how it works across all environments and easy to make changes with confidence whilst keeping high confidence that the fallback hosts will still work.

PR #731, opened 2019-02-13
[727] Hiring staff can now copy expired vacancies

robbiepaul:
## Trello card URL:

https://trello.com/c/m2eHnBMQ/727-hiring-staff-ui-user-can-make-a-copy-of-an-existing-job-on-the-expired-jobs-page-front-end

## Changes in this PR:

- Amended the `expired` vacancy factory to allow saving without validation
- Added 'Copy' button to the expired vacancies view
- Added new context for copying expired vacancies to feature spec
- Prevent dates from being set on expired vacancies

### Before

screenshot 2019-02-13 at 16 58 53

### After

screenshot 2019-02-13 at 16 58 11

### DIFF
--- a/app/form_models/copy_vacancy_form.rb
+++ b/app/form_models/copy_vacancy_form.rb
@@ -12,7 +12,7 @@ class CopyVacancyForm < VacancyForm
     self.job_title = vacancy.job_title
     self.starts_on = vacancy.starts_on
     self.ends_on = vacancy.ends_on
-    self.expires_on = vacancy.expires_on
+    self.expires_on = vacancy.expires_on.future? ? vacancy.expires_on : nil
     self.publish_on = vacancy.publish_on
   end
 
--- a/app/form_models/copy_vacancy_form.rb
+++ b/app/form_models/copy_vacancy_form.rb
@@ -10,14 +10,20 @@ class CopyVacancyForm < VacancyForm
   def initialize(vacancy:)
     self.vacancy = vacancy
     self.job_title = vacancy.job_title
-    self.starts_on = vacancy.starts_on
-    self.ends_on = vacancy.ends_on
-    self.expires_on = vacancy.expires_on.future? ? vacancy.expires_on : nil
-    self.publish_on = vacancy.publish_on
+    reset_date_fields if vacancy.expires_on.past?
   end
 
   def apply_changes!(params = {})
     vacancy.assign_attributes(params)
     vacancy
+  end
+
+  private
+
+  def reset_date_fields
+    self.starts_on  = nil
+    self.ends_on    = nil
+    self.expires_on = nil
+    self.publish_on = nil
   end
 end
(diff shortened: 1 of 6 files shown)


robbiepaul:
[Hiring staff UI: user can make a copy of an existing job on the 'expired jobs' page (front end)](https://trello.com/c/m2eHnBMQ/727-hiring-staff-ui-user-can-make-a-copy-of-an-existing-job-on-the-expired-jobs-page-front-end)

spec/factories/vacancies.rb
```diff
@@ -78,6 +78,7 @@
     end
 
     trait :expired do
+      to_create { |instance| instance.save(validate: false) }
```
tahb: Nice, I didn't know about this.

robbiepaul:
Yeah sure, I had the same concern yesterday and asked on the card: https://trello.com/c/m2eHnBMQ/727-hiring-staff-ui-user-can-make-a-copy-of-an-existing-job-on-the-expired-jobs-page-front-end#comment-5c644532036f2a878b54ea66

But I think you're right it will probably provide a better user experience not having to hit the validation error if we can

robbiepaul:
@tahb Added a commit that doesn't copy the expiry date on expired vacancies

app/form_models/copy_vacancy_form.rb
```diff
@@ -12,7 +12,7 @@ def initialize(vacancy:)
     self.job_title = vacancy.job_title
     self.starts_on = vacancy.starts_on
     self.ends_on = vacancy.ends_on
-    self.expires_on = vacancy.expires_on
+    self.expires_on = vacancy.expires_on.future? ? vacancy.expires_on : nil
```
tahb: I'm wondering if it would be simpler not to set any date fields if the vacancy is expired. The vacancy could have expired 1 day ago or more than a year ago so if we try to be helpful in pre populating the dates that _would_ pass validation I worry we'll have to keep a copy of the all the date validation logic here to do it right. 

For example the `starts_on` must be after the `ends_on` if present, and both must be after the `publish_on`.

Would it be simpler and safer to say if a listing has expired then we can't trust any of its dates so instead we ask the user fill out all the dates again, only 2 of which are mandatory.

Perhaps:

```
  def initialize(vacancy:)
    self.vacancy = vacancy
    self.job_title = vacancy.job_title
    prepopulate_date_fields
  end

  def prepopulate_date_fields
    return if vacancy.expires_on.past?

    self.starts_on = vacancy.starts_on
    self.ends_on = vacancy.ends_on
    self.expires_on = vacancy.expires_on
    self.publish_on = vacancy.publish_on
  end
```

robbiepaul: I [added a commit](https://github.com/dxw/teacher-vacancy-service/pull/731/commits/b719b8660b61e09a51dcf540ddabf004a54d9d42) that explicitly sets the dates to `nil` if the vacancy has expired.

I couldn't do your suggestion as the date fields get delegated the the vacancy instance.

tahb: Of course 🤦‍♂️

PR #736, opened 2019-02-18
[725] Caching and autoscaling to better support higher load

tahb:
## Trello card URL:
https://trello.com/c/ecGy8swS

## Changes in this PR:

Before this change a conceivable amount of concurrent requests to the home page starts to take 5-8 seconds, with searches and API requests introducing 503 responses. This has has become an issue as we have many more job listings on the service, a fair bit more traffic (with much more expected soon) arriving compared to 6 months ago and not having enough caching in place.

The investigation on a recent incident with one of our API endpoints has highlighted this situation which has led us here to review the general services ability to handle load.

On Edge (which has parity with Production) I have tested the request times and AWS resources being used in the following scenarios:

1. normal usage ✅
2. sudden load increase ✅
3. persistent load increase ✅
4. persistent load increase during a deployment ✅

My strategy is to _generously_ scale up the web services quickly and scale back down slowly. I understand that to start with we might be spinning up more instances and containers than we need if the thresholds are slightly out but I believe it's better to have more than we need and tweak it down later than to have not enough, which is the current problem.

* requests to the root path are cached by 15 minutes
* requests to the `/jobs` path including searches and sorting are cached by `60` seconds
* increase the number of web containers and ec2 instances autoscaling can utilise, and lower the thresholds to make them more responsive (lots more info on the commits themselves)
* deployments only use 100% of the desired count, ensuring if the desired count changes to 20/20 then it will not use more than 20 containers for the deployment, which is the maximum 6 medium instances can support, ensuring deployments can succeed.
* set logspout to daemon as it wasn’t always going onto autoscaled instances, sitting at 2/6 which won’t give us logs.

## Next steps:

### DIFF
--- a/terraform/modules/cloudfront/cloudfront.tf
+++ b/terraform/modules/cloudfront/cloudfront.tf
@@ -122,6 +122,29 @@ resource "aws_cloudfront_distribution" "default" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+  cache_behavior {
+    allowed_methods  = ["GET", "HEAD"]
+    cached_methods   = ["GET", "HEAD"]
+    target_origin_id = "${var.project_name}-${var.environment}-default-origin"
+
+    path_pattern = "/"
+
+    forwarded_values = {
+      query_string = false
+      headers      = ["Host", "Authorization"]
+
+      cookies {
+        forward = "none"
+      }
+    }
+
+    min_ttl     = 300 # 5 minutes
+    default_ttl = 900 # 15 minutes
+    max_ttl     = 3600 # 60 minutes
+
+    viewer_protocol_policy = "redirect-to-https"
+  }
+
   price_class = "PriceClass_100"
 
   restrictions {
--- a/terraform/modules/cloudfront/cloudfront.tf
+++ b/terraform/modules/cloudfront/cloudfront.tf
@@ -108,7 +108,7 @@ resource "aws_cloudfront_distribution" "default" {
 
     forwarded_values = {
       query_string = true
-      headers      = ["Host"]
+      headers      = ["Host", "Authorization"]
 
       cookies {
         forward = "none"
--- a/terraform/modules/cloudfront/cloudfront.tf
+++ b/terraform/modules/cloudfront/cloudfront.tf
@@ -145,6 +145,29 @@ resource "aws_cloudfront_distribution" "default" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+  cache_behavior {
+    allowed_methods  = ["GET", "HEAD"]
+    cached_methods   = ["GET", "HEAD"]
+    target_origin_id = "${var.project_name}-${var.environment}-default-origin"
+
+    path_pattern = "/jobs"
+
+    forwarded_values = {
+      query_string = true
+      headers      = ["Host", "Authorization"]
+
+      cookies {
+        forward = "none"
+      }
+    }
+
+    min_ttl     = 60 # 1 minute
+    default_ttl = 60 # 1 minute
+    max_ttl     = 300 # 5 minutes
+
+    viewer_protocol_policy = "redirect-to-https"
+  }
+
   price_class = "PriceClass_100"
 
   restrictions {
--- a/terraform/modules/core/core.tf
+++ b/terraform/modules/core/core.tf
@@ -398,14 +398,14 @@ resource "aws_autoscaling_policy" "ecs-autoscaling-down-policy" {
 }
 
 resource "aws_cloudwatch_metric_alarm" "average-reserved-cpu-high" {
-  alarm_name          = "${var.project_name}-${var.environment}-average-reserved-cpu-high"
+  alarm_name          = "${var.project_name}-${var.environment}-cluster-average-reserved-cpu-high"
   comparison_operator = "GreaterThanOrEqualToThreshold"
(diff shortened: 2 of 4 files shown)


terraform/modules/cloudfront/cloudfront.tf
```diff
@@ -108,7 +108,7 @@ resource "aws_cloudfront_distribution" "default" {
 
     forwarded_values = {
       query_string = true
-      headers      = ["Host"]
+      headers      = ["Host", "Authorization"]
```
tahb: This makes it easier to test on non-prod environments, and shouldn't make a different to prod as no auth headers are used.

terraform/modules/cloudfront/cloudfront.tf
```diff
@@ -122,6 +122,29 @@ resource "aws_cloudfront_distribution" "default" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+  cache_behavior {
+    allowed_methods  = ["GET", "HEAD"]
+    cached_methods   = ["GET", "HEAD"]
+    target_origin_id = "${var.project_name}-${var.environment}-default-origin"
+
+    path_pattern = "/"
```
tahb: `/` is the root path for the service where every user arrives too. This renders the same content as a request to `/jobs`, which is used when a user searches or sorts the content. For sorting and searching we want up to date data so this behaviours timings can be higher for root, I propose.

terraform/modules/cloudfront/cloudfront.tf
```diff
@@ -122,6 +122,52 @@ resource "aws_cloudfront_distribution" "default" {
...
+      query_string = true
+      headers      = ["Host", "Authorization"]
+
+      cookies {
+        forward = "none"
+      }
+    }
+
+    min_ttl     = 60 # 1 minute
+    default_ttl = 60 # 1 minute
```
tahb: This will stop a proportion of requests making it to our origin so Rails won't be able to record searches that are cached. I propose that the DOS reliability is worth the trade for losing a smaller percentage of audits for search data.

terraform/modules/core/core.tf
```diff
@@ -398,14 +398,14 @@ resource "aws_autoscaling_policy" "ecs-autoscaling-down-policy" {
...
-  alarm_name          = "${var.project_name}-${var.environment}-average-reserved-cpu-high"
+  alarm_name          = "${var.project_name}-${var.environment}-cluster-average-reserved-cpu-high"
   comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "2"
+  evaluation_periods  = "1"
   metric_name         = "CPUReservation"
   namespace           = "AWS/ECS"
   period              = "60"
   statistic           = "Average"
-  threshold           = "70"
```
tahb: This was never getting triggered at 70, I tested base resting and loaded usage of the CPU on edge and came up with 40 as a better average threshold.

terraform/modules/core/core.tf
```diff
@@ -398,14 +398,14 @@ resource "aws_autoscaling_policy" "ecs-autoscaling-down-policy" {
 }
 
 resource "aws_cloudwatch_metric_alarm" "average-reserved-cpu-high" {
-  alarm_name          = "${var.project_name}-${var.environment}-average-reserved-cpu-high"
+  alarm_name          = "${var.project_name}-${var.environment}-cluster-average-reserved-cpu-high"
   comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "2"
+  evaluation_periods  = "1"
```
tahb: Let's wait less time before scaling up instances. If EC2 instances have more than 40% CPU reserved for 60 seconds, we infer that ECS is adding new containers to respond to load and we should now add more EC2 instances to allow ECS to continue. We should scale up quickly, and scale down slowly.

terraform/modules/core/core.tf
```diff
@@ -453,15 +453,20 @@ resource "aws_autoscaling_group" "ecs-autoscaling-group" {
   ]
 
   depends_on = ["aws_vpc.vpc", "aws_launch_configuration.ecs-launch-configuration", "aws_security_group.default", "aws_security_group.ecs"]
+
+
+  lifecycle {
+    ignore_changes = ["desired_capacity"]
```
tahb: Ensure every time we deploy, we don't set the EC2 instance count back down to what's configured and change it from what the infrastructure load has calculated. This would undermine the autoscaling and make us more likely to serve 503.

terraform/modules/core/core.tf
```diff
@@ -506,14 +511,34 @@ resource "aws_appautoscaling_policy" "down" {
...
+    ClusterName = "${var.ecs_cluster_name}"
+    ServiceName = "${var.ecs_service_web_name}"
+  }
+
+  alarm_actions = ["${aws_appautoscaling_policy.up.arn}"]
+  ok_actions    = ["${aws_appautoscaling_policy.down.arn}"]
+}
+
+/* metric used for auto scale */
+resource "aws_cloudwatch_metric_alarm" "service_memory_high" {
```
tahb: Memory is actually quite high (90%) on production while CPU remains low, this should also cause containers to be added

tahb:

tahb:
bob has given me lightning feedback via Slack and I'm going to close this pull request and open 2 new ones instead with some iterations.

PR #739, opened 2019-02-18
Feature flag job alerts

pezholio:
## Trello card URL:

https://trello.com/c/zxL7s5Vn/738-feature-flag-job-alerts-in-frontend

## Changes in this PR:

This adds a feature flag to enable / disable job alerts. This gives us confidence that if something goes wrong when we go live, then we can easily turn alerts off, allowing us to deal with the problem without the added complication of more users signing up while we're trying to fix the problem. The flag is set by setting the env var `FEATURE_EMAIL_ALERTS` to either `true` or `false`

## Next steps:


app/controllers/subscriptions_controller.rb
```diff
@@ -46,4 +48,8 @@ def daily_subscription_params
   def search_criteria
     params.require(:search_criteria).permit(*VacancyFilters::AVAILABLE_FILTERS)
   end
+
+  def check_feature_flag
+    raise(ActionController::RoutingError, 'Not Found') unless EmailAlertsFeature.enabled?
```
tahb: We have a `not_found` method in ApplicationController you might be able to use to raise a 404 consistently, and with styling.

pezholio: Ah, nice, I did look for something like this, but only searching the codebase for `404`, should've checked the ApplicationController

PR #741, opened 2019-02-19
[729] Total job page views are calculated without Google Analytics

robbiepaul:
## Trello card URL:

https://trello.com/c/ZfIhptK0/729-total-job-page-views-are-calculated-without-google-analytics

## Changes in this PR:

- Adds [Redis Objects gem](https://github.com/nateware/redis-objects) to allow for atomic counters
- Page views are now tracked using a Redis counter
- Those counters are then persisted once a day with a rake task at 2am

## Next steps:


robbiepaul:
[Total job page views are calculated without Google Analytics](https://trello.com/c/ZfIhptK0/729-total-job-page-views-are-calculated-without-google-analytics)

config/initializers/redis.rb
```diff
@@ -0,0 +1,7 @@
+require 'connection_pool'
+
+redis_url = "#{ENV['REDIS_QUEUE_URL']}/0"
```
pezholio: This probably needs to go into the other Redis instance (at `ENV['REDIS_CACHE_URL]`). I'd probably name the initialiser something like `redis_objects.rb`, as we're using Redis in a bunch of places

app/services/vacancy_page_view.rb
```diff
@@ -0,0 +1,24 @@
...
+  def initialize(vacancy)
+    @vacancy = vacancy
+  end
+
+  def track
+    @vacancy.page_view_counter.increment
+  end
+
+  def persist!
+    @vacancy.total_pageviews = if @vacancy.total_pageviews.nil?
```
pezholio: We can probably just do `@vacancy.total_pageviews.to_i + @vacancy.page_view_counter.to_i` here I would have thought? As `nil.to_i` will just be `0`

app/controllers/vacancies_controller.rb
```diff
@@ -25,6 +25,7 @@ def show
     vacancy = Vacancy.listed.friendly.find(id)
     return redirect_to(job_path(vacancy), status: :moved_permanently) if old_vacancy_path?(vacancy)
     @vacancy = VacancyPresenter.new(vacancy)
+    TrackVacancyPageViewJob.perform_later(vacancy.id) unless authenticated?
```
pezholio: I like this. This is also something that I imagine we're not doing with GA, as it's tricky for GA to know when someone is authenticated

variables.tf
```diff
@@ -201,7 +201,7 @@ variable "vacancies_pageviews_refresh_cache_task_command" {
 
 variable "vacancies_pageviews_refresh_cache_task_schedule" {
   description = "vacancies_pageviews_refresh_cache schedule expression - https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
-  default     = "cron(0 08 * * ? *)"
+  default     = "cron(0 02 * * ? *)"
```
pezholio: Any reason why this is done at 2am, and not at midnight? I guess it won't matter too much if our daily counts are slightly off for a couple of hours, as I doubt (hope) very much no one is looking at vacancy stats in the wee small hours!

config/initializers/redis.rb
robbiepaul: Good spot, I've changed it to the Redis cache and renamed the initialiser

app/services/vacancy_page_view.rb
robbiepaul: Thanks for pointing that out, it was doing my head in. I was trying to do things like `@vacancy.total_pageviews.to_i += @vacancy.page_view_counter.to_i` before 😅

variables.tf
robbiepaul: We just thought it probably best to run this potentially intensive task at a time when less people are on the service. No real reason for 2am over 12am other than I supposed there will be even less people on at 2am

pezholio: I think 2am is fine in that case

app/controllers/vacancies_controller.rb
tahb: Super 👍 

It's not on the card (forgot) so I don't suggest we do it here but what we do get easily with GA is filtering known IP addresses out such as both offices. I imagine when we get to it we could easily extend this logic with the use of environment variables to hold some known IP addresses to check against the request IP, this would further improve the accuracy of this data and confident that it's real job seekers.

config/initializers/redis.rb
tahb: I wonder if there is advantage to be had by further separating this type of caching by using a different database eg. `/1`

Thinking aloud:

- we'd be able to connect to each redis database separately to interrogate the contents for debugging, we'd also be able to flush and delete keys with less risk to dumping the geolocation cache
- there are likely performance advantages for reads in redis when it doesn't have to go through +20k geolocation records to find the page view ones, and vice versa!

pezholio: I did wonder this. Worth trying in dev I reckon, but I can't see why it wouldn't work

pezholio: Or do we want to have a more sensible naming convention for the Redis databases? So `/geolocation` and `/pageviews` for example

tahb: Can we do meaningful names? I assumed from past examples and memory it was only integers!

pezholio: No, you're right, it is only integers. I guess we could refer to them by name internally, but I dunno if that would be that valuable

robbiepaul:
I've changed the Redis Objects initialiser to use a separate Redis database.

I also added a commit that saves the time that the page views were updated. We were doing this before and I thought it would be useful information to continue saving.

Ready for another look.

PR #743, opened 2019-02-20
Error messages from job alert creations should disappear if browser back used

pezholio:
## Trello card URL:

https://trello.com/c/F3G2fZFY/747-error-messages-from-job-alert-creations-should-disappear-if-browser-back-used

## Changes in this PR:

* Changes setting the flash message from `flash[:error]` to `flash.now[:error]`, which means the flash message is not persisted
* Removes the call to `flash.clear` when creating a subscription, as this is no longer needed

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -10,12 +10,11 @@ class SubscriptionsController < ApplicationController
   end
 
   def create
-    flash.clear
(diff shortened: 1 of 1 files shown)


tahb:
[Error messages from job alert creations should disappear if browser back used](https://trello.com/c/F3G2fZFY/747-error-messages-from-job-alert-creations-should-disappear-if-browser-back-used)

PR #746, opened 2019-02-21
(chore) Allow Postgres to auto update patch versions without conflicting with use of Terraform

tahb:
* This has already been applied to all environments, seemingly automatically by AWS and without this change our Terraform deployments will fail as we try to instruct it to go backwards.

screenshot 2019-02-21 at 10 51 56

* We can omit the patch version and set `9.6` to avoid this problem happening in future

### DIFF
--- a/terraform/modules/rds/rds.tf
+++ b/terraform/modules/rds/rds.tf
@@ -17,6 +17,7 @@ resource "aws_db_instance" "default" {
   backup_retention_period   = 14
   maintenance_window        = "Sun:00:00-Sun:03:00"
   multi_az                  = "${terraform.workspace == "production" ? "true" : "false"}"
+  auto_minor_version_upgrade = true
 
   tags {
     Name        = "${var.project_name}-${var.environment}"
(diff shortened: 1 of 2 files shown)


pezholio:
Ah, I noticed that. I assumed this was something we'd tried at some point in test. I guess it makes sense for Amazon to do minor upgrades automatically for security updates etc. Is there something we can do in Terraform to make the version automatically calculated? Or only specify the major version or something?

tahb:
@pezholio It appears it does allow that which should stop this coming up in future!

tahb:
@pezholio ready for another look 👍

terraform/modules/rds/rds.tf
```diff
@@ -17,6 +17,7 @@ resource "aws_db_instance" "default" {
   backup_retention_period   = 14
   maintenance_window        = "Sun:00:00-Sun:03:00"
   multi_az                  = "${terraform.workspace == "production" ? "true" : "false"}"
+  auto_minor_version_upgrade = true
```
pezholio: Boom, awesome!

PR #747, opened 2019-02-21
[725] Increase auto scaling

tahb:
## Trello card URL:
https://trello.com/c/ecGy8swS

## Changes in this PR:

Before this change a conceivable amount of concurrent requests to the home page starts to take 5-8 seconds, with searches and API requests introducing 503 responses. This has has become an issue as we have many more job listings on the service, a fair bit more traffic (with much more expected soon) arriving compared to 6 months ago and not having enough caching in place.

On Edge (which has parity with Production) I have tested the request times and AWS resources being used in the following scenarios:

1. normal usage ✅
2. sudden load increase ✅
3. persistent load increase ✅
4. persistent load increase during a deployment ✅

My strategy is to be more generous in scaling up the web services quickly and scale back down slowly. I understand that to start with we might be spinning up more instances and containers than we need if the thresholds are slightly out but I believe it's better to have more than we need and tweak it down later than to have not enough, which is the current problem.

## Next steps:

### DIFF
--- a/terraform/modules/core/core.tf
+++ b/terraform/modules/core/core.tf
@@ -466,8 +466,8 @@ resource "aws_appautoscaling_target" "target" {
   resource_id        = "service/${var.ecs_cluster_name}/${var.ecs_service_web_name}"
   scalable_dimension = "ecs:service:DesiredCount"
   role_arn           = "${aws_iam_role.ecs_autoscale_role.arn}"
-  min_capacity       = 3
-  max_capacity       = 10
+  min_capacity       = 6
+  max_capacity       = 20
 }
 
 resource "aws_appautoscaling_policy" "up" {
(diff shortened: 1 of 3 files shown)


tahb:

terraform/modules/core/core.tf
```diff
@@ -453,6 +453,10 @@ resource "aws_autoscaling_group" "ecs-autoscaling-group" {
   ]
 
   depends_on = ["aws_vpc.vpc", "aws_launch_configuration.ecs-launch-configuration", "aws_security_group.default", "aws_security_group.ecs"]
+
+  lifecycle {
```
rjw1: This probably needs a comment reflecting the need to comment it out when changing the config.

terraform/modules/core/core.tf
```diff
@@ -512,12 +512,31 @@ resource "aws_appautoscaling_policy" "down" {
 resource "aws_cloudwatch_metric_alarm" "web-cpu-utilisation-high" {
   alarm_name          = "${var.project_name}-${var.environment}-web-cpu-utilisation-high"
   comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "5"
+  evaluation_periods  = "2"
```
rjw1: Should the commit message say why we changed the number of evaluation periods.

PR #750, opened 2019-02-21
[725] POST, DELETE requests continue to work with 403 errors from Clo…

tahb:
…udfront

## Trello card URL:
https://trello.com/c/ecGy8swS

## Changes in this PR:

* Before this change trying to sign in or create a job listing were met with an unsettled cloud front error:
```
The request could not be satisfied.
This distribution is not configured to allow the HTTP request method that was used for this request. The distribution supports only cachable requests.
Generated by cloudfront (CloudFront)
Request ID: ZrAJUCiNfTwNUV0iFM398CaryU2CnS-ZP7NIZG5wV9dvAxNvuIoTjw==
```
* Default cache behaviour now respects cache-control headers instead of a different cache_behaviour block in Terraform. The hope was that request methods that did not match eg. POST would fall down the order to the default cache behaviour - apparently not!
* To make this change CloudFront does not allow all headers to be send, so we have to add ‘all’ the headers it used to allow as a white list for this change to work.
* There is a limit on how many headers that we can forward through Cloudfront when declaring it manually (this feels wrong) but these were the ones I chose to drop since they appear specific to Cloudfront and are unlikely to affect the app:
```
        "CloudFront-Forwarded-Proto",
        "CloudFront-Is-Desktop-Viewer",
        "CloudFront-Is-Mobile-Viewer",
        "CloudFront-Is-SmartTV-Viewer",
        "CloudFront-Is-Tablet-Viewer",
        "CloudFront-Viewer-Country",
```

### Before
screenshot 2019-02-21 at 18 02 45

## Next steps:

### DIFF
--- a/terraform/modules/cloudfront/cloudfront.tf
+++ b/terraform/modules/cloudfront/cloudfront.tf
@@ -40,16 +40,25 @@ resource "aws_cloudfront_distribution" "default" {
 
     forwarded_values {
       query_string = true
-      headers      = ["*"]
+      headers      = [
+        "Host",
+        "Authorization",
+        "Origin",
+        "Referer",
+        "Accept",
+        "Accept-Charset",
+        "Accept-DateTime",
+        "Accept-Encoding",
+        "Accept-Language"
+      ]
 
       cookies {
         forward = "all"
       }
     }
 
-    min_ttl     = 0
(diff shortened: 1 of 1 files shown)


tahb:

tahb:

PR #757, opened 2019-02-26
Data on published jobs should be downloadable rather than sent to a google sheet

pezholio:
## Trello card URL:

https://trello.com/c/BedHMPme/730-data-on-published-jobs-should-be-downloadable-rather-than-sent-to-a-google-sheet

## Changes in this PR:

* Creates a new `AuditData` model to store auditable events
* Rather than writing to a spreadsheet when publishing jobs, we create an `AuditData` record
* Add a task that takes all of today's `AuditData` records and appends them to a Google sheet
* A tweak that allows us to refer to worksheets in a spreadsheet by their GID, rather than by index (which can be changed by users when reordering worksheets, either accidentally or deliberately)

## Next steps


pezholio:
One thing that is missing here is setting up the scheduled job in the Terraform config, which I may need help with.

lib/spreadsheet_writer.rb
```diff
@@ -38,9 +39,13 @@ def session
     end
 
     def worksheet
-      @worksheet ||= session.spreadsheet_by_key(spreadsheet_id).worksheets[worksheet_pos]
+      @worksheet ||= use_gid ? spreadsheet.worksheet_by_gid(worksheet_pos) : spreadsheet.worksheets[worksheet_pos]
```
robbiepaul: 👍 looks good, is there any case where we don't want to look up a worksheet by `gid`?

searches.csv
```diff
@@ -0,0 +1,97 @@
+Timestamp,Number of results,Location,Radius,Keyword,Minimum Salary,Maximum salary,Working pattern,Phase,NQT,,
```
robbiepaul: I can't see this file being used anywhere, did you mean to commit it?

pezholio: Whoops! Yeah, I'll ding that

lib/spreadsheet_writer.rb
pezholio: Yeah, the other jobs that use SpreadsheetWriter currently lookup the worksheets by key. I'm going to slowly bring in the new stuff, but wanted to be backwardly compatible for now

robbiepaul:
Looks good to me, not sure how much I could help with setting up the scheduled job though. Happy to pair on it if you like

pezholio:
@robbiepaul Cool - let's maybe pair on it tomorrow? Will help me in my continued investigations into the world of Terraform 😄

pezholio:
@tahb I've added the Terraform config now (ably assisted by @robbiepaul). If you could give it a once over and lemme know if there's anything I'm missing, that'd be great 👍

terraform/modules/ecs/ecs.tf
```diff
@@ -192,6 +192,33 @@ data "template_file" "import_schools_container_definition" {
   }
 }
 
+/* import_schools task definition*/
```
tahb: Comment for a different task definition.

lib/spreadsheet_writer.rb
```diff
@@ -1,18 +1,31 @@
...
+    def append_row(row, last_pos = worksheet.num_rows, save = true)
       pos = last_pos + 1
       row.each_with_index do |cell, index|
         worksheet[pos, index + 1] = cell
       end
-      worksheet.save
+      worksheet.save if save
+    end
+
+    def last_row
```
tahb: Not sure where this method is called from but it seems to be generating errors https://rollbar.com/dxw/teacher-vacancies/items/409/

PR #765, opened 2019-02-27
Subscription emails are deleted after their last use

pezholio:
## Trello card URL:

https://trello.com/c/oL7yOC4I/728-431-email-alerts-subscription-emails-are-deleted-after-their-last-use

## Changes in this PR:

This PR uses the `SendDailyAlertEmailJob` job to run through all the subscriptions, and delete any that are expired before sending an alert. 

We may want to, in future, send a reminder email to users with subscriptions that are due to expire and allow them to continue to recieve emails after the initial three month period has expired, but this can come later.

### DIFF
--- a/spec/models/subscription_spec.rb
+++ b/spec/models/subscription_spec.rb
@@ -204,4 +204,18 @@ RSpec.describe Subscription, type: :model do
       end
     end
   end
+
+  describe 'expired?' do
+    context 'when the expires_on date is in the future' do
+      let(:subscription) { build_stubbed(:daily_subscription, expires_on: Time.zone.today + 1.month) }
+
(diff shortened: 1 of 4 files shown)


spec/models/subscription_spec.rb
```diff
@@ -204,4 +204,18 @@
       end
     end
   end
+
+  describe 'expired?' do
+    context 'when the expires_on date is in the future' do
+      let(:subscription) { create(:daily_subscription, expires_on: Time.zone.today + 1.month) }
```
robbiepaul: We could use `build_stubbed` in this spec instead of `create` since we don't need to actually create the record?

PR #766, opened 2019-02-28
[731] Feedback from hiring staff is sent to Google Sheets in a daily bulk upload

pezholio:
## Trello card URL:

## Changes in this PR:

This builds on #757 to add details of feedback responses to a Google Sheet on a daily basis, rather than as soon as they are recieved. This means we hit to Google Sheets API less often, and means we're queueing up fewer jobs

This is in draft at the moment, as #757 needs to be merged first


app/models/feedback.rb
```diff
@@ -8,12 +8,13 @@ class Feedback < ApplicationRecord
...
     [
-      created_at.iso8601.to_s,
+      Time.zone.now.to_s,
       user.oid,
       vacancy.id,
       vacancy.school.urn,
       rating,
-      comment
+      comment,
+      created_at.to_s
```
tahb: 👍

lib/tasks/update_spreadsheets.rake
```diff
@@ -2,5 +2,6 @@ namespace :spreadsheets do
   desc 'Updates Google spreadsheets with the latest audit data'
   task update: :environment do
     AddAuditDataToSpreadsheetJob.perform_later('vacancies')
+    AddFeedbackToSpreadsheetJob.perform_later
```
tahb: I wonder why we have a new job here as the existing job above looks to be configured to be reusable by the acceptance of the parameter. Could we save on an extra class here?

```
AddAuditDataToSpreadsheetJob.perform_later('feedback')
```

tahb: I've given the commits here a read through and now understand that we've refactored this to allow us with `AuditData` and other sources, like `Feedback`. 

Should we remove this line https://github.com/dxw/teacher-vacancy-service/blob/develop/app/models/audit_data.rb#L9 from this commit https://github.com/dxw/teacher-vacancy-service/pull/766/commits/09572933748c04e1e1baaa720128117b61e830aa which looks to have set an array of categories a bit prematurely.

robbiepaul: I've removed the feedback category from that commit 👍

tahb:

PR #770, opened 2019-02-28
[757] Don’t build dev and test gems into staging envs

tahb:
## Trello card URL:
In the investigation of https://trello.com/c/yeyEZCG7

## Changes in this PR:
* This increases parity with production which was discovered during an investigation into high web container memory usage. I noticed spring was running all over Edge with 4 servers and 8 clients, all using 4% of the disk. The way the Dockerfile works is that if it has a RAILS_ENV of edge, it bundles all the groups. This includes Spring, which is then loaded when bin/rails is called by `rails s` which is needed in development and test only. Switch the logic to whitelist the 2 known occasions when we DO want all the gems, this makes the code less fragile to new environments being created

## Before
Edge
screenshot 2019-02-28 at 11 46 38

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,10 +45,10 @@ RUN gem install bundler
 # bundle ruby gems based on the current environment, default to production
 RUN echo $RAILS_ENV
 RUN \
-  if [ "$RAILS_ENV" = "production" ]; then \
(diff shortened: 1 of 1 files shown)


tahb:

PR #771, opened 2019-02-28
[757] Update Ruby to 2.6.1

tahb:
## Trello card URL:
During investigation for https://trello.com/c/yeyEZCG7

## Changes in this PR:
* During the high memory investigation a similar issue was flagged around memory leaks in Ruby protobuffs: https://github.com/puma/puma/issues/1600 - while we aren’t on the same versions as mentioned, I updated, tested and checked the application can run. It's hard to tell if this change has had any improvements to memory usage as only so much can be done on Edge to replicate this Production only issue
* Rubocop required a change and autocorrect for new lines and a active record relationship method switch


tahb:

tahb:
Build is now failing after a rebase, woop :)

PR #773, opened 2019-03-04
[757] Align DB, Puma, Sidekiq and Redis connection count

tahb:
## Trello card URL:
https://trello.com/c/yeyEZCG7

## Changes in this PR:
* Everything uses 5 threads by default
* Before this change Sidekiq was using 10 workers which was more than the 5 default active record connections. This can lead to database connection issues for Sidekiq and they advice us to keep the 2 numbers as close as possible https://github.com/mperham/sidekiq/wiki/Advanced-Options#concurrency. From our autoscaling work we know that we are comfortable with each container having 5 puma threads and 5 db connections so we need Sidekiq to follow along for stability
* We can change this setting by a variable if we need to make future tweaks. I suggest we update variables.tf itself for a more permanent change rather than through .tfvars which suggests it's safe to tweak each per environment. A change to this number has to be tested against the resources the containers have available and how autoscaling responds due to any change in what resources are used when resting
* RAILS_MAX_THREADS variable used as the configurable variable for this. Sidekiq have built this in here https://github.com/mperham/sidekiq/pull/3807 however I've decided to use it more explicitly so it's clearer for us to see how they are all linked together

### Before
screenshot 2019-03-04 at 10 44 37

### After
screenshot 2019-03-04 at 10 44 09


erbridge:
Is there a reason this is being held out?

tahb:
No reason that I'm aware of.

PR #778, opened 2019-03-04
[757] ElasticSearch performance increase when only returning the first 10 results

tahb:
## Trello card URL:
https://trello.com/c/yeyEZCG7

## Changes in this PR:
* The size parameter allows you to configure the maximum amount of hits to be returned: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html
* This significantly improves response times as less is going over the wire and I presume ES is allowed to be more efficient
* No specific reason as to why this was first set to 1000 so I don’t believe I’m going to fall into an old trap

### Before
with 800 jobs
before 899ms

### After
with 800 jobs
after 429ms

### DIFF
--- a/app/services/elastic_search_finder.rb
+++ b/app/services/elastic_search_finder.rb
@@ -2,6 +2,10 @@ require 'elasticsearch'
 
 class ElasticSearchFinder
   def call(query, sort)
-    Vacancy.__elasticsearch__.search(size: 1000, query: query, sort: sort)
+    Vacancy.__elasticsearch__.search(size: size, query: query, sort: sort)
+  end
+
+  def size
+    Vacancy.default_per_page
   end
 end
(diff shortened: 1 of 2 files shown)


tahb:

app/services/elastic_search_finder.rb
```diff
@@ -2,6 +2,10 @@
 
 class ElasticSearchFinder
   def call(query, sort)
-    Vacancy.__elasticsearch__.search(size: 1000, query: query, sort: sort)
+    Vacancy.__elasticsearch__.search(size: size, query: query, sort: sort)
+  end
+
+  def size
+    Vacancy.default_per_page
```
tahb: Tap into the Kaminari configuration set on the Vacancy model with `paginates_per 10` to make it update automatically without being forgotten.

tahb:

PR #784, opened 2019-03-06
Fixes the published vacancy Google Sheet auditor

robbiepaul:
## Trello card URL:

https://trello.com/c/BedHMPme/730-data-on-published-jobs-should-be-downloadable-rather-than-sent-to-a-google-sheet

## Changes in this PR:

- We don't have a `default` queue so these jobs are never getting picked up by Sidekiq. By following convention for the rest of the project we give this job its own queue.
- First column in spreadsheet should be current time job ran at. We were getting duplicate audit records when setting it to be the record's `created_at`
- Convert `created_at` from audit record to string when saving to JSON. This allows the date to be formatted correctly when it is saved in Google Sheets

### DIFF
--- a/app/jobs/audit_published_vacancy_job.rb
+++ b/app/jobs/audit_published_vacancy_job.rb
@@ -1,4 +1,6 @@
 class AuditPublishedVacancyJob < ApplicationJob
+  queue_as :audit_published_vacancy
+
   def perform(vacancy_id)
     vacancy = Vacancy.find(vacancy_id)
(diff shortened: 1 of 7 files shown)


robbiepaul:
[Data on published jobs should be downloadable rather than sent to a Google Sheet](https://trello.com/c/BedHMPme/730-data-on-published-jobs-should-be-downloadable-rather-than-sent-to-a-google-sheet)

PR #786, opened 2019-03-07
[764] Fix vacancy publish_on validation

robbiepaul:
## Trello card URL:

https://trello.com/c/W4yWwhbW/764-fix-vacancy-validation

### The problem:

Any published vacancy can't be modified internally (eg. by a worker). The validation for checking `publish_on` date isn't in the past always fails.

### The solution:

We only check if `publish_on` isn't in the past if the value has changed, this allows us to save vacancies internally. We use ActiveModel's dirty methods for checking if the value has changed.

The presents a problem on the copy a vacancy page where the user could leave the `publish_on` date in the past, therefore passing validation. To fix this we added another check for the `publish_on` date in the `CopyVacancyForm` and reset the `publish_on` date if it has past to help the user not run in to the error.


robbiepaul:
[BUG: Fix vacancy validation](https://trello.com/c/W4yWwhbW/764-bug-fix-vacancy-validation)

PR #787, opened 2019-03-08
[729] Only update vacancies that have page view increments

tahb:
## Trello card URL:
https://trello.com/c/ZfIhptK0

## Changes in this PR:
* running this on Staging resulted in a single Vacancy update for every active listing which can be in the thousands, we need to optimise this to save on unnecessary resource use and logging storage
* avoid additional callbacks to elastic search to reindex for all records
### Before
screenshot 2019-03-08 at 09 43 30

### DIFF
--- a/app/services/vacancy_page_view.rb
+++ b/app/services/vacancy_page_view.rb
@@ -8,6 +8,8 @@ class VacancyPageView
   end
 
   def persist!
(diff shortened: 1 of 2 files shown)


tahb:
[Total job page views are calculated without Google Analytics](https://trello.com/c/ZfIhptK0/729-total-job-page-views-are-calculated-without-google-analytics)

PR #793, opened 2019-03-12
(768) Log the stored state and Accept header when the callback phase starts

robbiepaul:
## Trello card URL:

https://trello.com/c/stLxQgfi/768-investigate-intermittent-bug-stopping-hiring-staff-sign-in-on-specific-browsers-or-devices

## Changes in this PR:

- We're seeing a few occurrences of sign-in callback errors. To help us diagnose this we need to log the stored state as well as the state that comes back from DSI.
- From looking at the logs we can see multiple requests hitting our callback endpoint at the same time, the subsequent ones seem to have `*/*` as the value of the `Accept` header. This could be caused by erroneous software on the user's computer, it would be nice to see what values it can be when we see failures.

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -10,7 +10,8 @@ module OmniAuth
       def callback_phase
         Rollbar.log(:info, 'A sign-in callback was started',
                     stored_state: session['omniauth.state'],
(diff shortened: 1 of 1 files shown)


robbiepaul:
[Investigate intermittent bug stopping hiring staff sign in on specific browsers or devices](https://trello.com/c/stLxQgfi/768-investigate-intermittent-bug-stopping-hiring-staff-sign-in-on-specific-browsers-or-devices)

PR #794, opened 2019-03-12
(fix) Audit worksheet GIDs are now interpolated strings

robbiepaul:
## Trello card URL:

## Changes in this PR:

I noticed during the most recent `terraform plan` on staging that these values weren't getting passed in to the container definitions. This commit should fix that by making the strings interpolated so they are resolved when the container definition is generated.

### DIFF
--- a/terraform/container-definitions/web_container_definition.json
+++ b/terraform/container-definitions/web_container_definition.json
@@ -131,14 +131,6 @@
         "value": "${audit_spreadsheet_id}"
       },
       {
-        "name": "AUDIT_VACANCIES_WORKSHEET_GID",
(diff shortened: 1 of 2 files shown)


robbiepaul:

terraform/container-definitions/web_container_definition.json
```diff
@@ -132,11 +132,11 @@
       },
       {
         "name": "AUDIT_VACANCIES_WORKSHEET_GID",
```
pezholio: The GIDs actually I don't think _need_ to be in the web container definition, so they can probably be binned

robbiepaul: I've removed the GIDs from the web container definition 👍

PR #795, opened 2019-03-13
Redirect auth/failure to DSI login

pezholio:
## Trello card URL:

https://trello.com/c/ikmTQ91m/769-fix-404-error-as-dsi-is-sometimes-redirecting-to-auth-failure-when-credentials-are-invalid

## Changes in this PR:

When users use the back button after logging in, DfE signin redirects the user to /auth/failure on our app, which returns a 404. This redirects that path to the `HiringStaff::IdentificationsController#create` action, which then in turn prompts the user to log in again.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -35,6 +35,7 @@ Rails.application.routes.draw do
            path: '/dfe/sessions',
            controller: 'hiring_staff/sign_in/dfe/sessions'
   get '/auth/dfe/callback', to: 'hiring_staff/sign_in/dfe/sessions#create'
+  get '/auth/failure', to: 'hiring_staff/sign_in/dfe/sessions#new'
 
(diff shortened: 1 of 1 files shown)


pezholio:
[Fix 404 error as DSI is sometimes redirecting to auth/failure when credentials are invalid](https://trello.com/c/ikmTQ91m/769-fix-404-error-as-dsi-is-sometimes-redirecting-to-auth-failure-when-credentials-are-invalid)

robbiepaul:
When testing this locally I get `OmniAuth::Strategies::OpenIDConnect::CallbackError`

Screenshot 2019-03-13 at 14 43 32

That's after I've signed in successfully and I'm on `/school`, then I hit the back button - is that correct?

pezholio:
Yep, I see the same thing. Lemme see if there's a better way to handle this

PR #796, opened 2019-03-13
(731) Allow feedback user to be nil

robbiepaul:
## Trello card URL:

## Changes in this PR:

We only introduced the feedback->user relationship recently and if we want to backfill older feedback items on to Google Sheets we need to allow for `user` to be nil.


robbiepaul:

PR #799, opened 2019-03-13
(fix) add validations to vacancy form model

pezholio:
## Trello card URL:

https://trello.com/c/gbTNiXRm/789-bug-when-a-hiring-staff-copied-a-job

## Changes in this PR:

This adds a validation to make sure that `publish_on` is present when submitting a job copy. It's required anyway, but the form model validations get run before the model validations themselves, and the `publish_on_must_not_be_in_the_past` method assumes that `publish_on` is always present. 

I initially added some other validations to the form model, but then realised they weren't necessary as they were covered by the Vacancy model, but I think the tests are useful to have around anyway.

### DIFF
--- a/app/form_models/copy_vacancy_form.rb
+++ b/app/form_models/copy_vacancy_form.rb
@@ -7,10 +7,11 @@ class CopyVacancyForm < VacancyForm
            :publish_on_dd, :publish_on_mm, :publish_on_yyyy,
            :errors, to: :vacancy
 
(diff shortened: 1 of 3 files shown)


pezholio:
[Bug when a hiring staff copied a job](https://trello.com/c/gbTNiXRm/789-bug-when-a-hiring-staff-copied-a-job)

tahb:
Rebasing

PR #804, opened 2019-03-14
(chore) Temporarily turn off audit search data Google Sheet job

robbiepaul:
## Trello card URL:

https://trello.com/c/OOdZ8z3b/796-temporarily-turn-off-audit-search-data-until-we-decide-what-to-do-with-it

## Changes in this PR:

It's the last remaining Google Sheet job that uses index instead of GID. We should decide what to do with it then take our time to refactor out using the index in Google Sheets.

For now its having unintended consequences overwriting whichever sheet is in position 3 and the worksheet's positions seem to randomly change so we decided to turn it off for now.

### DIFF
--- a/app/jobs/audit_search_event_job.rb
+++ b/app/jobs/audit_search_event_job.rb
@@ -3,6 +3,7 @@ class AuditSearchEventJob < SpreadsheetWriterJob
   queue_as :audit_search_event
 
   def perform(data)
+    return if WORKSHEET_POSITION.present? # Temporarily turn off this auditing until we decide what to do with it
     return unless AUDIT_SPREADSHEET_ID
 
     write_row(data, WORKSHEET_POSITION)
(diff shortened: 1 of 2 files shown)


robbiepaul:
[Temporarily turn off audit search data Google Sheet until we decide what to do with it](https://trello.com/c/OOdZ8z3b/796-temporarily-turn-off-audit-search-data-google-sheet-until-we-decide-what-to-do-with-it)

app/jobs/audit_search_event_job.rb
```diff
@@ -3,6 +3,7 @@ class AuditSearchEventJob < SpreadsheetWriterJob
   queue_as :audit_search_event
 
   def perform(data)
+    return if WORKSHEET_POSITION.present? # Temporarily turn off this auditing until we decide what to do with it
```
pezholio: I can see why you've done this, but is there a risk that if we remove that variable the code will kick into life and then start erroring? It's a pretty low risk, so probably fine

robbiepaul: Yea Rubocop complained if I just put `return`. I hope the comment adds enough context to stop that from happening

pezholio: Yeah, it should be fine, and even if it isn't, all we'll get are errors

PR #809, opened 2019-03-18
Job seeker alerts

tahb:
This is a big feature branch that has been worked on beside develop and is constructed of many Trello cards under the 'job seeker alerts' label.

## Trello card URL:
https://trello.com/c/JNvYZhth

## Changes in this PR:
- feature flagged
- job seekers can set up daily alerts for all new listings that match their search for up to 3 months
- job seekers can unsubscribe at any time via the email
- emails sent by GOV.UK Notify at 8am every day
- job alert emails include links back to every individual listing and the service

## Next steps:


Gemfile
```diff
@@ -57,6 +57,8 @@ gem 'high_voltage', '~> 3.1'
 gem 'google_drive', require: false
 gem 'google-api-client'
 gem 'browser'
+gem 'notifications-ruby-client'
```
pezholio: We probably don't need this gem to be specified as it's a dependency of `mail-notify`

app/models/subscription.rb
```diff
@@ -0,0 +1,76 @@
+class Subscription < ApplicationRecord
+  include Auditor::Model
+
+  enum status: %i[active trashed]
```
pezholio: I don't think we need this anymore, we loop over all subscriptions and deleted them when they're expired

app/models/subscription.rb
```diff
@@ -0,0 +1,76 @@
...
+  enum frequency: %i[daily]
+
+  has_many :alert_runs
+
+  validates :email, email_address: { presence: true }
+  validates :frequency, presence: true
+  validates :search_criteria, uniqueness: { scope: %i[email expires_on frequency] }
+
+  scope :ongoing, -> { active.where('expires_on >= current_date') }
```
pezholio: Likewise, I don't think this is used anymore

app/services/vacancy_filters.rb
```diff
@@ -30,6 +30,10 @@ def to_hash
     }
   end
 
+  def only_active_to_hash
```
pezholio: There's some code here that could be used in `any?` to DRY up the logic I think

config/sidekiq.yml
```diff
@@ -1,4 +1,6 @@
 :queues:
+  - [ daily_alert_email, 1 ]
```
pezholio: I don't think this queue gets used anymore, as the daily alert email will go to the `mailers` queue by default

lib/services/notify.rb
```diff
@@ -0,0 +1,37 @@
+require 'notifications/client'
+
+class Notify
```
pezholio: We don't need this class anymore, as all the Notify grunt work is handled by `mail-notify` and the mailers

config/sidekiq.yml
tahb: I think it's still in there for use https://github.com/dxw/teacher-vacancy-service/blob/feature/job-seeker-alerts/app/jobs/send_daily_alert_email_job.rb#L2 and nothing is using the `mailers` queue. 

Is there a reason that we're using a more generically named queue? I'm wondering how this will extend if we wanted to send another _less important_ email to users, perhaps when a job listing has expired with some stats and a call for feedback as has been discussed already.

pezholio: Ah, right you are. The reason we're using the `mailers` queue is because that's what `deliver_later` uses by default. I think it's easy to configure though.

app/models/subscription.rb
tahb: I believe this is still being used here: https://github.com/dxw/teacher-vacancy-service/blob/feature/job-seeker-alerts/app/services/subscription_finder.rb#L10

config/sidekiq.yml
tahb: On testing at the moment there are queues for both names:

```
daily_alert_email
mailers
```

I'm going to see if I can set this with an explicit queue name of priority 3.

tahb: Oh I think I get it.. queue `daily_alert_email` is for the jobs that enqueue the emails to be sent via `mailers`. Will see if I can find clearer names to avoid future confusion.

app/models/subscription.rb
tahb: I see, we can bosh the concept of `ongoing` and `status` because we ended up hard deleting records instead of the original plan to soft delete.

tahb:
@pezholio Ready for another pass now 🚀

app/jobs/send_daily_alert_email_job.rb
```diff
@@ -11,11 +11,11 @@ def perform
 
       Rails.logger.info("Sidekiq: Sending vacancy alerts for #{vacancies.count} vacancies")
 
-      AlertMailer.daily_alert(s.id, vacancies.pluck(:id)).deliver_later
+      AlertMailer.daily_alert(s.id, vacancies.pluck(:id)).deliver_later(queue: :email_daily_alerts)
```
pezholio: Sweet - can we get a test in for this to make sure the job gets sent to the right queue?

app/services/vacancy_filters.rb
```diff
@@ -31,12 +31,12 @@ def to_hash
   end
 
   def only_active_to_hash
-    to_hash.delete_if { |_, v| v.blank? }
+    to_hash.delete_if { |_, v| v.blank? || v.nil? }
```
pezholio: Can we just do `v.empty?` for this?

app/services/vacancy_filters.rb
```diff
@@ -30,9 +30,13 @@ def to_hash
...
 
+  def only_active_to_hash
+    to_hash.delete_if { |_, v| v.blank? || v.nil? }
+  end
+
   def any?
-    filters = to_hash
-    filters.delete_if { |k, v| k.eql?(:radius) || v.nil? }
+    filters = only_active_to_hash
+    filters.delete_if { |k, _| k.eql?(:radius) }
```
pezholio: This `k.eql?(:radius)` seems a bit redundant now, as all we're doing is casting the radius to a string. It might be better to just cast the radius to a string if it's present in the initializer. Something like `radius = args[:radius].to_s if args[:radius].present?`

app/jobs/send_daily_alert_email_job.rb
tahb: Quite right 👍

app/services/vacancy_filters.rb
tahb: There is one method that handles both, after testing I think it is in fact `blank?` rather than `empty?` as the test suite is very unhappy when only using `empty?` so I've switched to `blank?` only!

Screenshot 2019-03-19 at 09 55 43

Screenshot 2019-03-19 at 09 56 23

PR #810, opened 2019-03-18
Address noisy test output

tahb:
## Changes in this PR:
- update shoulda-matchers deprecation warning
- I had to remove everything rom config/application.yml to remove the double ENV load
- change the way we use SystemExit to notify us of problems to a way that allows us to be told about them only when we need to take action

### Before
Screenshot 2019-03-18 at 18 34 38

### After
Screenshot 2019-03-18 at 19 39 22

### DIFF
--- a/lib/analytics.rb
+++ b/lib/analytics.rb
@@ -9,7 +9,7 @@ class Analytics
   attr_reader :service, :path, :start_date, :end_date
 
   def initialize(path, start_date = '30daysAgo', end_date = 'today')
-    abort('No Google API key set.') if api_key_empty?
+    return if api_key_empty?
 
     @service = API::AnalyticsService.new
     @start_date = start_date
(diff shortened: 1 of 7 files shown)


lib/performance_platform.rb
```diff
@@ -5,7 +5,9 @@ class Base
     attr_reader :headers
 
     def initialize(token = nil)
-      token || abort('No token set. Note that this task should only be executed in production.')
+      return unless Rails.env.production?
```
pezholio: Nice. Much better to be explicit, just in case we do set the token in another environment

lib/analytics.rb
```diff
@@ -9,7 +9,7 @@ class Analytics
   attr_reader :service, :path, :start_date, :end_date
 
   def initialize(path, start_date = '30daysAgo', end_date = 'today')
-    abort('No Google API key set.') if api_key_empty?
+    return if api_key_empty?
```
pezholio: This is better. Raising `SystemExit` is probably a bit too much of a blunt instrument anyway

PR #813, opened 2019-03-19
Address performance warnings

pezholio:
## Trello card URL:

https://trello.com/c/mYKQZS1v/766-address-performance-warnings-to-ensure-we-can-scale-up-to-more-jobs

## Changes in this PR:

I've had a stab at addressing some of the issues highlighted by Skylight. There's probably more that can be surfaced, but this is a good start I think:

* When we hit the homepage of TV, we're always hitting ElasticSearch, which probably isn't necessary. If there are no queries, then we just default to getting all vacancies from the database (paginated of course)
* When listing vacancies, the `count` query gets hit a number of times. We're now memoizing the `total_count` var in the presenter
* I noticed some DB queries were running a lot slower than expected, so I installed [lol_dba](https://github.com/plentz/lol_dba) locally to see if we had any missing indexes (which we did), so I've added a migration to add some that we were missing.

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -17,8 +17,12 @@ class VacanciesController < ApplicationController
   def index
     @filters = VacancyFilters.new(search_params)
     @sort = VacancySort.new.update(column: sort_column, order: sort_order)
-    records = Vacancy.public_search(filters: @filters, sort: @sort).page(page_number).records(includes: [:school])
-    audit_search_event(records, @filters) if @filters.any?
+
+    records = if @filters.any?
+                Vacancy.public_search(filters: @filters, sort: @sort).page(page_number).records(includes: [:school])
+              else
+                Vacancy.live.order(@sort.column => @sort.order).includes(:school).page(page_number)
+              end
 
     @vacancies = VacanciesPresenter.new(records, searched: @filters.any?)
 
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -26,6 +26,8 @@ class VacanciesController < ApplicationController
 
     @vacancies = VacanciesPresenter.new(records, searched: @filters.any?)
 
+    audit_search_event if @filters.any?
+
     expires_in 5.minutes, public: true
   end
 
@@ -55,10 +57,10 @@ class VacanciesController < ApplicationController
     request.path != job_path(vacancy) && !request.format.json?
   end
 
-  def audit_search_event(records, filters)
+  def audit_search_event
     AuditSearchEventJob.perform_later([Time.zone.now.iso8601.to_s,
-                                       records.total_count,
-                                       *filters.to_hash.values])
+                                       @vacancies.total_count,
+                                       *@filters.to_hash.values])
   end
 
   def id
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -17,16 +17,7 @@ class VacanciesController < ApplicationController
   def index
     @filters = VacancyFilters.new(search_params)
     @sort = VacancySort.new.update(column: sort_column, order: sort_order)
-
-    records = if @filters.any?
-                Vacancy.public_search(filters: @filters, sort: @sort).page(page_number).records(includes: [:school])
-              else
-                Vacancy.live.order(@sort.column => @sort.order).includes(:school).page(page_number)
-              end
-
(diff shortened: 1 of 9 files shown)


pezholio:
[Address performance warnings to ensure we can scale up to more jobs](https://trello.com/c/mYKQZS1v/766-address-performance-warnings-to-ensure-we-can-scale-up-to-more-jobs)

pezholio:
[Address performance warnings to ensure we can scale up to more jobs](https://trello.com/c/mYKQZS1v/766-address-performance-warnings-to-ensure-we-can-scale-up-to-more-jobs)

pezholio:
[Address performance warnings to ensure we can scale up to more jobs](https://trello.com/c/mYKQZS1v/766-address-performance-warnings-to-ensure-we-can-scale-up-to-more-jobs)

app/controllers/vacancies_controller.rb
```diff
@@ -17,8 +17,12 @@ class VacanciesController < ApplicationController
   def index
     @filters = VacancyFilters.new(search_params)
     @sort = VacancySort.new.update(column: sort_column, order: sort_order)
-    records = Vacancy.public_search(filters: @filters, sort: @sort).page(page_number).records(includes: [:school])
-    audit_search_event(records, @filters) if @filters.any?
+
+    records = if @filters.any?
```
tahb: Good idea. I'd propose moving this logic into a new service object to keep on-top of controllers getting too big, perhaps `VacancyFinder`?

app/controllers/vacancies_controller.rb
```diff
@@ -26,6 +26,8 @@ def index
 
     @vacancies = VacanciesPresenter.new(records, searched: @filters.any?)
 
+    audit_search_event if @filters.any?
```
tahb: Surprised to see we are adding this line in in this commit, is this right?

app/controllers/vacancies_controller.rb
```diff
@@ -55,10 +57,10 @@ def old_vacancy_path?(vacancy)
     request.path != job_path(vacancy) && !request.format.json?
   end
 
-  def audit_search_event(records, filters)
+  def audit_search_event
     AuditSearchEventJob.perform_later([Time.zone.now.iso8601.to_s,
-                                       records.total_count,
-                                       *filters.to_hash.values])
+                                       @vacancies.total_count,
```
tahb: 👍

app/presenters/vacancies_presenter.rb
```diff
@@ -70,6 +70,10 @@ def next_api_url
 
   private
 
+  def total_count
+    @total_count ||= model.total_count
```
tahb: 👍

app/presenters/vacancies_presenter.rb
```diff
@@ -21,6 +21,10 @@ def each(&block)
     decorated_collection.each(&block)
   end
 
+  def any?
```
tahb: 🎉 👍

app/controllers/vacancies_controller.rb
pezholio: Sounds like a plan. I'll look into it 👍

app/controllers/vacancies_controller.rb
pezholio: Yeah, I agonised over this line a bit, as I initially put in in the finer, but I think it's OK here. I might actually stick it in the service object now

pezholio:
@tahb I haven't deployed it to testing yet, because I was under the impression that we'd need a decent corpus of data to be able to see if it made a difference

tahb:
@pezholio The strategy I used for the previous performance work (which wasn't perfect) was to simulate load to various endpoints on Edge from my machine for an hour or so to compare before and after.

app/controllers/vacancies_controller.rb
```diff
@@ -17,16 +17,7 @@ class VacanciesController < ApplicationController
...
-    records = if @filters.any?
-                Vacancy.public_search(filters: @filters, sort: @sort).page(page_number).records(includes: [:school])
-              else
-                Vacancy.live.order(@sort.column => @sort.order).includes(:school).page(page_number)
-              end
-
-    @vacancies = VacanciesPresenter.new(records, searched: @filters.any?)
-
-    audit_search_event if @filters.any?
+    @vacancies = VacanciesFinder.new(@filters, @sort, page_number).vacancies
```
tahb: This commit is a great refactor 🌟

tahb:
Rebasing

PR #816, opened 2019-03-20
[801] Email subject clarity for job alerts

tahb:
## Trello card URL:
https://trello.com/c/DQjIIKLC

## Changes in this PR:
- job alert emails are grouped by their subscription in inboxes, rather than a single group of all job alerts
- alert and confirmation emails for job alerts are consistent

### Before
Screenshot 2019-03-20 at 14 21 12

### After
Screenshot 2019-03-20 at 14 15 31

### DIFF
--- a/app/mailers/alert_mailer.rb
+++ b/app/mailers/alert_mailer.rb
@@ -12,7 +12,11 @@ class AlertMailer < ApplicationMailer
     @unsubscribe_token = subscription.token
     @vacancies = VacanciesPresenter.new(vacancies, searched: false)
 
-    subject_key = @vacancies.many? ? 'alerts.email.daily.subject.many' : 'alerts.email.daily.subject.one'
+    subject_key = if @vacancies.many?
+                    'job_alerts.alert.email.daily.subject.many'
+                  else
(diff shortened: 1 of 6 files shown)


tahb:
[Iterate subject line for job seeker alerts](https://trello.com/c/DQjIIKLC/801-iterate-subject-line-for-job-seeker-alerts)

pezholio:
Do you think we need to force the user to specify a reference at this point now, or do we think that needs more research?

tahb:
@pezholio Michael and I discussed that and agree that it would be best to add a new card to make the reference mandatory, and remove the machine generated value completely.

We thought it made sense for it to be a separate piece of work that could be added after we're in production behind the feature flag, this way round allows us to test the big immediate issue and to get the pipeline  unblocked. Card coming soon!

@robbiepaul also had a further idea that concerns in adding another mandatory field could be improved in future by interrogating the `search_criteria` to make a sensible default reference.

pezholio:
Yeah, being able to make a sensible default reference from the search criteria would be awesome. I guess the risk is that if there's a lot of search criteria it could get ridiculously long though

PR #819, opened 2019-03-21
Fix autogenerated subscriptions

pezholio:
If we use this rake task, we need to make sure subscriptions have `expires_on` dates and `search_criteria`, otherwise things don’t work as expected


tahb:
Rebasing

PR #820, opened 2019-03-21
[806] Anonymise email addresses used for job alerts in the logs

tahb:
## Changes in this PR:
* stop them going to Papertrail or any future logging service as we have no need for this information and it makes it impractical to clear up when we no longer need to for sending alerts.
* arguably we should use a more specific parameter key here to avoid any unexpected filtering for other controllers in the future that may have `email: true` for example. The current `email` field is baked into the form via the `Subscription` model rather than an intermediary form object so this can't be changed super quickly. I've decided to kick this can down the road, there's a low likelihood we add a new parameter like this and a low impact if we didn't have this information in our logs - we might be hindered in debugging for a time

### Before
Screenshot 2019-03-21 at 16 02 16

### After
Screenshot 2019-03-21 at 16 02 48

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += %i[password email]


tahb:
[Anonymise user email addresses used in job alerts in Papertrail](https://trello.com/c/MPAmVeuN/806-anonymise-user-email-addresses-used-in-job-alerts-in-papertrail)

tahb:
I spent more than 20 minutes trying to see if I could drive with with tests without making a mess, then settled on the notion that we should be confident enough with Rails config, not to have to test Rails itself. 

That said if you can figure out how to turn on logging to stdout _only_ for this test and complete this test, please let me know:

```
require 'rails_helper'

RSpec.describe 'Subscription#create', type: :request do
  before(:each) do
    Rails.application.configure do
      logger           = ActiveSupport::Logger.new(STDOUT)
      logger.formatter = config.log_formatter
      config.logger    = ActiveSupport::TaggedLogging.new(logger)
    end
  end

  context 'when an email address is in the parameters' do
    it 'filters it out of the logs' do
      search_criteria = { radius: '20' }.to_json
      params = {
        subscription: {
          subscription_email_address: 'email@example.com',
          search_criteria: search_criteria
        }
      }

      expect { post subscriptions_path, params: params }.to_not output('foo@example.com',).to_stderr

      expect(response.status).to eq(200)
    end
  end
end
```

PR #821, opened 2019-03-21
[809] Track job alert emails sent by the service

tahb:
## Trello card URL:
https://trello.com/c/Odgbp1nK

## Changes in this PR:
- refactor existing stats variables to make this change simple to make
- using a simple count of the AlertRuns we already have rather than add a new record via the Public Activity gem or our own AuditData for Google sheets - this may not work for ever but it makes it very cheap to meet the need

### After
Screenshot 2019-03-21 at 17 30 21

Screenshot 2019-03-21 at 17 30 08

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -3,6 +3,8 @@ class StatsController < ApplicationController
     @stats = PublicActivity::Activity.order(:key)
                                      .group(:key).count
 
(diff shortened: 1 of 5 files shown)


tahb:
[Expose the sending of job alerts to the stats page for monitoring](https://trello.com/c/Odgbp1nK/809-expose-the-sending-of-job-alerts-to-the-stats-page-for-monitoring)

PR #827, opened 2019-03-26
Make bin/drebuild work properly

erbridge:
## Changes in this PR:

- `bin/drebuild` now echos out the commands it's running to make debugging easier.
- The config for development database and elasticsearch were moved to `docker-compose.yml`. `env_file` overrides `environment`, so both the test and development images were using the same DB and ES (the ones specified in the `.env` file). Specifying `DATABASE_URL` and `ELASTICSEARCH_URL` in `docker-compose.env` will still override the defaults set in `docker-compose.yml` and `docker-compose.test.yml`.
- `bin/drebuild` now explicitly sets up the development database by running `bin/dsetup` (which feels badly named as the rest of the `bin/d*` scripts run docker commands, but `bin/dsetup` runs `rake` directly). Previously it relied on the test setup setting up the dev database. This had the useful side-effect of giving ES time to start before seeding the database and creating the index (at least on my machine).

## Next steps:

Recommend removing `DATABASE_URL` and `ELASTICSEARCH_URL` from `docker-compose.env`.

### DIFF
--- a/bin/drebuild
+++ b/bin/drebuild
@@ -1,29 +1,25 @@
 #!/bin/sh
 
+run_with_pretty_print()
+{
+  echo "\n\033[36m===> $1\033[0m\n"
+  eval $1
+}
+
 echo "Remove all unused images before we build new ones…"
-echo "\n\033[36m===> docker image prune -f\033[0m\n"
-docker image prune -f
+run_with_pretty_print 'docker image prune -f'
 
 echo "Rebuilding the test server…"
-echo "\n\033[36m===> docker-compose --file=docker-compose.test.yml down -v --remove-orphans\033[0m\n"
-docker-compose --file=docker-compose.test.yml down -v --remove-orphans
-echo "\n\033[36m===> docker-compose --file=docker-compose.test.yml build\033[0m\n"
-docker-compose --file=docker-compose.test.yml build
-echo "\n\033[36m===> bin/dtest-server\033[0m\n"
-bin/dtest-server
+run_with_pretty_print 'docker-compose --file=docker-compose.test.yml down -v --remove-orphans'
+run_with_pretty_print 'docker-compose --file=docker-compose.test.yml build'
+run_with_pretty_print 'bin/dtest-server'
 echo "Rebuilt and started the testing server for TVS."
 
 echo "Rebuilding the web server…"
-echo "\n\033[36m===> docker-compose down -v --remove-orphans\033[0m\n"
-docker-compose down -v --remove-orphans
-echo "\n\033[36m===> docker-compose build\033[0m\n"
-docker-compose build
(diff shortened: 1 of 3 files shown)


bin/drebuild
```diff
@@ -1,18 +1,29 @@
 #!/bin/sh
 
 echo "Remove all unused images before we build new ones…"
+echo "\n\033[36m===> docker image prune -f\033[0m\n"
```
tahb: The colour and line breaks make watching this script run easier to follow and undoubtedly debug. I do think there is a cost of visual noise when working with this file though, feels like I'd have to strip the echo lines to get to grips with the order of things and then add them back in.

Would deduplicating this into a method help with this? Something like so:
```
#!/bin/sh

run_with_pretty_print()
{
  echo "\n\033[36m===> $1\033[0m\n"
  eval $1
}

run_with_pretty_print 'docker image prune -f'
…
```

bin/drebuild
```diff
@@ -1,18 +1,29 @@
...
 docker image prune -f
 
 echo "Rebuilding the test server…"
+echo "\n\033[36m===> docker-compose --file=docker-compose.test.yml down -v --remove-orphans\033[0m\n"
 docker-compose --file=docker-compose.test.yml down -v --remove-orphans
+echo "\n\033[36m===> docker-compose --file=docker-compose.test.yml build\033[0m\n"
```
tahb: As we are now using `bin/dtest-server` I think this command is being run twice, I noticed as I was getting the output from entrypoint.sh twice in the output.

docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
     container_name: teacher-vacancy-service_web_1
     environment:
       RAILS_ENV: "development"
+      DATABASE_URL: "postgres://postgres@db:5432/tvs_development?template=template0&pool=5&encoding=unicode"
```
tahb: A neat way to address the problem, I'm in favour of setting these variables here rather than docker-compose.env with an attempt to overwrite only in test, much cleaner and easier to follow what's going on 👍 

I've been playing around with it to test in my environment and still get the same error. I've spent some further time trying to poke it to see if I can fix it but no joy.

```
ActiveRecord::EnvironmentMismatchError: You are attempting to modify a database that was last run in `test` environment.
You are running in `development` environment. If you are sure you want to continue, first set the environment using:

        bin/rails db:environment:set RAILS_ENV=development
```

Requires a bit more digging I think and happy to screen share to help debug it.

erbridge: Just to check: did you remove `DATABASE_URL` (and `ELASTICSEARCH_URL`) from your `docker-compose.env`?

tahb: Weird it works now, they were both omitted before though!

erbridge: Odd. Glad it's working now, though.

PR #828, opened 2019-03-27
[811] Match feedback array items regardless of order

robbiepaul:
## Trello card URL:

https://trello.com/c/7xii0viN/811-intermittent-test-failure-for-feedback

## Changes in this PR:

From the failure output it looks like the arrays are in different orders which would result in a failure.

```
expect([1,2,3,4]).to eq([4,3,1,2]) #failure
```

However we can use the `match_array` matcher to match the items of the array match regardless of order.

```
expect([1,2,3,4]).to match_array([4,3,1,2]) #pass
```

I don't think the order matters in this spec so it should be fine.

It's worth mentioning, I couldn't recreate the error, this is a best guess at a fix


robbiepaul:
[Intermittent test failure for Feedback?](https://trello.com/c/7xii0viN/811-intermittent-test-failure-for-feedback)

PR #829, opened 2019-03-27
Pin terraform and provider minor versions

erbridge:
## Changes in this PR:

Terraform and provider versions are now pinned to their minor versions. The versions were provided by @tahb from his local environment.

## Next steps:

Local `terraform init` may be required if local versions don't match the constraints.

### DIFF
--- a/terraform.tf
+++ b/terraform.tf
@@ -1,5 +1,10 @@
 provider "aws" {
   region = "${var.region}"
+  version = "~> 1.36.0"
+}
+
+provider "template" {
+  version = "~> 1.0.0"
 }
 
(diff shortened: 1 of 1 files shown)


erbridge:
Not sure I fully followed that @tahb, but it looks like the Cloudfront reordering was due to an updated `aws` provider. I no longer see reordering when I plan. I also wonder if getting everyone on the same versions will resolve some of the trashing of default values you said we see regularly.

erbridge:
@pezholio would you mind giving this a try locally to check it works for someone who isn't Tom or new to the project?

PR #836, opened 2019-04-01
[817] Send Performance Platform data in GMT rather than UTC 

tahb:
## Trello card URL:
https://trello.com/c/aj6ikgAo

## Changes in this PR:
* Thanks to a test failing today it's brought our attention to a problem with how we are sending data to the performance platform, this is fixed and the driver for this work
* Previously this was using UTC and now we have moved to BST this
continuously fails the test `spec/jobs/performance_platform_transactions_queue_job_spec.rb:17`
* Currently (and fortunately) AWS has scheduled this job to happen
around 14:10 GMT so this does not looked to have caused an issue on production. However should this job change schedule to around midnight there could be issues where it's always trying to run the job on listings created 2 days ago!
* This ran the day after the clocks ran with the correct date and
without error ([the numbers generally in PP look very odd though](https://www.performance.service.gov.uk/data/teaching-jobs-job-listings/transactions-by-channel?duration=52&collect=count%3Asum&group_by=channel&period=day&&format=json)):
```
Apr 01 14:10:33 Production ecs-tvs2_production_worker-48-tvs2productionweb-fadce1ed94d2e3dd2100: Performed PerformancePlatformTransactionsQueueJob (Job ID: 1b2f5acf-0bdb-425e-89e6-7495dc24f378) from Sidekiq(performance_platform) in 489.38ms
```
* There is nothing I can find in the documentation of the Performance Platform that specifies we must use UTC [1], however it accepts a the iso8601 format which implies that whatever is using the timestamp is zone aware which gives me some confidence that this won’t break it
* There is nothing I can find in the original commit on the subject of using UTC or GMT [2]
* `TransactionAuditor` and `Vacancy.published_on_count` both expect a Date object however they were receiving a Time object, so this change fixes some behaviour. `published_on_count` caused a particular problem in returning an incorrect 0 count when doing a date lookup with a time value (perhaps this has been contributing to weird data in PP):
```
[2] pry(#<PerformancePlatformTransactionsQueueJob>)> Vacancy.pluck(:publish_on)
=> [Mon, 01 Apr 2019, Mon, 01 Apr 2019, Mon, 01 Apr 2019]
[4] pry(#<PerformancePlatformTransactionsQueueJob>)> Vacancy.published_on_count(date)
=> 0
[5] pry(#<PerformancePlatformTransactionsQueueJob>)> date
=> Mon, 01 Apr 2019 00:00:00 BST +01:00
```
* The fact that the rake task injects a date parameter into this job that is 1 day behind of the current date took me down the wrong path of removing this wasn’t counting all the jobs published on a given day. It seems this has been done as there are 2 different ways the job is being called, short of refactoring this all I’ve tried to be more expressive with the test variables. In particular that the runtime of the task and the date_to_upload dates are different values.

[1] https://performance-platform.readthedocs.io/en/latest/api/write-api.html
https://github.com/alphagov/performanceplatform-client.py
[2] https://github.com/dxw/teacher-vacancy-service/commit/e6632ccd3f7ec4cb4f605354dbfc4190a6a9bbd4#diff-c995bd0b523d3323e79766b163320d94

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -4,7 +4,7 @@ class Feedback < ApplicationRecord
 
   validates :rating, presence: true
 
-  scope :published_on, (->(date) { where('date(created_at) = ?', date) })
+  scope :published_on, (->(date) { where(created_at: date.all_day) })
 
   def to_row
     [
--- a/spec/jobs/performance_platform_feedback_queue_job_spec.rb
+++ b/spec/jobs/performance_platform_feedback_queue_job_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe PerformancePlatformFeedbackQueueJob, type: :job do
     expect(PerformancePlatform::UserSatisfaction).to receive(:new)
       .with('user-satisfaction-token')
       .and_return(user_satisfaction)
-    expect(user_satisfaction).to receive(:submit).with(ratings, date.utc.iso8601)
+    expect(user_satisfaction).to receive(:submit).with(ratings, date.iso8601)
 
     perform_enqueued_jobs { job }
   end
--- a/spec/jobs/performance_platform_feedback_queue_job_spec.rb
+++ b/spec/jobs/performance_platform_feedback_queue_job_spec.rb
@@ -3,8 +3,8 @@ require 'rails_helper'
 RSpec.describe PerformancePlatformFeedbackQueueJob, type: :job do
   include ActiveJob::TestHelper
 
-  subject(:date) { Date.current.beginning_of_day.in_time_zone }
-  subject(:job) { described_class.perform_later(date.to_s) }
+  subject(:date_to_upload) { Date.current.beginning_of_day.in_time_zone - 1.day }
+  subject(:job) { described_class.perform_later(date_to_upload.to_s) }
 
   it 'queues the job' do
     expect { job }.to change(ActiveJob::Base.queue_adapter.enqueued_jobs, :size).by(1)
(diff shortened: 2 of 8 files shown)


tahb:
[Intermittent performance platform test, clocks changed?](https://trello.com/c/aj6ikgAo/817-intermittent-performance-platform-test-clocks-changed)

spec/jobs/performance_platform_transactions_queue_job_spec.rb
```diff
@@ -14,15 +24,18 @@
...
-  it 'executes perform' do
+  it 'enqueues a job to send all published job from yesterday' do
     stub_const('PP_TRANSACTIONS_BY_CHANNEL_TOKEN', 'not-nil')
 
     pp = double(:performance_platform)
 
-    published = create_list(:vacancy, 3, :published, publish_on: date)
+    job_1_published_yesterday = build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false)
+    job_2_published_yesterday = build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false)
+    jobs_published_yesterday = [job_1_published_yesterday, job_2_published_yesterday]
```
pezholio: Probably best to add some jobs published today and further in the past to check they don't get submitted

app/models/feedback.rb
```diff
@@ -4,7 +4,7 @@ class Feedback < ApplicationRecord
 
   validates :rating, presence: true
 
-  scope :published_on, (->(date) { where('date(created_at) = ?', date) })
+  scope :published_on, (->(date) { where(created_at: date.all_day) })
```
tahb: I did not know of this, nicee 👌

pezholio: Yeah, it's basically a shorthand for `date.beginning_of_day..date.end_of_day` (which is a mouthful). New in Rails > 5.1

spec/jobs/performance_platform_feedback_queue_job_spec.rb
```diff
@@ -19,16 +19,27 @@
...
+    create_list(:feedback, 4, rating: 4, created_at: two_days_ago)
+
+    rating3 = create_list(:feedback, 2, rating: 3, created_at: date_to_upload)
+    rating5 = create_list(:feedback, 3, rating: 5, created_at: date_to_upload)
+
+    ratings = {
+      1 => 0,
+      2 => 0,
+      3 => rating3.count,
+      4 => 0,
```
tahb: 👍

spec/jobs/performance_platform_transactions_queue_job_spec.rb
```diff
@@ -14,15 +24,24 @@
...
 
-    published = create_list(:vacancy, 3, :published, publish_on: date)
+    jobs_published_yesterday = [
+      build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false),
+      build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false)
+    ]
 
     expect(PerformancePlatform::TransactionsByChannel).to receive(:new).with('not-nil').and_return(pp)
-    expect(pp).to receive(:submit_transactions).with(published.count, date.utc.iso8601)
+    expect(pp).to receive(:submit_transactions).with(jobs_published_yesterday.count, date_to_upload.iso8601)
```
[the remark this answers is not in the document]
pezholio: Yeah, that's a good shout, gives us more confidence. I'll have a look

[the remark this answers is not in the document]
pezholio: Hmmm... Doesn't `and_return` just specify a return value, rather than expect a return value? I think a better way to do this would be break it out into a method and test that method.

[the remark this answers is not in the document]
pezholio: TBH, I'm not a fan of this much logic in a job anyway. This really should be in a testable class

PR #837, opened 2019-04-01
[803] Make job alert references required and generate defaults from search criteria

erbridge:
## Trello card URL:

https://trello.com/c/npmtxxuy/803-reference-numbers-for-job-alerts-should-not-be-machine-generated-but-a-mandatory-human-reference-so-that-the-alert-reference-is

## Changes in this PR:

- Removes the reference setter in favour of requiring presence.
- Update the reference hint copy.
- Generate a default reference from keyword and location search criteria, if present. If neither are present, leave the field blank for the user to fill. Location and keyword were chosen as 93% of searches include location and 44% keyword, the top two optional fields.
- Default to an empty hash when requesting a hash of search criteria in the case where the cirteria is missing or malformed.

### Before

Screenshot 2019-04-04 at 14 12 13

### After

Screenshot 2019-04-04 at 14 08 50
Screenshot 2019-04-04 at 14 08 20
Screenshot 2019-04-04 at 14 07 31

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -12,6 +12,8 @@ class Subscription < ApplicationRecord
 
   scope :ongoing, -> { where('expires_on >= current_date') }
 
+  after_initialize :default_reference
+
   def self.encryptor
     key_generator_secret = SUBSCRIPTION_KEY_GENERATOR_SECRET
     key_generator_salt = SUBSCRIPTION_KEY_GENERATOR_SALT
@@ -68,5 +70,42 @@ class Subscription < ApplicationRecord
 
   def expired?
     expires_on < Time.zone.today
+  end
+
+  private
+
+  def needs_reference?
+    new_record? && reference.blank?
+  end
+
+  def keyword_reference_part
+    search_criteria_to_h if @search_criteria_hash.nil?
+
+    @search_criteria_hash['keyword'].strip.split(/\s+/).join(' ') if @search_criteria_hash.key?('keyword')
+  end
+
+  def location_reference_part
+    search_criteria_to_h if @search_criteria_hash.nil?
+
+    has_location = ['location', 'radius'].all? { |key| @search_criteria_hash.key?(key) }
+
+    "within #{@search_criteria_hash['radius']} miles of #{@search_criteria_hash['location'].strip}" if has_location
+  end
+
+  def default_reference
+    return unless needs_reference?
+
+    keyword_part = keyword_reference_part
+    location_part = location_reference_part
+
+    return if keyword_part.blank? && location_part.blank?
+
+    self.reference = if keyword_part.present?
+                       "#{keyword_part.upcase_first} jobs"
+                     else
+                       'Jobs'
+                     end
+
+    self.reference += " #{location_part}" if location_part.present?
   end
 end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -36,14 +36,10 @@ class Subscription < ApplicationRecord
   end
 
   def search_criteria_to_h
-    parsed_criteria = (JSON.parse(search_criteria) if search_criteria.present?)
-    @search_criteria_hash = if parsed_criteria.is_a?(Hash)
-                              parsed_criteria
-                            else
-                              {}
-                            end
+    parsed_criteria = JSON.parse(search_criteria) if search_criteria.present?
+    parsed_criteria.is_a?(Hash) ? parsed_criteria : {}
   rescue JSON::ParserError
-    @search_criteria_hash = {}
+    {}
   end
 
   def token(expiration_in_days: 2)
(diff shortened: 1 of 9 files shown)


erbridge:
[Reference numbers for job alerts should not be machine generated but a mandatory human reference so that the alert reference is clear and understandable](https://trello.com/c/npmtxxuy/803-reference-numbers-for-job-alerts-should-not-be-machine-generated-but-a-mandatory-human-reference-so-that-the-alert-reference-is)

app/models/subscription.rb
```diff
@@ -69,4 +71,41 @@ def create_alert_run
   def expired?
     expires_on < Time.zone.today
   end
+
+  private
+
+  def needs_reference?
+    new_record? && reference.blank?
```
pezholio: This is probably moot, as `after_initialize` callback only gets called on a new record

app/models/subscription.rb
```diff
@@ -69,4 +71,41 @@ def create_alert_run
...
+
+  def location_reference_part
+    search_criteria_to_h if @search_criteria_hash.nil?
+
+    has_location = ['location', 'radius'].all? { |key| @search_criteria_hash.key?(key) }
+
+    "within #{@search_criteria_hash['radius']} miles of #{@search_criteria_hash['location'].strip}" if has_location
+  end
+
+  def default_reference
```
pezholio: I think this would be better broken out into a service object (in `app/services`). Will keep the model code nice and tidy and allow it to be more easily testable.

spec/models/subscription_spec.rb
```diff
@@ -61,6 +54,32 @@
...
+          location: 'Somewhere', radius: 30, keyword: 'some keywords'
+        }.to_json)
+
+        expect(subscription.reference).to eq('Some keywords jobs within 30 miles of Somewhere')
+      end
+    end
+
+    context 'when no common search criteria is provided' do
+      it 'does not set a default reference' do
+        subscription = Subscription.new(search_criteria: { radius: 20, phase: 'primary' }.to_json)
```
pezholio: Could we possibly add support for phases, or some other filters too? Would be nice if we could catch the vast majority of cases

app/models/subscription.rb
```diff
@@ -34,7 +34,14 @@ def self.find_and_verify_by_token(token)
   end
 
   def search_criteria_to_h
-    @search_criteria_hash = JSON.parse(search_criteria)
+    parsed_criteria = (JSON.parse(search_criteria) if search_criteria.present?)
```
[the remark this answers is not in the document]
erbridge: This went through so many iterations to get it working :). I'll look at doing that. Definitely like that memoization solution.

app/models/subscription.rb
erbridge: The Stack Overflow I read that told me how to do this suggested it. I will do more reading.

spec/models/subscription_spec.rb
erbridge: I think we cover most with the two we're checking, and it makes the references a lot more complex to generate if we turn some on and off depending on what's there. If we don't it gets very long. Not sure what the best approach is.

app/models/subscription.rb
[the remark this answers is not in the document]
erbridge: I decided to remove the memoization. The function isn't pure, so updates could cause a mismatch between the hash and the source data.

app/models/subscription.rb
erbridge: According to https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html, `after_initialize` is also called after a find. We don't want to add references to existing records, so I think this is required.

app/models/subscription.rb
erbridge: Done.

app/services/subscription_reference_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+
+  def keyword_part
+    @search_criteria['keyword'].strip.split(/\s+/).join(' ') if @search_criteria.key?('keyword')
+  end
+
+  def location_part
+    has_location = ['location', 'radius'].all? { |key| @search_criteria.key?(key) }
```
pezholio: This would be more expressive if it was a method called `has_location?`. You could also have the first line as a guard clause: `return unless has_location?`

app/services/subscription_reference_generator.rb
```diff
@@ -0,0 +1,32 @@
+class SubscriptionReferenceGenerator
+  def initialize(search_criteria:)
+    update(search_criteria)
+  end
+
+  def update(search_criteria)
```
pezholio: I'm not sure we need this method here (unless it's being used somewhere I can't see) - by convention we set instance variables in the initializer

app/services/subscription_reference_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+    @search_criteria = search_criteria
```
pezholio: Another point on instance variables, if you add `attr_reader :search_criteria`, you'll be able to access the instance variable as a standard variable, just another convention we use elsewhere, so would be good to be consistent.

app/services/subscription_reference_generator.rb
erbridge: Done.

app/services/subscription_reference_generator.rb
erbridge: Yeah. I put it in because I thought I would use it but didn't. Removing it.

app/services/subscription_reference_generator.rb
erbridge: Ah. Good to know.

PR #839, opened 2019-04-02
[819] Rails receives the correct client IP address

tahb:
## Trello card URL:
https://trello.com/c/IvJtATi7

## Changes in this PR:
* Change of logic to ensure we trust the current IP address of cloudfront. Through trusting we enable the Rails method `request.remote_ip` to access the correct client IP address from the X-Forwarded-For header. With that, it is used as the IP in logs, and forwarded to Rollbar
* The current IP address by our CloudFront install is not in the GLOBAL and is part of eu-west-2 region which is the regions our environments are configured to use. Found as the value for `region` within the relevant .tfvars. Presumably this has been moved by AWS and we’ve only just noticed the change

## Screenshots of UI changes:
Screenshot 2019-04-02 at 12 10 52

### Before
Screenshot 2019-04-02 at 17 00 52

### After
Screenshot 2019-04-02 at 17 00 42

### DIFF
--- a/lib/modules/aws_ip_ranges.rb
+++ b/lib/modules/aws_ip_ranges.rb
@@ -1,6 +1,7 @@
 module AWSIpRanges
   # Used based on this AWS instruction: https://forums.aws.amazon.com/ann.jspa?annID=2051
   PATH = 'https://ip-ranges.amazonaws.com/ip-ranges.json'.freeze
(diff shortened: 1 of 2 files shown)


tahb:
[Correct client IP address is no longer being received in our logs or in Rollbar](https://trello.com/c/IvJtATi7/819-correct-client-ip-address-is-no-longer-being-received-in-our-logs-or-in-rollbar)

PR #842, opened 2019-04-10
[834] Terraform uses newer version to deploy

tahb:
## Trello card URL:
https://trello.com/c/SobmxM2m

## Changes in this PR:
* Changing to 0.11.13 as we have at least deployed out staging environment with this version, causing it to be locked from further deployments with earlier versions.
* Although this is only a patch version change, the release logs for 12 and 13 don’t have anything alarming: https://github.com/hashicorp/terraform/releases/tag/v0.11.12, https://github.com/hashicorp/terraform/releases/tag/v0.11.13

### Before
Screenshot 2019-04-10 at 14 06 22

### After

Screenshot 2019-04-10 at 14 15 24
Screenshot 2019-04-10 at 14 19 21

No new unexpected changes presented from a `plan`:
Screenshot 2019-04-10 at 14 20 09

## Next steps:

### DIFF
--- a/terraform.tf
+++ b/terraform.tf
@@ -12,7 +12,7 @@ Store infrastructure state in a remote store (instead of local machine):
 https://www.terraform.io/docs/state/purpose.html
 */
 terraform {
-  required_version = "~> 0.11.11"
+  required_version = "~> 0.11.13"
 
   backend "s3" {
     bucket  = "terraform-state-002"


tahb:
[Bump Terraform to 0.11.13 to reenable staging deployments](https://trello.com/c/SobmxM2m/834-bump-terraform-to-01113-to-reenable-staging-deployments)

terraform.tf
```diff
@@ -12,7 +12,7 @@ Store infrastructure state in a remote store (instead of local machine):
 https://www.terraform.io/docs/state/purpose.html
 */
 terraform {
-  required_version = "~> 0.11.11"
+  required_version = "~> 0.11.13"
```
erbridge: We're going to have this problem again as soon as a new version of Terraform is released. Would it be better to pin to an exact version, rather than the minor version?

tahb: Very true, it will happen again.

Seems reasonable to lock it exactly. For gems we'd like an easy and fast way to get bug and security fixes so the minor lock makes sense, Terraform isn't running any code in production per say and it's hard to imagine what an urgent security fix would look like!

tahb:
Closing this as discussed the important lock is the aws provider as that has the real potential to break something.

I believe the next steps are to review if a major lock is valuable, or to remove the lock entirely and to document the decision that we must continuously update Terraform as a team, as we might do with Docker.

PR #843, opened 2019-04-11
Improve bin scripts

erbridge:
### `bin/drebuild`

- Exit on any errors.
- Take the Docker environments down before cleaning images. This means that we're really starting fresh as intended. It also prevents the down after the test build from taking out the test server.
- Pretty print the stage logging.
- Stop leaving orphan web images during a rebuild by adding `--rm` to the `docker-compose run`s (again).

### `bin/dspec`

- Stop relying on specific docker image names for running specs.

### `docker-compose.env.sample`

- Add a missing env var: `NOTIFY_SUBSCRIPTION_DAILY_TEMPLATE`.


bin/dspec
```diff
@@ -4,8 +4,8 @@ set -e
 if [ $# -eq 0 ]
 then
   echo "No arguments supplied. Defaults to 'rake default'"
-  docker exec -it teacher-vacancy-service_test_1 rake default $@
+  docker-compose --file=docker-compose.test.yml exec test rake default $@
```
tahb: Much better 👍

PR #844, opened 2019-04-11
(WIP) User location search radius should be geographically not as the crow flies

pezholio:
## Trello card URL:

https://trello.com/c/Zi8Eu8w9/513-user-location-search-radius-should-be-geographically-not-as-the-crow-flies

## Changes in this PR:

This PR uses the [Google Distance Matrix API](https://developers.google.com/maps/documentation/distance-matrix/start) to check if vacancies are really in an area before presenting them to a user. This can occur if a user is searching a large radius in somewhere like Cardiff where there are vacancies over the Bristol Channel in somewhere like Weston-Super-Mare. 

It works like so:

* User enters a location and radius
* We geocode the location, and search for schools within that radius as the crow flies (using the [GeoKit gem](https://github.com/geokit/geokit), which, under the hood, generates a Postgres query using the [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula)
* We then pass the user's location and the latlngs to the Distance Matrix API, filtering any schools that are not actually reachable by land
* We then pass the IDs of those schools to the ElasticSearch query, so we know that only vacancies at those schools will be returned

The only drawback with this is that we are now making an extra API call to Google when a user carries out a distance search, which may come with extra costs / a performance hit. Might be worth getting this up on Edge and see if we notice any change in Skylight.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -157,6 +157,12 @@ GEM
       representable (~> 3.0)
       retriable (>= 2.0, < 4.0)
       signet (~> 0.10)
+    google_business_api_url_signer (0.1.3)
+      activesupport (>= 3.2.0)
+    google_distance_matrix (0.6.1)
+      activemodel (>= 3.2.13, < 5.3)
+      activesupport (>= 3.2.13, < 5.3)
+      google_business_api_url_signer (~> 0.1.3)
     google_drive (3.0.2)
       google-api-client (>= 0.11.0, < 0.29.0)
       googleauth (>= 0.5.0, < 1.0.0)
@@ -516,6 +522,7 @@ DEPENDENCIES
   geocoder
   geokit-rails
   google-api-client
+  google_distance_matrix
   google_drive
   gov_uk_date_fields!
   haml-rails
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -143,6 +143,10 @@ GEM
     friendly_id (5.2.5)
       activerecord (>= 4.0.0)
     geocoder (1.5.1)
+    geokit (1.13.1)
+    geokit-rails (2.3.1)
+      geokit (~> 1.5)
+      rails (>= 3.0)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     google-api-client (0.25.0)
@@ -510,6 +514,7 @@ DEPENDENCIES
   figaro
   friendly_id
   geocoder
+  geokit-rails
   google-api-client
   google_drive
   gov_uk_date_fields!
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,5 +1,8 @@
 require 'breasal'
 class School < ApplicationRecord
+  acts_as_mappable lat_column_name: :latitude,
+                   lng_column_name: :longitude
+
   include Auditor::Model
 
   belongs_to :school_type, optional: false
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -38,6 +38,10 @@ class School < ApplicationRecord
     GoogleDistanceMatrix::Place.new(lat: latitude, lng: longitude)
   end
 
+  def geolocation?
+    (latitude && longitude).present?
+  end
+
   private
 
   def set_geolocation_from_easting_and_northing
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -34,16 +34,21 @@ class School < ApplicationRecord
   private
 
   def set_geolocation_from_easting_and_northing
-    if easting && northing
-      wgs84 = Breasal::EastingNorthing.new(
-        easting: easting.to_i,
-        northing: northing.to_i,
-        type: :gb
-      ).to_wgs84
+    self.latitude = wgs84[:latitude]
+    self.longitude = wgs84[:longitude]
+  end
 
-      geolocation = [wgs84[:latitude], wgs84[:longitude]]
-    end
+  def wgs84
+    return {} if invalid_coords?
 
-    self.geolocation = geolocation
+    Breasal::EastingNorthing.new(
+      easting: easting.to_i,
+      northing: northing.to_i,
+      type: :gb
+    ).to_wgs84
+  end
+
+  def invalid_coords?
+    easting.blank? || northing.blank?
   end
 end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -34,6 +34,10 @@ class School < ApplicationRecord
     set_geolocation_from_easting_and_northing
   end
 
+  def place
+    GoogleDistanceMatrix::Place.new(lat: latitude, lng: longitude)
+  end
+
   private
 
   def set_geolocation_from_easting_and_northing
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -45,11 +45,13 @@ class School < ApplicationRecord
   private
 
   def set_geolocation_from_easting_and_northing
+    wgs84 = easting_northing_to_latlng
+
     self.latitude = wgs84[:latitude]
     self.longitude = wgs84[:longitude]
   end
 
-  def wgs84
+  def easting_northing_to_latlng
     return {} if invalid_coords?
 
     Breasal::EastingNorthing.new(
--- a/app/services/schools_in_area.rb
+++ b/app/services/schools_in_area.rb
@@ -8,13 +8,23 @@ class SchoolsInArea
   end
 
   def results
+    schools.each_slice(25).map do |school_slice|
+      filter_schools_by_distance(school_slice)
+    end.flatten
+  end
+
+  private
+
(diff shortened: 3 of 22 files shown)


app/models/school.rb
```diff
@@ -31,19 +34,28 @@ def northing=(northing)
...
-      wgs84 = Breasal::EastingNorthing.new(
-        easting: easting.to_i,
-        northing: northing.to_i,
-        type: :gb
-      ).to_wgs84
+    self.latitude = wgs84[:latitude]
+    self.longitude = wgs84[:longitude]
+  end
+
+  def wgs84
```
erbridge: What is `wgs84`?

app/views/hiring_staff/vacancies/_show.html.haml
```diff
@@ -78,11 +78,11 @@
           %dd
             = link_to("#{@vacancy.school.name} website (opens in a new window)", @vacancy.school.url, class: 'govuk-link wordwrap', target: '_blank')
 
-      - if @vacancy.school.geolocation
+      - if @vacancy.school.latitude && @vacancy.school.longitude
```
erbridge: We check these a lot. Perhaps refactor to have `.geolocation?` or something?

db/migrate/20190408143457_add_lat_and_lng_to_schools.rb
```diff
@@ -0,0 +1,7 @@
+class AddLatAndLngToSchools < ActiveRecord::Migration[5.2]
+  def change
+    add_column :schools, :latitude, :float
+    add_column :schools, :longitude, :float
+    remove_column :schools, :geolocation
```
erbridge: Do we need a task to migrate the data over, too?

pezholio: I think (@tahb, correct me if I'm wrong) we should be able to run the `data:schools:import` job on deploy, which will update the schools from the source data. The only issue is there may be a small window where location searches may not work properly.

pezholio:
Gaah, so this hasn't made it very far. Just tried with a full corpus of schools, and we're already hitting limitations. There's a limit of 25 destinations per request, so even with a very small area, there's going to be more than 25 schools nearby, so I think this might not be a goer in its current form :(

erbridge:
We could do requests in batches of 25, right? Any way to do the requests concurrently so we wouldn't have to wait for a round trip to trigger the next one?

pezholio:
Yeah, we could do it in batches, but I'm worried that would slow things down quite considerably. I'm looking into other alternatives, like [Graphhopper](https://www.graphhopper.com) at the moment

pezholio:
Right, so I've made some tweaks, so now we only fetch schools that actually have vacancies, rather than fetching all schools, which was a bit of a blunt instrument. I also had some problems with VCR, so have refactored the tests a bit too and removed the dependency on VCR that I introduced.

I'd still be keen to test this against a corpus of real live data on Edge or something. Not sure of the best way to do this though, as the real litmus test would be the amount of data.

Gemfile.lock
```diff
@@ -468,6 +478,7 @@ GEM
     validate_url (1.0.6)
       activemodel (>= 3.0.0)
       public_suffix
+    vcr (4.0.0)
```
erbridge: `vcr` is still here.

app/services/schools_in_area.rb
```diff
@@ -0,0 +1,45 @@
+class SchoolsInArea
+  attr_accessor :lat, :lng, :radius
```
erbridge: Should this be `attr_reader` since we're memoizing based on these values, so if they change, our data will become invalid?

app/services/vacancy_search_builder.rb
```diff
@@ -79,15 +74,23 @@ def keyword_query
...
-        }
-      }
+      bool: {
+        filter: {
+          terms: {
+            school_id: schools_in_area,
+          },
+        },
+      },
     }
```
erbridge: This looks a little weird to me. This results in something that looks like:
```
{
  bool: {
    must: [
      bool: {
        filter: {
          terms: {
            school_id: schools_in_area,
          }
        }
      },
      ...
    ]
  }
}
```
Whereas, `filter` used to be set on the first `bool`. Was that intentional?

app/views/vacancies/_vacancies_for_map.json.jbuilder
```diff
@@ -1,7 +1,7 @@
-json.array! vacancies.select(&:school_geolocation) do |vacancy|
+json.array! vacancies.select do |vacancy|
```
erbridge: Should this be filtering by ones that have `schoool.geolocation? == true`?

app/services/schools_in_area.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def matrix
+    @matrix ||= begin
+      matrix = GoogleDistanceMatrix::Matrix.new
+      matrix.origins << GoogleDistanceMatrix::Place.new(lat: lat, lng: lng)
+      schools.each do |s|
+        matrix.destinations << s.place
+      end
+      matrix
```
erbridge: Do we need to guard against there being more than 25? Perhaps log something so we hear about it, and fall back to a basic radius search? Or make more API calls?

spec/lib/schools_in_area_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+RSpec.describe 'SchoolsInArea' do
```
erbridge: We should probably also test what happens when there are more than 25 schools to search.

spec/presenters/school_vacancies_presenter_spec.rb
```diff
@@ -35,7 +35,8 @@
 
     it 'returns sorted draft vacancies' do
       sort.update(column: 'job_title', order: 'desc')
-      expect(presenter.vacancies).to eq(draft_vacancies.sort_by { |vacancy| vacancy['job_title'] }.reverse)
+      vacancy_ids = draft_vacancies.sort_by { |vacancy| vacancy['job_title'] }.reverse.pluck(:id)
+      expect(presenter.vacancies.pluck(:id)).to eq(vacancy_ids)
```
erbridge: Think we could get these pulled out into their own PR?

app/services/schools_in_area.rb
pezholio: Good catch 👍

spec/presenters/school_vacancies_presenter_spec.rb
pezholio: Yep, we don't need this now 👍

app/services/schools_in_area.rb
pezholio: Yep, I've done this now - breaking them down into batches of 25 and making subsequent calls. I'm still curious / concerned about how this will work in practice. Are we going to end up with really poor performance on areas with lots of schools?

app/services/vacancy_search_builder.rb
pezholio: I think (and correct me if I'm wrong) that we were searching by all the search criteria, and then filtering by distance after doing the initial search. Now we're doing the reverse and the school ids are now things to be searched on, rather than filtered by. My ElasticSearch knowledge is not the best, so I could be wrong...

app/views/vacancies/_vacancies_for_map.json.jbuilder
pezholio: Good catch. I dunno what made me delete that line. That said, I don't even thing we use this partial, so I've binned it

app/services/vacancy_search_builder.rb
erbridge: The difference between `must` and `filter` is only whether the score is important. `must` uses the score, while `filter` discards it. If they're on the same level, I believe they're run concurrently, and it's a bit of a simpler read than having nested `bool` queries, IMO. I _think_ the functionality is the same, though, but I'm not sure.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html

app/views/vacancies/_vacancies_for_map.json.jbuilder
erbridge: :+1:

app/services/schools_in_area.rb
erbridge: If we log something with the number of schools, or some note about having to batch them, maybe that would help us track if it's occuring at least?

pezholio:
[User location search radius should be geographically not as the crow flies](https://trello.com/c/Zi8Eu8w9/513-user-location-search-radius-should-be-geographically-not-as-the-crow-flies)

adrianclay:
Closing as this isn't a current priority.  Can re-open if we want to restart work on this.

PR #847, opened 2019-04-11
Terraform cli versioning strategy

tahb:
* Lock to the major version to be explicit about Terraform usage when files may include breaking changes
* Passing reference to the requirement of AWS credentials being set up locally too

### DIFF
--- a/documentation/deployment.md
+++ b/documentation/deployment.md
@@ -20,7 +20,7 @@ There are currently 4 long standing environments that the team have been using, 
 As we use Terraform to make 95% of our changes to the infrastructure they too should be made in the form of pull requests into the relevant git branch. [You can read about what's not covered in our set up docs](../terraform/README.md).
 
 ### Before you start
-Get a local copy of the Terraform variables from the 1Password vault and move them into `/workspace-variables`, eg. `/workspace-variables/edge.tfvars`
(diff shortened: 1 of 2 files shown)


documentation/deployment.md
```diff
@@ -46,3 +51,8 @@ Do you want to perform these actions in workspace "testing"?
 
 #### Step 3
 Once the deployment has finished, review the output of the script to ensure everything succeeded by checking AWS to make sure the expected changes were made.
+
+## Upgrading Terraform
+We have locked Terraform to its major version to protect the team from accidentally running a deployment from a version not compatible with the Terraform files we have. This will hopefully give us explicit feedback on version issues **before** being presented with a barrage of Terraform errors throughout all the files during the last phase of a production deployment.
```
erbridge: Terraform uses the pre-release form of semver, where minor version changes are actually major releases. So if we're pinning, we should pin to minor version. See e.g. 0.12's draft upgrade guide: https://www.terraform.io/upgrade-guides/0-12.html

tahb: Good spot 

> Terraform v0.12 will be a major release focused on configuration language improvements and thus will include some changes that you'll need to consider when upgrading.

tahb:
Took some time but I feel more informed now!

PR #850, opened 2019-04-15
[818] Refactor Performance Platform jobs

davidwinter:
## Trello card URL:

https://trello.com/c/hkhh0Vfc/818-refactor-performance-platform-jobs

## Changes in this PR:

* Made the `PerformancePlatform` class a little more consistent for both Transactions and Feedback. Both now have a `submit` method in the base class, with variations in the subclasses, such as data manipulation.
* Add `PerformancePlatformSender` which takes the bulk of the logic out of the jobs, and dry's things up to work for both transactions and feedback.
* Integrate `PerformancePlatformSender` into the background jobs for transactions and feedback.

### DIFF
--- a/lib/performance_platform.rb
+++ b/lib/performance_platform.rb
@@ -26,11 +26,11 @@ module PerformancePlatform
       '/data/teaching-jobs-job-listings/transactions-by-channel'
     end
 
-    def data_map(timestamp, period, data)
+    def data_map(timestamp, period, count)
       { _timestamp: timestamp,
         service: 'teaching_jobs_listings',
         channel: 'digital',
-        count: data,
+        count: count,
         dataType: 'transactions-by-channel',
         period: period }.to_json
     end
@@ -43,15 +43,15 @@ module PerformancePlatform
       '/data/teaching-jobs-job-listings/user-satisfaction'
     end
 
-    def data_map(timestamp, period, data)
+    def data_map(timestamp, period, rating_counts)
       { _timestamp: timestamp,
-        rating_1: data[1],
-        rating_2: data[2],
-        rating_3: data[3],
-        rating_4: data[4],
-        rating_5: data[5],
+        rating_1: rating_counts[1],
+        rating_2: rating_counts[2],
+        rating_3: rating_counts[3],
+        rating_4: rating_counts[4],
+        rating_5: rating_counts[5],
         service: 'teaching_jobs_listings',
-        total: data.values.inject(:+),
+        total: rating_counts.values.inject(:+),
         dataType: 'user-satisfaction',
         period: period }.to_json
     end
--- a/lib/performance_platform.rb
+++ b/lib/performance_platform.rb
@@ -11,56 +11,49 @@ module PerformancePlatform
       @headers = { 'Authorization' => "Bearer #{token}",
                    'Content-Type' => 'application/json' }
     end
+
+    def submit(data, timestamp = Time.zone.now.iso8601, period = 'day')
+      HTTParty.post("#{PP_DOMAIN}#{transaction_endpoint}",
+                    body: data_map(timestamp, period, data),
+                    headers: headers)
+    end
   end
 
   class TransactionsByChannel < Base
-    def submit_transactions(count, timestamp = Time.zone.now.iso8601, period = 'day')
-      HTTParty.post("#{PP_DOMAIN}#{transaction_endpoint}",
-                    body: data(timestamp, period, count),
-                    headers: headers)
-    end
-
     private
 
     def transaction_endpoint
       '/data/teaching-jobs-job-listings/transactions-by-channel'
     end
 
-    def data(timestamp, period, count)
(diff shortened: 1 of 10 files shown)


lib/performance_platform.rb
```diff
@@ -11,56 +11,49 @@ def initialize(token = nil)
...
       '/data/teaching-jobs-job-listings/transactions-by-channel'
     end
 
-    def data(timestamp, period, count)
+    def data_map(timestamp, period, data)
       { _timestamp: timestamp,
         service: 'teaching_jobs_listings',
         channel: 'digital',
-        count: count,
+        count: data,
```
erbridge: We know what the data is here. I think it would be clearer to keep it named `count`.

lib/performance_platform.rb
```diff
@@ -11,56 +11,49 @@ def initialize(token = nil)
...
-      data
+    def data_map(timestamp, period, data)
+      { _timestamp: timestamp,
+        rating_1: data[1],
+        rating_2: data[2],
+        rating_3: data[3],
+        rating_4: data[4],
+        rating_5: data[5],
+        service: 'teaching_jobs_listings',
+        total: data.values.inject(:+),
```
erbridge: Similar to above, we know what the data is here. Can we name it `rating_counts` or similar to make that clear?

lib/tasks/performance_platform.rake
```diff
@@ -8,13 +8,13 @@ namespace :performance_platform do
...
-    yesteday = Date.current.beginning_of_day.in_time_zone - 1.day
-    PerformancePlatformFeedbackQueueJob.perform_later(yesteday.to_s)
+    yesterday = Date.current.beginning_of_day.in_time_zone - 1.day
+    PerformancePlatformTransactionsQueueJob.perform_later(yesterday.to_s)
   end
 
   task submit_user_satisfaction: :environment do
     yesterday = Date.current.beginning_of_day.in_time_zone - 1.day
-    PerformancePlatformTransactionsQueueJob.perform_later(yesterday.to_s)
+    PerformancePlatformFeedbackQueueJob.perform_later(yesterday.to_s)
```
erbridge: :+1:

spec/jobs/performance_platform_transactions_queue_job_spec.rb
```diff
@@ -25,22 +16,10 @@
...
-    build(:vacancy, :past_publish, publish_on: today).save(validate: false)
-
-    jobs_published_yesterday = [
-      build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false),
-      build(:vacancy, :past_publish, publish_on: date_to_upload).save(validate: false)
-    ]
+    pp = instance_double(PerformancePlatformSender::Transactions)
 
-    expect(PerformancePlatform::TransactionsByChannel).to receive(:new).with('not-nil').and_return(pp)
-    expect(pp).to receive(:submit_transactions).with(jobs_published_yesterday.count, date_to_upload.iso8601)
```
erbridge: Have we lost the test of this functionality?

spec/jobs/performance_platform_feedback_queue_job_spec.rb
```diff
@@ -15,31 +15,10 @@
...
-      2 => 0,
-      3 => rating3.count,
-      4 => 0,
-      5 => rating5.count
-    }
-
-    expect(PerformancePlatform::UserSatisfaction).to receive(:new)
-      .with('user-satisfaction-token')
-      .and_return(user_satisfaction)
-    expect(user_satisfaction).to receive(:submit).with(ratings, date_to_upload.iso8601)
```
erbridge: Have we lost the test of this functionality?

spec/jobs/performance_platform_transactions_queue_job_spec.rb
davidwinter: For the job classes yes. These specific tests have now moved into the `PerformancePlatformSender` class spec instead.

spec/jobs/performance_platform_feedback_queue_job_spec.rb
davidwinter: As above.

spec/jobs/performance_platform_transactions_queue_job_spec.rb
erbridge: My read of those suggests that we're no longer testing that the data is collected correctly in those tests.

lib/performance_platform.rb
davidwinter: 👍

lib/performance_platform.rb
davidwinter: 👍

spec/jobs/performance_platform_transactions_queue_job_spec.rb
davidwinter: You're right @erbridge - just pushed a commit to fix this. Nice catch 👍

PR #853, opened 2019-04-15
[839] Remove keyword as a search option and replace with subject and job title

erbridge:
## Trello card URL:

https://trello.com/c/iI5dMkXS/839-remove-keyword-as-a-search-option-and-replace-with-subject-and-job-title

### Add search filters for subject and job title

Job title is a simple match filter on the job_title field. It
searches for all given words in the fields with automatic
fuzziness according to Elasticsearch's defaults.

Subject is a more complex query. It searches the three subject
fields in addition to the job title. Querying job title is needed
because subjects are optional, and job title might be the only
place the subject is specified. The query scoring is weighted
towards main subject, then the two supporting subjects, then job
title. It is an "or" query, meaning results containing any of the
search terms will match.

This also refactors the keyword query so we can reuse the fallback
to the null query if the filter is empty.

Alert references now use subject and job title rather than keyword.

Elasticsearch documentation on fuzziness:
https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness

### Remove keyword filter from search

This refactors keyword filter to only be accessible by alerts by
creating a subclass of `VacancyFilters` called
`VacancyAlertFilters` that extends functionality. The hope is that
one day, we can remove the keyword filter from this, once there
are no existing suibscriptions using it.

This also changes most of the tests to use subject as the primary
filter over keyword, given that is it now the most complex filter.

There is also a slight change in how we test stopwords in
searches. A search for "Part" was being used while a subject
included "Art". With the keyword search, a prefix length is set,
meaning searches for "Art" had to start with "A". The new search
fields are more flexible than that (hoping to catch more typos),
which was causing "Part" searches to find "Art". This is because
the stopwords analyzer (correctly) isn't being used on the subject
fields, so the test was succeeding by coincidence.

### Misc extras

- Pull the radius label into the i18n doc.
- Use `VacancyFilters::AVAILABLE_FILTERS` rather than repeating them everywhere.

### Before

image

image

### After

image

image


erbridge:
[Remove 'keyword' as a search option and replace with 'subject' and  'job title'](https://trello.com/c/iI5dMkXS/839-remove-keyword-as-a-search-option-and-replace-with-subject-and-job-title)

spec/features/job_seekers_can_filter_vacancies_spec.rb
```diff
@@ -56,6 +56,50 @@
     end
   end
 
+  scenario 'Filterable by subject', elasticsearch: true do
+    headmaster = create(:vacancy, :published, job_title: 'Headmaster')
```
pezholio: These can probably be set in a `context` block using `let`s, just to DRY up the tests a bit

config/locales/en.yml
```diff
@@ -200,6 +200,7 @@ en:
       title: 'Search job listings'
       location: 'Location'
       location_hint: 'Postcode'
+      radius: 'Radius'
```
pezholio: 👍

app/services/vacancy_alert_filters.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize(args)
+    args = ActiveSupport::HashWithIndifferentAccess.new(args)
+
+    @keyword = args[:keyword]
+
+    super(args)
+  end
+
+  def to_hash
+    super().merge(keyword: keyword)
```
pezholio: I don't think you need the `()` if there's no args to pass to `super`

app/services/vacancy_alert_filters.rb
```diff
@@ -0,0 +1,17 @@
+class VacancyAlertFilters < VacancyFilters
+  AVAILABLE_FILTERS = %i[keyword].concat(superclass::AVAILABLE_FILTERS).freeze
```
pezholio: Niiiiice 👍

config/locales/en.yml
```diff
@@ -204,7 +204,7 @@ en:
       subject: 'Subject'
       subject_hint: 'e.g. Maths'
       job_title: 'Job title'
-      job_title_hint: 'e.g. Principal'
+      job_title_hint: 'e.g. Leader'
```
pezholio: Do we know this is a term teachers commonly use?

app/controllers/vacancies_controller.rb
```diff
@@ -41,9 +41,7 @@ def params
   private
 
   def search_params
-    params.permit(:subject, :job_title, :location, :radius,
-                  :minimum_salary, :maximum_salary, :phase,
-                  :phase, :working_pattern, :newly_qualified_teacher).to_hash
+    params.permit(*VacancyFilters::AVAILABLE_FILTERS).to_hash
```
pezholio: All for DRYing things up 👍

app/services/vacancy_alert_filters.rb
erbridge: My understanding of what I read was that `super.merge()` would try to call the `merge` method on the superclass. Did I get that wrong?

config/locales/en.yml
erbridge: Lots of jobs seem to use it 🤷‍♀️. Most of the searches seem to be for "teacher" which I'd like to discourage.

app/services/vacancy_alert_filters.rb
pezholio: Nah, that's right, I just don't think you need the brackets as there's no arguments to the method in the superclass. I could be wrong like (I frequently am 😆 )

spec/features/job_seekers_can_filter_vacancies_spec.rb
erbridge: Done!

PR #855, opened 2019-04-15
[696] Redesign the feedback form

erbridge:
## Trello card URL:

https://trello.com/c/fG04DWAP/696-re-design-the-feedback-form-to-gather-most-relevant-data-and-to-consider-ways-to-promote-more-responses-from-hiring-staff-resear

## Changes in this PR:

- Replace the SurveyMonkey form with an internally managed one called `GeneralFeedback`.
- Rename `Feedback` to `VacancyPublishFeedback`.
- Combine user satisfaction from `GeneralFeedback` with that of `VacancyPublishFeedback` for Performance Platform.
- Export `GeneralFeedback` to a spreadsheet as part of the `update_spreadsheets` rake task.
- Refactor feedback views to use shared partials.
- Rename the audit key for `VacancyPublishFeedback`.

### After

Screen Shot 2019-04-15 at 18 24 07

## Next steps:

- [ ] Add a new sheet to the audit spreadsheet
- [ ] Add the GID of the new sheet to the `tfvars` files
- [ ] Deploy Terraform


erbridge:
[Re-design the feedback form to gather most relevant data and to consider ways to promote more responses from hiring staff research](https://trello.com/c/fG04DWAP/696-re-design-the-feedback-form-to-gather-most-relevant-data-and-to-consider-ways-to-promote-more-responses-from-hiring-staff-resear)

lib/tasks/performance_platform.rake
```diff
@@ -3,8 +3,8 @@ require 'performance_platform'
 namespace :performance_platform do
   desc 'Performance Platform transaction submission'
   task submit: :environment do
-    Rake::Task['performance_platform:submit_transactions'].invoke
     Rake::Task['performance_platform:submit_user_satisfaction'].invoke
+    Rake::Task['performance_platform:submit_transactions'].invoke
```
pezholio: So what was happening before? Was the call to `performance_platform:submit_transactions` stopping `performance_platform:submit_user_satisfaction` from being invoked?

app/helpers/feedback_helper.rb
```diff
@@ -1,8 +1,8 @@
 module FeedbackHelper
   def visit_purpose_options
-    [['Find a job in teaching', 1],
-     ['List a teaching job on the service', 2],
-     ['Something else (tell us in the box below)', 3]]
+    [['Find a job in teaching', :find_teaching_job],
```
pezholio: These should probably be translations I reckon

docker-compose.env.sample
```diff
@@ -14,10 +14,13 @@ PP_USER_SATISFACTION_TOKEN=
 GOOGLE_API_JSON_KEY=
 GOOGLE_ANALYTICS_PROFILE_ID=
 AUDIT_SPREADSHEET_ID=
+AUDIT_VACANCIES_WORKSHEET_GID=
+AUDIT_FEEDBACK_WORKSHEET_GID=
+AUDIT_EXPRESS_INTEREST_WORKSHEET_GID=
```
pezholio: Nice spot 👍

db/migrate/20190410110530_remove_email_from_general_feedbacks.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveEmailFromGeneralFeedbacks < ActiveRecord::Migration[5.2]
```
pezholio: This stuff be better to be rebased into the initial migrations, rather than added as a seperate commit if we're never going to use it, just to keep the commit history and migrations tidy

app/views/hiring_staff/terms_and_conditions/show.html.haml
```diff
@@ -5,7 +5,7 @@
     - if @terms_and_conditions_form.errors.any?
       #errors.govuk-error-summary{role: 'alert', tabindex: '-1'}
         %h2.govuk-error-summary__title
-          = t('errors.feedback.errors_present')
+          = t('errors.terms_and_conditions.errors_present')
```
pezholio: 👍  That would've bugged me if I saw it 😆

lib/tasks/performance_platform.rake
erbridge: Ah no. I swapped these because I also fixed the fact that the tasks were running the wrong commands, and wanted to preserve the order they actually ran in. The fix for the mis-labelled tasks went in with the PP refactor.

app/helpers/feedback_helper.rb
erbridge: :+1:

db/migrate/20190410110530_remove_email_from_general_feedbacks.rb
erbridge: The problems with doing that is that it would require squashing a bunch of the commit history to make sure all commits refer to a working project state that can be revisited without too much pain (e.g. when bisecting). This PR went through a few different versions, hence the long commit history. I'm happy to try and squash the history a bunch to make it a clearer path to the current version if you think it's important.

lib/tasks/performance_platform.rake
erbridge: (see 50888c5)

db/migrate/20190410110530_remove_email_from_general_feedbacks.rb
pezholio: Fair enough. It's probably not worth the pain to rebase 👍

PR #857, opened 2019-04-16
Revise the reference hint text

erbridge:
## Trello card URL:

https://trello.com/c/npmtxxuy/803-reference-numbers-for-job-alerts-should-not-be-machine-generated-but-a-mandatory-human-reference-so-that-the-alert-reference-is

## Changes in this PR:

New reference text based on feedback from Fiona.

### DIFF
--- a/config/locales/activerecord.yml
+++ b/config/locales/activerecord.yml
@@ -65,4 +65,4 @@ en:
         reference: 'Your reference'
     hints:
       defaults:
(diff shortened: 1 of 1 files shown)


erbridge:
[Reference numbers for job alerts should not be machine generated but a mandatory human reference so that the alert reference is clear and understandable](https://trello.com/c/npmtxxuy/803-reference-numbers-for-job-alerts-should-not-be-machine-generated-but-a-mandatory-human-reference-so-that-the-alert-reference-is)

PR #861, opened 2019-04-17
[821] Make it clearer as to why a deleted job posting is no longer available on the service

erbridge:
## Trello card URL:

https://trello.com/c/fBHon4UL/821-make-it-clearer-as-to-why-a-deleted-job-posting-is-no-longer-available-on-the-service

## Changes in this PR:

Add a new error view with more specific copy if the vacancy the user is trying to get to was trashed.

### Before

Screen Shot 2019-04-17 at 14 22 10

### After

Screen Shot 2019-04-17 at 14 22 24


erbridge:
[Make it clearer as to why a deleted job posting is no longer available on the service](https://trello.com/c/fBHon4UL/821-make-it-clearer-as-to-why-a-deleted-job-posting-is-no-longer-available-on-the-service)

app/controllers/vacancies_controller.rb
```diff
@@ -24,7 +24,14 @@ def index
   end
 
   def show
-    vacancy = Vacancy.listed.friendly.find(id)
+    begin
+      vacancy = Vacancy.listed.friendly.find(id)
+    rescue ActiveRecord::RecordNotFound
+      raise unless Vacancy.trashed.friendly.exists?(id)
```
pezholio: I like the implementation, but can we get some tests for this?

pezholio: You can probably use Rspec's [render template matcher](https://relishapp.com/rspec/rspec-rails/docs/matchers/render-template-matcher) in a controller test for this

app/controllers/vacancies_controller.rb
```diff
@@ -24,7 +24,14 @@ def index
...
+
+      return render '/errors/trashed_vacancy_found'
```
pezholio: Also, can you add `status: :not_found` here too, so the server returns the correct status code?

erbridge: :+1:

PR #865, opened 2019-04-18
Pin to latest instead of a specific task revisions

erbridge:
## Changes in this PR:

My hunch is that our deployment process is creating new task
definitions, but Terraform doesn't pick up on them because it's a
computed property of a task definiton. Because we were pinning to
specific task revisions, we were rolling back to the last revision
Terraform knew about, which was the last revision it created. This
change makes it so that we pin to latest, which will always be the
most recent definition.

The revisions are pinned to latest by default, but are
overridable.

### Before

image

### After

image

## Next steps:

### DIFF
--- a/terraform/modules/ecs/ecs.tf
+++ b/terraform/modules/ecs/ecs.tf
@@ -43,7 +43,7 @@ resource "aws_ecs_service" "web" {
   name            = "${var.ecs_service_web_name}"
   iam_role        = "${aws_iam_role.ecs_role.arn}"
   cluster         = "${aws_ecs_cluster.cluster.id}"
-  task_definition = "${aws_ecs_task_definition.web.arn}"
+  task_definition = "${replace("${aws_ecs_task_definition.web.arn}", "/:${aws_ecs_task_definition.web.revision}$$/", ":${var.ecs_service_web_task_revision}")}"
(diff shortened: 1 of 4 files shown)


terraform/modules/ecs/ecs.tf
```diff
@@ -43,7 +43,7 @@ resource "aws_ecs_service" "web" {
   name            = "${var.ecs_service_web_name}"
   iam_role        = "${aws_iam_role.ecs_role.arn}"
   cluster         = "${aws_ecs_cluster.cluster.id}"
-  task_definition = "${aws_ecs_task_definition.web.arn}"
+  task_definition = "${replace("${aws_ecs_task_definition.web.arn}", "/:${aws_ecs_task_definition.web.revision}$$/", ":${var.ecs_service_web_task_revision}")}"
```
pezholio: Do we need that `$$` at the end of the regex, or is that a typo?

erbridge: `$$` escapes to `$`, i.e. ensuring the string ends with the revision. Revisions are numbers, and the ARN also includes numbers earlier on, so we want to make sure we're replacing the correct instance of the number.

pezholio: Ah, it's an escape character, of course!

PR #868, opened 2019-04-23
Move to Ordnance Survey for geocoding locations

pezholio:
## Trello card URL:

https://trello.com/c/BbfW4tBz/850-move-to-os-for-geocoding-locations

## Changes in this PR:

I've forked the [Geocoder gem](https://github.com/dxw/geocoder/tree/add-os-name-support) to support the [Ordnance Survey OpenNames API](https://apidocs.os.uk/docs/os-names-overview) when searching by location. 

This PR uses that fork and tweaks the config a bit, as well as adding a Terraform environment variable for the API Key.

This should cut down on the amount of calls to Google as we've just been billed a not insignificant amount for use of the Geocoding API. It makes sense for use to use a free, government backed service where we can.

## Next steps:


pezholio:
[Move to Ordnance Survey for geocoding locations](https://trello.com/c/BbfW4tBz/850-move-to-ordnance-survey-for-geocoding-locations)

docker-compose.env.sample
```diff
@@ -9,6 +9,7 @@ AUTHORISATION_SERVICE_URL=
 AUTHORISATION_SERVICE_TOKEN=
 GOOGLE_DRIVE_JSON_KEY=
 GOOGLE_GEOCODING_API_KEY=
+OS_GEOCODING_API_KEY=
```
erbridge: Can we be verbose about using Ordinance Survey instead of OS for clarity (since in most computer contexts, OS is Operating System)?

docker-compose.env.sample
```diff
@@ -9,6 +9,7 @@ AUTHORISATION_SERVICE_URL=
 AUTHORISATION_SERVICE_TOKEN=
 GOOGLE_DRIVE_JSON_KEY=
 GOOGLE_GEOCODING_API_KEY=
```
erbridge: If we're not using this anymore, can we drop it?

terraform.tf
```diff
@@ -147,6 +147,7 @@ module "ecs" {
   audit_subscription_creation_worksheet_gid = "${var.audit_subscription_creation_worksheet_gid}"
   domain                           = "${var.domain}"
   google_geocoding_api_key         = "${var.google_geocoding_api_key}"
```
erbridge: Same. If we're not using this, can we remove it from the Terraform?

config/initializers/geocoder.rb
```diff
@@ -1,11 +1,10 @@
 redis = Redis.new(url: ENV['REDIS_CACHE_URL'])
-key = ENV.fetch('GOOGLE_GEOCODING_API_KEY', '')
+key = ENV.fetch('OS_GEOCODING_API_KEY', '')
 
 Geocoder.configure(
-  lookup: :google,
+  lookup: :os_names,
   api_key: key,
   timeout: 5,
-  units: :mi,
```
erbridge: Does OS default to miles or something?

docker-compose.env.sample
pezholio: I'd rather keep it around for now, as we might end up using it in #844

config/initializers/geocoder.rb
pezholio: Whoops - dunno why I deleted this. It's not actually relevant to the Geocoding part of the gem, it's used for calculations (https://github.com/alexreisner/geocoder#geospatial-calculations), which we don't use, but may use in future.

variables.tf
```diff
@@ -352,6 +352,8 @@ variable "google_drive_json_key" {
 
 variable "audit_spreadsheet_id" {}
 variable "google_geocoding_api_key" {}
+variable "variable "google_geocoding_api_key" {}
+" {}
```
erbridge: This looks bad.

config/initializers/geocoder.rb
```diff
@@ -1,8 +1,8 @@
 redis = Redis.new(url: ENV['REDIS_CACHE_URL'])
-key = ENV.fetch('GOOGLE_GEOCODING_API_KEY', '')
+key = ENV.fetch('ORDNANCE_SURVEY_API_KEY', '')
 
 Geocoder.configure(
-  lookup: :google,
+  lookup: :os_names,
```
erbridge: Have we got enough testing to be confident that this replacement Just Works?

pezholio:
Ready for another look now. I've written tests in the gem, so these should be sufficient to give us confidence that switching providers will work 👍

docker-compose.env.sample
erbridge: Ok, fair enough.

variables.tf
pezholio: Gaaaaaah

config/initializers/geocoder.rb
pezholio: Ah, I'd just typed that in the comment as you were typing this. There's tests in the gem, so these should be sufficient to give us confidence that switching providers will work  👍

erbridge: :+1:

lib/geocoding.rb
```diff
@@ -1,13 +1,12 @@
 class Geocoding
   attr_reader :params, :location
 
-  def initialize(location, params = { params: { region: 'uk' } })
```
erbridge: Does OS support Northern Ireland? I know we don't support schools there, but is it something that we'd need to be aware of if we do?

PR #866, opened 2019-04-18
[848] Listings dont appear in the published tab on their last day before expiry

erbridge:
## Trello card URL:

https://trello.com/c/LMjeNx09/848-listings-dont-appear-in-the-published-tab-on-their-last-day-before-expiry

## Changes in this PR:

Make jobs that expire today count as published.

**This was a hotfix and is already merged into `master` (#863).**

**DO NOT REBASE**

PR #858, opened 2019-04-16
[843] Add missing punctuation to subscription confirmation

erbridge:
## Trello card URL:

https://trello.com/c/WArlK1Ge/843-job-alerts-confirmation-page

PR #849, opened 2019-04-11
Remove maximum salary as a search option for jobseekers

pezholio:
## Trello card URL:

https://trello.com/c/tiKHkebb/837-remove-maximum-salary-as-a-search-option-for-jobseekers

## Changes in this PR:

This removes the option for jobseekers to search by maximum salary. This will have an impact on jobseekers who have already set up an alert with that filter, so I've made sure our alert searches still carry out the maximum salary searches, at least for now.

## Next steps:

- [ ] Terraform deployment required?

- [ ] New development configuration to be shared?

PR #845, opened 2019-04-11
[836] Revert temporary task to backfill historic "more info" clicks

erbridge:
## Trello card URL:

https://trello.com/c/4bjIV6qM/836-revert-temporary-task-to-backfill-historic-more-info-clicks

## Changes in this PR:

Removes the backfill task and the related Terraform config.

## Next steps:

- [x] Terraform deployment required?

PR #869, opened 2019-04-23
Production Release #71

pezholio:
It's been a while since we deployed to production because of the national launch, so this is a bit of a big one. 

Changes in Release 71:

✨ Make Job alert references required, and generate a sensible default (#837)
✨ Users can unsubscribe from a job alert at any time (#840)
✨ Enable sorting functionality in hiring staff backend (#835)
✨ Send jobseeker alert data to Google Sheets (#841)
✨ Removing maximum salary as a search option (#849)
✨ Replace keyword search with job title and subject (#853)

🐛 Send the correct datetimes to the performance platform (#836)
🐛 Make sure we're sending the correct client IP address to the server (#839)
🐛 Add missing env vars to Sidekiq in development (#848)
🐛 Fix flickering failing tests (#856, #875)
🐛 Add missing punctuation to subscription confirmation text (#858)
🐛 Remove some stray Terraform config that we didn't need (#859)
🐛 Make sure listings appear in the hiring staff backend on the day they expire (#866) (already pushed to master as a hotfix)
🐛 Use live data for ECS task definitions for service pinning (#876)

🔨 Improve bin scripts (#843)
🔨 Revert temporary backfill task for more info clicks (#845)
🔨 Change scaling policy in Terraform (#846)
🔨 Change the Terraform versioning policy to be more explicit (#847)
🔨 Refactor performance platform jobs (#850)
🔨 Fix the last Rubocop TODO (#852)
🔨 Security update for Nokogiri (#867)

Phew!


erbridge:
>  🐛 Fix flickering failing tests (#856)

Just a note that that PR didn't actually fix them.

pezholio:
I'll change that to "attempt to". I think I know what the problem is though.

app/services/vacancy_filters.rb
```diff
@@ -9,7 +9,8 @@ def initialize(args)
 
     @location = args[:location]
     @radius = args[:radius].to_s if args[:radius].present?
-    @keyword = args[:keyword]
+    @subject = args[:subject]
+    @job_title = args[:job_title]
     @minimum_salary = args[:minimum_salary]
     @maximum_salary = args[:maximum_salary]
```
erbridge: This line should have moved into `VacancyAlertFilters`. Oops.

app/services/vacancy_search_builder.rb
```diff
@@ -187,14 +194,27 @@ def match_all_hash
...
+        type: 'best_fields',
+        fields: %w[subject.name^3 first_supporting_subject.name^2 second_supporting_subject.name^2 job_title],
+        operator: 'or',
+        minimum_should_match: 1,
+        fuzziness: 'AUTO'
+      },
+    }
+  end
+
+  def job_title_multi_match(job_title)
```
erbridge: Sigh. Badly named method since it's not a multi match.

variables.tf
```diff
@@ -158,6 +168,11 @@ variable "send_job_alerts_daily_email_command" {
   default     = ["rake", "verbose", "daily_emails:send"]
 }
 
+variable "backfill_jobseeker_alert_data_command" {
+  description = "The Entrypoint for the backfill_jobseeker_alert_data task"
+  default     = ["rake", "verbose", "jobseeker_alerts:statistics:backfill:alert_data"]
+}
```
erbridge: We also need to run this after merging. Did we confirm that it works on staging?

pezholio: Nope. We probably need to do a staging deploy and check that out 👍

erbridge: I think we should always be doing those as soon as we merge to `develop`.

erbridge:
Task that need to happen:

Before deploy:
- [ ] Create worksheet for `audit_subscription_creation_worksheet_gid` and add to `production.tfvars`.

After deploy:
- [ ] Run `backfill_jobseeker_alert_data_command` task.

PR #875, opened 2019-04-23
Fix flickering tests

pezholio:
## Trello card URL:

https://trello.com/c/8EMdqNfY/835-fix-flickering-failing-tests

## Changes in this PR:

Finally found the issue behind this. The fix is to remove all whitespace from the titles when using `sort_by`, which mimics the behaviour of Postgres's `ORDER BY`

### DIFF
--- a/spec/presenters/school_vacancies_presenter_spec.rb
+++ b/spec/presenters/school_vacancies_presenter_spec.rb
@@ -1,8 +1,8 @@
 require 'rails_helper'
 RSpec.describe SchoolVacanciesPresenter do
   let(:school) { create(:school) }
-  let(:sort) { VacancySort.new }
   let(:type) { 'published' }
+  let(:sort) { VacancySort.new }
   let(:presenter) { described_class.new(school, sort, type) }
 
   let!(:draft_vacancies) { create_list(:vacancy, 5, :draft, school: school) }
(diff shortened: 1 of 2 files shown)


spec/presenters/school_vacancies_presenter_spec.rb
```diff
@@ -18,78 +18,86 @@
   end
   let!(:published_vacancies) { create_list(:vacancy, 7, :published, school: school) }
 
+  let(:vacancy_ids) do
+    vacancies.sort_by { |vacancy| vacancy['job_title'].delete(' ') }.reverse.pluck(:id)
+  end
+
```
erbridge: This is specifically for when it's sorted by `column: 'job_title', order: 'desc'`, and isn't general. I'm not sure it makes sense to pull out to here.

pezholio: Ah, well spotted. I just got refactor-happy there :D

pezholio: I've wrapped the sorting stuff in a `describe` block to be clearer about what we're doing here

pezholio:
Ready for another go round now, I've also stopped using the `let!`s to set up state, as these get called for each spec, slowing the execution time down dramatically.

spec/presenters/school_vacancies_presenter_spec.rb
```diff
@@ -1,25 +1,77 @@
...
+      3.times do
+        expired_vacancy = build(:vacancy, :expired, school: @school)
+        expired_vacancy.save(validate: false)
+        expired_vacancies << expired_vacancy
+      end
+      expired_vacancies
     end
-    expired_vacancies
+    @published_vacancies = create_list(:vacancy, 7, :published, school: @school)
   end
```
erbridge: :+1:

PR #876, opened 2019-04-23
[853] Use live data for ECS task definitions for service pinning

erbridge:
## Trello card URL:

https://trello.com/c/nlsIBurV/853-fix-issue-with-terraform-ecs-task-pinning

## Changes in this PR:

Terraform `data` resources fetch the current state of the live system, rather than relying on the Terraform state. This means we can create new task definitions outside of Terraform and have it
pick them up in a plan. Our build and deploy process does this, which was causing an issue where `terraform apply`s were rolling back to earlier versions if there was no change to the task definition.

After merging this, we should never see changes to the `task_definition` for `module.ecs.aws_ecs_service.*` unless we're changing the connected task definition in the same change.

This may cause issues on a fresh system deployment, as there is no matching task definition at that point.

https://github.com/hashicorp/terraform/issues/16380#issuecomment-355846367

## Next steps:

### DIFF
--- a/terraform/modules/ecs/ecs.tf
+++ b/terraform/modules/ecs/ecs.tf
@@ -43,7 +43,7 @@ resource "aws_ecs_service" "web" {
   name            = "${var.ecs_service_web_name}"
   iam_role        = "${aws_iam_role.ecs_role.arn}"
   cluster         = "${aws_ecs_cluster.cluster.id}"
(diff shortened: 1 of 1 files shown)


erbridge:
[Fix issue with Terraform ECS task pinning](https://trello.com/c/nlsIBurV/853-fix-issue-with-terraform-ecs-task-pinning)

PR #881, opened 2019-04-26
(fix) Setting of db:environment:set for development

davidwinter:
## Trello card URL:

https://trello.com/c/cgiojdXK/864-fix-ar-environment-mismatch-error

## Changes in this PR:

* Ensures that the correct environment is set in the AR internal metadata by calling `db:environment:set`. This was causing an error when doing initial setup.
* Refactored `bin/drebuild` to use `bin/drake` for `db:seed` task


davidwinter:
@pezholio out of interest, had you, or are you aware of anyone else that had experienced this issue?

pezholio:
@davidwinter From time to time, yeah, so good to get this resolved 👍

PR #885, opened 2019-04-26
(feature) formatting total job count

mingoland:
## Changes in this PR:
- Added delimiters to total job count

### Before
Screenshot 2019-04-26 at 15 34 59

### After
Screenshot 2019-04-26 at 15 34 29


mingoland:
[Format the total job count so that larger numbers are clearer (i.e. 1000 will instead be 1,000)](https://trello.com/c/8a5nfbI7/854-format-the-total-job-count-so-that-larger-numbers-are-clearer-ie-1000-will-instead-be-1000)

PR #886, opened 2019-04-29
[855] Create an SSH key for ECS from a shared public key

erbridge:
## Trello card URL:

https://trello.com/b/vo4AYL7a/dfe-teaching-vacancies-product-development

## Changes in this PR:

Rather than manually creating (or uploading) an SSH key in AWS to be used for sign in, we create one using Terraform. We still need to generate and share the private key, but I suggest that it goes into our secrets management system.

## Next steps:

- [ ] Generate new SSH key: `ssh-keygen -t rsa`
- [ ] Add public key to `tfvars`
- [ ] Add private key to secret store
- [ ] Deploy Terraform


erbridge:
Key is generated for staging. Private is uploaded, and public is ready to share in `tfvars` (if you want to plan).

erbridge:
[Future proof the way we are using SSH keys](https://trello.com/c/Tcjms2oM/855-future-proof-the-way-we-are-using-ssh-keys)

PR #894, opened 2019-05-08
Add Ordnance Survey API key to worker

pezholio:
## Trello card URL

https://trello.com/c/YxYwCYBk/871-fix-rollbar-errors-in-development-when-running-senddailyalertemailjob

## Changes in this PR:

After deploying #868 - I noticed some strange errors were cropping up when the worker was running the `SendDailyAlertEmailJob`. This is because I only had the Ordnance Survey API key set up in the web container, and not the worker. To complicate matters further, the exact nature of the error was being masked, because `ENV.fetch` was defaulting to an empty string, so the Geocoder gem assumed there was a valid API key present, and so requests were getting passed through to the API, and the repsonse wasn't what we were expecting. I've fixed this by making sure the API key's default value is `nil` in the initializer now.

## Next steps:

### DIFF
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -1,5 +1,5 @@
 redis = Redis.new(url: ENV['REDIS_CACHE_URL'])
-key = ENV.fetch('ORDNANCE_SURVEY_API_KEY', '')
+key = ENV.fetch('ORDNANCE_SURVEY_API_KEY', nil)
 
 Geocoder.configure(
   lookup: :os_names,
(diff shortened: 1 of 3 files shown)


config/initializers/geocoder.rb
```diff
@@ -1,5 +1,5 @@
 redis = Redis.new(url: ENV['REDIS_CACHE_URL'])
-key = ENV.fetch('ORDNANCE_SURVEY_API_KEY', '')
+key = ENV.fetch('ORDNANCE_SURVEY_API_KEY', nil)
```
erbridge: :+1:

pezholio:
[Fix Rollbar errors in development when running SendDailyAlertEmailJob](https://trello.com/c/YxYwCYBk/871-fix-rollbar-errors-in-development-when-running-senddailyalertemailjob)

PR #913, opened 2019-05-16
(Hotfix) [885] Fix phase JSON invalid Array to String conversion

erbridge:
Alerts already parse the full JSON when they get passed in to the filter, while the param from a search is a string.

This is blocking the daily alerts from sending.

Resolves https://rollbar.com/dfe/teacher-vacancies/items/476/

**This is a hotfix directly on to `master`**

### DIFF
--- a/app/services/vacancy_filters.rb
+++ b/app/services/vacancy_filters.rb
@@ -13,8 +13,8 @@ class VacancyFilters
     @job_title = args[:job_title]
     @minimum_salary = args[:minimum_salary]
     @newly_qualified_teacher = args[:newly_qualified_teacher]
-    @working_pattern = extract_working_pattern(args)
-    @phases = extract_phases(args)
+    @working_pattern = extract_working_pattern(args[:working_pattern])
+    @phases = extract_phases(args[:phases])
   end
 
   def to_hash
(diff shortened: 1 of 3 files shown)


app/services/vacancy_filters.rb
```diff
@@ -57,13 +57,15 @@ def any?
...
+    working_pattern if Vacancy.working_patterns.include?(working_pattern)
   end
 
-  def extract_phases(params)
-    return if params[:phases].blank?
+  def extract_phases(phases)
+    return if phases.blank?
 
-    JSON.parse(params[:phases]).select { |phase| School.phases.include?(phase) }.presence
+    phases = JSON.parse(phases) if phases.is_a?(String)
```
adrianclay: It looks like we have this ` JSON.parse ` functionality for "phases" in both the controller and this class.

What is encoding the phases as JSON?

spec/jobs/send_daily_alert_email_job_spec.rb
```diff
@@ -5,7 +5,15 @@
 
   subject(:job) { described_class.perform_later }
 
-  let!(:subscription) { create(:subscription, frequency: :daily) }
```
adrianclay: Minor follow up question, can a subscription not have a `search_criteria`?

app/services/vacancy_filters.rb
erbridge: When we save a `Subscription`, we encode the `search_criteria` as JSON. We parse this JSON in full, meaning `phases` is now an array.

When we get a new search, we get a series of string parameters. This means that `phases` is a string representation of an array that needs parsing.

Probably we should do that parsing earlier.

spec/jobs/send_daily_alert_email_job_spec.rb
erbridge: Apparently so.

app/services/vacancy_filters.rb
erbridge: Created a card to look into this a bit more: https://trello.com/c/fPWAOZKD/887-refactor-search-criteria

erbridge:
[Phase JSON invalid Array to String conversion](https://trello.com/c/Upi1cY70/885-phase-json-invalid-array-to-string-conversion)

PR #914, opened 2019-05-16
[885] Fix phase JSON invalid Array to String conversion

erbridge:
Only parse phases as JSON if they're a string.

Alerts already parse the full JSON when they get passed in to the filter, while the param from a search is a string.

This is blocking the daily alerts from sending.

Resolves https://rollbar.com/dfe/teacher-vacancies/items/476/

**This was a hotfix and is cherry-picked from `master`**

### DIFF
--- a/app/services/alert_result_finder.rb
+++ b/app/services/alert_result_finder.rb
@@ -3,8 +3,8 @@ class AlertResultFinder
 
   attr_reader :query, :results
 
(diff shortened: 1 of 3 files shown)


erbridge:
[Phase JSON invalid Array to String conversion](https://trello.com/c/Upi1cY70/885-phase-json-invalid-array-to-string-conversion)

PR #915, opened 2019-05-16
Validate the presence of search criteria on a subscription

erbridge:
## Changes in this PR:

This makes `search_criteria` required on subscriptions.

### DIFF
--- a/spec/factories/subscriptions.rb
+++ b/spec/factories/subscriptions.rb
@@ -4,6 +4,14 @@ FactoryBot.define do
     email { Faker::Internet.email }
     reference { Faker::Lorem.sentence }
     frequency { :daily }
+    search_criteria do
+      {
+        location: 'EC1A 1AA',
+        radius: 20,
+        working_pattern: 'full_time',
+        phases: ['primary', 'secondary']
+      }.to_json
+    end
 
     factory :daily_subscription do
       frequency { :daily }
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -8,7 +8,7 @@ class Subscription < ApplicationRecord
   validates :email, email_address: { presence: true }
   validates :reference, presence: true
   validates :frequency, presence: true
-  validates :search_criteria, uniqueness: { scope: %i[email expires_on frequency] }
+  validates :search_criteria, presence: true, uniqueness: { scope: %i[email expires_on frequency] }
 
   scope :ongoing, -> { where('expires_on >= current_date') }
 
--- a/spec/geocoder_helper.rb
+++ b/spec/geocoder_helper.rb
@@ -1,0 +1,21 @@
+RSpec.configure do |config|
+  config.before(:suite) do
+    Geocoder.configure(lookup: :test)
+  end
+
+  config.before(:each) do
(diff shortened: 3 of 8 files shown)


spec/factories/subscriptions.rb
```diff
@@ -4,6 +4,14 @@
     email { Faker::Internet.email }
     reference { Faker::Lorem.sentence }
     frequency { :daily }
+    search_criteria do
+      {
+        location: 'EC1A 1AA',
+        radius: 20,
+        working_pattern: 'full_time',
+        phases: ['primary', 'secondary']
+      }.to_json
```
adrianclay: I don't think the call to `to_json` is necessary based on a bit of playing.  However, given we use this in the `SubscriptionController` makes sense to be consistent.

erbridge: I want to make sure that we're using the data type that's expected so we don't build to unexpected behaviour.

adrianclay: From the [example usage](https://edgeguides.rubyonrails.org/active_record_postgresql.html#json-and-jsonb) of `jsonb` it doesn't include manual calls to `to_json`.  So it looks like ActiveRecord is not expecting a string.

I'm now wondering if calling `to_json` causes unexpected behaviour.  I'll give it a test and see.

adrianclay: Having had a look, calling `to_json` means the hash is being doubly converted to JSON inside of `psql`.  Logically makes sense once you realise what's going on. We will need to continue with the double encoding tech debt but is a potential gotcha and source of confusion going forward


Currently:
```
                  id                  |        email        | frequency | expires_on |              search_criteria               |          reference          |         created_at         |         upda
ted_at         
--------------------------------------+---------------------+-----------+------------+--------------------------------------------+-----------------------------+----------------------------+-------------
---------------
 e98b5014-a6fa-4bb4-818e-5e55cf61be48 | adrian@madetech.com |         0 | 2019-11-17 | "{\"location\":\"xxx\",\"radius\":\"20\"}" | Jobs within 20 miles of xxx | 2019-05-17 08:47:45.045288 | 2019-05-17 0
8:47:45.045288
```


Without the additional `.to_json`
```
                  id                  |        email        | frequency | expires_on |           search_criteria           |          reference          |         created_at         |         updated_at 
        
--------------------------------------+---------------------+-----------+------------+-------------------------------------+-----------------------------+----------------------------+--------------------
--------
 e98b5014-a6fa-4bb4-818e-5e55cf61be48 | adrian@madetech.com |         0 | 2019-11-17 | {"radius": "20", "location": "xxx"} | Jobs within 20 miles of xxx | 2019-05-17 08:47:45.045288 | 2019-05-17 08:50:23
.214391
```

erbridge: Fun... If we could undo that, it would sinmplify a lot of the search criteria parsing that goes on. It should be possible if we support both for a release and do a data migration to JSON encode once.

erbridge:
@adrianclay please re-review as I had to make some changes to the tests to get this working.

tatyree:
This has been sitting for seven months.  Closing.

PR #917, opened 2019-05-21
[891] Speed up Docker rebuilds

erbridge:
## Trello card URL:

https://trello.com/c/FVvOzK4O/891-speed-up-docker-rebuilds

## Changes in this PR:

- Switch to explicitly including directories in Docker images.
- Switch to only precompiling in production environments.
- Move the lambda payload into `terraform/`.


erbridge:
[Speed up docker rebuilds](https://trello.com/c/FVvOzK4O/891-speed-up-docker-rebuilds)

PR #925, opened 2019-05-23
Amend job alert confirmation email and confirmation page text

pezholio:
## Trello card URL:

https://trello.com/c/AfFF070i/870-amend-job-alert-confirmation-email-and-confirmation-page-text

## Changes in this PR:

Tweaks some of the content in the job alert confirmation email to: 

- confirm that link at bottom offers both unsubscribe and resubscribe
- fix the representation of the expiry date ("the 2 of November" is incorrect)
- make sure it refers to 'job alerts' rather than 'email notifications'

The confirmation page text has also been tweaked.

### DIFF
--- a/app/helpers/notify_view_helper.rb
+++ b/app/helpers/notify_view_helper.rb
@@ -4,7 +4,7 @@ module NotifyViewHelper
   end
 
   def unsubscribe_link(token)
-    link_text = I18n.t('subscriptions.email.unsubscribe_link_text')
+    link_text = t('subscriptions.email.unsubscribe_link_text')
     notify_link(subscription_unsubscribe_url(subscription_id: token), link_text)
   end
 end
--- a/app/helpers/notify_view_helper.rb
+++ b/app/helpers/notify_view_helper.rb
@@ -4,7 +4,7 @@ module NotifyViewHelper
   end
 
   def unsubscribe_link(token)
-    link_text = t('subscriptions.email.unsubscribe_link_text')
+    link_text = I18n.t('subscriptions.email.unsubscribe_link_text')
     notify_link(subscription_unsubscribe_url(subscription_id: token), link_text)
   end
 end
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -1,11 +1,7 @@
 class SubscriptionMailer < ApplicationMailer
   def confirmation(subscription_id)
-    @subscription = Subscription.find(subscription_id)
-
-    @subscription_reference = @subscription.reference
-    @search_criteria = SubscriptionPresenter.new(@subscription).filtered_search_criteria
-    @expires_on = @subscription.expires_on
-    @unsubscribe_token = @subscription.token
+    subscription = Subscription.find(subscription_id)
+    @subscription = SubscriptionPresenter.new(subscription)
 
     view_mail(
       NOTIFY_SUBSCRIPTION_CONFIRMATION_TEMPLATE,
--- a/app/mailers/subscription_mailer.rb
+++ b/app/mailers/subscription_mailer.rb
@@ -1,16 +1,16 @@
 class SubscriptionMailer < ApplicationMailer
   def confirmation(subscription_id)
-    subscription = Subscription.find(subscription_id)
+    @subscription = Subscription.find(subscription_id)
 
-    @subscription_reference = subscription.reference
-    @search_criteria = SubscriptionPresenter.new(subscription).filtered_search_criteria
-    @expires_on = subscription.expires_on
-    @unsubscribe_token = subscription.token
+    @subscription_reference = @subscription.reference
+    @search_criteria = SubscriptionPresenter.new(@subscription).filtered_search_criteria
+    @expires_on = @subscription.expires_on
+    @unsubscribe_token = @subscription.token
 
     view_mail(
       NOTIFY_SUBSCRIPTION_CONFIRMATION_TEMPLATE,
-      to: subscription.email,
-      subject: I18n.t('job_alerts.confirmation.email.subject', reference: subscription.reference),
+      to: @subscription.email,
+      subject: I18n.t('job_alerts.confirmation.email.subject', reference: @subscription.reference),
     )
   end
 end
--- a/app/views/alert_mailer/daily_alert.text.haml
+++ b/app/views/alert_mailer/daily_alert.text.haml
@@ -19,4 +19,4 @@ Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))
 
 \---
 
-= t('subscriptions.email.unsubscribe_text_html', unsubscribe_link: unsubscribe_link(@unsubscribe_token))
+= t('subscriptions.email.unsubscribe_text_html', unsubscribe_link: unsubscribe_link(@subscription.token))
--- a/app/views/alert_mailer/daily_alert.text.haml
+++ b/app/views/alert_mailer/daily_alert.text.haml
@@ -15,7 +15,7 @@
   \---
   \
 \
-Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} -  the free service for schools in England to list teaching roles and for jobseekers to find them.
+Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - the free service for schools in England to list teaching roles and for jobseekers to find them.
 
 \---
 
--- a/app/views/alert_mailer/daily_alert.text.haml
+++ b/app/views/alert_mailer/daily_alert.text.haml
@@ -1,5 +1,5 @@
 \# #{t('app.title')}
-\## #{t('job_alerts.alert.email.daily.summary', count: @vacancies.count)}
+\## #{t('job_alerts.alert.email.daily.summary', count: @vacancies.count, reference: @subscription.reference)}
 \
 \---
 \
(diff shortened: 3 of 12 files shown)


app/views/subscription_mailer/confirmation.text.haml
```diff
@@ -1,13 +1,13 @@
 \# #{t('app.title')}
-\# #{@subscription_reference}
+\## #{t('subscriptions.email.confirmation.subject', reference: @subscription_reference)}
```
erbridge: Can we use `@subscription.reference` for this and save ourselves a level of indirection?

app/mailers/subscription_mailer.rb
```diff
@@ -1,16 +1,16 @@
 class SubscriptionMailer < ApplicationMailer
   def confirmation(subscription_id)
-    subscription = Subscription.find(subscription_id)
+    @subscription = Subscription.find(subscription_id)
```
erbridge: Why not pass the presenter itself to the view?

app/views/subscription_mailer/confirmation.text.haml
```diff
@@ -1,13 +1,13 @@
...
 \
-= t('subscriptions.email.confirmation.subheading')
+= t('subscriptions.email.confirmation.subheading', email: @subscription.email)
 \
 - @search_criteria.each_pair do |k, v|
   * #{[k&.titleize, v].reject(&:blank?).join(': ')}
 \
-= t('subscriptions.email.confirmation.expiry_text_html', distance: '6 months', date: l(@expires_on))
+= t('subscriptions.email.confirmation.expiry_text_html', date: l(@expires_on))
```
erbridge: Can we use `@subscription.expires_on` for this and save ourselves a level of indirection?

app/views/subscription_mailer/confirmation.text.haml
```diff
@@ -1,13 +1,13 @@
...
 
 \---
 
-= t('subscriptions.email.unsubscribe_text_html', link: unsubscribe_link(@unsubscribe_token))
\ No newline at end of file
+= t('subscriptions.email.unsubscribe_text_html', unsubscribe_link: unsubscribe_link(@unsubscribe_token))
```
erbridge: Can we use `@subscription.token` for this and save ourselves a level of indirection?

app/mailers/subscription_mailer.rb
```diff
@@ -1,16 +1,16 @@
...
 
-    @subscription_reference = subscription.reference
-    @search_criteria = SubscriptionPresenter.new(subscription).filtered_search_criteria
-    @expires_on = subscription.expires_on
-    @unsubscribe_token = subscription.token
+    @subscription_reference = @subscription.reference
+    @search_criteria = SubscriptionPresenter.new(@subscription).filtered_search_criteria
```
erbridge: If we pass the presenter to the views, can we make the view responsible for getting the `filtered_search_criteria`?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_04_15_121556) do
+ActiveRecord::Schema.define(version: 2019_05_16_213012) do
```
erbridge: This slipped in @pezholio.

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
```diff
@@ -2,6 +2,9 @@
 
 RSpec.feature 'A job seeker can subscribe to a job alert' do
   before { allow(EmailAlertsFeature).to receive(:enabled?) { true } }
+  let(:header) do
+    ActionView::Base.full_sanitizer.sanitize(I18n.t('subscriptions.confirmation.header'))
+  end
```
erbridge: You can avoid this by using `expect(page.html).to include(I18n.t('subscriptions.confirmation.header'))`, which is arguably a better test, because it ensures we get the structured header.

db/schema.rb
pezholio: Ugh. That'll teach me to blindly commit everything

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
erbridge: See this.

pezholio:
[Amend job alert confirmation email and confirmation page text](https://trello.com/c/AfFF070i/870-amend-job-alert-confirmation-email-and-confirmation-page-text)

pezholio:
[Amend job alert confirmation email and confirmation page text](https://trello.com/c/AfFF070i/870-amend-job-alert-confirmation-email-and-confirmation-page-text)

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
pezholio: Weird, I thought I'd fixed that

pezholio:
I've tidied up the mailers now. No idea why I'd done it that way in the first place. There were also a bunch of unused instance vars in the daily alert mailer, which I've dinged too.

app/views/alert_mailer/daily_alert.text.haml
```diff
@@ -19,4 +19,4 @@ Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))
 
 \---
 
-= t('subscriptions.email.unsubscribe_text_html', link: unsubscribe_link(@unsubscribe_token))
+= t('subscriptions.email.unsubscribe_text_html', unsubscribe_link: unsubscribe_link(@unsubscribe_token))
```
erbridge: This should probably also be `@subscription.token`, then it could be removed from the mailer.

app/views/subscriptions/confirm.html.haml
```diff
@@ -3,16 +3,14 @@
 .govuk-grid-row
   .govuk-grid-column-two-thirds
     .govuk-panel.govuk-panel--confirmation
-      %h1.govuk-panel__title= t('subscriptions.confirmation.header')
+      %h1.govuk-panel__title= t('subscriptions.confirmation.header').html_safe
```
adrianclay: Why was this change required?

app/helpers/notify_view_helper.rb
```diff
@@ -4,7 +4,7 @@ def notify_link(url, text = url)
   end
 
   def unsubscribe_link(token)
-    link_text = t('subscriptions.email.unsubscribe_link_text')
+    link_text = I18n.t('subscriptions.email.unsubscribe_link_text')
```
adrianclay: Why is this change necessary?

app/views/alert_mailer/daily_alert.text.haml
```diff
@@ -15,8 +15,8 @@
   \---
   \
 \
-Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - the free service for schools in England to list teaching roles and for jobseekers to find them.
+Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} -  the free service for schools in England to list teaching roles and for jobseekers to find them.
```
adrianclay: Looks like there is an additional space on this line, not sure if that was intentional.

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
```diff
@@ -45,7 +45,7 @@
         fill_in 'subscription[email]', with: 'jane.doe@example.com'
         click_on 'Subscribe'
 
-        expect(page).to have_content(I18n.t('subscriptions.confirmation.header'))
+        expect(page.html).to include(I18n.t('subscriptions.confirmation.header'))
```
adrianclay: Why did these need to change?  The comit messages didn't provide any clarity for me.

app/views/subscriptions/confirm.html.haml
Renny4Real: Git lens highlighted this change as a part of commit Stuart authored.
 
It looks like it was meant for a change in wording for the subscription header which initially had HTML in it. The change is also a part of the same commit [here](https://github.com/DFE-Digital/teacher-vacancy-service/pull/925/commits/adcc3e0ee4ab00b8eb2883cb936777849650eef1#diff-37c9b835927e3c363588f172de9bb6faR306)
Will remove the `html.safe` since we no longer use HTML in the `subscriptions.confirmation.header` translation file.

app/helpers/notify_view_helper.rb
Renny4Real: We made this change because one of the tests complained about it. 
Ran the test suite again (as of the time of writing this) with this change removed. 
Tests work fine. 
Must have been left over from brainstorming how to get the tests to pass.

app/views/alert_mailer/daily_alert.text.haml
Renny4Real: Thanks for spotting this. 
Not intentional. 
Can see that this was pushed as part of a WIP commit: ea6fda, so that someone else could continue where I left off since I had to quickly work on another feature.

spec/features/job_seekers_can_subscribe_to_a_job_alert_spec.rb
Renny4Real: This was also part of the commit Stuart authored: (adcc3e0) and is possibly related to the previous change that had HTML in the subscription header (also in the same commit). 
Have tested this without using `page.html` and `include()` in the `expect` block and all seems well.

spec/mailers/subscription_spec.rb
```diff
@@ -45,4 +47,8 @@
     expect(body_lines[12]).to match(/#{I18n.t('subscriptions.email.unsubscribe_text_html')}/)
     expect(body_lines[14]).to match(%r{http:\/\/localhost:3000\/subscriptions\/#{subscription.token}\/unsubscribe})
   end
+
+  def escape_single_quotes(unescaped_string)
```
Renny4Real: This looks great. Wonder if we can use this method for the other places where we would like to escape the string in our tests.

dushan-madetech: If so we could probably extract this to somewhere that enables us to reuse it without duplicating code

adrianclay: Yep.  I'll leave that to be an exercise for a later PR, but agreed we should pull it out if it's needed in multiple places.

PR #929, opened 2019-05-24
Remove remote working as an option

erbridge:
## Trello card URL:

https://trello.com/c/4owuXmAO/762-update-working-pattern-options-so-there-is-more-flexibility-around-types-of-role-reflecting-the-needs-of-jobseekers-schools

## Changes in this PR:

Remove remote working as an option. The guidance includes it as a footnote, but suggests that it's uncommon. We think that including it would be misleading without further explanation.

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -6,8 +6,7 @@ class Vacancy < ApplicationRecord
     'part_time' => 100,
     'job_share' => 101,
     'compressed_hours' => 102,
-    'staggered_hours' => 104,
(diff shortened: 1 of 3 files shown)


erbridge:
[Update 'working pattern' options so there is more flexibility around types of role reflecting the needs of jobseekers/schools](https://trello.com/c/4owuXmAO/762-update-working-pattern-options-so-there-is-more-flexibility-around-types-of-role-reflecting-the-needs-of-jobseekers-schools)

PR #926, opened 2019-05-23
Production Release #75

Renny4Real:
## Changes in release 75:

(feature) Optimise mobile pagination (#922)
(feature) Update working pattern (#920, #928, #929)

(chore) Update documentation to refer to new secrets repository (#906)
(chore) Add basic documentation for re-running `tvs2_production_web_update_spreadsheets_task`(#916)
(chore) Map the Sidekiq docker container volume to the local filesystem (#923)
(chore) Speed up Docker rebuilds (#917)
(chore) Add a .terraform-version file to support tfenv (#924)
(chore) Change the Docker maintainer to be DfE (#927)

## Next steps:

- [ ] Deploy Terraform
- [ ] Run the migrate working pattern rake task
- [ ] Run the reindex vacancies rake task
- [ ] Check that when creating a job, working patterns options are available to select.
- [ ] Check that job listings contain working patterns in their description.
- [ ] Create card to remove the migrate working pattern rake task
- [ ] Create card to remove defunct `working_pattern` from `Vacancy`

**Note: Wait to deploy until 4pm**

PR #932, opened 2019-05-28
Revert working patterns functionality as it is broken on Production.

Renny4Real:
Migrate working pattern script breaks on production due to incorrect format of `starts_on` date.

This was introduced alongside the Working patterns changes in #926.

By reverting the working patterns functionality, we intend to bring production to a working state.

We recreated the broken working patterns column locally, and this branch fixes the issue.

## Next steps:

- [ ]  Run the reindex vacancies rake task


Renny4Real:
# Update

We noticed an issue where any vacancies created between deploy time and now would cause further errors down the line.

We decided to fix the record in production with the incorrect year.

Fortunately this was the only record affected, and the migrate working patterns completed successfully.

PR #942, opened 2019-06-19
Fix copy job listing bug

dushan-madetech:
## Trello card URL:

https://trello.com/c/IQmSekv5

## Changes in this PR:

- Add test to check for vacancy fields not being copied e.g. page views
- Add fix for copying vacancy bug

### DIFF
--- a/spec/services/copy_vacancy_spec.rb
+++ b/spec/services/copy_vacancy_spec.rb
@@ -28,25 +28,51 @@ RSpec.describe CopyVacancy do
     end
 
     context 'not all fields are copied' do
-      it 'should not copy fields that should be unique' do
-        vacancy = create(:vacancy,
-                         job_title: 'Maths teacher',
-                         slug: 'maths-teacher',
-                         weekly_pageviews: 4,
-                         total_pageviews: 4,
-                         weekly_pageviews_updated_at: Time.zone.today - 5.days,
-                         total_pageviews_updated_at: Time.zone.today - 5.days,
-                         total_get_more_info_clicks: 6,
-                         total_get_more_info_clicks_updated_at: Time.zone.today - 5.days)
+      let(:vacancy) do
+        create(:vacancy,
+               job_title: 'Maths teacher',
+               slug: 'maths-teacher',
+               weekly_pageviews: 4,
+               total_pageviews: 4,
+               weekly_pageviews_updated_at: Time.zone.today - 5.days,
+               total_pageviews_updated_at: Time.zone.today - 5.days,
+               total_get_more_info_clicks: 6,
+               total_get_more_info_clicks_updated_at: Time.zone.today - 5.days)
+      end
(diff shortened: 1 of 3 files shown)


spec/services/copy_vacancy_spec.rb
```diff
@@ -26,5 +26,28 @@
...
+                         slug: 'maths-teacher',
+                         weekly_pageviews: 4,
+                         total_pageviews: 4,
+                         weekly_pageviews_updated_at: Time.zone.today - 5.days,
+                         total_pageviews_updated_at: Time.zone.today - 5.days,
+                         total_get_more_info_clicks: 6,
+                         total_get_more_info_clicks_updated_at: Time.zone.today - 5.days)
+
+        result = described_class.new(vacancy).call
+        expect(Vacancy.find(result.id).slug).to_not eq('maths_teacher')
```
adrianclay: The slug above has a dash?

spec/services/copy_vacancy_spec.rb
```diff
@@ -26,5 +26,28 @@
...
+        expect(Vacancy.find(result.id).weekly_pageviews).to eq(0)
+        expect(Vacancy.find(result.id).weekly_pageviews_updated_at).to_not eq(Time.zone.today - 5.days)
```
adrianclay: What do we want the `updated_at` to be? Maybe investigate how the field is used, then make a call on what a sensible value is?

spec/services/copy_vacancy_spec.rb
```diff
@@ -26,5 +26,28 @@
 
       Timecop.return
     end
+
+    context 'not all fields are copied' do
+      it 'should not copy fields that should be unique' do
```
adrianclay: I might separate out the slug part and the counters into separate `it` blocks.  They feel like separate enough tests, with different reasons to fail.

spec/services/copy_vacancy_spec.rb
dushan-madetech: Fixed in https://github.com/DFE-Digital/teacher-vacancy-service/pull/942/commits/39703520053b37b6e1a9095598d71df36b0bfd75

spec/services/copy_vacancy_spec.rb
dushan-madetech: Set to current time using Time.zone.now, see https://github.com/DFE-Digital/teacher-vacancy-service/pull/942/commits/39703520053b37b6e1a9095598d71df36b0bfd75 for this change

spec/services/copy_vacancy_spec.rb
dushan-madetech: Separated the test into multiple individual tests, see https://github.com/DFE-Digital/teacher-vacancy-service/pull/942/commits/39703520053b37b6e1a9095598d71df36b0bfd75 for this change

PR #944, opened 2019-06-20
(Fix) Implement UI changes from staging review for jobs awaiting feedback tab

Renny4Real:
## Changes in this PR:
- Fixed column header width for select boxes
- Fixed header links which would previously redirect to another tab rather than sort the column on the currently active tab.
- Removed the `'Please tell us...' intro text` when there are no jobs awaiting feedback.
- Reduced the font size for `'There are no jobs...' no awaiting feedback text`

### Before
**1.** image

------------------------------------------------------------------------------------------------------

**2.** image (1)

------------------------------------------------------------------------------------------------------

### After
**1.**  Screenshot 2019-06-20 at 12 35 48

------------------------------------------------------------------------------------------------------

**2.** Screenshot 2019-06-20 at 13 05 11

### DIFF
--- a/app/assets/stylesheets/components/awaiting_feedback_vacancies.scss
+++ b/app/assets/stylesheets/components/awaiting_feedback_vacancies.scss
@@ -1,0 +1,3 @@
+.job-title-table-header {
+  width: 195px;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -28,6 +28,7 @@ $govuk-image-url-function: 'image-url';
 @import 'components/select';
 @import 'components/tabs';
 @import 'components/checkboxes';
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/components/awaiting_feedback_vacancies.scss
```diff
@@ -0,0 +1,3 @@
+.job-title-table-header {
```
adrianclay: Minor comment, do we want to use a more specific class name.  This only applies to the awaiting feedback table headings.

Renny4Real: I only used this with the `job-title` heading which is why I named it specific.
Didn't want to imply that this would be useful for all the awaiting feedback table headings.
Could change this if this styling is used in more than one place.

PR #949, opened 2019-07-01
(Fix) Add working patterns to Vacancy Spreadsheet

Renny4Real:
Since it went live, working pattern data was not being written to the spreadsheet.

This change adds an `add_vacancy_to_spreadsheet` job that fills in the missing working patterns data in the published vacancy sheet.

The code might require some refactoring but this should be timeboxed.

### DIFF
--- a/lib/add_vacancies_to_spreadsheet.rb
+++ b/lib/add_vacancies_to_spreadsheet.rb
@@ -12,24 +12,24 @@ class AddVacanciesToSpreadsheet < ExportToSpreadsheet
   end
 
   def to_row(vacancy)
-    {
-      id: vacancy.id,
-      slug: vacancy.slug,
-      created_at: vacancy.created_at.to_s,
-      status: vacancy.status,
-      publish_on: vacancy.publish_on,
-      expires_on: vacancy.expires_on,
-      starts_on: vacancy.starts_on,
-      ends_on: vacancy.ends_on,
-      weekly_hours: vacancy.weekly_hours,
-      flexible_working: vacancy.working_patterns.join(','),
-      school_urn: vacancy.school.urn,
-      school_county: vacancy.school.county
-    }
+    [
+      Time.zone.now.to_s,
+      vacancy.id,
+      vacancy.slug,
+      vacancy.created_at.to_s,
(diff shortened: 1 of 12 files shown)


lib/add_vacancies_to_spreadsheet.rb
```diff
@@ -0,0 +1,35 @@
...
+      weekly_hours: vacancy.weekly_hours,
+      flexible_working: vacancy.working_patterns.join(','),
+      school_urn: vacancy.school.urn,
+      school_county: vacancy.school.county
+    }
+  end
+
+  def present(vacancy)
+    row = to_row(vacancy)
+    row.values.unshift(Time.zone.now.to_s)
```
adrianclay: Be nice to not have the call to `values` here, and also have the time as part of the other fields.

There is currently some mental gymnastics requried to read this.

spec/support/shared_examples/export_to_spreadsheet.rb
```diff
@@ -59,9 +58,9 @@
     end
 
     it 'adds all the data to the spreadsheet' do
-      data = (existing_data + new_data).map(&:to_row)
+      data = (existing_data + new_data).map(&subject.method(:present))
```
adrianclay: I'm thinking it would also be nice to use a similar "expected_spreadsheet_rows" as above.

I find having the expected row as part of the spec much more readable.

adrianclay: Or maybe, the spec names aren't clear as to what functionality they care about.  It's quite possible that this spec shouldn't be concerned about the specific presentation of the data, in which case can we test what it does care about without needing to be affected by potential presentational changes in the future.

Renny4Real: Changed the test setup so that hopefully it's a lot clearer:

The initial data variable was the sum of `existing data` and `new data` being transformed to be appended to a blank worksheet. Renamed the data variable to be `full_worksheet_data` and have also used the `expected_spreadsheet_rows` variable (renamed: `expected_new_spreadsheet_rows`) in the setup.

PR #950, opened 2019-07-01
Feature/946 setup basic smoke test for circleci

C-gyorfi:
Pull Request template

## Trello card URL:
https://trello.com/c/gxKqZ7aC/946-high-level-smoke-tests-for-key-jobseeker-journey-through-tvs-and-to-send-alerts-if-triggered-8

## Changes in this PR:
Add two test to visit the vacancy page, run a search and visit the first result 
Add Circle CI config file with scheduled smoke tests

## Next steps:
- Add more tests for better coverage
- Discover caching with Circle CI to avoid rebuilding before every test run and minimise test time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,28 @@
+version: 2
+jobs:
+  run_smoke_test:
+    working_directory: ~/teacher-vacancy-service
+    docker:
+      - image: circleci/ruby:2.6.1-node-browsers-legacy
+        environment:
+        RAILS_ENV: test
+    steps:
+      - checkout
+      - run: gem install bundler
+      - run: bundle install
+      - run: bundle exec rspec --tag smoke_test spec/smoke_test/
+
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+      - run_smoke_test
+  scheduled-smoke-tests:
+    triggers:
+      - schedule:
+          cron: "5,10,15,20,25,30,35,40,45,50,55,0 * * * *"
+          filters:
+            branches:
+              only: master
+    jobs:
+      - run_smoke_test
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,1 @@
---require spec_helper
+--require spec_helper --tag ~smoke_test
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -40,7 +40,7 @@ class VacanciesController < ApplicationController
 
     @vacancy = VacancyPresenter.new(vacancy)
 
-    VacancyPageView.new(vacancy).track unless authenticated?
+    VacancyPageView.new(vacancy).track unless authenticated? || smoke_test?
 
     expires_in 5.minutes, public: true
   end
@@ -125,5 +125,9 @@ class VacanciesController < ApplicationController
 
   def set_headers
     response.set_header('X-Robots-Tag', 'noarchive')
+  end
+
+  def smoke_test?
+    cookies[:smoke_test] != nil
   end
 end
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -23,7 +23,7 @@ class VacanciesController < ApplicationController
     @filters = VacancyFilters.new(search_params.to_hash)
     @sort = VacancySort.new.update(column: sort_column, order: sort_order)
     @vacancies = VacanciesFinder.new(@filters, @sort, page_number).vacancies
-
+    AuditSearchEventJob.perform_later(audit_row) if valid_search?
     expires_in 5.minutes, public: true
   end
 
@@ -129,5 +129,13 @@ class VacanciesController < ApplicationController
 
   def smoke_test?
     cookies[:smoke_test] != nil
+  end
+
+  def valid_search?
+    @filters.any? && !smoke_test?
+  end
+
+  def audit_row
+    { total_count: @vacancies.total_count }.merge(@filters.audit_hash)
   end
 end
--- a/spec/controllers/vacancies_controller_spec.rb
+++ b/spec/controllers/vacancies_controller_spec.rb
@@ -60,6 +60,33 @@ RSpec.describe VacanciesController, type: :controller do
           subject
         end
       end
+
+      context 'search audiotor' do
+        let(:params) do
+          {
+            subject: 'foo'
+          }
+        end
+
+        it 'should call the search auditor' do
+          expect(AuditSearchEventJob).to receive(:perform_later)
+
+          subject
+        end
+
+        it 'should not call the search auditor if its a smoke test' do
+          cookies[:smoke_test] = 1
+          expect(AuditSearchEventJob).to_not receive(:perform_later)
+
+          subject
+        end
+
+        it 'should not call the search auditor if no search parameters are given' do
+          expect(AuditSearchEventJob).to_not receive(:perform_later)
+
+          get :index
+        end
+      end
     end
 
     context 'feature flagging' do
--- a/spec/controllers/vacancies_controller_spec.rb
+++ b/spec/controllers/vacancies_controller_spec.rb
@@ -61,15 +61,22 @@ RSpec.describe VacanciesController, type: :controller do
         end
       end
 
-      context 'search audiotor' do
+      context 'search auditor' do
         let(:params) do
           {
-            subject: 'foo'
+            job_title: 'Should have three match'
           }
         end
 
-        it 'should call the search auditor' do
-          expect(AuditSearchEventJob).to receive(:perform_later)
+        it 'should call the search auditor', elasticsearch: true do
+          3.times { create(:vacancy, job_title: 'Should have three match') }
+          Vacancy.__elasticsearch__.client.indices.flush
+          expect(AuditSearchEventJob).to receive(:perform_later).with(
+            hash_including(
+              total_count: 3,
+              job_title: 'Should have three match'
+            )
+          )
 
           subject
         end
--- a/spec/controllers/vacancies_controller_spec.rb
+++ b/spec/controllers/vacancies_controller_spec.rb
@@ -114,5 +114,25 @@ RSpec.describe VacanciesController, type: :controller do
         expect(response).to have_http_status(:not_found)
       end
     end
+
+    context 'when using cookies' do
+      let(:vacancy) { create(:vacancy) }
+      let(:params) { { id: vacancy.slug } }
+      let(:vacancy_page_view) { instance_double(VacancyPageView) }
+
+      it 'should call the track method if cookies not set' do
+        expect(VacancyPageView).to receive(:new).with(vacancy).and_return(vacancy_page_view)
+        expect(vacancy_page_view).to receive(:track)
+
+        subject
+      end
+
+      it 'should not call the track method if smoke_test cookies set' do
+        expect(VacancyPageView).not_to receive(:new).with(vacancy)
+        cookies[:smoke_test] = '1'
+
+        subject
+      end
+    end
   end
 end
--- a/spec/smoke_test/job_seekers_can_view_homepage_spec.rb
+++ b/spec/smoke_test/job_seekers_can_view_homepage_spec.rb
@@ -1,0 +1,27 @@
+require 'browser_test_helper'
+require 'i18n_helper'
+
+RSpec.describe 'Page availability', js: true, smoke_test: true do
+  context 'Job seeker visits vacancy page' do
+    it 'should ensure users can search and view a job vacancy page' do
+      page = Capybara::Session.new(:poltergeist)
+      page.driver.set_cookie('smoke_test', '1', domain: 'teaching-vacancies.service.gov.uk')
+
+      page.visit 'https://teaching-vacancies.service.gov.uk/'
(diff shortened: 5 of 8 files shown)


spec/smoke_test/job_seekers_can_view_homepage_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'browser_test_helper'
+require 'pry'
```
dushan-madetech: Did you mean to leave the require pry in?

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+RSpec.describe 'Page availability', js: true do
+  context 'Job seeker visits home page' do
+    it 'Page should be available' do
+      page = Capybara::Session.new(:poltergeist)
+      page.visit 'https://teaching-vacancies.service.gov.uk/'
+      expect(page).to have_content('Find a job in teaching')
+    end
+
+    it 'Job search should return a result' do
```
davidwinter: Can we merge this into the above `it` assertion block so that we have a single test for the user journey being tested? That way, we only visit the homepage the one, assert on that, then move onto the next page etc? If at any point the assertion fails, the test will too.

C-gyorfi:
[High level smoke tests for key jobseeker journey through TVS, and to send alerts if triggered [8]](https://trello.com/c/gxKqZ7aC/946-high-level-smoke-tests-for-key-jobseeker-journey-through-tvs-and-to-send-alerts-if-triggered-8)

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
dushan-madetech:
```suggestion
    it 'should return a result when performing a job search' do
```
perhaps this might sound better?

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'browser_test_helper'
+require 'pry'
+
+RSpec.describe 'Page availability', js: true do
+  context 'Job seeker visits home page' do
+    it 'Page should be available' do
```
davidwinter: With `rspec` I think it's important to ensure that when you back the test within the `it` block that it reads well. For example, 'it should ensure users can search and view a job vacancy page' would be:

```
it 'should ensure users can search and view a job vacancy page' do
```

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      page = Capybara::Session.new(:poltergeist)
+      page.visit 'https://teaching-vacancies.service.gov.uk/'
+      expect(page).to have_content('Find a job in teaching')
```
davidwinter: With Rails apps that use i18n, it's a good idea to use the i18n helper methods to reference a translation key rather than the translated text, just in case in future the translation changes, so that the tests don't break. So, would recommend using:

```
expect(page).to have_content(I18n.t('jobs.heading')
```

Further reading: https://guides.rubyonrails.org/i18n.html

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
C-gyorfi: yep, thanks! Resolved in a255f3e

davidwinter:
Also, I would wait until all feedback has been addressed, and once the PR has been approved, to rebase the commits in the branch, and if it makes sense, squash them into one commit with the commit naming convention this project uses, which can be found here: https://github.com/dxw/playbook/blob/master/guides/development-workflow.md#convention-1

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
C-gyorfi: Yes that makes sense, resolved this in cd60e72

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
C-gyorfi: Yes it reads much better, resolved this in cd60e72

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
C-gyorfi: Sure! resolved this in cd60e72

app/controllers/vacancies_controller.rb
```diff
@@ -126,4 +126,8 @@ def specific_phases?
   def set_headers
     response.set_header('X-Robots-Tag', 'noarchive')
   end
+
+  def smoke_test?
+    cookies[:smoke_test] != nil
```
davidwinter: Thought of potentially a better way to handle this with just:

```
cookies[:smoke_test].present?
```

spec/controllers/vacancies_controller_spec.rb
```diff
@@ -114,5 +114,26 @@
...
+      it 'should call the track method if cookies not set' do
+        expect(VacancyPageView).to receive(:new).with(vacancy).and_return(vacancy_page_view)
+        expect(vacancy_page_view).to receive(:track)
+
+        subject
+      end
+
+      it 'should not call the track method if smoke_test cookies set' do
+        expect(VacancyPageView).not_to receive(:new).with(vacancy)
+        expect(vacancy_page_view).not_to receive(:track)
```
davidwinter: I think this expectation is redundant and having the one above it should be sufficient.

.gitignore
```diff
@@ -27,6 +28,7 @@
 /config/application.yml
 
 /spec/fixtures/temp_schools_data.csv
+/spec/smoke_test/cookies.txt
```
davidwinter: Can the `cookies.txt` file be deleted, and then we don't need to add it to the `.gitignore` file?

.circleci/config.yml
```diff
@@ -0,0 +1,28 @@
...
+    working_directory: ~/teacher-vacancy-service
+    docker:
+      - image: circleci/ruby:2.6.1-node-browsers-legacy
+        environment:
+        RAILS_ENV: test
+    steps:
+      - checkout
+      - run: gem install bundler
+      - run: bundle install
+      - run: bundle exec rspec --tag smoke_test spec/smoke_test/
```
davidwinter: Is there a need for the directory path `spec/smoke_test/` or is having the `--tag smoke_test` argument sufficient? Rspec should scan all possible tests and would only use any with `smoke_test: true` tagged anyway. I guess it may speed up file traversing when searching for files to run.

.rspec
```diff
@@ -1 +1 @@
---require spec_helper
+--require spec_helper --tag ~smoke_test
```
davidwinter: Am just curious, did this work? And mean that the smoke tests weren't run by default?

app/controllers/vacancies_controller.rb
C-gyorfi: I was looking for this solution but it throws undefined method error

.circleci/config.yml
C-gyorfi: In this case, it directly looks into the spec/smoke_test folder instead of searching through all spec files in the project, the later also causing errors since most of the test requires the test server to be up

.rspec
C-gyorfi: yep, smoke tests are excluded, really neat solution thanks for that!

.gitignore
C-gyorfi: Yeh it actually deleted, but forgot to remove from .gitignore, resolved in 36439d1

spec/controllers/vacancies_controller_spec.rb
C-gyorfi: Resolved in 36439d1

davidwinter:
All looking good 👍 One last thing I can think of is if the test fails, where will notifications go? I know that we can set up Slack integration to have it alert in the relevant Slack channel, but is that sufficient? Should we have it send emails too to the product team I wonder? Only a blocker to this pull request if it requires some notification related changes to the `.circleci/config.yml` file.

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  context 'Job seeker visits vacancy page' do
+    it 'should ensure users can search and view a job vacancy page' do
+      page = Capybara::Session.new(:poltergeist)
+      page.driver.set_cookie('smoke_test', '1', domain: 'teaching-vacancies.service.gov.uk')
+
+      page.visit 'https://teaching-vacancies.service.gov.uk/'
+      expect(page).to have_content(I18n.t('jobs.heading'))
+
+      page.within '.filters-form' do
+        page.fill_in I18n.t('jobs.filters.subject'), with: 'Maths'
```
C-gyorfi: This is going to inflate the stats we have for the number of searches performed on the website in the audit spreadsheet.  We should agree on the approach for whether to keep this search data being recorded or not.

spec/controllers/vacancies_controller_spec.rb
```diff
@@ -60,6 +60,33 @@
...
+
+      context 'search audiotor' do
+        let(:params) do
+          {
+            subject: 'foo'
+          }
+        end
+
+        it 'should call the search auditor' do
+          expect(AuditSearchEventJob).to receive(:perform_later)
```
adrianclay: You can also assert against the value of the arguments given.  Otherwise we don't have any confidence about the `audit_row` method.

spec/controllers/vacancies_controller_spec.rb
```diff
@@ -60,6 +60,33 @@
...
+
+          subject
+        end
+
+        it 'should not call the search auditor if its a smoke test' do
+          cookies[:smoke_test] = 1
```
adrianclay: Forgive my knowledge of rails.  Does this get reset between tests?

spec/controllers/vacancies_controller_spec.rb
```diff
@@ -60,6 +60,33 @@
           subject
         end
       end
+
+      context 'search audiotor' do
```
adrianclay: Typo with "auditor".

app/controllers/vacancies_controller.rb
```diff
@@ -130,4 +130,12 @@ def set_headers
   def smoke_test?
     cookies[:smoke_test] != nil
   end
+
+  def valid_search?
+    @filters.any? && !smoke_test?
+  end
+
+  def audit_row
+    { total_count: @vacancies.total_count }.merge(@filters.audit_hash)
```
adrianclay: Does `VacanciesFinder` have a `total_count` method?

spec/controllers/vacancies_controller_spec.rb
C-gyorfi: Yes it does, according to my tests

app/controllers/vacancies_controller.rb
C-gyorfi: Yes, I mean VacanciesFinder.vacancies has!

spec/controllers/vacancies_controller_spec.rb
C-gyorfi: Thanks for pointing this out, fixed in 3701a28

spec/controllers/vacancies_controller_spec.rb
C-gyorfi: Resolved in 3701a28

spec/smoke_test/job_seekers_can_view_homepage_spec.rb
C-gyorfi: Resolved in fe4dbb8

app/controllers/vacancies_controller.rb
adrianclay: Ahh I'm such a fool.  Good point.

spec/controllers/vacancies_controller_spec.rb
adrianclay: That assertion still doesn't give any confidence about `total_count` being populated.  I believe we should add something which tests that?  Or remove it if it isn't important.

C-gyorfi: This been resolved in 1fc5397

spec/controllers/vacancies_controller_spec.rb
```diff
@@ -61,15 +61,22 @@
...
-            subject: 'foo'
+            job_title: 'Should have three match'
           }
         end
 
-        it 'should call the search auditor' do
-          expect(AuditSearchEventJob).to receive(:perform_later)
+        it 'should call the search auditor', elasticsearch: true do
+          3.times { create(:vacancy, job_title: 'Should have three match') }
+          Vacancy.__elasticsearch__.client.indices.flush
```
adrianclay: Eeek, we should move this into a nice place, but I'm happy to approve this for now.

C-gyorfi: I was surprised its not part of the teardown, but I found this solution at a few other spec files so I assumed its fine. I would suggest fixing this in another PR since this could involve some refactoring in other spec files e.g: `job_seekers_can_search_vacancies_spec.rb`

PR #961, opened 2019-07-10
(chore) changing server error content

C-gyorfi:
##  Trello card URL: 
https://trello.com/c/fR9rlofE/935-review-the-page-which-is-shown-when-there-is-an-error-with-the-dfe-authorisation-application-3
## Changes in this PR:
Change the content of internal server error


app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -20,7 +20,7 @@ def create
   rescue Authorisation::ExternalServerError => error
     Rollbar.log(:error, error)
     respond_to do |format|
```
adrianclay: Does re-raising the exception have the same effect without having a "respond_to" ?

PR #963, opened 2019-07-12
Fix change organisation link display

dushan-madetech:
## Trello card URL:

https://trello.com/c/FcMjGS2t/933-only-show-the-change-organisation-link-when-the-logged-in-user-is-invited-into-multiple-organisations-5

## Changes in this PR:

- show change organisation link to user only if they are part of multiple organisations (regardless of whether the DfE sign in feature flag is toggled on or off)

### DIFF
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -32,7 +32,7 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
   end
 
   def update_session(authorisation_permissions)
-    session.update(session_id: user_id, urn: school_urn, multiple_schools: authorisation_permissions.many?)
+    session.update(session_id: user_id, urn: school_urn, multiple_schools: authorisation_permissions.many_schools?)
     audit_successful_authorisation
   end
 
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -12,6 +12,8 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
 
   def create
     Rails.logger.warn("Hiring staff signed in: #{user_id}")
+    audit_successful_authentication
+
     if DfeSignInAuthorisationFeature.enabled?
       perform_dfe_sign_in_authorisation
     else
@@ -29,13 +31,8 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
     render 'user-not-authorised'
   end
 
-  def update_session(authorisation)
-    session.update(session_id: user_id, urn: school_urn, multiple_schools: authorisation.many?)
-    audit_successful_authorisation
-  end
-
-  def update_non_dsi_session(school_urn, permissions)
-    session.update(session_id: user_id, urn: school_urn, multiple_schools: permissions.many?)
+  def update_session(authorisation_permissions)
+    session.update(session_id: user_id, urn: school_urn, multiple_schools: authorisation_permissions.many?)
     audit_successful_authorisation
   end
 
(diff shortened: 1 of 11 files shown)


app/services/authorisation.rb
```diff
@@ -37,6 +38,14 @@ def authorised?
     role_ids.include?(dfe_sign_in_service_access_role_id)
   end
 
+  def many?
+    org_api_path = "/users/#{user_id}/organisations"
+    response = JSON.parse(get_dfe_sign_in_api_response(org_api_path).body)
+    return true if response.count > 1
```
adrianclay: This and the line below can be simplifed to returning the expression `response.count > 1`.

spec/services/authorisation_spec.rb
```diff
@@ -110,4 +110,20 @@
...
+      described_class.new(
+        organisation_id: '939eac36-0777-48c2-9c2c-b87c948a9ee0',
+        user_id: '161d1f6a-44f1-4a1a-940d-d1088c439da7'
+      )
+    end
+
+    before(:each) do
+      stub_sign_in_with_multiple_organisations
+    end
+    it 'should check if the user is a member of multiple organisations' do
```
adrianclay: We've only got 1 test at the moment to cover this functionality.  I have no cofidence that:

- The `user_id` above is being passed into the URL
- The method doesn't always return true

Can we get some triangulation going?

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -34,13 +36,8 @@ def not_authorised
...
 
-  def update_session
-    session.update(session_id: user_id, urn: school_urn)
-    audit_successful_authorisation
-  end
-
-  def update_non_dsi_session(school_urn, permissions)
-    session.update(session_id: user_id, urn: school_urn, multiple_schools: permissions.many?)
+  def update_session(authorisation_permissions)
+    session.update(session_id: user_id, urn: school_urn, multiple_schools: authorisation_permissions.many?)
```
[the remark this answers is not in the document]
C-gyorfi: Many permissions? I'm not a fan of this name either, but reading with the class name almost makes sense to me.

spec/services/authorisation_spec.rb
C-gyorfi: Added some test to check for sign in with a single organisation and also for different user id in 414ede9

app/services/authorisation.rb
C-gyorfi: Resolved this in 414ede9

app/services/authorisation.rb
```diff
@@ -37,6 +38,14 @@ def authorised?
     role_ids.include?(dfe_sign_in_service_access_role_id)
   end
 
+  def many?
+    org_api_path = "/users/#{user_id}/organisations"
+    response = JSON.parse(get_dfe_sign_in_api_response(org_api_path).body)
```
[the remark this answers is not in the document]
C-gyorfi: Made some changes in 414ede9 to fix this, although this might not be the cleanest way to solve this...

spec/services/authorisation_spec.rb
```diff
@@ -115,15 +115,31 @@
...
-    before(:each) do
-      stub_sign_in_with_multiple_organisations
+    context 'user is a member of multiple organisations' do
+      let(:user_id) { '161d1f6a-44f1-4a1a-940d-d1088c439da7' }
+      it 'should find that the user is a member of multiple organisations' do
+        stub_sign_in_with_multiple_organisations(user_id: user_id)
+        expect(subject.many?).to be(true)
+      end
+
+      let(:user_id) { 'another_user_id' }
```
adrianclay: There are 2 lets for the same variable in the same context.  Not sure this is behaving the way you expect.

app/services/authorisation.rb
```diff
@@ -40,8 +40,8 @@ def authorised?
 
   def many?
     org_api_path = "/users/#{user_id}/organisations"
-    response = JSON.parse(get_dfe_sign_in_api_response(org_api_path).body)
-    return true if response.count > 1
+    response = get_dfe_sign_in_api_response(org_api_path)
+    return JSON.parse(response.body).count > 1 if response.code.eql?('200')
 
     false
```
adrianclay: Do you have a test which covers this line?

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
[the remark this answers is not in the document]
C-gyorfi: `many_schools?`

app/services/authorisation.rb
C-gyorfi: Resolved this in 5db5651

spec/services/authorisation_spec.rb
C-gyorfi: Resolved this in 5db5651

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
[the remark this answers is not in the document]
C-gyorfi: Resolved this in 5db5651

PR #969, opened 2019-07-23
(chore) Set a specific user to a published vacancy

Renny4Real:
As part of a feature to nudge users into giving feedback on expired vacancies that are older than 2 weeks. 
We needed to:
- Associate a published vacancy with a user id.
- Identify which user to target with this feature i.e `publisher_user`
- Identify where in the application we would retrieve the user information.

### DIFF
--- a/app/controllers/hiring_staff/vacancies/publish_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/publish_controller.rb
@@ -1,9 +1,10 @@
 class HiringStaff::Vacancies::PublishController < HiringStaff::Vacancies::ApplicationController
   def create
     vacancy = Vacancy.find(vacancy_id)
+    user = current_user
     return redirect_to school_job_path(vacancy.id), notice: I18n.t('jobs.already_published') if vacancy.published?
 
-    if PublishVacancy.new(vacancy).call
+    if PublishVacancy.new(vacancy, user).call
       Auditor::Audit.new(vacancy, 'vacancy.publish', current_session_id).log
       AuditPublishedVacancyJob.perform_later(vacancy.id)
       update_google_index(vacancy) if vacancy.listed?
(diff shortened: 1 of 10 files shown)


app/controllers/hiring_staff/vacancies/publish_controller.rb
```diff
@@ -1,9 +1,10 @@
 class HiringStaff::Vacancies::PublishController < HiringStaff::Vacancies::ApplicationController
   def create
     vacancy = Vacancy.find(vacancy_id)
+    user = current_user
     return redirect_to school_job_path(vacancy.id), notice: I18n.t('jobs.already_published') if vacancy.published?
 
-    if PublishVacancy.new(vacancy).call
+    if PublishVacancy.new(vacancy, user).call
```
adrianclay: I _think_ inlining this variable would look nicer, but not 100% sure!

db/migrate/20190722142115_change_foreign_key_for_vacancies.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeForeignKeyForVacancies < ActiveRecord::Migration[5.2]
```
adrianclay: I'm not sure this is needed, as this column was only created for you locally.

Give me a shout and I can discuss this in more detail.

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -464,6 +464,16 @@
     end
 
     context '#publish' do
+      scenario 'adds the current user as a contact for feedback on the published vacancy' do
+        current_user = User.find_by(oid: session_id)
+        vacancy = create(:vacancy, :draft, school: school)
+
+        expect(PublishVacancy).to receive(:new).with(vacancy, current_user).and_call_original
```
adrianclay: I'd prefer this feature spec not to mock out this method.  I'd prefer it to be asserting against what you actually care about, i.e. the `publisher_user_id` column being populated.

PR #974, opened 2019-07-24
Remove service rating section on general feedback page

Renny4Real:
This is the first slice towards a redesign of the general feedback page.

### Before

Screenshot 2019-07-25 at 09 48 40

------------------------------------------------------------------------------------------------------

### After
Screenshot 2019-07-24 at 16 02 58

### DIFF
--- a/app/assets/stylesheets/components/feedback_form.scss
+++ b/app/assets/stylesheets/components/feedback_form.scss
@@ -5,12 +5,3 @@
     margin-top: 0;
   }
 }
-
-.general-feedback-form-group {
-  margin-top: 50px;
-  margin-bottom: 0px;
-
-  &:first-child {
-    margin-top: 0;
-  }
-}
--- a/app/assets/stylesheets/components/feedback_form.scss
+++ b/app/assets/stylesheets/components/feedback_form.scss
@@ -5,3 +5,12 @@
     margin-top: 0;
   }
 }
+
+.general-feedback-form-group {
+  margin-top: 50px;
+  margin-bottom: 0px;
+
+  &:first-child {
+    margin-top: 0;
+  }
+}
--- a/app/views/general_feedback/new.html.haml
+++ b/app/views/general_feedback/new.html.haml
@@ -5,7 +5,7 @@
   .govuk-grid-row
     .govuk-grid-column-two-thirds
 
-      .govuk-form-group.general-feedback-form-group
+      .govuk-form-group.feedback-form-group
         %fieldset.govuk-fieldset
           %legend.govuk-fieldset__legend.govuk-fieldset__legend--m
             %h2.govuk-fieldset__heading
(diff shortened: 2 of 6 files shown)


app/views/general_feedback/new.html.haml
```diff
@@ -23,6 +23,15 @@
...
                 input_html: { class: 'form-control form-control-4-4', rows: '5' }
 
-      = render partial: '/feedback/fields', locals: { form: f }
+      .govuk-form-group.general-feedback-form-group
+        %fieldset.govuk-fieldset
+          %legend.govuk-fieldset__legend.govuk-fieldset__legend--m
+            %h2.govuk-fieldset__heading
+              = t('feedback.comment_label')
+
+      = f.input :comment, as: :text,
```
adrianclay: I think the indentation here is off, which is causing the text box to be placed outside the fieldset.

You can read up on how fieldsets should be used here:
https://design-system.service.gov.uk/components/fieldset/

app/assets/stylesheets/components/feedback_form.scss
```diff
@@ -5,3 +5,12 @@
     margin-top: 0;
   }
 }
+
+.general-feedback-form-group {
```
adrianclay: I'm suprised custom CSS is required, using the design system correctly usually means you don't need to make any modifications out of the box.  I suspect that if you fix the fieldset issue below the CSS won't be required.

spec/features/general_feedback_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+RSpec.feature 'Giving general feedback for the service' do
+  before(:each) { visit new_feedback_path }
+
+  context 'when viewing the visiting purpose section' do
+    it 'displays the section heading' do
+      expect(page).to have_content(I18n.t('general_feedback.visit_purpose_legend'))
```
adrianclay: How much value does having these tests which assert against page content bring?

Renny4Real: I feel like asserting against content gives me more confidence that on the desired page (i.e that the test setup is correct). 
Although it might be enough to just assert on the name of the page.

spec/features/general_feedback_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+RSpec.feature 'Giving general feedback for the service' do
+  before(:each) { visit new_feedback_path }
+
+  it 'displays the general feedback page' do
+    expect(page).to have_content('Give feedback')
```
adrianclay: Minor: Not blocking this PR, it's nicer to use the internationalised versions of these strings, to reduce flakiness when the translation string fails.

Renny4Real: Sure, I changed it at the last moment.
(Wasn't sure if the internationalised version helped with the readability of the test at first glance.)

PR #978, opened 2019-07-29
Email hiring staff to request feedback

C-gyorfi:
## Trello card URL:
https://trello.com/c/w6XuSGrN/970-e-mail-hiring-staff-2-weeks-after-a-vacancy-expires-to-ask-them-to-provide-feedback-8

## Changes in this PR:
- Add new mailer to prompt hiring staff for feedback on expired vacancies which are older than 2 weeks
- Create a new job for sending out feedback prompt emails
- Add rake task to schedule emails
- Add unit tests and end to end test accordingly

## Screenshots of UI changes:
Screenshot 2019-07-29 at 10 44 10

## Next steps:

- [ ] Check new cron job 
- [ ] Start recording user email on sign in

### DIFF
--- a/app/jobs/send_expired_vacancy_feedback_email_job.rb
+++ b/app/jobs/send_expired_vacancy_feedback_email_job.rb
@@ -3,9 +3,8 @@ class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
 
   def perform
     expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
-      FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
-
       unless user.email.nil?
+        FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later
         Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{users_vacancies.count} vacancies")
       end
     end
--- a/app/jobs/send_expired_vacancy_feedback_email_job.rb
+++ b/app/jobs/send_expired_vacancy_feedback_email_job.rb
@@ -1,0 +1,16 @@
+class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
+  queue_as :email_feedback_prompt
+
+  def perform
+    expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
+      FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
+    end
+  end
+
+  private
+
+  def expired_vacancies
+    Vacancy.where(expires_on: Time.zone.now - 2.weeks, hired_status: nil)
+           .where.not(publisher_user: nil)
+  end
+end
--- a/app/jobs/send_expired_vacancy_feedback_email_job.rb
+++ b/app/jobs/send_expired_vacancy_feedback_email_job.rb
@@ -4,6 +4,7 @@ class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
   def perform
     expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
       FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
+      Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{users_vacancies.count} vacancies") unless user.email.nil?
     end
   end
 
--- a/app/jobs/send_expired_vacancy_feedback_email_job.rb
+++ b/app/jobs/send_expired_vacancy_feedback_email_job.rb
@@ -4,7 +4,10 @@ class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
   def perform
     expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
       FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
-      Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{users_vacancies.count} vacancies") unless user.email.nil?
+
+      unless user.email.nil?
+        Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{users_vacancies.count} vacancies")
+      end
     end
   end
 
--- a/variables.tf
+++ b/variables.tf
@@ -164,6 +164,17 @@ variable "send_job_alerts_daily_email_task_schedule" {
   default     = "cron(0 08 * * ? *)"
 }
 
+variable "send_feedback_prompt_email_task_command" {
+  description = "The Entrypoint for the send_feedback_prompt_email task"
+  default     = ["rake", "verbose", "feedback_prompt_email:send"]
+}
+
+variable "send_feedback_prompt_email_task_schedule" {
+  description = "send_feedback_prompt_email schedule expression - https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  default     = "cron(0 09 * * ? *)"
+}
+
+
 variable "import_schools_task_command" {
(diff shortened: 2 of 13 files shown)


variables.tf
```diff
@@ -164,6 +164,17 @@ variable "send_job_alerts_daily_email_task_schedule" {
...
 }
 
+variable "send_feedback_prompt_email_task_command" {
+  description = "The Entrypoint for the send_feedback_prompt_email task"
+  default     = ["rake", "verbose", "feedback_prompt_email:send"]
+}
+
+variable "send_feedback_prompt_email_task_schedule" {
+  description = "send_feedback_prompt_email schedule expression - https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  default     = "cron(0 09 * * ? *)"
```
C-gyorfi: Set this to run at 9 every day but this was just my best guess at a suitable time. Could this be a different time and ideally when should this cron be run?

variables.tf
```diff
@@ -164,6 +164,17 @@ variable "send_job_alerts_daily_email_task_schedule" {
...
+}
+
+
```
C-gyorfi: Extra empty line

C-gyorfi:
[E-mail hiring staff 2 weeks after a vacancy expires to ask them to provide feedback [8]](https://trello.com/c/w6XuSGrN/970-e-mail-hiring-staff-2-weeks-after-a-vacancy-expires-to-ask-them-to-provide-feedback-8)

variables.tf
dushan-madetech:
```suggestion
```

app/jobs/send_expired_vacancy_feedback_email_job.rb
```diff
@@ -0,0 +1,16 @@
+class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
+  queue_as :email_feedback_prompt
```
adrianclay: Creating a new queue name here will require the creation of the queue config inside ‘sidekiq.yml’ configuration file.  Otherwise the job won’t be picked up by a worker and run.

You can test this locally by running the `rake` task locally and seeing the worker logs don’t pick anything up.

If you’re also using sidekiq to perform this asynchronously, you way want to look at the retry system that sidekiq has in place.  If there was an issue with this worker which caused it to die midway through, we wouldn’t want some users to get multiple emails.  It would be safer that if the job died, that it didn’t get rescheduled for the same day.

variables.tf
adrianclay: Could be at any time 🙂.

9am sounds sensible to me, but maybe ask in the #tv_product room for any suggestions otherwise.

Also be aware, what does this config do in regard to summer time? Is it 9 UTC or 9 local time?

app/jobs/send_expired_vacancy_feedback_email_job.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def perform
+    expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
+      FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
```
adrianclay: Did we want to put some logging in, similar to the other Jobs? Can be useful for debugging in the future.

variables.tf
dushan-madetech: @adrianclay The cron times are in UTC as per https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html

app/jobs/send_expired_vacancy_feedback_email_job.rb
dushan-madetech: Are we looking to log errors or just log emails being sent e.g. adding this line `Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{expired_vacancies.count} vacancies")` after the each do block

app/jobs/send_expired_vacancy_feedback_email_job.rb
dushan-madetech: I've been doing some digging around this and it seems like we can't really configure the job to not be retried using Sidekiq's retry configuration. We're using Sidekiq through Rails Active Job i.e. 
```ruby 
config.active_job.queue_adapter = :sidekiq
``` 
As per the Sidekiq wiki entry for [Active Job](https://github.com/mperham/sidekiq/wiki/Active-Job#customizing-error-handling) we can't make use of sidekiq's retry system. It seems we can add a rescue from or rescue that doesn't use retries when Active Job https://github.com/rails/activejob/issues/47#issuecomment-73525302 but it seems to rely on rescuing exceptions

C-gyorfi: When I run `bin/drake feedback_prompt_email:send` locally, seems like its queuing the job without any issues, this is the output:
`[ActiveJob] Enqueued SendExpiredVacancyFeedbackEmailJob (Job ID: d042ed5b-2cd7-4346-bbe5-2e0b30d8f91a) to Sidekiq(email_feedback_prompt)`

Any idea why?

adrianclay: So that output is saying that the job is being enqueued.

>Otherwise the job won’t be picked up by a worker and run.

Did you see any output which suggested that a worker picked up and executed the job?

A way to get round the Retrys would be for the Job to be exectued syncronously, e.g. by using `perform_now` inside of the `rake` task.

C-gyorfi: I've just executed syncronously with the output:
`[ActiveJob] [SendExpiredVacancyFeedbackEmailJob] [0f654051-bf62-4ae8-90e6-7773d2ae4533] Performed SendExpiredVacancyFeedbackEmailJob (Job ID: 0f654051-bf62-4ae8-90e6-7773d2ae4533) from Sidekiq(email_feedback_prompt) in 291.07ms`

no change needed in `sidekiq.yml` 🤔

adrianclay: Yep, if executing synchronously.

app/jobs/send_expired_vacancy_feedback_email_job.rb
dushan-madetech: Added a fix in https://github.com/DFE-Digital/teacher-vacancy-service/pull/978/commits/0bd426bf0b8b2b66536f4ff3b2cf049cb7c0c702

app/jobs/send_expired_vacancy_feedback_email_job.rb
C-gyorfi: Changed this to synchronous in d56a02a

app/jobs/send_expired_vacancy_feedback_email_job.rb
C-gyorfi: @dushan-madetech Rubocop is crying for the long line 😢

dushan-madetech:
> Looks good, but can you address the odd commit history before merging?

@adrianclay will perform an interactive rebase with @C-gyorfi to fix up the history before merging

app/jobs/send_expired_vacancy_feedback_email_job.rb
```diff
@@ -4,7 +4,10 @@ class SendExpiredVacancyFeedbackEmailJob < ApplicationJob
   def perform
     expired_vacancies.group_by(&:publisher_user).each do |user, users_vacancies|
       FeedbackPromptMailer.prompt_for_feedback(user.email, users_vacancies).deliver_later unless user.email.nil?
-      Rails.logger.info("Sidekiq: Sending feedback prompt emails for #{users_vacancies.count} vacancies") unless user.email.nil?
+
+      unless user.email.nil?
```
adrianclay: You can move line 6 into this block and remove the additional unless at the end of that line, minor refactoring.

C-gyorfi: Resolved in 38041b7

PR #980, opened 2019-07-29
(chore) Add domains to the linker parameter

C-gyorfi:
We want to be able to track user journey before they get onto
the TVS website, there are currently 2 other .gov.uk sites that are
related to TVS, the linker parameter should treat them as part of
TVS.

### DIFF
--- a/app/views/shared/_gtag.html.haml
+++ b/app/views/shared/_gtag.html.haml
@@ -4,4 +4,7 @@
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
-  gtag('config', "#{ENV['GOOGLE_ANALYTICS']}", { 'anonymize_ip': true, 'linker': { 'accept_incoming': 'true' } });
+  gtag('config', "#{ENV['GOOGLE_ANALYTICS']}", { 'anonymize_ip': true, 'linker': {
+     'domains': ['gov.uk/find-teaching-job', 'gov.uk/guidance/list-a-teaching-job-at-your-school-on-teaching-vacancies'],
+    'accept_incoming': 'true' }
+  });


C-gyorfi:
[As a user of the Teaching Vacancies Google Analytics account, I need to know where users started their journey before the gov.uk Teaching Vacancies pages so that I can assess how effective our SEO, PPC and comms efforts are [3]](https://trello.com/c/M0b8Rsez/961-as-a-user-of-the-teaching-vacancies-google-analytics-account-i-need-to-know-where-users-started-their-journey-before-the-govuk-t)

app/views/shared/_gtag.html.haml
```diff
@@ -4,4 +4,7 @@
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
-  gtag('config', "#{ENV['GOOGLE_ANALYTICS']}", { 'anonymize_ip': true, 'linker': { 'accept_incoming': 'true' } });
+  gtag('config', "#{ENV['GOOGLE_ANALYTICS']}", { 'anonymize_ip': true, 'linker': {
+     'domains': ['gov.uk/find-teaching-job', 'gov.uk/guidance/list-a-teaching-job-at-your-school-on-teaching-vacancies'],
```
adrianclay: These look like URLs and not domains to me 🤔 .

adrianclay: Also the documentation says that you only need to add this config if we're the site which links to those domains (which we aren't).

C-gyorfi: In that case, this is something we should add to `/find-teaching-job` and `/guidance/list-a-teaching-job-at-your-school-on-teaching-vacancies` sites, however, the domain is gov.uk in each of these cases. Feels like this is an all or nothing situation

adrianclay: I’m not sure what you’re saying 🤷‍♂️.

C-gyorfi: According to your suggestion, we need to add this code to the sites which link to TVS? But the "domain" itself is `gov.uk` in each case 🤯

adrianclay: Yep, if you see Isabelle’s comment in the Trello card there are changes being made to GOV.uk

PR #983, opened 2019-08-01
Add canonical tagging

dushan-madetech:
## Trello card URL:
https://trello.com/c/Uriikjvs/983-canonical-tagging-and-pages-with-pagination

## Changes in this PR:
- Added canonical URL to head using request URL

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -17,6 +17,8 @@
     %link{href: asset_path("govuk-apple-touch-icon-167x167.png"), rel: "apple-touch-icon", sizes: "167x167"}/
     %link{href: asset_path("govuk-apple-touch-icon-152x152.png"), rel: "apple-touch-icon", sizes: "152x152"}/
     %link{href: asset_path("govuk-apple-touch-icon.png"), rel: "apple-touch-icon"}/
+    - if content_for? :head_links
+      = yield(:head_links) 
     /[if !IE 8]
     = stylesheet_link_tag 'application', media: 'all'
     /[if lte IE 8]
--- a/app/views/vacancies/index.html+phone.haml
+++ b/app/views/vacancies/index.html+phone.haml
@@ -1,4 +1,5 @@
 - content_for :page_title_prefix, t('jobs.heading')
+= render partial: 'head_links'
 
 %h1.govuk-heading-xl
   = t('jobs.heading')
--- a/app/views/vacancies/index.html.haml
+++ b/app/views/vacancies/index.html.haml
@@ -1,4 +1,5 @@
 - content_for :page_title_prefix, t('jobs.heading')
+= render partial: 'head_links'
 
 %h1.govuk-heading-xl
   = t('jobs.heading')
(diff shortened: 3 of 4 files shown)


dushan-madetech:
Not adding next/prev rel links as these are no longer used by Google for indexing. As mentioned in the Trello card there are 2 options around canonical URLs with search parameters. I've implemented the second option but the first commit represents the first option. Can roll it back to that after discussion with SEO agency for the best approach

dushan-madetech:
> I'm not sure this canonical link is adding the value that we believe it is, I thought it was to do with duplicate content. I'm not sure how adding a metatag which just displays what the `request.url` is would add any value.

It's treating the pages of the paginated search as different pages - so it's not treating the different pages of the search as duplicate content, instead each page of the search is seen as unique. Each search with different parameters is also not being treated as duplicated content but is seen as unique. I will clarify with the agency when I speak to them

app/views/layouts/application.html.haml
```diff
@@ -17,6 +17,9 @@
     %link{href: asset_path("govuk-apple-touch-icon-167x167.png"), rel: "apple-touch-icon", sizes: "167x167"}/
     %link{href: asset_path("govuk-apple-touch-icon-152x152.png"), rel: "apple-touch-icon", sizes: "152x152"}/
     %link{href: asset_path("govuk-apple-touch-icon.png"), rel: "apple-touch-icon"}/
+    %link{href: request.url, rel: "canonical"}/
```
adrianclay: I’d prefer having the canonical link as only part of the vacancies listing page. From that call, I believe the intention is to add duplicate content for this and have them redirect to a single canonical version.

app/views/vacancies/index.html.haml
```diff
@@ -1,4 +1,10 @@
 - content_for :page_title_prefix, t('jobs.heading')
+- unless (path_to_next_page @vacancies).nil?
+  = content_for :for_head_link do
```
adrianclay: I find that `content_for :for_head_link` reads a bit odd, the double for especially.  `content_for :head_links` maybe?

dushan-madetech: Thank you, I really like the suggested name so I've changed it in https://github.com/DFE-Digital/teacher-vacancy-service/pull/983/commits/6ce0868d306fb803f739a17b146644b12c153e2b

app/views/layouts/application.html.haml
dushan-madetech: That makes sense -  I will move this to the appropriate view

dushan-madetech: Resolved this in d0d5072

app/views/vacancies/index.html+phone.haml
```diff
@@ -1,4 +1,12 @@
 - content_for :page_title_prefix, t('jobs.heading')
+= content_for :head_links do
```
adrianclay: Could the duplication between the 2 views be fixed with a partial?

dushan-madetech: Good spot, I didn't actually know that this is what partials were for. I've made the change in https://github.com/DFE-Digital/teacher-vacancy-service/pull/983/commits/82ce1e8f17dbcef3e8eb058da54d0d36ccf09b92

PR #907, opened 2019-05-14
Understand the impact of teaching vacancies on schools

pezholio:
## Trello card URL:

https://trello.com/c/jau6X6EC/804-epic-understand-the-impact-of-teaching-vacancies-on-schools

## Changes in this PR:

This updates the Hiring Staff UI to add a section that nudges users to provide information on expired jobs - asking them if they posted the vacancy anywhere else / if they filled the vacancy.

This is still in user testing, so subject to change, but opened a draft PR to get feedback

## Screenshots of UI changes:

image

PR #984, opened 2019-08-01
Add user participation request section to general feedback

Renny4Real:
### Changes in this PR:
This is the second slice of a feature to redesign the general feedback form for the service.
First slice: (#974)

### Before
Screenshot 2019-08-01 at 12 29 14

### After
Screenshot 2019-08-01 at 12 28 33

## Next steps:

- [ ] Check the new section works on staging? (Criteria: The user should be required to give a valid email address only when they select the **Yes** radio button.)

- [ ] Add new headers to the General Feedback worksheet in the Staging audit spreadsheet for the user response and email columns. (NOTE: Inform Business Analyst when adding these headers to the Production audit spreadsheet.)

- [ ] Check that running the update spreadsheet task on staging updates the Staging version of General feedback worksheet with user response data and email address data.

### DIFF
--- a/app/models/general_feedback.rb
+++ b/app/models/general_feedback.rb
@@ -23,7 +23,9 @@ class GeneralFeedback < ApplicationRecord
       visit_purpose_comment,
       rating,
       comment,
-      created_at.to_s
+      created_at.to_s,
+      user_participation_response,
+      email
     ]
   end
 end
--- a/app/models/general_feedback.rb
+++ b/app/models/general_feedback.rb
@@ -8,8 +8,13 @@ class GeneralFeedback < ApplicationRecord
   validates :visit_purpose_comment, length: { maximum: 1200 }, if: :visit_purpose_comment?
   validates :comment, length: { maximum: 1200 }, if: :comment?
   validates :user_participation_response, presence: true
+  validates :email, presence: true, if: :user_is_interested?
 
   scope :published_on, (->(date) { where(created_at: date.all_day) })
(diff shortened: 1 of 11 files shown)


spec/features/general_feedback_spec.rb
```diff
@@ -6,13 +6,55 @@
...
+
+      click_on 'Submit feedback'
+
+      expect(page).to have_content('Your feedback has been successfully submitted')
+    end
+
+    scenario 'must have an email when participation response is Yes' do
+      choose 'Find a job in teaching'
+      fill_in 'general_feedback_comment', with: 'Keep going!'
+      choose('general_feedback_user_participation_response_interested')
```
dushan-madetech: You could extract this to a method for added clarity in the spec, perhaps something like `choose_yes`

dushan-madetech: similarly for choose no

app/models/general_feedback.rb
```diff
@@ -2,21 +2,30 @@ class GeneralFeedback < ApplicationRecord
   include Auditor::Model
 
   enum visit_purpose: %i[other_purpose find_teaching_job list_teaching_job]
+  enum user_participation_response: %i[interested not_interested]
```
adrianclay: It is best practice to add the specific integer values for each value, see discussion here: 

https://github.com/DFE-Digital/teacher-vacancy-service/pull/907#discussion_r294832864

PR #993, opened 2019-08-06
Make page titles more descriptive (accessibility fix)

dushan-madetech:
## Trello card URL:
https://trello.com/c/fEW99Kpc/184-make-page-titles-descriptive-1

## Changes in this PR:
- Add gov.uk to page titles of every page

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -8,7 +8,7 @@
     = render partial: 'shared/open_graph'
     = render partial: 'shared/meta'
     %meta{charset: "utf-8"}/
-    %title #{content_for :page_title_prefix} — #{t('app.title')}
+    %title #{content_for :page_title_prefix} — #{t('app.title')} #{t('app.govuk')}
     %meta{content: "width=device-width, initial-scale=1", name: "viewport"}/
     %meta{content: "blue", name: "theme-color"}/
     %link{href: asset_path("favicon.ico"), rel: "shortcut icon", type: "image/x-icon"}/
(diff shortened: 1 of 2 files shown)


dushan-madetech:
I've just added this in the format 'gov.uk' but not sure if it should maybe be 'GOV.UK' instead? If so I can easily change it. Was debating between adding it directly or adding it by adding something to the yaml file so that it can be used that way instead e.g. t(app.service-domain) = 'GOV.UK' but added it directly. I can also change this if necessary

app/views/layouts/application.html.haml
```diff
@@ -8,7 +8,7 @@
     = render partial: 'shared/open_graph'
     = render partial: 'shared/meta'
     %meta{charset: "utf-8"}/
-    %title #{content_for :page_title_prefix} — #{t('app.title')}
+    %title #{content_for :page_title_prefix} — #{t('app.title')} #{"gov.uk"}
```
C-gyorfi: Just wondering how does this look like? There is a `—` between `#{content_for :page_title_prefix} — #{t('app.title')}` but not before `#{"gov.uk"}`

dushan-madetech: It just has an empty space between the two, here's an example: 
image. Hope that provides more clarity 😄

dushan-madetech: I can  amend it to also add the `-` between Teaching Vacancies and gov.uk if this is desired

C-gyorfi: https://design-system.service.gov.uk/styles/ suggests using `GOV.UK`

C-gyorfi: I dont think we need `-` between Teaching Vacancies and gov.uk

dushan-madetech:
Amended to use localization to get GOV.UK into the title

PR #994, opened 2019-08-06
remove id duplication (accessibility fix)

dushan-madetech:
## Trello card URL:
https://trello.com/c/3s4j1pR9/971-fix-duplicate-id-attributes-found-on-some-pages-problematic-for-assistive-technologies-1

## Changes in this PR:
- amend duplicate id

### DIFF
--- a/app/views/hiring_staff/vacancies/_show.html.haml
+++ b/app/views/hiring_staff/vacancies/_show.html.haml
@@ -13,10 +13,10 @@
       .govuk-grid-column-one-half
         .share-url.mb1.display-none
           %h2.govuk-heading-m.mt0 Share this job
-          #{link_to t('jobs.view_public_link'), @vacancy.share_url, class: 'govuk-link', id: 'share-url', 'aria-label': t('jobs.view_public_link')}, or #{link_to t('jobs.copy_public_url'), '#', class: 'govuk-link copy-to-clipboard', 'data-clipboard-text': @vacancy.share_url, 'aria-label': t('jobs.copy_public_url')}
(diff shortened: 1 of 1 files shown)


app/views/hiring_staff/vacancies/_show.html.haml
```diff
@@ -16,7 +16,7 @@
           #{link_to t('jobs.view_public_link'), @vacancy.share_url, class: 'govuk-link', id: 'share-url', 'aria-label': t('jobs.view_public_link')}, or #{link_to t('jobs.copy_public_url'), '#', class: 'govuk-link copy-to-clipboard', 'data-clipboard-text': @vacancy.share_url, 'aria-label': t('jobs.copy_public_url')}
         .share-url-no-js.mb1
           %h2.govuk-heading-s.mt0 Share this job
-          #{link_to "#{t('jobs.view_public_link')} (#{@vacancy.share_url})", job_path(@vacancy), class: 'govuk-link', id: 'share-url'}
+          #{link_to "#{t('jobs.view_public_link')} (#{@vacancy.share_url})", job_path(@vacancy), class: 'govuk-link', id: 'share-url-no-js'}
```
adrianclay: Is the `id` even valuable for these links?  If not might be easier to remove the `id`s?

dushan-madetech: conversely, is there any harm in leaving the `id`s on the elements?

adrianclay: Only that other developers will come along and wonder why this element has an ID, when it doesn't seem necessary 😛

dushan-madetech: amended this in https://github.com/DFE-Digital/teacher-vacancy-service/pull/994/commits/a09302c03e19a75cd6abf7dadc523090fc6a5bf4

PR #996, opened 2019-08-07
fix for attribute of labels for month input fields (accessibility fix)

dushan-madetech:
## Trello card URL:
https://trello.com/c/YiYqRJyA/974-provide-descriptive-labels-to-form-elements-missing-them-2
## Changes in this PR:
- Change gemfile to use fixed version of gov-uk-date-fields gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,8 +42,7 @@ gem 'roadie-rails'
 gem 'simple_form'
 gem 'rails-html-sanitizer', '~> 1.0.4', '>= 1.0.4' # Must be above this version due to CVE-2018-3741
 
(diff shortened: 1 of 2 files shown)


dushan-madetech:
> I'm willing to give the other option one more try (with help, of course): https://trello.com/c/t41kRgCV
> 
> Before merging this.

I'd prefer to get this merged in so that a fix is released in this sprint then we can do a production release with a bunch of accessibility fixes (+anything else that gets picked up). I'm also leaning towards merging this now as we've already had a timeboxed attempt at the ideal fix - ideally we should point it up as we now have a seperate trello card and then we can work on it in this sprint or a future sprint

PR #1004, opened 2019-08-12
Stop using forked version gem

Renny4Real:
From a maintenance standpoint, we stopped using the forked version gem and used the [original gem](https://github.com/ministryofjustice/gov_uk_date_fields).

### Next Steps
- [ ] Check all the forms where there are date fields are rendered properly.

#### Forms Checklist:
When creating a new vacancy:
- [ ] Job Specification
- [ ] Application Details

When editing a vacancy:
- [ ] Job Specification
- [ ] Application Details

When copying a vacancy:
- [ ] Job Specification
- [ ] Application Details

### DIFF
--- a/app/views/hiring_staff/vacancies/application_details/new.html.haml
+++ b/app/views/hiring_staff/vacancies/application_details/new.html.haml
@@ -26,16 +26,10 @@
 
       %div.govuk-form-group#publish_on
         = f.gov_uk_date_field :publish_on,
-                              legend_text: t('jobs.publication_date'),
-                              legend_class: 'govuk-label',
-                              form_hint_text: t('jobs.form_hints.publication_date',
-                                              date: l(Date.today, format: :hinttext))
+                              legend_options: { page_heading: false, class: "govuk-label" }
 
       %div.govuk-form-group#expires_on
         = f.gov_uk_date_field :expires_on,
-                              legend_text: t('jobs.deadline_date'),
-                              legend_class: 'govuk-label',
-                              form_hint_text: t('jobs.form_hints.deadline_date',
-                                                date: l(Date.today + 2.months, format: :hinttext))
+                              legend_options: { page_heading: false, class: "govuk-label" }
 
       = f.button :submit, t('buttons.save_and_continue')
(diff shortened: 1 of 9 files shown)


app/views/hiring_staff/vacancies/application_details/new.html.haml
```diff
@@ -26,16 +26,12 @@
 
       %div.govuk-form-group#publish_on
         = f.gov_uk_date_field :publish_on,
-                              legend_text: t('jobs.publication_date'),
-                              legend_class: 'govuk-label',
-                              form_hint_text: t('jobs.form_hints.publication_date',
-                                              date: l(Date.today, format: :hinttext))
+                              i18n_attribute: :publish_on,
```
adrianclay: I remember you saying that the `i18n_attribute` was optional and that it will use the attribute name if not given.  With the cases where the `i18n_attribute` and the attribute name are the same can we just remove the `i18n_attribute` definition?

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -150,11 +150,11 @@
           expect(page).to have_content(I18n.t('activerecord.errors.models.vacancy.attributes.contact_email.blank'))
         end
 
-        within_row_for(element: 'legend', text: I18n.t('jobs.deadline_date')) do
+        within_row_for(element: 'span', text: I18n.t('jobs.deadline_date')) do
           expect(page).to have_content(I18n.t('activerecord.errors.models.vacancy.attributes.expires_on.blank'))
         end
 
-        within_row_for(element: 'legend', text: I18n.t('jobs.publication_date')) do
+        within_row_for(element: 'span', text: I18n.t('jobs.publication_date')) do
```
adrianclay: Does the gem now generate a `fieldset` without a `legend`?  This doesn't sound great.  I'm not aware of how a screen-reader would react in this scenario, but it feels dodgy.

app/views/hiring_staff/vacancies/application_details/new.html.haml
Renny4Real: Thanks for pointing this out. Yes, we can remove them.

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
Renny4Real: Thanks.
The fieldset does have a legend. 
However, after comparing the forms on development and staging, have realised that what should have been the legend text was written as hint text instead. My mistake!

PR #1008, opened 2019-08-22
Add expiry time to vacancy schema

C-gyorfi:
## Trello card URL:
https://trello.com/c/nDuQJ51p/1010-add-time-to-vacancy-deadline-so-that-jobseekers-know-how-long-they-have-to-apply-and-dont-click-on-expired-jobs-sliced

## Changes in this PR:
- Changed the scope queries in the vacancy model, so that
expiry_time is used in queries when it's not nil, while
tolerates the absence of this value. So that existing vacancies
that have no expiry_time can be expired based on their expires_on date,
just like the old behaviour. Added tests accordingly.

- Changed the VacancyPresenter to used expiry_time when its 
not null.

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -123,13 +123,27 @@ class Vacancy < ApplicationRecord
 
   acts_as_gov_uk_date :starts_on, :ends_on, :publish_on, :expires_on
 
-  scope :applicable, (-> { where('expires_on >= ?', Time.zone.today) })
+  scope :applicable, (-> { applicable_by_date.or(applicable_by_time) })
+  scope :applicable_by_time, (-> { where.not(expiry_time: nil).where('expiry_time >= ?', Time.zone.now) })
+  scope :applicable_by_date, (-> { where(expiry_time: nil).where('expires_on >= ?', Time.zone.today) })
   scope :active, (-> { where(status: %i[published draft]) })
   scope :listed, (-> { published.where('publish_on <= ?', Time.zone.today) })
   scope :published_on_count, (->(date) { published.where(publish_on: date.all_day).count })
   scope :pending, (-> { published.where('publish_on > ?', Time.zone.today) })
-  scope :expired, (-> { published.where('expires_on < ?', Time.zone.today) })
-  scope :live, (-> { published.where('publish_on <= ?', Time.zone.today).where('expires_on >= ?', Time.zone.today) })
+  scope :expired, (-> { expired_by_time.or(expired_by_date) })
+  scope :expired_by_time, (-> { published.where.not(expiry_time: nil).where('expiry_time < ?', Time.zone.now) })
+  scope :expired_by_date, (-> { published.where(expiry_time: nil).where('expires_on < ?', Time.zone.today) })
+  scope :live, (-> { live_by_date.or(live_by_time) })
+  scope :live_by_time, (lambda {
+                          published.where.not(expiry_time: nil)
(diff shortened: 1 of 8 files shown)


C-gyorfi:
[Change the vacancy model [5]](https://trello.com/c/ZquLp0uY/1019-change-the-vacancy-model-5)

app/models/vacancy.rb
```diff
@@ -123,13 +123,27 @@ class Vacancy < ApplicationRecord
 
   acts_as_gov_uk_date :starts_on, :ends_on, :publish_on, :expires_on
 
-  scope :applicable, (-> { where('expires_on >= ?', Time.zone.today) })
+  scope :applicable, (-> { applicable_by_date.or(applicable_by_time) })
+  scope :applicable_by_time, (-> { where.not(expiry_time: nil).where('expiry_time >= ?', Time.zone.now) })
```
Renny4Real: I could be wrong but could we condense the SQL queries to use only one where statement?

C-gyorfi: Possibly, although the first `where` is a `where.not` 🤔

db/schema.rb
```diff
@@ -216,6 +216,7 @@
     t.integer "hired_status"
     t.datetime "stats_updated_at"
     t.uuid "publisher_user_id"
+    t.datetime "expiry_time"
     t.index ["expires_on"], name: "index_vacancies_on_expires_on"
```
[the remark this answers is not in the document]
dushan-madetech: That seems like a useful thing to have, we'll get that column indexed as suggested. Thanks! 😄

spec/models/vacancy_spec.rb
```diff
@@ -312,17 +312,34 @@
...
+        end
+      end
 
-        results = Vacancy.applicable
-        expect(results).to include(expires_today)
-        expect(results).to include(expires_future)
-        expect(results).to_not include(expired)
+      context 'when expiry time given' do
+        it 'finds current vacancies' do
+          expired_earlier_today = build(:vacancy, expires_on: Time.zone.today,
```
Renny4Real: This could be made easier to read i.e. rather than defining one expired vacancy, doing some actions on it and then defining the second one. Maybe define both before doing the actions?

e.g `expired_earlier_today` and `expires_later_today` could be built/created in the context block using `let`.

app/models/vacancy.rb
Renny4Real: We could probably have something like `where(condition AND NOT condition)`. 
There are also other places where we're doing `.where(condition).where(condition)`. 
Could these be written as `where(condition AND condition)`?

spec/models/vacancy_spec.rb
C-gyorfi: Resolved in 3d78ebf

db/schema.rb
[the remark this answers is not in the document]
C-gyorfi: Resolved in 2a700b8

app/models/vacancy.rb
C-gyorfi: Resolved in 2a700b8

db/schema.rb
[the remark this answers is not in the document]
adrianclay: Did you manually update the `schema.rb`?  You should give the top of this file a read if so, and create a migration instead.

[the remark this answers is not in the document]
C-gyorfi: I've added to the previous migration file, but I just realised it won't work

PR #1013, opened 2019-08-23
(chore) display expiry time in vacancy list

dushan-madetech:
## Trello card URL:
https://trello.com/c/uPDUKS9Q/1017-job-seekers-can-view-expiry-time-2

## Changes in this PR:
- Display vacancy expiry time in job listings view

### Before
Screenshot 2019-08-23 at 11 50 34

### After
Screenshot 2019-08-23 at 11 50 40

### DIFF
--- a/app/views/vacancies/_vacancy.html.haml
+++ b/app/views/vacancies/_vacancy.html.haml
@@ -12,7 +12,7 @@
     - if vacancy.expiry_time.nil?
       %dd= format_date(vacancy.expires_on)
     - else 
-      %dd= format_date(vacancy.expires_on) + " at " + vacancy.expiry_time.strftime("%H:%M%P")
+      %dd= format_date(vacancy.expires_on) + " at " + vacancy.expiry_time.strftime('%-l:%M%P')
     %dt= t('jobs.starts_on')
     %dd= format_date(vacancy.starts_on)
     - if vacancy.working_patterns?
--- a/app/views/vacancies/_vacancy.html.haml
+++ b/app/views/vacancies/_vacancy.html.haml
@@ -9,7 +9,10 @@
     %dd.double
       = format_date(vacancy.publish_on)
     %dt= t('jobs.expires_on')
-    %dd= format_date(vacancy.expires_on)
+    - if vacancy.expiry_time.nil?
+      %dd= format_date(vacancy.expires_on)
+    - else 
+      %dd= format_date(vacancy.expires_on) + " at " + vacancy.expiry_time.strftime("%H:%M%P")
     %dt= t('jobs.starts_on')
     %dd= format_date(vacancy.starts_on)
     - if vacancy.working_patterns?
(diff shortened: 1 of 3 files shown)


dushan-madetech:
Accepting thoughts on this. I noticed that the `format_date()` method came from the DateHelper and I was tempted to add a `format_time()` method to either that helper or a new TimeHelper to achieve this but decided not to. Keen to hear if that might have been a better approach. Also conscious that it may look weird if we display time for some vacancies but not for others, but haven't added anything to address this.

C-gyorfi:
> Accepting thoughts on this. I noticed that the `format_date()` method came from the DateHelper and I was tempted to add a `format_time()` method to either that helper or a new TimeHelper to achieve this but decided not to. Keen to hear if that might have been a better approach. Also conscious that it may look weird if we display time for some vacancies but not for others, but haven't added anything to address this.

I would suggest not to display any expiry time if it doesn't exist,

dushan-madetech:
I'm conscious that there are no tests for this, @adrianclay I'd be keen to hear your thoughts before I proceed with merging

spec/support/vacancy_helpers.rb
```diff
@@ -134,6 +134,7 @@ def verify_vacancy_list_page_details(vacancy)
     expect(page.find('.vacancy')).to have_content(vacancy.publish_on)
     expect(page.find('.vacancy')).to have_content(vacancy.starts_on)
     expect(page.find('.vacancy')).to have_content(vacancy.working_patterns)
+    expect(page.find('.vacancy')).to have_content(vacancy.expiry_time)
```
C-gyorfi: It could be sensible to add a similar assertion to `verify_all_vacancy_details` and `verify_vacancy_show_page_details` tests

app/views/vacancies/_vacancy.html.haml
```diff
@@ -9,7 +9,10 @@
     %dd.double
       = format_date(vacancy.publish_on)
     %dt= t('jobs.expires_on')
-    %dd= format_date(vacancy.expires_on)
+    - if vacancy.expiry_time.nil?
+      %dd= format_date(vacancy.expires_on)
+    - else 
+      %dd= format_date(vacancy.expires_on) + " at " + vacancy.expiry_time.strftime("%H:%M%P")
```
C-gyorfi: To get the desired format e.g `9:10` we need to use `'%-l':'%-M%P`

dushan-madetech: Good spot, will correct that 👍

spec/support/vacancy_helpers.rb
```diff
@@ -134,6 +134,12 @@ def verify_vacancy_list_page_details(vacancy)
     expect(page.find('.vacancy')).to have_content(vacancy.publish_on)
     expect(page.find('.vacancy')).to have_content(vacancy.starts_on)
     expect(page.find('.vacancy')).to have_content(vacancy.working_patterns)
+
+    # rubocop:disable Style/GuardClause
+    unless vacancy.expiry_time.nil?
+      expect(page.find('.vacancy')).to have_content(vacancy.expiry_time.strftime('%H:%M%P'))
```
C-gyorfi: This might need to be updated according to the previous comment

dushan-madetech: fixed in b4a7425

app/views/vacancies/_vacancy.html.haml
dushan-madetech: fixed in eea1018

PR #1017, opened 2019-08-29
Expiry time changes

dushan-madetech:
## Changes in this PR:

- Update vacancy factory
- Alert mailer also sorts by expiry time


app/form_models/application_details_form.rb
```diff
@@ -10,6 +10,6 @@ def disable_editing_publish_on?
   end
 
   def completed?
-    application_link && contact_email && publish_on && expires_on
+    application_link && contact_email && publish_on && expires_on && expiry_time
```
adrianclay: Why was this change made?  I'm not sure I understand what this method does.

C-gyorfi: Yeah I think this change was unnecessary, removed in 0fdb4fc

PR #1022, opened 2019-09-11
Add accessibility statement to the site.

Renny4Real:
We are required to add an accessibility statement (based on our most recent accessibility audit) to the service's website.
We want our service users to be able to view this accessibility statement from the site footer.

### Further work
This does not include a link to download the complete report. Further work could be done around this if this becomes a necessary requirement. At the moment of writing, it is not.

## Next steps:

- [ ] Check that date of deployment to the live site matches the date that the document was published and updated _(check the last line of the accessibility statement)_  as this may need altering.

### DIFF
--- a/app/views/pages/accessibility.html.haml
+++ b/app/views/pages/accessibility.html.haml
@@ -1,0 +1,68 @@
+- content_for :page_title_prefix  do
+  =t('static_pages.accessibility.page_title')
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    %h1.govuk-heading-xl=t('static_pages.accessibility.page_title')
+
+    %h2.govuk-heading-l=t('static_pages.accessibility.title')
+    %p.govuk-body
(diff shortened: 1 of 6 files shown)


app/views/pages/accessibility.html.haml
```diff
@@ -0,0 +1,68 @@
...
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    %h1.govuk-heading-xl=t('static_pages.accessibility.page_title')
+
+    %h2.govuk-heading-l=t('static_pages.accessibility.title')
+    %p.govuk-body
+      =t('static_pages.accessibility.mission.opening_text')
+      %ul.govuk-list.govuk-list--bullet
+        -t('static_pages.accessibility.mission.list').each do |li|
+          %li=li.html_safe
```
adrianclay: I'm not sure I'm a fan of this internationalisation, it feels really painful to have this separation.  What value does using it provide I wonder 🤔 .

PR #1028, opened 2019-09-18
Move tag manager snippet into codebase

dushan-madetech:
## Trello card URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&projectKey=TEVA&modal=detail&selectedIssue=TEVA-154

## Changes in this PR:

- Move Tag Manager code into codebase and add tests


app/views/layouts/application.html.haml
```diff
@@ -29,6 +28,7 @@
     = javascript_include_tag 'application'
     %meta{content: asset_path("govuk-opengraph-image.png"), property: "og:image"}/
     = csrf_meta_tags
+    = render partial: 'shared/google_tag_manager_head'
```
adrianclay: Was moving this to the bottom intentional?  It isn't clear from this PR _why_ this has happened.

spec/features/google_tag_manager_can_anonomise_page_url_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      visit jobs_path
+
+      within '.filters-form' do
+        fill_in 'location', with: 'CH52DD'
+        select 'Within 5 miles'
+        page.find('.govuk-button[type=submit]').click
+      end
+
+      expect(evaluate_script('dataLayer')).to include(include('event' => 'parametersRemoved', 'dePIIedURL' => '/jobs'))
```
adrianclay: I think this commit should also contain some more context as to why we're putting this URL into the `dataLayer`.

spec/javascripts/google_tag_manager_url_snippet_spec.js
```diff
@@ -0,0 +1,21 @@
+describe("Google Tag Manager snippet", function () {
+
+  it("should remove params from the URL", function () {
```
adrianclay: Can we get these tests to cover more of the production code?  It's currently only testing lines 9-11.

app/views/layouts/application.html.haml
dushan-madetech: This was originally done when the code was moved across to the codebase. The decision to move it was based on this article http://labs.wrprojects.com/how-to-use-google-tag-manager-with-rails-and-turbolinks/ - but it's no longer necessary

dushan-madetech: Will revert this change

spec/features/google_tag_manager_can_anonomise_page_url_spec.rb
dushan-madetech: Will amend the commit to contain more context

Renny4Real: Why are there two includes? (New to me)

dushan-madetech: These are nested within the dataLayer which is an array as attributes of an object within the dataLayer.

The outer include says look at any object in the array.
The inner include says only care about these 2 attributes within each object.

PR #1037, opened 2019-09-26
Add Dfe Sign-in API gateway

C-gyorfi:
We want to get service users from the DSI API to synchronise user
data with a google sheet.

## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-47

### DIFF
--- a/lib/dfe_sign_in_api.rb
+++ b/lib/dfe_sign_in_api.rb
@@ -1,4 +1,7 @@
 module DFESignIn
+  class ExternalServerError < StandardError; end
+  class ForbiddenRequestError < StandardError; end
+
   class API
     def users(page: 1)
       token = generate_jwt_token
@@ -6,7 +9,11 @@ module DFESignIn
         api_url(page),
         headers: { 'Authorization' => "Bearer #{token}" }
       )
-      JSON.parse(response)
+
+      raise ExternalServerError if response.code.eql?(500)
+      raise ForbiddenRequestError if response.code.eql?(403)
+
+      JSON.parse(response) if response.code.eql?(200)
     end
 
     private
--- a/lib/dfe_sign_in_api.rb
+++ b/lib/dfe_sign_in_api.rb
@@ -1,0 +1,35 @@
+module DFESignIn
+  class API
+    BASE_URL = DFE_SIGN_IN_URL
+
+    attr_accessor :dfe_sign_in_password
+
+    def initialize
+      self.dfe_sign_in_password = DFE_SIGN_IN_PASSWORD
+    end
+
+    def users(page: 1)
+      token = generate_jwt_token
+      response = HTTParty.get(
+        api_url(page),
+        headers: { 'Authorization' => "Bearer #{token}" }
+      )
+      JSON.parse(response)
+    end
+
+    private
+
+    def api_url(page = 1, records = 25)
+      "#{DFE_SIGN_IN_URL}/users?page=#{page}&pageSize=#{records}"
+    end
+
+    def generate_jwt_token
+      payload = {
+        iss: 'schooljobs',
+        exp: (Time.now.getlocal + 60).to_i,
+        aud: 'signin.education.gov.uk'
+      }
+      JWT.encode(payload, dfe_sign_in_password, 'HS256')
+    end
+  end
+end
--- a/lib/dfe_sign_in_api.rb
+++ b/lib/dfe_sign_in_api.rb
@@ -1,13 +1,5 @@
 module DFESignIn
   class API
-    BASE_URL = DFE_SIGN_IN_URL
-
-    attr_accessor :dfe_sign_in_password
-
-    def initialize
-      self.dfe_sign_in_password = DFE_SIGN_IN_PASSWORD
-    end
-
     def users(page: 1)
       token = generate_jwt_token
       response = HTTParty.get(
@@ -19,8 +11,8 @@ module DFESignIn
 
     private
 
-    def api_url(page = 1, records = 25)
-      "#{DFE_SIGN_IN_URL}/users?page=#{page}&pageSize=#{records}"
+    def api_url(page = 1)
+      "#{DFE_SIGN_IN_URL}/users?page=#{page}&pageSize=25"
     end
 
     def generate_jwt_token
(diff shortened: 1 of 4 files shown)


spec/lib/dfe_sign_in_api_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+require 'dfe_sign_in_api'
+
+RSpec.describe DFESignIn::API do
+  describe '#users' do
+    it 'returns the users from the API ' do
+      stub_api_response_for_page(1)
+      response = described_class.new.users(page: 1)
```
Renny4Real: Since the default value of the page argument is 1, could we rewrite `response = described_class.new.users(page: 1)` to be `response = described_class.new.users`?
This way this current test describes making a successful API call while the other test below is testing making a successful API call for a **specific page**.

lib/dfe_sign_in_api.rb
```diff
@@ -0,0 +1,35 @@
+module DFESignIn
+  class API
+    BASE_URL = DFE_SIGN_IN_URL
```
adrianclay: Is this used?

lib/dfe_sign_in_api.rb
```diff
@@ -0,0 +1,35 @@
...
+      response = HTTParty.get(
+        api_url(page),
+        headers: { 'Authorization' => "Bearer #{token}" }
+      )
+      JSON.parse(response)
+    end
+
+    private
+
+    def api_url(page = 1, records = 25)
```
adrianclay: This is a private method, with the records always set to 25.  Is this param useful?

lib/dfe_sign_in_api.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    attr_accessor :dfe_sign_in_password
+
+    def initialize
+      self.dfe_sign_in_password = DFE_SIGN_IN_PASSWORD
+    end
+
+    def users(page: 1)
+      token = generate_jwt_token
+      response = HTTParty.get(
```
adrianclay: How does this gateway behave if there is a 500 error from the remote service?

Can this API return any other response code?

lib/dfe_sign_in_api.rb
```diff
@@ -0,0 +1,35 @@
+module DFESignIn
+  class API
+    BASE_URL = DFE_SIGN_IN_URL
+
+    attr_accessor :dfe_sign_in_password
```
adrianclay: What was the thinking behind making this an `attr_accessor`?

lib/dfe_sign_in_api.rb
```diff
@@ -0,0 +1,35 @@
+module DFESignIn
+  class API
```
adrianclay: Do we imagine moving the other DfESignIn methods into this class, or is this only for the `/users` endpoint?

spec/lib/dfe_sign_in_api_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        {
+          iss: 'schooljobs',
+          exp: (Time.now.getlocal + 60).to_i,
+          aud: 'signin.education.gov.uk'
+        },
+        'test-password',
+        'HS256'
+      ).and_return(jwt_token)
+
+      stub_api_response_with_token(jwt_token)
```
adrianclay: Is this required in conjunction with the `expect` below?  Can this line be removed?

lib/dfe_sign_in_api.rb
C-gyorfi: I believe it's open for extensions 🤔

Renny4Real: 🤔

spec/lib/dfe_sign_in_api_spec.rb
C-gyorfi: Refactored this in 935caaf

lib/dfe_sign_in_api.rb
C-gyorfi: Changed this to an instance variable in 935caaf

lib/dfe_sign_in_api.rb
C-gyorfi: It's not useful atm, I forgot to remove, thanks for pointing this out, resolved in  935caaf

lib/dfe_sign_in_api.rb
C-gyorfi: Nope, removed! 935caaf

lib/dfe_sign_in_api.rb
C-gyorfi: Updated in 51186ec to tolerate these errors

spec/lib/dfe_sign_in_api_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    context 'when the external response status is 500' do
+      before { stub_api_response_with_external_error(1) }
+
+      it 'raises an external server error' do
+        expect { described_class.new.users }.to raise_error(DFESignIn::ExternalServerError)
+      end
+    end
+
+    context 'when the response was 403' do
+      it 'returns a JSON with success equal to false' do
```
adrianclay: Looks like it raises an error as opposed to what you describe,

spec/lib/dfe_sign_in_api_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      it 'sets a token in the header of the request' do
+        Timecop.freeze(Time.zone.now) do
+          payload = {
+            iss: 'schooljobs',
+            exp: (Time.now.getlocal + 60).to_i,
+            aud: 'signin.education.gov.uk'
+          }
+
+          expected_token = JWT.encode(payload, DFE_SIGN_IN_PASSWORD, 'HS256')
```
adrianclay: Minor, this test method is quite large, did you think about refactoring?

Renny4Real: Thanks. Forgot about this. 👍

spec/lib/dfe_sign_in_api_spec.rb
Renny4Real: Thanks for spotting this. Changed the expectation, forgot the description.

PR #1038, opened 2019-09-26
Reset the session[:vacancy_attributes] between edits

adrianclay:
## Jira ticket URL:

https://dfedigital.atlassian.net/browse/TEVA-176

## Changes in this PR:

Previously this wasn't being cleared which caused:
  * Data from one Vacancy being shown when editing another vacancy
  * Data from a draft Vacancy not being loaded and shown to the user
    if the session was empty, resulting in the vacancy details
    having to be re-entered.

As far as I can tell this defect was introduced as part of 77badb7

In the long term it would be nicer if the session wasn't being used to
store Vacancy attributes, as it would avoid these issues.

### Before

draft_vacancy_broken_small

### After

draft_vacancy_small

### DIFF
--- a/spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
+++ b/spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
@@ -1,0 +1,61 @@
+require 'rails_helper'
+RSpec.feature 'Hiring staff can edit a draft vacancy' do
+  let(:school) { create(:school) }
+  before do
+    stub_hiring_staff_auth(urn: school.urn)
+  end
+
+  context 'with job specification completed' do
+    let!(:vacancy) do
+      pay_scales = create_list(:pay_scale, 3)
+      VacancyPresenter.new(build(:vacancy, :complete,
+                                 job_title: 'Draft vacancy',
+                                 school: school,
+                                 min_pay_scale: pay_scales.sample,
+                                 max_pay_scale: pay_scales.sample,
+                                 working_patterns: ['full_time', 'part_time']))
+    end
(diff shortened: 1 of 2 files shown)


spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      visit edit_school_job_path(id: draft_vacancy.id)
+
+      expect(page).to have_content('Step 2 of 3')
+      expect(page).to have_content(draft_vacancy.education)
+      expect(page).to have_content(draft_vacancy.qualifications)
+    end
+
+    context 'when editing a different vacancy' do
+      # We use the session to store vacancy attributes, make sure it doesn't leak between edits.
+      before do
```
C-gyorfi: This could fit on a single line `before { edit_a_published_vacancy }`

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      visit edit_school_job_path(id: draft_vacancy.id)
+
+      expect(page).to have_content('Step 2 of 3')
+      expect(page).to have_content(draft_vacancy.education)
+      expect(page).to have_content(draft_vacancy.qualifications)
+    end
+
+    context 'when editing a different vacancy' do
+      # We use the session to store vacancy attributes, make sure it doesn't leak between edits.
```
C-gyorfi: I wonder if this comment could be described in the context instead, although its quite long 🤔

PR #1040, opened 2019-09-30
Add class that requests and exports user data per page to spreadsheet

Renny4Real:
This would add a class that would:
- Communicate with the `DFE:SignIn:API` class to make a request for a single page to the DSI external API
- Add data from the response to a spreadsheet. (By communicating with the `Spreadsheet::Writer` Class)
- Do this as many times as there are pages. _(We receive a paginated response from the API)_

**Any feedback/suggestions would be appreciated**


lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,63 @@
...
+      raise response_error(response) if users_nil_or_empty?(response)
+
+      rows = response_to_rows(response)
+      @worksheet.append_rows(rows)
+
+    rescue StandardError => e
+      Rails.logger.warn("DSI API failed to respond at page #{page} with error: #{e.message}")
+    end
+  end
+
```
C-gyorfi: The methods below could be `Private`?

lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,63 @@
...
+  def total_page_number
+    num = DFESignIn::API.new.users.with_indifferent_access[:numberOfPages]
+    return 1 if num.nil?
```
C-gyorfi: Could we change this to `return 0 if num.nil?`

lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,63 @@
...
+        user['organisation']['regionCode'],
+        user['organisation']['createdAt'],
+        user['organisation']['updatedAt']
+      ]
+    end
+  end
+
+  def users_nil_or_empty?(response)
+    response.with_indifferent_access[:users].nil? ||
+      response.with_indifferent_access[:users].first.empty?
```
C-gyorfi: Do we need the `with_indifferent_access` ?

lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def response_error(response)
```
C-gyorfi: Maybe we could reanem this to `error message` ?

C-gyorfi: resolved in 4a7ceed

lib/add_dsi_users_to_spreadsheet.rb
C-gyorfi: resolved in 4a7ceed

lib/add_dsi_users_to_spreadsheet.rb
C-gyorfi: resolved in 4a7ceed

lib/add_dsi_users_to_spreadsheet.rb
C-gyorfi: resolved in 4a7ceed

lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,65 @@
...
+    (1..total_page_num).each do |page|
+      response = DFESignIn::API.new.users(page: page)
+      raise error_message_for(response) if users_nil_or_empty?(response)
+
+      rows = response_to_rows(response)
+      @worksheet.append_rows(rows)
+
+    rescue StandardError => e
+      Rails.logger.warn("DSI API failed to respond at page #{page} with error: #{e.message}")
+    end
```
dushan-madetech: You're clearing all the rows first so in the event the API errors and fails to return anything would this leave the sheet empty? If so, would it be possible to rewrite this to only clear rows if some data was successfully returned by the API?

dushan-madetech:
@C-gyorfi @Renny4Real Overall this looks good. Not going to approve this just yet - I want to know if the current behaviour clears the spreadsheet when the API fails to respond i.e. responds with an error. It looks like it does but I might have overlooked something. If it does then you need to confirm if that behaviour is acceptable or if the spreadsheet should only be cleared when data is returned by the API. Once I've heard back from you about this I'll be happy to give this an approval

dushan-madetech:
Also the acceptance criteria for the JIRA ticket seems to indicate that this should be running nightly - I'm guessing in the form of a job. Is that going to be covered/is covered in this branch or will you do a separate branch for that?

lib/add_dsi_users_to_spreadsheet.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def users_nil_or_empty?(response)
+    response['users'].nil? ||
+      response['users'].first.empty?
+  end
+
+  def error_message_for(response)
+    return response['message'] if response['message']
+
+    'DSI API failed to respond with users'
```
dushan-madetech: could this maybe work as `response['message'] ||= 'DSI API failed to respond with users'`?

C-gyorfi:
> Also the acceptance criteria for the JIRA ticket seems to indicate that this should be running nightly - I'm guessing in the form of a job. Is that going to be covered/is covered in this branch or will you do a separate branch for that?

Yeah, we are going to create a rake task for that. It's not covered in this PR, the Rake task will run this use-case.

lib/add_dsi_users_to_spreadsheet.rb
C-gyorfi: I don't think we want to assign a value to `response['message']` when its nil, so that using `||` seems better for me. We fixed this in b71d0d2

lib/add_dsi_users_to_spreadsheet.rb
C-gyorfi: We resolved this in b71d0d2

dushan-madetech:
@C-gyorfi @Renny4Real see the following: https://github.com/DFE-Digital/teacher-vacancy-service/commit/b71d0d2d88f9688d3d7a0ac888b312477ec7a98a#r35342218

C-gyorfi:
> @C-gyorfi @Renny4Real Github was being weird so see the other comment I left - I pasted a link to it. Otherwise this looks good to me 👍
> 
> One final thought - what happens in there are say 10 pages of user response data and the API errors while you're making a request to get the users on page 6? Would the spreadsheet be cleared?

We would get data for all the pages except for page 6. An error message would have been logged when page 6 failed so that we would see it. But the loop should carry on as normal to the last page.

dushan-madetech:
> > @C-gyorfi @Renny4Real Github was being weird so see the other comment I left - I pasted a link to it. Otherwise this looks good to me 👍
> > One final thought - what happens in there are say 10 pages of user response data and the API errors while you're making a request to get the users on page 6? Would the spreadsheet be cleared?
> 
> We would get data for all the pages except for page 6. An error message would have been logged when page 6 failed so that we would see it. But the loop should carry on as normal to the last page.

Is that okay with the rest of team? Would there be a something on the google sheet to indicate that some data is missing?

Renny4Real:
> > > @C-gyorfi @Renny4Real Github was being weird so see the other comment I left - I pasted a link to it. Otherwise this looks good to me 👍
> > > One final thought - what happens in there are say 10 pages of user response data and the API errors while you're making a request to get the users on page 6? Would the spreadsheet be cleared?
> > 
> > 
> > We would get data for all the pages except for page 6. An error message would have been logged when page 6 failed so that we would see it. But the loop should carry on as normal to the last page.
> 
> Is that okay with the rest of the team? Would there be a something on the google sheet to indicate that some data is missing?

We've spoken with Steven. 
He doesn't feel that there is value in knowing which pages failed but that's he's made aware that data is missing.
Ideally, we (the developers) would want to know if anything did happen so that we could troubleshoot and fix the problem accordingly asap.

Google sheets also save versions so this may come in handy. e.g Resetting the spreadsheet to the version before the job was run while the error is investigated.

PR #1042, opened 2019-10-03
Add rake task to sync dsi users to spreadsheet

C-gyorfi:
Add the add_dfe_users_to_spreadsheet job to the update spreadsheet task 
- Update date helper to format the dates into a readable format
- Reload worksheet after clearing
- Update terraform vars

## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&selectedIssue=TEVA-44


Renny4Real:
@adrianclay @dushan-madetech The PR fails because of rubocop. (We will fix this)
There is also a failing test around DateTime because of time zones. ( We are also looking at how to fix this and we can give more context if you're interested). 
Please review the code that we currently have.

lib/dfe_sign_in_api.rb
```diff
@@ -13,7 +13,7 @@ def users(page: 1)
       raise ExternalServerError if response.code.eql?(500)
       raise ForbiddenRequestError if response.code.eql?(403)
 
-      JSON.parse(response) if response.code.eql?(200)
+      JSON.parse(response&.body) if response.code.eql?(200)
```
adrianclay: Do you have a spec covering this change?

C-gyorfi: Its difficult to test with `webmock` as the tests are passing with both `response` and `response.body`

Renny4Real: I think we found this because we got an `HTTParty::Response` Error when the job ran in the local env.
Maybe we should have a test around not getting this particular error? Not sure.

PR #1053, opened 2019-10-09
Add format salary helper method to vacancy model

C-gyorfi:
## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-215

## Changes in this PR:
- Add `format_salary` helper method to vacancy model, to allow any correct salary formats, including comma and decimal point.
- Update hint text for salary field

### Before
Screenshot 2019-10-09 at 13 09 51

### After
Screenshot 2019-10-09 at 13 09 27

## Next steps:
- [ ] Content review

### DIFF
--- a/spec/models/vacancy_spec.rb
+++ b/spec/models/vacancy_spec.rb
@@ -186,6 +186,35 @@ RSpec.describe Vacancy, type: :model do
           .to eq(['must not be more than £200000'])
         expect(job.errors.messages[:maximum_salary]).to be_empty
       end
+
+      it 'can accept salary format with comma' do
+        job = build(:vacancy, minimum_salary: '20,000')
+
+        expect(job.valid?).to be true
+        expect(job.minimum_salary).to eq '20000'
+      end
+
+      it 'can tolerate salary format with decimal points at the end' do
+        job = build(:vacancy, minimum_salary: '20,000.')
+
+        expect(job.valid?).to be true
+        expect(job.minimum_salary).to eq '20000'
+      end
+
+      it 'can accept salary format with two decimal points' do
+        job = build(:vacancy, minimum_salary: '20,000.23')
+
+        expect(job.valid?).to be true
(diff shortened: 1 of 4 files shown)


dushan-madetech:
@C-gyorfi what happens if you put in 1 decimal place E.g. 25000.4? Could you also write a test for that?

C-gyorfi:
> @C-gyorfi what happens if you put in 1 decimal place E.g. 25000.4? Could you also write a test for that?

Good shout, I've updated the regex in the validator in 8ab6118

spec/models/vacancy_spec.rb
```diff
@@ -186,6 +186,41 @@
...
+
+        expect(job.valid?).to be true
+        expect(job.minimum_salary).to eq '20000'
+      end
+
+      it 'can accept salary format with one decimal point' do
+        job = build(:vacancy, minimum_salary: '25000.4')
+        expect(job.valid?).to be true
+        expect(job.minimum_salary).to eq '25000.4'
+      end
```
dushan-madetech: this seems to go against the acceptance criteria of the ticket. I think that this should be rejected based on the acceptance criteria

C-gyorfi: Oh, in that case, the behaviour was correct before this test!

spec/models/vacancy_spec.rb
```diff
@@ -186,6 +186,45 @@
...
+
+      it 'can accept salary format with comma' do
+        job = build(:vacancy, minimum_salary: '20,000')
+
+        expect(job.valid?).to be true
+        expect(job.minimum_salary).to eq '20000'
+      end
+
+      it 'rejects salary format with incorrect comma placement' do
+        incorrect_salary_format = ['20,,000', '200,00', ',200'].sample
```
Renny4Real: Yay for `.sample`!

PR #1084, opened 2019-10-24
Remove user satisfaction from performance platform

adrianclay:
## Jira ticket URL:
https://dfedigital.atlassian.net/browse/TEVA-253

## Changes in this PR:
We're going to stop asking for this data from users as part of the "Create a job" journey.

We will continue asking for user satisfaction, but via surveys outside the service.

## Next steps:
- [ ] New development configuration to be shared


Renny4Real:
Just making a note here (please comment if incorrect):
Not merging this because of terraform work being done on Staging.

PR #1099, opened 2019-10-29
Remove rating from hiring staff feedback form

cesidio:
## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&projectKey=TEVA&modal=detail&selectedIssue=TEVA-253

### DIFF
--- a/app/models/vacancy_publish_feedback.rb
+++ b/app/models/vacancy_publish_feedback.rb
@@ -6,7 +6,6 @@ class VacancyPublishFeedback < ApplicationRecord
   belongs_to :vacancy
   belongs_to :user
 
-  validates :rating, presence: true
   validates :comment, length: { maximum: 1200 }, if: :comment?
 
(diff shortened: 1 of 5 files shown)


app/models/vacancy_publish_feedback.rb
```diff
@@ -17,7 +16,7 @@ def to_row
       user&.oid,
       vacancy.id,
       vacancy.school.urn,
-      rating,
+      nil, # we used to send rating, for now we just stop sending it
```
adrianclay: Might be worth keeping this line, so that if we re-exported from blank then where ratings did exist, they would still get output.

cesidio:
> I was just looking at this on my local. Main thought is the form looks a bit odd.
> image
> 
> vs
> 
> The general feedback form:
> image

Yes I am aware of that, would like to do that work in another PR.

PR #1110, opened 2019-11-01
Error quick fixes

Renny4Real:
- This fixes the error anchor link for minimum salary
- This adds commas to the character limit error messages on the feedback forms.

Before:
Screenshot 2019-11-01 at 12 29 41

After:
Screenshot 2019-11-01 at 12 22 17

### DIFF
--- a/config/locales/activerecord.yml
+++ b/config/locales/activerecord.yml
@@ -74,7 +74,7 @@ en:
               invalid: Enter an email address in the correct format, like name@example.com
             comment:
               blank: Enter your feedback
-              too_long: Feedback must not be more than %{count} characters
+              too_long: Feedback must not be more than 1,200 characters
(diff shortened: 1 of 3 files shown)


config/locales/activerecord.yml
```diff
@@ -74,7 +74,7 @@ en:
               invalid: Enter an email address in the correct format, like name@example.com
             comment:
               blank: Enter your feedback
-              too_long: Feedback must not be more than %{count} characters
+              too_long: Feedback must not be more than 1,200 characters
```
adrianclay: Shame to lose the dynamism, I wonder if `number_with_delimiter` could have helped.

Renny4Real: Thanks, I can look at trying this later.

PR #1113, opened 2019-11-04
Edit error content on the Job Specification form

Renny4Real:
This introduces the desired changes for the errors on the job specification form. See the list of job specification changes [here](https://docs.google.com/spreadsheets/d/1-MeFAToarUINg-8Sif4K0ZTPudowE6FU_mf3Hhf_iOk/edit#gid=0&range=37:37).

Since a lot of the errors across the site are currently being rewritten to not use the field name, the current idea is to later add `format: "{message}"` to the `activerecord.yml` so that we wouldn't have to add a lot of logic to the HAML files to separate the errors from the attribute names (as suggested by @C-gyorfi ).

### UI Screenshots 

_(Please note that these still have their attribute names. These will be removed when all the error content changes are ready to be merged into develop, for now, this PR would be merged into a feature branch)_

Screenshot 2019-11-06 at 17 06 40
Screenshot 2019-11-06 at 17 09 07
Screenshot 2019-11-06 at 17 09 29
Screenshot 2019-11-06 at 17 09 56

### Next Steps
- [ ] Change the global formatting to use `format: "{message}"` (Separate PR)

**Any feedback would be appreciated**

### DIFF
--- a/app/validators/salary_validator.rb
+++ b/app/validators/salary_validator.rb
@@ -7,15 +7,16 @@ class SalaryValidator < ActiveModel::EachValidator
 
   SALARY_FORMAT = /^(\d+|\d{1,3}(,\d{3})*)(\.\d{2})?$/.freeze
   MIN_SALARY_ALLOWED = 0
-  MAX_SALARY_ALLOWED = 200000
+  MAX_SALARY_LIMIT = 200000
 
   def validate_each(record, attribute, value)
-    return error_message(record, attribute, cant_be_blank_message) if check_presence? && value.blank?
+    return error_message(record, attribute, blank_minimum_salary_message) if check_presence? && value.blank?
     return error_message(record, attribute, invalid_format_message) if value[SALARY_FORMAT].nil?
 
     salary = converted_salary(value)
     return error_message(record, attribute, must_be_higher_than_min_allowed_message) if less_than_min_allowed?(salary)
-    return error_message(record, attribute, must_be_less_than_max_salary_message) if salary > MAX_SALARY_ALLOWED
+    return error_message(record, attribute, must_be_less_than_max_salary_message(attribute)) \
+      if salary >= MAX_SALARY_LIMIT
   end
 
   private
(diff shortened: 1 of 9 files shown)


config/locales/activerecord.yml
```diff
@@ -1,5 +1,10 @@
+dictionary:
+  job_title_blank: &job_title_blank Enter a job title
+  job_description_blank: &job_description_blank Enter a job description
+  minimum_salary_blank: &minimum_salary_blank Enter a minimum salary
 en:
   errors:
+    # format: "%{message}"
```
dushan-madetech: Not sure if you meant to leave this in

app/validators/salary_validator.rb
```diff
@@ -7,15 +7,16 @@ class SalaryValidator < ActiveModel::EachValidator
 
   SALARY_FORMAT = /^(\d+|\d{1,3}(,\d{3})*)(\.\d{2})?$/.freeze
   MIN_SALARY_ALLOWED = 0
-  MAX_SALARY_ALLOWED = 200000
+  MAX_SALARY_LIMIT = 200000
```
Renny4Real: Changed this to MAX_SALARY_LIMIT, as one of the desired changes was to make 199,999 the salary limit but the error message would still display 200,000 in the error message. See [here](https://docs.google.com/spreadsheets/d/1-MeFAToarUINg-8Sif4K0ZTPudowE6FU_mf3Hhf_iOk/edit?disco=AAAAD5CLJT4).

config/locales/activerecord.yml
Renny4Real: Yep. What to show what @C-gyorfi and I were thinking of implementing. 
Didn't want any other tests (especially unrelated to this PR) to fail so left this commented out.

app/validators/salary_validator.rb
```diff
@@ -7,15 +7,16 @@ class SalaryValidator < ActiveModel::EachValidator
...
 
   def validate_each(record, attribute, value)
```
C-gyorfi: I wonder where did we call this method, I looked up but I couldn't find anything but the method. Is this overriding some rails validations? If so, is this used on any other fields than minimum salary fields?

config/locales/activerecord.yml
```diff
@@ -1,5 +1,10 @@
+dictionary:
+  job_title_blank: &job_title_blank Enter a job title
```
C-gyorfi: Although I believe it's a good practice using aliases I'm wondering if we could rewrite these in a more organised way. something like ```dictionary.errors.messages.job_title.blank: Enter a job title ```

config/locales/activerecord.yml
dushan-madetech: Got it 😄

config/locales/activerecord.yml
```diff
@@ -15,6 +30,14 @@ en:
       year_invalid: must include a 4-digit year
       email:
         invalid: is not a valid email address
+  activemodel:
+    errors:
+      models:
+        job_specification_form:
+          attributes:
+            <<: *vacancy_errors_labels
```
dushan-madetech: what does the `<<` do here? Just curious about it 😄

C-gyorfi: "its a yaml thing dushan!" - Adrian

C-gyorfi: https://stackoverflow.com/questions/6651275/what-do-the-mean-in-this-database-yml-file

dushan-madetech: Thank you!

dushan-madetech:
Seems good to me 🙂

config/locales/activerecord.yml
Renny4Real: @dushan-madetech In this case, `<<` adds all the keys under the defined `vacancy_errors_labels` to the attributes under `job_specification_form`. It reduces having to rewrite the same keys over and over again which would take up lines.

PR #1114, opened 2019-11-04
Fix error messages on copy vacancy form

C-gyorfi:
## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-268

## Changes in this PR:
- move date format validations out from the gov_uk_date_fields gem (it used it's on date validator method with hard coded strings, in the future, we would like to use more descriptive error messages)
- Switched off gov_uk_date_field_errors when using gov_uk_date_fields gem
- update localisation file with the new error messages
- update tests accordingly
- Update the translation path for blank errors in failing tests

### After
image

## Next steps:
- [ ] update error formatting in activerecords.yml using ` format: "%{message}"`

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -121,7 +121,8 @@ class Vacancy < ApplicationRecord
   delegate :name, to: :school, prefix: true, allow_nil: false
   delegate :geolocation, to: :school, prefix: true, allow_nil: true
 
-  acts_as_gov_uk_date :starts_on, :ends_on, :publish_on, :expires_on
+  acts_as_gov_uk_date :starts_on, :ends_on, :publish_on,
+                      :expires_on, error_clash_behaviour: :omit_gov_uk_date_field_error
 
   scope :applicable, (-> { applicable_by_date.or(applicable_by_time) })
   scope :applicable_by_time, (-> { where('expiry_time IS NOT NULL AND expiry_time >= ?', Time.zone.now) })
--- a/app/validators/date_format_validator.rb
+++ b/app/validators/date_format_validator.rb
@@ -1,12 +1,30 @@
 class DateFormatValidator < ActiveModel::Validator
   def validate(record)
     options[:fields].each do |field|
-      date = record.send(field)
-      next if date.blank?
-
-      match = date.strftime('%Y-%m-%d').match(/^(?<y>\d*)\-/)
-
-      record.errors.add(field, I18n.t('errors.messages.year_invalid')) if match[:y].length > 4
+      validate_date_fields(field, record)
     end
+  end
+
+  private
+
+  def validate_date_fields(field, record)
+    day = record.send(field.to_s + '_dd')
+    month = record.send(field.to_s + '_mm')
+    year = record.send(field.to_s + '_yyyy')
+    return if day.blank? || month.blank? || year.blank?
+    begin
+      Date.parse("#{day}-#{month}-#{year}")
+    rescue
+      record.errors.add(field, invalid_field_error(field))
+    end
+    record.errors.add(field, invalid_year_error(field)) if year.length > 4
+  end
+
+  def invalid_field_error(field)
+    I18n.t("activerecord.errors.models.vacancy.attributes.#{field}.invalid")
+  end
+
+  def invalid_year_error(field)
+    I18n.t("activerecord.errors.models.vacancy.attributes.#{field}.invalid_year")
   end
 end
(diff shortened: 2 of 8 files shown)


config/locales/activerecord.yml
```diff
@@ -1,5 +1,6 @@
 en:
   errors:
+    # format: "%{message}" #Add this once all other errors updated
```
dushan-madetech: @adrianclay @cesidio I don't suppose you know of a way to achieve the same effect without changing the global error format?

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
```diff
@@ -153,31 +153,31 @@
...
+        expect(page).to have_content('Enter the date the role will be listed')
       end
     end
 
     context 'when publish on date is in the past' do
       let(:new_attributes) { { publish_on: 1.day.ago } }
 
       it 'shows an error' do
-        expect(page).to have_content("Publish on can't be before today")
+        expect(page).to have_content('Date role will be listed must be in the future')
```
dushan-madetech: This should be changed to `Date role will be listed must be either today or a date in the future` as the above error is not representative of how the field behaves

app/validators/date_format_validator.rb
```diff
@@ -1,12 +1,26 @@
...
+    day = record.send(field.to_s + '_dd')
+    month = record.send(field.to_s + '_mm')
+    year = record.send(field.to_s + '_yyyy')
+    return if day.blank? || month.blank? || year.blank?
+    begin
+      Date.parse("#{day}-#{month}-#{year}")
+    rescue
+      record.errors.add(field, invalid_field_error(field))
     end
+    record.errors.add(field, I18n.t('errors.messages.year_invalid')) if year.length > 4
```
dushan-madetech: I'm wondering if we should do more to validate the year? I.e. 2050 is a valid year but realistically should we be allowing that?

app/models/vacancy.rb
```diff
@@ -121,7 +121,8 @@ class Vacancy < ApplicationRecord
   delegate :name, to: :school, prefix: true, allow_nil: false
   delegate :geolocation, to: :school, prefix: true, allow_nil: true
 
-  acts_as_gov_uk_date :starts_on, :ends_on, :publish_on, :expires_on
+  acts_as_gov_uk_date :starts_on, :ends_on, :publish_on,
+                      :expires_on, error_clash_behaviour: :omit_gov_uk_date_field_error
```
dushan-madetech: It would be helpful if your commit message explained why you added `error_clash_behaviour: :omit_gov_uk_date_field_error` here 😄

config/locales/activerecord.yml
C-gyorfi: @dushan-madetech we did some hacky stuff in the haml files before to avoid this. Since we want to update most of the error messages to a more descriptive instruction, I think changing the global formatting makes more sense

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
C-gyorfi: I would argue with this. Even if it's today it must be in the future, therefore I'm happy with this message

app/validators/date_format_validator.rb
C-gyorfi: I wasn't trying to keep the current behaviour since this ticket is about the content update, but if you think its worth to review this, I can get some input for the team and do it in a separate ticket :)

config/locales/activerecord.yml
cesidio: > @adrianclay @cesidio I don't suppose you know of a way to achieve the same effect without changing the global error format?

Having developed several governmental websites, I always did that. If you have a look at how GOV.UK design system dictates how error message should be written:

https://design-system.service.gov.uk/components/error-message

you realise that you definitely don't want the standard rails format:
```
format: "%{attribute} %{message}"
```

spec/features/hiring_staff_can_copy_a_vacancy_spec.rb
C-gyorfi: Sorry I missed that its an ongoing discussion, I updated this in 45aac81

app/validators/date_format_validator.rb
dushan-madetech: Might just be worth chucking into an icebox :)

C-gyorfi: https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&view=planning&selectedIssue=TEVA-277

PR #1119, opened 2019-11-07
Add class for writing approver data to google sheet

dushan-madetech:
We need to write the approver data to a Google sheet in order for our
analysts to be able to access it. The approver data will be written to
the same spreadsheet file, just a different worksheet. Hence the
introduction of only one new env var (the worksheet GID).


config/initializers/google_drive_vars.rb
```diff
@@ -12,3 +12,4 @@
 
 DSI_USER_SPREADSHEET_ID = ENV['DSI_USER_SPREADSHEET_ID']
 DSI_USER_WORKSHEET_GID = ENV['DSI_USER_WORKSHEET_GID']
+DSI_APPROVER_WORKSHEET_GID = ENV['DSI_APPROVER_WORKSHEET_GID']
```
C-gyorfi: This will require some terraform changes.

dushan-madetech:
> Did you put any thought into reducing the duplication between this and `add_dsi_users_to_spreadsheet.rb`? They look very similar.

Yes, as mentioned I was going to do the refactoring of this once the functionality of the ticket is in production. This was because I wanted to avoid blocking the BA work that needs this ticket to be completed

config/initializers/google_drive_vars.rb
dushan-madetech: Yep, the terraform changes will be made in the next PR

PR #1124, opened 2019-11-08
Remove invalid year error from DateValidator

C-gyorfi:
## Changes in this PR:
Going forward we don't want to use confusing error messages. Michael Nottingham suggested removing the `date must include a 4-digit year` errors and use the generic invalid year errors instead. 
- Removed the error from DateValidator class
- Removed `invalid year` errors from the translation file
- Update specs accordingly

### DIFF
--- a/spec/validators/date_format_validator_spec.rb
+++ b/spec/validators/date_format_validator_spec.rb
@@ -10,96 +10,72 @@ RSpec.describe DateFormatValidator do
       end
     end
 
-    context 'a record with a malformed starts_on year' do
-      let(:vacancy) { FactoryBot.build(:vacancy, starts_on: Date.parse('12-01-202018')) }
-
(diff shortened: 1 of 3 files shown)


spec/validators/date_format_validator_spec.rb
```diff
@@ -10,96 +10,72 @@
...
+        }
       end
-    end
-
-    context 'a record with a malformed ends_on year' do
-      let(:vacancy) { FactoryBot.build(:vacancy, ends_on: Date.parse('12-01-202018')) }
 
-      it 'shows an invalid year error' do
+      it 'shows an invalid date format error' do
+        tested_field = [:starts_on, :ends_on, :publish_on, :expires_on].sample
```
Renny4Real: Sample!! Yay! 👍

PR #1127, opened 2019-11-11
Fix meta tags for public pages

cesidio:
## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&projectKey=TEVA&modal=detail&selectedIssue=TEVA-245&quickFilter=28


app/views/shared/_meta.html.haml
```diff
@@ -1,7 +1,2 @@
 %meta{name: 'title', content: "#{content_for :page_title_prefix}".strip}/
-- if content_for :page_description
```
C-gyorfi: I like that you removed this logic from the haml files 👍

PR #1136, opened 2019-11-13
Increase page size for DSI api requests

C-gyorfi:
## Changes in this PR:
When writing approvers data we hit google sheet API rate limit, breaking the data into larger packages should solve this issue

### DIFF
--- a/lib/dfe_sign_in_api.rb
+++ b/lib/dfe_sign_in_api.rb
@@ -4,6 +4,8 @@ module DFESignIn
   class UnknownResponseError < StandardError; end
 
   class API
+    PAGE_SIZE = 1000
+
     def users(page: 1)
       perform_request('/users', page)
     end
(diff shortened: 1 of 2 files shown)


lib/dfe_sign_in_api.rb
```diff
@@ -17,7 +17,7 @@ def approvers(page: 1)
     def perform_request(endpoint, page)
       token = generate_jwt_token
       response = HTTParty.get(
-        "#{DFE_SIGN_IN_URL}#{endpoint}?page=#{page}&pageSize=25",
+        "#{DFE_SIGN_IN_URL}#{endpoint}?page=#{page}&pageSize=1000",
```
cesidio: I think that this should be a constant

C-gyorfi: Thanks, updated in e42be48

PR #1142, opened 2019-11-15
Refactor spreadsheet writer class

C-gyorfi:
## Changes in this PR:

- The spreadsheet writer class uses #num_rows method from the google drive gem to find the last row when clearing the spreadsheet. This method is expensive as it maps all the data from the spreadsheet to find the first empty row. This causes a failure in clearing and updating the sheet when there are a large amount of data. Replacing #num_rows method to #max_rows solves this problem. https://www.rubydoc.info/gems/google_drive/2.1.2/GoogleDrive%2FWorksheet:num_rows

- While the `approvers` endpoint can respond with 1000 records, `users` endpoint failed to respond (with `External Server` error) when requesting 1000 records per page.  Change `USERS_ PAGE_SIZE` back to the default 25 records


lib/dfe_sign_in_api.rb
```diff
@@ -4,22 +4,22 @@ class ForbiddenRequestError < StandardError; end
...
-    PAGE_SIZE = 1000
+    PAGE_SIZE = 25
 
     def users(page: 1)
       perform_request('/users', page)
     end
 
     def approvers(page: 1)
-      perform_request('/users/approvers', page)
+      perform_request('/users/approvers', page, 1000)
```
cesidio: This number could be a constant too, like `APPROVERS_PAGE_SIZE`

adrianclay: 🤔

>  Requesting 1000 records per page results in External Server error, change this back to the default 25 records

Either your message is wrong, or this diff is wrong.

C-gyorfi: @adrianclay This is the `approver` endpoint which works fine with 1000 request, but the `users` endpoint (line 10) doesn't. I realised the message wasn't clear enough, so I updated it! 

@cesidio I was wondering if there is any value making this a constant since we are not using it anywhere else. I changed it in a8ee632, and I hope it reads better now 😃

PR #1151, opened 2019-11-18
Delete UrlValidator Class

C-gyorfi:
- UrlValidator seems to be dead code. There is only one URL validation in the job listing journey which uses the `validates_url` [gem](https://github.com/perfectline/validates_url). The UrlValidator class is not being referenced or tested anywhere

- Add the `validates_url`gem to the gemfile, to make it an explicit dependency

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ gem 'simple_form'
 gem 'rails-html-sanitizer', '~> 1.3.0' # Must be above this version due to CVE-2018-3741
 
 gem 'gov_uk_date_fields'
-
+gem 'validate_url', '~> 1.0.6'
 gem 'xml-sitemap'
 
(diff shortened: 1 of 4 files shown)


adrianclay:
👍 on removing this.  I don't think Rails comes with a UrlValidator by default, and instead I believe we're using the [validates_url](https://github.com/perfectline/validates_url) gem (based on what I can see from the `Gemfile.lock`).

Is there a way you can check that?

Further if we are depending on this gem for our production code, can we add an explicit dependency on it in the Gemfile?  It's currently a transitive dependency.

C-gyorfi:
> I don't think Rails comes with a UrlValidator by default

You are right it's the `validates_url` gem:
Screenshot 2019-11-19 at 13 03 03

Updated the PR message!

PR #1154, opened 2019-11-19
Fix capybara warnings

cesidio:


spec/features/hiring_staff_can_sign_in_with_dfe_spec.rb
```diff
@@ -31,7 +31,7 @@
     sign_in_user
 
     expect(page).to have_content(I18n.t('static_pages.not_authorised.title'))
-    expect(page).to have_content(options['email'])
+    expect(page).to have_content(options['email']) if options['email']
```
C-gyorfi: Does this mean we only run this assertion if options['email'] `not nil`, is there a clearer way to assert this? 🤔

cesidio: what's not clear about it? it is quite a common pattern in our specs

C-gyorfi: I may be wrong, but as far as I understand this gives a false positive in case the email is nil

PR #1160, opened 2019-11-21
Re-raise errors in dsi_api_response_to_spreadsheet usecase

C-gyorfi:
Handled exceptions don't show up in Rollbar. As suggested by Adrian Clay, leaving these exceptions unhandled would allow us to know when a scheduled job failed. This commit changes the behaviour of the use-case, when there is an error on a given page, the spreadsheet writer task will terminate on the page.


C-gyorfi:
> Looks good.
> I have an idea about what this is doing.
> (Gonna read up on re-raising exceptions)

Yes, I changed the code to re-raise exception because handled exceptions are not showing up in rollbar. I could remove the whole exception handling, but I believe it's still valuable to have some detailed logs in papertrail. Hope this makes sense but happy to discuss it :)

PR #1162, opened 2019-11-25
Fix sorting bug after searching with multiple education phases

Renny4Real:
This stemmed from a review done on PR #1161 after it was merged into develop.
The fix in that PR did not work when a search included multiple education phases in its search parameters.

This commit takes into account searches with multiple education phases as well as single-phase searches.

## Next steps:

- [ ] Check sort functionality & searching by phase(s) both work in Staging.

### DIFF
--- a/app/views/vacancies/_sorting_options.html.haml
+++ b/app/views/vacancies/_sorting_options.html.haml
@@ -1,9 +1,10 @@
 .govuk-grid-row.sortable-links
   .govuk-grid-column-one-half
     = form_tag('', method: :get) do
-      - search_criteria.each do |filter, value|  
+      - search_criteria.each do |filter, value|
         - if filter == :phases
-          = hidden_field_tag "phases[]", value
+          - phases.each do | phase |
+            = hidden_field_tag "phases[]", phase
         - else
           = hidden_field_tag filter, value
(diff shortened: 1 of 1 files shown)


app/views/vacancies/_sorting_options.html.haml
```diff
@@ -1,9 +1,10 @@
 .govuk-grid-row.sortable-links
   .govuk-grid-column-one-half
     = form_tag('', method: :get) do
-      - search_criteria.each do |filter, value|  
+      - search_criteria.each do |filter, value|
         - if filter == :phases
-          = hidden_field_tag "phases[]", value
+          - phases.each do | phase |
+            = hidden_field_tag "phases[]", phase
```
adrianclay: [StackOverflow](https://stackoverflow.com/a/10343089) suggests you can use `multiple: true` config option to automagically add the `[]`.

Renny4Real: @dushan-madetech and I tried something similar to this.
Couldn't get it to work with `multiple: true`.

Renny4Real: I've also tried the examples in the link you suggested above.

Renny4Real: Would it be okay to get this merged this in since:
- It's behaving as desired
- Its other half is already in develop
?

PR #1172, opened 2019-11-27
Disable/enable radius filter selection when searching by location

Renny4Real:
When a user has performed a location category search, we want to disable the radius filter as they would receive all vacancies that fall within a given category for their location.
If the user wants to redefine their search e.g use a postcode or different location, the radius filter would be re-enabled upon clicking the location text field in JS-enabled browsers.
For JS-Disabled browsers, the user would have to click on the search button, the new search results page would have the radius filter enabled.

Jira ticket URL: https://dfedigital.atlassian.net/browse/TEVA-322?atlOrigin=eyJpIjoiMDNiY2Y5ZTc0OGVjNGQ0YThiYTc1YmU4MWIxYTMwZGUiLCJwIjoiaiJ9

### When a location Category Search is done _(Note disabled Radius Button)_:
Screenshot 2019-11-27 at 12 47 09

### When a user clicks the location field afterwards _(JS-enabled browsers)_:
Screenshot 2019-11-27 at 12 47 32

## Next steps:
- [ ] Add content change for location category results
- [ ] Check it works on the Staging environment

### DIFF
--- a/spec/features/job_seekers_can_filter_vacancies_spec.rb
+++ b/spec/features/job_seekers_can_filter_vacancies_spec.rb
@@ -14,11 +14,11 @@ RSpec.feature 'Filtering vacancies' do
       allow(LocationCategory).to receive(:include?).with('enfield').and_return(false)
       expect(Geocoder).to receive(:coordinates).with('enfield')
                                                .and_return([51.6622925, -0.1180655])
-      enfield_vacancy = create(:vacancy, :published,
+      camden_vacancy = create(:vacancy, :published,
                                school: build(:school, name: 'St James School',
                                                       town: 'Enfield',
                                                       geolocation: '(51.6580645, -0.0448643)'))
-      penzance_vacancy = create(:vacancy, :published, school: build(:school, name: 'St James School',
+      victoria_vacancy = create(:vacancy, :published, school: build(:school, name: 'St James School',
                                                                              town: 'Penzance'))
 
       Vacancy.__elasticsearch__.client.indices.flush
@@ -30,8 +30,8 @@ RSpec.feature 'Filtering vacancies' do
         page.find('.govuk-button[type=submit]').click
       end
 
-      expect(page).to have_content(enfield_vacancy.job_title)
-      expect(page).not_to have_content(penzance_vacancy.job_title)
+      expect(page).to have_content(camden_vacancy.job_title)
+      expect(page).not_to have_content(victoria_vacancy.job_title)
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/vacancies_helper.rb
```diff
@@ -72,6 +72,11 @@ def pay_scale_options
   def subject_options
     @subject_options ||= Subject.all
   end
+
+  def disable_radius?
```
cesidio: I would just check `@filters.location_category_search?` from the view, this helper method just invokes it. Also `@filters.location_category_search?` returns true or false, so no point in doing the short circuit here

Renny4Real: 👍

spec/features/job_seekers_can_filter_vacancies_spec.rb
```diff
@@ -14,11 +14,11 @@
       allow(LocationCategory).to receive(:include?).with('enfield').and_return(false)
       expect(Geocoder).to receive(:coordinates).with('enfield')
                                                .and_return([51.6622925, -0.1180655])
-      enfield_vacancy = create(:vacancy, :published,
+      camden_vacancy = create(:vacancy, :published,
```
cesidio: this change is wrong

spec/features/job_seekers_can_filter_vacancies_spec.rb
```diff
@@ -14,11 +14,11 @@
...
                                school: build(:school, name: 'St James School',
                                                       town: 'Enfield',
                                                       geolocation: '(51.6580645, -0.0448643)'))
-      penzance_vacancy = create(:vacancy, :published, school: build(:school, name: 'St James School',
+      victoria_vacancy = create(:vacancy, :published, school: build(:school, name: 'St James School',
```
cesidio: this change is wrong

spec/features/job_seekers_can_filter_vacancies_spec.rb
```diff
@@ -30,28 +30,48 @@
         page.find('.govuk-button[type=submit]').click
       end
 
-      expect(page).to have_content(enfield_vacancy.job_title)
-      expect(page).not_to have_content(penzance_vacancy.job_title)
+      expect(page).to have_content(camden_vacancy.job_title)
+      expect(page).not_to have_content(victoria_vacancy.job_title)
```
cesidio: this change is wrong

Renny4Real: Thanks for pointing these out. ☝️ 
Fixed in 38a44c4

PR #1173, opened 2019-11-27
Increase page size for users endpoint in DSI:API gateway

C-gyorfi:
With the current page size(25) it takes 4hours+ to complete the spreadsheet writer task. The endpoint has a limitation of maximum 100 records per page. Updating the page size will reduce the duration of `dsi_users_to_spreadsheet` task.

### DIFF
--- a/lib/dfe_sign_in_api.rb
+++ b/lib/dfe_sign_in_api.rb
@@ -4,7 +4,7 @@ module DFESignIn
   class UnknownResponseError < StandardError; end
 
   class API
-    USERS_PAGE_SIZE = 25
+    USERS_PAGE_SIZE = 100
     APPROVERS_PAGE_SIZE = 1000
 
     def users(page: 1)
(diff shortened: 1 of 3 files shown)


lib/dfe_sign_in_api.rb
```diff
@@ -4,7 +4,7 @@ class ForbiddenRequestError < StandardError; end
   class UnknownResponseError < StandardError; end
 
   class API
-    USERS_PAGE_SIZE = 25
+    USERS_PAGE_SIZE = 100
```
dushan-madetech: Did DSI say they weren't going to upgrade the limit beyond 100?

C-gyorfi: I'm still waiting for their reply 🤔

C-gyorfi: It's in their backlog now, but it is uncertain when this change going to happen

PR #1175, opened 2019-11-27
Add specific headings for location category search result pages

Renny4Real:
When a user searches by a location category, we want to display specific landing pages for specific location categories.
This commit edits the heading on the page to be specific as a step towards this.

### Before
Screenshot 2019-11-27 at 18 42 33

### After
Screenshot 2019-11-27 at 18 35 56

Screenshot 2019-11-27 at 18 35 41

## Next steps:

- [ ] Check on Staging

### DIFF
--- a/app/views/vacancies/index.html+phone.haml
+++ b/app/views/vacancies/index.html+phone.haml
@@ -1,4 +1,7 @@
-- content_for :page_title_prefix, t('jobs.heading')
+- if @filters.location_category_search?
+  = content_for :page_title_prefix, t('jobs.location_search_heading', location: format_location_name(@filters.location))
+- else
+  = content_for :page_title_prefix, t('jobs.heading')
 = render partial: 'head_links'
 
 %h1.govuk-heading-xl
@@ -9,7 +12,7 @@
 
   .govuk-grid-column-two-thirds
     .govuk-grid-column-full.vacancies-count{class: @vacancies.user_search? ? 'search-made' : ''}
-      %p.govuk-heading-l.mb0= @vacancies.total_count_message
+      %p.govuk-heading-l.mb0= @filters.location_category_search? ? @vacancies.total_count_message_with_location(format_location_name(@filters.location)) : @vacancies.total_count_message
       - if @vacancies.user_search?
         %span.clear-search.nobreak= link_to t('jobs.filters.clear_filters'), root_path, class: 'govuk-link'
         = render(partial: 'subscribe_link', locals: { search_criteria: @filters.only_active_to_hash }) if EmailAlertsFeature.enabled?
--- a/app/views/vacancies/index.html.haml
+++ b/app/views/vacancies/index.html.haml
@@ -1,4 +1,7 @@
-- content_for :page_title_prefix, t('jobs.heading')
+- if @filters.location_category_search?
+  = content_for :page_title_prefix, t('jobs.location_search_heading', location: format_location_name(@filters.location))
+- else
+  = content_for :page_title_prefix, t('jobs.heading')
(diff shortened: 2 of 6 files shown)


app/views/vacancies/index.html+phone.haml
```diff
@@ -9,7 +9,7 @@
 
   .govuk-grid-column-two-thirds
     .govuk-grid-column-full.vacancies-count{class: @vacancies.user_search? ? 'search-made' : ''}
-      %p.govuk-heading-l.mb0= @vacancies.total_count_message
+      %p.govuk-heading-l.mb0= @filters.location_category_search? ? @vacancies.total_count_message_with_location(@filters.location) : @vacancies.total_count_message
```
C-gyorfi: Is it possible to move this logic into `VacanciesPresenter` ?

app/views/vacancies/index.html.haml
```diff
@@ -8,7 +8,7 @@
     = render 'filters'
   .govuk-grid-column-two-thirds
     .govuk-grid-column-full.vacancies-count{class: @vacancies.user_search? ? 'search-made' : ''}
-      %p.govuk-heading-l.mt0.mb1.inline-block#vacancy-results= @vacancies.total_count_message
+      %p.govuk-heading-l.mt0.mb1.inline-block#vacancy-results= @filters.location_category_search? ? @vacancies.total_count_message_with_location(@filters.location) : @vacancies.total_count_message
```
C-gyorfi: This is identical to the change in `index.html+phone.haml`, moving this logic into `VacanciesPresenter` could reduce duplication

app/views/vacancies/index.html+phone.haml
Renny4Real: Not sure. Since `@filters` is available to the view, it was easier to pass in the `location name` and also check if a `location category search` was performed from within the view.

I did try to put all the logic in the presenter but my first attempt had me trying to make the `location name` and `location_category_seach?` values accessible by the `Vacancy Presenter` (which meant passing them in from another class as variables) and then changing everywhere that `Vacancy Presenter` was called.

**I'm open to any suggestions.**

C-gyorfi: Yep, looks like you would need to initialise the `VacancyPresenter` with the `filters` to be able to do that, not sure which route is better. I think both solutions are acceptable

config/initializers/titleize_exceptions.rb
```diff
@@ -0,0 +1,14 @@
+class String
+  def uncapitalize_words
+      words = ['and', 'the', 'of', 'upon']
```
cesidio: there is 4 space indentation here

Renny4Real: 👍

config/initializers/titleize_exceptions.rb
```diff
@@ -0,0 +1,14 @@
+class String
```
cesidio: I suggest not to monkeypatch the class string, I think that this could be just a regular rails helper

Renny4Real: 👍

Renny4Real: Fixed in 9199d80

config/initializers/titleize_exceptions.rb
Renny4Real: Fixed in 9199d80

Renny4Real:
@dushan-madetech Are you okay with these changes getting merged into `develop` and the WIP commit you're working on for `content changes only when a location category search is done` be done as a separate branch?

PR #1185, opened 2019-12-04
Add home page with partials

cesidio:
## Jira ticket URL:
https://dfedigital.atlassian.net/browse/TEVA-332

### DIFF
--- a/app/views/pages/home/_location_categories.html.haml
+++ b/app/views/pages/home/_location_categories.html.haml
@@ -1,0 +1,1 @@
+%h1.govuk-heading-xl Location Categories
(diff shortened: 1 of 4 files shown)


app/views/pages/home/_location_categories.html.haml
```diff
@@ -0,0 +1 @@
+%h1.govuk-heading-xl Location Categories
```
cesidio: Note, we haven't used any `I18n` yet, but this should be fixed in future tickets.

dushan-madetech: That's fine 👍

PR #1197, opened 2019-12-10
Run without docker tt

tatyree:
A set of small changes to make it easier to get development and test versions of the app running outside of the docker containers. 

## Changes in this PR: As per the commit message titles.


dushan-madetech:
These seem like sensible quality of life changes to me, will leave this to @cesidio @adrianclay to approve

adrianclay:
I tested it locally, and all worked well. 🎉 

I'd only question the `--fail-fast` option, will this affect the pipeline too?  I _think_ I'd rather see a full list of failures than have the feedback faster.  Though I'm not strongly opinioned about this, so happy to merge and change if it turns out to be a pain.

tatyree:
> I tested it locally, and all worked well. 🎉
> 
> I'd only question the `--fail-fast` option, will this affect the pipeline too? I _think_ I'd rather see a full list of failures than have the feedback faster. Though I'm not strongly opinioned about this, so happy to merge and change if it turns out to be a pain.

Ah, yes, of course. Probably the best way to deal with this is to put in a separate .rspec config for the CI build.  I've done it before, but I'l have to refresh my memory.  I'll merge this for now and add a ticket in a bit.

PR #1199, opened 2019-12-10
Export Vacancy Table Data into Big Query Warehouse

Renny4Real:
Jira ticket URL: https://dfedigital.atlassian.net/browse/TEVA-26?atlOrigin=eyJpIjoiMDFiMjc4MjRjNzZlNDRlZjg2ZDNkMTI4NTQ5YmI4MTEiLCJwIjoiaiJ9

### Next steps:
- [ ] Terraform deployment is required
- [ ]  Test that code works in production


lib/export_vacancy_records_to_big_query.rb
```diff
@@ -0,0 +1,118 @@
+require 'google/cloud/bigquery'
+
+class ExportVacancyRecordsToBigQuery
+  attr_reader :dataset
+
+  def initialize(bigquery: Google::Cloud::Bigquery.new)
+    @dataset = bigquery.dataset ENV.fetch('BIG_QUERY_DATASET')
+  end
+
+  # rubocop:disable Metrics/AbcSize
```
dushan-madetech: Are we really getting any value from this rubocop metric? It feels like in the codebase we're just disabling and enabling it for the sake of dismissing it

lib/export_vacancy_records_to_big_query.rb
```diff
@@ -0,0 +1,118 @@
...
+        schema.string 'publisher_user_id', mode: :nullable
+      end
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
+
+
+  private
+
+  # TODO: Check what the last updated line and append new data after that.
```
dushan-madetech: Did you mean to leave this in?

Renny4Real: Not really. Will remove this comment and add a JIRA ticket instead.
Thanks!

PR #1237, opened 2019-12-16
Lighten containers

tatyree:
Advisory at this point.  DO NOT MERGE.
I'm cutting weight/time off of the containers.  This is a first effort and not fully working yet.  I'm putting this PR up so anyone who wants to can take a look/comment.


Dockerfile
```diff
@@ -1,82 +1,63 @@
...
+      -newkey rsa:2048 \
+      -keyout /etc/ssl/selfsigned.key \
+      -out /etc/ssl/selfsigned.crt;
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
 EXPOSE 3000
-
-ENTRYPOINT ["/docker-entrypoint.sh"]
-CMD ["rails", "server"]
+CMD sh -c "wait-for db:5432 -- rails server"
```
adrianclay: I don't feel good about using `wait-for` in a production environment.  If the database died, I'd prefer the user to see the rails 500 page, plus the service should recover more quickly once the db was back.

spec/features/job_seekers_can_view_vacancies_spec.rb
```diff
@@ -242,8 +242,9 @@
 
   context 'when the user is on mobile', js: true do
     scenario 'they should see the \'Show more filters\' link' do
-      page.driver.add_header('User-Agent', USER_AGENTS['MOBILE_CHROME'])
-      visit jobs_path
+      inspect_requests(inject_headers: { 'User-Agent' => USER_AGENTS['MOBILE_CHROME'] }) do
```
adrianclay: I found some code on the [ChromeDriver](http://chromedriver.chromium.org/mobile-emulation) website which looks simpler:

```
mobile_emulation = {

    "deviceMetrics" => { "width" => 360, "height" => 640, "pixelRatio" => 3.0 },

    "userAgent" => "Mozilla/5.0 (Linux; Android 4.2.1; en-us; Nexus 5 Build/JOP40D) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.166 Mobile Safari/535.19"

}

caps = Selenium::WebDriver::Remote::Capabilities.chrome("chromeOptions" => mobile_emulation)
```

Dockerfile
tatyree: Thank you for the feedback :)

`wait-for` only applies at container startup and, in its default configuration, introduces a 15 second timeout (counted down at 1 second intervals) to allow the database container to start and accept connections.  Once the rails server is running it will error as expected if the db connection disappears.  `wait-for`'s timeout duration is configurable, but 15 seconds should be sufficient.

`docker-entrypoint.sh` attempts something similar. However, it only checks that the `DATABASE_URL` environment variable is set, not that postgres is actually accepting connections.  It would be sensible to refactor it to use `wait-for` as well although I hadn't gotten to that yet.

spec/features/job_seekers_can_view_vacancies_spec.rb
tatyree: Unfortunately, multiple different variations on that weren't working as expected.  However, when I went to copy the stackoverflow version that I'd been attempting just now (very similar to what you've pasted here), I noticed that I'd missed another version of the answer: 

https://stackoverflow.com/questions/38969593/rails-simulate-mobile-device-with-capybara

```bash 
Capybara.register_driver :iphone do |app|
  Capybara::Selenium::Driver.new(app,
    :browser => :chrome,
    :desired_capabilities => Selenium::WebDriver::Remote::Capabilities.chrome(
      'chromeOptions' => {
        'mobileEmulation' => {
          'deviceMetrics' => { 'width' => 360, 'height' => 640, 'pixelRatio' => 3.0 },
          'userAgent' => "Mozilla/5.0 (Linux; Android 4.2.1; en-us; Nexus 5 Build/JOP40D) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.166 Mobile Safari/535.19"
        }
      }
    )
  )
end
```
I might give that a try.  

However, the GOV.UK design toolkit is bootstrap based and fully responsive. We really shouldn't be delivering different views based on user agent to accomplish something that could be done in a much more straightforward way using breakpoints on a responsive page.

tatyree:
I'm closing this PR.  I'll leave the branch in place so we can refer to it if we need to when we're moving to GOV.UK PAAS.

PR #1258, opened 2019-12-23
Add rake task to sync vacancy records in big query

Renny4Real:
Jira ticket URL: https://dfedigital.atlassian.net/browse/TEVA-27?atlOrigin=eyJpIjoiOTI2MjMxNTI0MjFiNDRiZDg1MzY5MTI3ZWRkYWJjNmMiLCJwIjoiaiJ9

## Next steps:

- [ ] Staging & production terraform deployment is required

### DIFF
--- a/lib/tasks/update_database_records_in_big_query.rake
+++ b/lib/tasks/update_database_records_in_big_query.rake
@@ -1,0 +1,8 @@
+namespace :database_records do
+  desc 'Updates Big Query Datasets with latest database records'
+  namespace :in_big_query do
+    task update: :environment do
+      ExportVacancyRecordsToBigQueryJob.perform_later
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/tasks/update_database_records_in_big_query.rake
```diff
@@ -0,0 +1,8 @@
+namespace :database_records do
```
adrianclay: I think I'd prefer namespace `:big_query`, then task `:export`.  I'm not sure what the `database_records` adds.

variables.tf
```diff
@@ -204,6 +204,16 @@ variable "update_dsi_spreadsheets_task_schedule" {
   default     = "cron(0 02 * * ? *)"
 }
 
+variable "update_database_records_in_big_query_task_command" {
```
adrianclay: I'm not sure I understand what value having this indirection here and below adds. 🤔

I know the other tasks do it, but I'm not sure _why_.

PR #1270, opened 2020-01-10
Move from Sprockets to Webpacker

cesidio:


app/frontend/src/details.js
```diff
@@ -18,9 +17,12 @@ function addDynamicSummaryTextForExpandedAndCollapsedDetailsTag(detailsElement)
     }
 }
 
-
```
tatyree: Jquery is a massive, insecure, slow way to do this.  We should probably install a modern JS framework and let it handle all of this.  I'm most experienced with [Vue](https://vue.js), and [React](https://react.js) is the mode widely used, but there are several others.  Won't block the PR on this, but we should discuss and get something in the roadmap to review and decide.

app/frontend/src/details.js
```diff
@@ -18,9 +17,12 @@ function addDynamicSummaryTextForExpandedAndCollapsedDetailsTag(detailsElement)
...
 $( document ).ready(function() {
     $('details').each(function(_, details) {
         addDynamicSummaryTextForExpandedAndCollapsedDetailsTag(details);
     });
 });
+
```
tatyree: If you change to the code splitting imports as shown in my comment on application.js, you shouldn't need these as the `const`s are declared there.  It'll mean explicitly exporting the functions, though.

app/views/layouts/application.html.haml
```diff
@@ -109,5 +108,4 @@
                 %a.govuk-footer__link{href: "https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/", rel: "license"} Open Government Licence v3.0
           .govuk-footer__meta-item
             %a.govuk-footer__link.govuk-footer__copyright-logo{href: "https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/"} © Crown copyright
-  :javascript
-    window.GOVUKFrontend.initAll()
+    = javascript_pack_tag 'application'
```
tatyree: I don't know how this will behave at build time given that it's an integration between webpack and rails. If you refactor `application.js` to use code splitting and this was a modern Single Page App, I'd expect webpack to inject a collection of separate `<script src` and `<link preload` tags, one for each `import(...` declaration. Webpack would use the comments to add the name to the file, and would also automatically include a cachebusting hash.

config/environments/staging.rb
```diff
@@ -25,11 +25,11 @@
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  # config.assets.js_compressor = :uglifier
```
tatyree: Can we just remove these instead of commenting them out.

config/webpack/test.js
```diff
@@ -0,0 +1,5 @@
+process.env.NODE_ENV = process.env.NODE_ENV || 'development'
```
tatyree: Shouldn't 'development' here be 'test'?

spec/teaspoon_env.rb
```diff
@@ -12,7 +12,7 @@
 
   # Paths that will be appended to the Rails assets paths
   # Note: Relative to `config.root`.
-  config.asset_paths = ['spec/javascripts', 'spec/javascripts/stylesheets']
+  # config.asset_paths = ['spec/javascripts', 'spec/javascripts/stylesheets']
```
tatyree: Again, let's just remove comments like this. They just turn in to lint over time.

Gemfile.lock
```diff
@@ -89,9 +89,9 @@ GEM
       xpath (~> 3.2)
     cliver (0.3.2)
     coderay (1.1.2)
-    coffee-rails (4.2.2)
+    coffee-rails (5.0.0)
```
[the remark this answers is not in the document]
cesidio: coffee-rails and sprockets are still required by teaspoon

app/views/teaspoon/suite/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<!-- TODO: remove after removing Teaspoon  -->
```
cesidio: Teaspoon doesn't support webpacker so we had to override the view to call
```
javascript_pack_tag 'application'
```

PR #1271, opened 2020-01-13
Filter out web crawlers visits from tracking.

Renny4Real:
There was a large discrepancy found between the total number of page views, in the database and GA, for vacancies published in Autumn 2019.
Investigating the case based on the hypothesis that the value in the database was incorrect led to the discovery that our application’s in-built tracker did not ignore visits by web crawlers.
Given the number of both good and bad web crawlers, this could have contributed to inflating the total number of page views in the database.

This commit adds another condition for the tracker to verify before attempting to log the visit as a page view.
Any suggestion or feedback would be appreciated.

Jira ticket URL: https://dfedigital.atlassian.net/browse/TEVA-381

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -46,7 +46,7 @@ class VacanciesController < ApplicationController
 
     @vacancy = VacancyPresenter.new(vacancy)
 
-    VacancyPageView.new(vacancy).track unless authenticated? || smoke_test?
+    VacancyPageView.new(vacancy).track unless non_tracked_visit?
 
     expires_in 5.minutes, public: true
   end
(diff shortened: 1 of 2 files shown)


app/controllers/vacancies_controller.rb
```diff
@@ -46,7 +46,7 @@ def show
 
     @vacancy = VacancyPresenter.new(vacancy)
 
-    VacancyPageView.new(vacancy).track unless authenticated? || smoke_test?
+    VacancyPageView.new(vacancy).track unless authenticated? || smoke_test? || browser.bot?
```
dushan-madetech: could you extract this check to some method like `not_tracked_case?` or something like that - might look cleaner if the list of cases we shouldn't track were to grow for some reason. I'll leave the method naming to you as my one was a poor example

Renny4Real: Thanks!

dushan-madetech: For clarity - `authenticated? || smoke_test? || browser.bot` is the bit I would like to see extracted

Renny4Real:
@tatyree I agree.

PR #1272, opened 2020-01-13
Add rake script to update subscriptions to new working pattern

dushan-madetech:
Subscriptions that currently exist in the database have the old style
of working pattern based on only being able to select one, however the
subscriptions code was update to handle arrays for working patterns as
we wanted users to be be able to search for multiple working patterns.
The result is that subscriptions created before the mutliple working
patterns change would no longer receive emails.

This change adds a rake task to update all the subscriptions in the
database to make themcompatible with the working pattern changes
(specifically those relatedto job alerts).

### DIFF
--- a/lib/tasks/update_subscriptions.rake
+++ b/lib/tasks/update_subscriptions.rake
@@ -2,6 +2,8 @@ class UpdateSubscriptionsWithNewWorkingPatterns
   def self.run!
     Subscription.all.each do |s|
       search_criteria_hash = s.search_criteria_to_h
+      # Old format hash = {"subject"=>"english", "working_pattern"=>"full_time"}
+      # New format hash = {"subject"=>"english", "working_patterns"=>["full_time"]}
       search_criteria_hash.transform_keys! { |k| k.gsub(/working_pattern\b/, 'working_patterns') }
                           .transform_values! {
                             |v| search_criteria_hash.key(v) == 'working_patterns' && v.is_a?(String) ? [v] : v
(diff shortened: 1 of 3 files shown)


lib/tasks/update_subscriptions.rake
```diff
@@ -0,0 +1,14 @@
+namespace :subscription_records do
+  desc 'converts old working pattern subscriptions to array style'
+  task convert_subscriptions: :environment do
+    Subscription.all.each do |s|
+      search_criteria_hash = s.search_criteria_to_h
+      search_criteria_hash.transform_keys! { |k| k.gsub(/working_pattern\b/, 'working_patterns') }
```
tatyree: Could you please include a comment explaining the old format of the keys and giving a quick example of how one would change?  The word boundary marker makes me realise I don't know what a valid and complete example of a key might look like.

spec/lib/tasks/update_subscriptions_rake_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  let!(:nil_criteria_subscription) do
+    create(:subscription, frequency: :daily, search_criteria: nil)
+  end
+
+  before do
+    UpdateSubscriptionsWithNewWorkingPatterns.run!
+  end
+
+  it 'Converts a string working pattern to an array working pattern in search criteria' do
```
tatyree: Not a problem here, but you might want to have a look at the `Rspec` `change` pattern matcher: 

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher

It's ideal for these kinds of specs if you're writing more of them in the future :)

PR #1274, opened 2020-01-20
Index home page on root path

dushan-madetech:
When we switched over to the new homepage we introduced a bug where we
set a noindex header on the homepage - this was because the homepage was
accessible through two paths, the root path (/) and the home page path
(/pages/home), but we were setting the noindex header for both paths.
This was bad as it would negatively impact our visibility on search
engines.

This commit allows the home page to be indexed with the root path, but a
noindex header is still set on the home page path (/pages/home). Some
tests were also added to the test suite to test for this.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :check_staging_auth, except: :check
   before_action :set_headers
   before_action :detect_device_format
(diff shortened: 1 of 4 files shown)


dushan-madetech:
> 👍 and :shipit:
> 
> but wouldn't this have been much simpler if we just put the excluded duplicate of the path in `robots.txt`?

@tatyree no, because the noindex header was set regardless of path

PR #1278, opened 2020-01-20
Add ADR to reflect decision to use Big Query

Renny4Real:
After a discussion on the Big Query work we were doing in TVS, a suggestion was made to create an ADR which would document the decision to replace Google Sheets with Big Query.

**Any suggestions or feedback would be appreciated.**

### DIFF
--- a/documentation/adr/0002_replace_google_sheets_with_big_query.md
+++ b/documentation/adr/0002_replace_google_sheets_with_big_query.md
@@ -1,14 +1,75 @@
 # Replace Google Sheets with Big Query as a reporting database
+
 Date: 20/01/2020
 
-## Status
-approved
+## Prologue (Summary)
 
-## Context
-We need to replace google sheets as a reporting database due to us approaching the spreadsheet limit.
+As a result of TVS' Production Data Spreadsheets approaching the [5 million cell limit](https://support.google.com/drive/answer/37603?hl=en). We decided to replace Google Sheets with Google Big Query as TVS' reporting database. With our current volume of data, we would be operating within the free tier of the pricing model; however we accept that we could fall outside the boundaries of this free tier in future.
 
-## Decision
-To use Google Big Query instead of Google Sheets and export our data into Big Query tables.
+### Status: **Approved**
 
-# Consequences
-For the volume of data we currently have at the time of writing, we can take advantage of Big Query's pricing model (Pay for above 1TB query usage / 10GB storage, our DB currently holds ~ 2 GB of data). However, the volume of the data stored in Big Query could increase over time and we would therefore fall within the boundaries of the pricing model.
+## Discussion (Context)
+
+In our codebase, we were using the Google Sheets API to write data to Google Sheets - effectively using Google Sheets as a reporting database.
+
+Advantage:
+
+- It was free.
+
+Disadvantages:
+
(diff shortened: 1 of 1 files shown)


documentation/adr/0002_replace_google_sheets_with_big_query.md
```diff
@@ -0,0 +1,14 @@
...
+approved
+
+## Context
+We need to replace google sheets as a reporting database due to us approaching the spreadsheet limit.
+
+## Decision
+To use Google Big Query instead of Google Sheets and export our data into Big Query tables.
+
+# Consequences
+For the volume of data we currently have at the time of writing, we can take advantage of Big Query's pricing model (Pay for above 1TB query usage / 10GB storage, our DB currently holds ~ 2 GB of data). However, the volume of the data stored in Big Query could increase over time and we would therefore fall within the boundaries of the pricing model.
```
stevenleggdfe: 'could therefore fall outside the boundaries of the free tier of the pricing model in future' rather than 'we would therefore fall...' might be clearer?

documentation/adr/0002_replace_google_sheets_with_big_query.md
```diff
@@ -0,0 +1,14 @@
+# Replace Google Sheets with Big Query as a reporting database
+Date: 20/01/2020
+
+## Status
+approved
+
+## Context
+We need to replace google sheets as a reporting database due to us approaching the spreadsheet limit.
```
stevenleggdfe: Is it worth filling in some more context here? e.g. 'We were using the Google Sheets API to write data to Google Sheets - effectively using Google Sheets as a reporting database. This had a number of disadvantages: Google Sheets have a space limit, which meant we had filled up some tables already and lost access to these datasets; we were projected to run out of space in the vacancies table in Q1 2020; accidental edits to spreadsheets were common, resulting in substantial maintenance effort; and Google Sheets formulae were more difficult to maintain and hand over than SQL queries.'

documentation/adr/0002_replace_google_sheets_with_big_query.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Decision
+To use Google Big Query instead of Google Sheets and export our data into Big Query tables.
```
stevenleggdfe: Should we also add some notes here about the options we explored that were *not* Big Query, and why we went with Big Query instead of them?

documentation/adr/0002_replace_google_sheets_with_big_query.md
```diff
@@ -0,0 +1,62 @@
...
+
+The options we looked at were:
+- Creating a read replica of the production DB
+- Data streaming from the API directly into the dashboard
+- Creating a new DB solely for the purposes of reporting
+- Using a data warehouse e.g Google Big Query, AWS RedShift
+
+The table below describes the results of investigating each option:
+
+|  Option                                                                                     |  Do nothing \(keep the Google Sheet\)            |  Read replica of the production database |  API\-to\-dashboard ‘data streaming’ |  PostgreSQL reporting database      |  Big Query data warehouse                     |  Amazon Redshift data warehouse                                               |   |   |
```
Renny4Real: I need to remove extra columns added by mistake

PR #1279, opened 2020-01-20
Google drive api av spike

tatyree:
No code in this spike per se.  

I've documented my findings about the simplest way we can interact with the Google Drive API in order to work out if a file has a virus. It probably isn't the best or most efficient way, but we can refine and refactor as we learn more.  

https://dfedigital.atlassian.net/browse/TEVA-445

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ workspace-variables/*.tfvars
 # Certificates
 config/localhost/https/*.key
 config/localhost/https/*.crt
+
+# GCP development credentails
+serviceAccount.json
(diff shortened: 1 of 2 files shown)


dushan-madetech:
This looks well documented to me 👍.  It seems weird to allow the upload of a malicious file to google drive, I wonder why it works that way rather than preventing the upload in the first place 😕

tatyree:
> Documentation as a whole looks good. I'm a little worried about this method of detecting virus becoming flaky in the future. I think increased monitoring around this would be useful.

It's a reasonable concern. It would be sensible to include logging whenever this occurs. I'll have another trawl through the google api docs to see if I'm missing a method that can just alert us.

PR #1282, opened 2020-01-22
Upload table data as CSV files into Cloud Storage

Renny4Real:
This change iterates through a list of tables and converts each one into a self-named CSV file.

The CSV files are saved into a temporary folder called `csv_exports` within the container and then uploaded into a bucket in Google Cloud Storage.

Once these files have been uploaded, they are deleted from within the container.

Jira ticket URL: https://dfedigital.atlassian.net/browse/TEVA-461


lib/export_tables_to_cloud_storage.rb
```diff
@@ -0,0 +1,66 @@
...
+  def export_csv_files
+    Dir.mkdir(Rails.root.join('tmp/csv_export')) unless File.exist?(Rails.root.join('tmp/csv_export'))
+
+    TABLES.each do |table|
+      file = Rails.root.join("tmp/csv_export/#{table.underscore}.csv")
+      records = table.constantize.all
+
+      next unless records.any?
+
+      CSV.open(file, 'w') do |csv|
```
tatyree: :+1: this is a clever way to solve this so it all stays in Ruby.  When I was first thinking about it, I had imagined it as a `.sql` that we ran using the `psql` client.  This keeps it closer to the main code and makes it all very explicit.

config/initializers/big_query.rb
```diff
@@ -7,8 +7,4 @@
     config.project_id  = 'teacher-vacancy-service'
     config.credentials = JSON.parse(BIG_QUERY_API_JSON_KEY)
   end
-
-  # rubocop:disable Lint/UselessAssignment
```
tatyree: Do we want to remove this now?   Per TEVA-460: "Note: We do NOT want to delete the previous Big Query work until all the new work for Big Query is completed"

config/initializers/google_cloud_storage.rb
```diff
@@ -0,0 +1,10 @@
+require 'google/cloud/storage'
+
+CLOUD_STORAGE_API_JSON_KEY = ENV['CLOUD_STORAGE_API_JSON_KEY']
+
+if CLOUD_STORAGE_API_JSON_KEY
+  Google::Cloud::Storage.configure do |config|
+    config.project_id = 'teacher-vacancy-service'
```
tatyree: I appreciate this is unlikely to change, but can we please make it an ENV as well?

spec/lib/export_tables_to_cloud_storage_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+require 'export_tables_to_cloud_storage'
+require 'fileutils'
+
+RSpec.describe ExportTablesToCloudStorage do
+  before do
+    FileUtils.rm_rf(Rails.root.join('tmp/csv_export'))
```
tatyree: Don't worry about interacting with the filesystem, just mock it out in exactly the same way you've done with the GCP buckets.

Probably something like `expect (fileutil_mock).to receive(:rm_rf).with('tmp/csv_export')`.  It'll remove the need to clear existing temp files and make the tests faster.

spec/lib/export_tables_to_cloud_storage_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    subject.run!
+  end
+
+  it 'removes the folder from the file system' do
+    expect(bucket_stub).to receive(:create_file).at_least(:once)
+
+    subject.run!
+
+    expect(File.exist?(Rails.root.join('tmp/csv_export'))).to be false
```
tatyree: See comment on mocking, above.

config/initializers/big_query.rb
Renny4Real: We can remove this later. It would fit the context better.

Note: We define the `big_query` variable **here** and in our **Big Query class**. 
So the definition above wasn't being used anywhere.

PR #1284, opened 2020-01-24
ADR outlining the argument to move away from gitflow and the steps to do so

tatyree:
This is a first draft version of the document.  If everyone is happy with it, then we start to do it and change the status.

### DIFF
--- a/documentation/adr/0003_replace_gitflow_with_simple_git_workflow.md
+++ b/documentation/adr/0003_replace_gitflow_with_simple_git_workflow.md
@@ -4,9 +4,47 @@ Date: 23/01/2020
 
 ## Prologue (Summary)
 
+We currently use a [gitflow](https://datasift.github.io/gitflow/IntroducingGitFlow.html) model for branching and
+release. It is intended to keep long-lived feature branches isolated from master at the cost of additional complexity in
+keeping the development branch and the master branch in sync. Given the simplicity of the service, our collective desire
+to cut down on irregular maintenance tasks (like merging develop into master manually to release new code), and our
+desire to increase the velocity at which we can release new features, we would benefit from moving to a [simplified git
+workflow](https://www.atlassian.com/git/articles/simple-git-workflow-is-simple).
+
 ### Status: **TBD**
 
 ## Discussion (Context)
+
+Changing our git workflow is straightforward. All we need is agreement between the devs on changing to a different
+approach. However, complexity arises owing to the setup of our [AWS
+Codebuild](https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects?region=eu-west-2) environment. In
+addition to the project, which is attached to our GitHub repo, there are five different build projects:
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects/tvs2-pull-requests/history?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects/tvs2-production-codebuild/history?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects/tvs2-testing-codebuild/history?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects/tvs2-edge-codebuild/history?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/projects/tvs2-staging-codebuild/history?region=eu-west-2
+
+and four different pipelines: 
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/tvs2-production-pipeline/view?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/tvs2-staging-pipeline/view?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/tvs2-edge-pipeline/view?region=eu-west-2
+
+https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/tvs2-testing-pipeline/view?region=eu-west-2
+
+They are each triggered by self-named branches in GitHub. At the time of writing, `edge` hasn't be triggered in 10 days
+and `testing` hasn't been triggered for three months. 
+
+We need to plan to remove all of these that correspond to branches we intend to stop using.
+
 
 ## Solution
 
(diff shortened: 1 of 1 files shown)


documentation/adr/0003_replace_gitflow_with_simple_git_workflow.md
```diff
@@ -0,0 +1,74 @@
...
+1) Brief all the team to ensure they understand their responsibilities: 
+  1) All code must have at least happy path tests (we have started explicitly putting this in acceptance criteria)
+  1) All new code that changes existing production workflows must be set up in such a way that it can be switched on and
+  off on production (feature flags).
+  1) Any epic that includes feature flags must include an end-of-epic ticket to remove the feature flagging code after
+  successful launch.
+1) Review, expand and better expose our error monitoring tools so we can keep track of our defect rate as we increase
+our release velocity.
+1) Adapt the staging pipeline to allow it to be built on-demand. For example it might build stating if sees a push with
+the tag `staging`. This will allow devs to build staging for sign off by Product and other stakeholders.
```
stevenleggdfe: I'd add here that staging will remain important to test new Tag Manager configurations for changes that are not yet enabled in production.

adrianclay: From my experience having the staging deploy as a part of the pipeline allows you to confirm the code can be deployed is useful, together with some smoke tests to confirm that deploy.
```
Build -> Test -> Staging -> Smoke test staging -> Prod
```

tatyree: > I'd add here that staging will remain important to test new Tag Manager configurations for changes that are not yet enabled in production.

I'm not suggesting we don't have a staging branch at all, I'm saying that we change the workflow so it is not a necessary step towards release. 

There are a few ways to do it; one that I've used a lot is set up a watcher that looks for commits with names that end in `-staging`. When you push one, the deploy logic rebuilds staging with your so-named branch.  Another is to have the deploy logic look for a `staging` tag on a branch and do the same thing if it is present.  Another is simply to keep a version of the staging branch and manually merge to it when you want to check/test something. 

So, with the branch-naming example, assume you are working on a new GTM branch, you'd name it `some-gtm-thing-staging`.  Then, every time you pushed it, the continuous deploy code would automatically rebuild the staging environment with your branch. To avoid people overwriting each others' staging builds-which isn't always a problem-the deploy logic can be configured to build a new, short-lived staging environment  named after each branch.  That is outside the scope of this change, however.

tatyree: > From my experience having the staging deploy as a part of the pipeline allows you to confirm the code can be deployed is useful, together with some smoke tests to confirm that deploy.
> 
> ```
> Build -> Test -> Staging -> Smoke test staging -> Prod
> ```

Aren't the main tests sufficient to ensure the code can be deployed? We have integration tests in the suite that require the app to run before they will pass don't we?  

The smoke tests can be re-targeted at production. They can also be configured to trigger an automatic rollback on failure-or a series of retries, each of which fails-although both of these tasks is outside the scope of this ADR.

stevenleggdfe: @tatyree I think there's two possible gaps here which I'm not sure we've agreed how they'd fit into this process. Perhaps if we agreed them we'd all feel more comfortable about this?

1. At what point in the process - and for what types of changes - do we do manual testing/QA? And how do we ensure that it does get done when it needs to? I find that automated testing picks up loads of possible bugs, but misses big common sense things that a human can spot in an instant.
2. How do we ensure that staging stays up to date enough to be a place where non-developers can get themselves up to speed on the hiring staff journey when they aren't allowed a DfE Sign-in account on production? Support need this to answer queries, I need this to set up tags, people designing new functions need the latest screens to refer to, comms need to be able to see the latest version of what they're telling schools about etc.

tatyree: @stevenleggdfe I think we're diverging somewhat from the intent of this particular ADR (or perhaps I need to make it clearer in the document?)- this is about simplifying the git workflow to remove the **necessity** of the staging branch. It doesn't mean we can't have a staging branch, nor does it mean we can't build other branches on 'rush boxes' (throw-away testing environments) for testing/QA/presentation purposes.  Instead, it means that we do not have to do so prior to release. 

All that said, to increase the efficiency/speed of our release process, I would suggest that manual QA as a step is added ad hoc to the acceptance criteria when it is felt it is needed.  As when we added `and product, ux and content have signed off` to the UD tickets.  The branch associated with that ticket can then be merged with staging when it's ready for review and the current staging auto-build logic will still work exactly as it does now.   It is worth noting here that not every sign off needs a full review of the running site. 

If we don't feel that this is a sufficient solution, there are other approaches that would still increase our release speed/efficiency, but these are somewhat out-of-scope as regards the original intent of the document.

adrianclay: > Aren't the main tests sufficient to ensure the code can be deployed? We have integration tests in the suite that require the app to run before they will pass don't we?

Yep, deploying to staging is providing very little confidence about the quality of the app, instead, giving confidence whether your application is deployable.

I appreicate we're getting into the weeds, and happy to discuss outside this ADR but I'd be worried about a workflow without an automatically updated staging environment.  There is an expectation from the wider team (product, marketing, digital marketing agency) that staging is a reliable non-production test environment for them.

tatyree: > Yep, deploying to staging is providing very little confidence about the quality of the app, instead, giving confidence whether your application is deployable.

Apologies, but you've lost me there.  What is the difference?  If the tests run in the same containerised environment that will constitute the release environment and have integration tests, how could the application fail to be deployable if the tests pass? 

> I appreicate we're getting into the weeds, and happy to discuss outside this ADR but I'd be worried about a workflow without an automatically updated staging environment. There is an expectation from the wider team (product, marketing, digital marketing agency) that staging is a reliable non-production test environment for them.

That's simply a case of managing expectations. If someone needs to sign something off before release - be that product, marketing or the agency (do they have any say in signoff?  If so, why?) - whoever owns that ticket builds an environment for them to test, either as an aforementioned 'rush box', or by merging their code with staging.

tatyree:
> I am happy to approve this and would suggest separating the conversation about environments from this one which specific about the usage of git.
> 
> Kill long-living branch -> good ✅
> What environments are needed and what they're for -> move to separate conversation/ADR

Thanks @ddippolito.  I'm going to merge this one then.  @adrianclay and @stevenleggdfe, let's grab the stakeholders and anyone else who wants to be involved and decide how we deal with environments.

PR #1288, opened 2020-01-27
Import table csv files to big query

dushan-madetech:
## Jira ticket URL:
https://dfedigital.atlassian.net/browse/TEVA-460

## Changes in this PR:
- Import CSV files from Google Cloud Storage to Google Big Query

### DIFF
--- a/lib/import_csv_to_big_query.rb
+++ b/lib/import_csv_to_big_query.rb
@@ -1,28 +1,25 @@
 require 'google/cloud/bigquery'
 require 'rollbar'
 require 'export_tables_to_cloud_storage'
-require 'active_support'
-include ActiveSupport::Inflector
 
 class ImportCSVToBigQuery
   def load(bigquery: Google::Cloud::Bigquery.new)
-    dataset  = bigquery.dataset ENV['BIG_QUERY_DATASET']
+    dataset = bigquery.dataset ENV['BIG_QUERY_DATASET']
 
     ExportTablesToCloudStorage::TABLES.each do |table_csv|
-      import_csv_uri  = "gs://tvs_staging/csv_export/#{underscore(table_csv)}.csv"
+      import_csv_uri = "gs://#{ENV['CLOUD_STORAGE_BUCKET']}/csv_export/#{underscore(table_csv)}.csv"
       table_id = import_csv_uri.split('/').last.sub('.csv', '')
 
       load_job = dataset.load_job table_id, import_csv_uri, skip_leading: 1, autodetect: true, write: 'truncate'
 
-      load_job.wait_until_done!  # Waits for table load to complete.
+      load_job.wait_until_done! # Waits for table load to complete.
 
       table = dataset.table(table_id)
       if table.nil?
         Rollbar.log(:error, "The #{table_id} table failed to load to Big Query")
       else
         Rollbar.log(:info, "Loaded #{table.rows_count} rows to table #{table.id}")
-      end 
+      end
     end
   end
-
 end
--- a/lib/import_csv_to_big_query.rb
+++ b/lib/import_csv_to_big_query.rb
@@ -1,0 +1,24 @@
+require "google/cloud/bigquery"
+
+class ImportCSVToBigQuery
+
+  def load
+    bigquery = Google::Cloud::Bigquery.new
+    dataset  = bigquery.dataset ENV.fetch('BIG_QUERY_DATASET')
+    csv_files = ['detailed_school_type', 'leadership', 'region', 'school', 'school_type', 'vacancy']
+    csv_files.each do |file|
+      import_csv_uri  = "gs://tvs_staging/csv_export/#{file}.csv"
+      table_id = import_csv_uri.split("/").last.sub(".csv","")
+
+      load_job = dataset.load_job table_id, import_csv_uri, skip_leading: 1, autodetect: true, write: 'truncate'
+      
+      puts "Starting job #{load_job.job_id}"
+
+      load_job.wait_until_done!  # Waits for table load to complete.
+      puts "Job finished."
+
+      table = dataset.table(table_id)
+      puts "Loaded #{table.rows_count} rows to table #{table.id}"
+    end
+  end
+end
--- a/lib/import_csv_to_big_query.rb
+++ b/lib/import_csv_to_big_query.rb
@@ -7,8 +7,8 @@ class ImportCSVToBigQuery
     dataset = bigquery.dataset ENV['BIG_QUERY_DATASET']
 
     ExportTablesToCloudStorage::TABLES.each do |table_csv|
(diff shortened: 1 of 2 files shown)


lib/import_csv_to_big_query.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'export_tables_to_cloud_storage'
+require 'active_support'
+include ActiveSupport::Inflector
+
+class ImportCSVToBigQuery
+  def load(bigquery: Google::Cloud::Bigquery.new)
+    dataset  = bigquery.dataset ENV['BIG_QUERY_DATASET']
+
+    ExportTablesToCloudStorage::TABLES.each do |table_csv|
+      import_csv_uri  = "gs://tvs_staging/csv_export/#{underscore(table_csv)}.csv"
```
cesidio: You are hardcoding the bucket name here `tvs_staging ` you should take it from the environment variable `CLOUD_STORAGE_BUCKET`

dushan-madetech: Good spot! Will get this fixed up 😄

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+require 'rails_helper'
+
+RSpec.describe ImportCSVToBigQuery do
+  let(:table) { double('table').as_null_object }
+  let(:loaded_job) { double('loaded job').as_null_object }
+  let(:dataset) { double('dataset').as_null_object }
+  let(:bigquery_client) { double('Bigquery Client', dataset: dataset) }
+  let(:export_tables) { class_double(ExportTablesToCloudStorage, TABLES: %w[TestTable]) }
+
+  it 'loads the dataset from ENV' do
```
david-mears-dfe: Does this mean 'dataset name' or 'dataset'?

lib/import_csv_to_big_query.rb
```diff
@@ -0,0 +1,25 @@
...
+require 'rollbar'
+require 'export_tables_to_cloud_storage'
+
+class ImportCSVToBigQuery
+  def load(bigquery: Google::Cloud::Bigquery.new)
+    dataset = bigquery.dataset ENV['BIG_QUERY_DATASET']
+
+    ExportTablesToCloudStorage::TABLES.each do |table_csv|
+      import_csv_uri = "gs://#{ENV['CLOUD_STORAGE_BUCKET']}/csv_export/#{table_csv.underscore}.csv"
+      table_id = import_csv_uri.split('/').last.sub('.csv', '')
```
cesidio: Isn't it easier to just use `table_csv.underscore` here instead of doing this? You could put that in a local variable to not to repeat it

dushan-madetech: I've updated this line to use `table_csv.underscore` but leaving it assigned to `table_id` for clarity that it represents that parameter in big query

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'import_csv_to_big_query'
+require 'rails_helper'
```
adrianclay: Is this required, looks like you've completely decoupled from Rails at this point.

spec/lib/import_csv_to_big_query_spec.rb
dushan-madetech: dataset in big query - this might look something like `staging_tvs_dataset`

dushan-madetech: The dataset in big query

dushan-madetech: Which is just a collection of tables etc. in big query

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+
+  it 'loads the bucket endpoint from ENV' do
+    expect(ENV).to receive(:[]).with('CLOUD_STORAGE_BUCKET').and_return('test').at_least(:once)
+    allow(ENV).to receive(:[]).with('BIG_QUERY_DATASET')
+    subject.load(bigquery: double('BigqueryClient').as_null_object)
+  end
+
+  it 'connects to the correct data set' do
```
adrianclay: I'm not sure this is asserting that it connects to the correct dataset, I'd imagine this and `loads the dataset from ENV` be combined?

Similar to:
` expect(bigquery_client).to receieve(:dataset).with('test_dataset') `

At the moment, the test to me is more `it 'connects to a data set'`.

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    expect(bigquery_client).to receive(:dataset)
+    subject.load(bigquery: bigquery_client)
+  end
+
+  it 'processes at least one table' do
+    expect(ExportTablesToCloudStorage::TABLES).to receive(:each)
```
adrianclay: I'd say this creates some unnecessary tight coupling between the behaviour of the test and the implementation.  

If I wanted to assert that the production code was iterating over multiple records, I'd create a let for the result of `ExportTablesToCloudStorage::TABLES` into an array and assert that `load_table` was called twice.

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+  end
+
+  it 'returns the loaded table for reporting purposes' do
+    allow(bigquery_client).to receive(:dataset).and_return(dataset)
+    expect(dataset).to receive(:table).with('vacancy')
+    subject.load(bigquery: bigquery_client)
+  end
+
+  context 'dataset#table' do
```
adrianclay: I'm not sure I understand the context naming convention here, in RSpec I'm used to the context describing the situation in which the test is executing.

i.e.
```
context 'when some rows exported' do
  it 'reports the table rows count' {...}
end
```

spec/lib/import_csv_to_big_query_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      expect(dataset).to receive(:load_job).with(anything, anything, hash_including(skip_leading: 1))
+      subject.load(bigquery: bigquery_client)
+    end
+
+    it 'sets autodetect: true when loading the dataset' do
+      expect(dataset).to receive(:load_job).with(anything, anything, hash_including(autodetect: true))
+      subject.load(bigquery: bigquery_client)
+    end
+
+    it 'sets write: "truncate" when loading the dataset' do
```
adrianclay: I like these single asserts 👍

spec/lib/import_csv_to_big_query_spec.rb
dushan-madetech: The tests fail without it, I think it's partly because the `underscore` method is a rails method

PR #1291, opened 2020-01-30
Set up feature flag for upload documents feature

dushan-madetech:
We want to hide the upload documents feature flag from users until we're
ready for it to go live and feature flagging it is the simplest way to
do this.

## Next steps:


docker-compose.env.sample
```diff
@@ -38,3 +38,4 @@ BIG_QUERY_DATASET=
 CLOUD_STORAGE_API_JSON_KEY=
 CLOUD_STORAGE_BUCKET=
 GOOGLE_CLOUD_PLATFORM_PROJECT_ID=
+FEATURE_UPLOAD_DOCUMENTS=true
```
tatyree: While you're adding to this file, could you put these in alphabetical order, please?

dushan-madetech: Will do this as part of the upload docs work I do, thanks for the one liner command for this 😄

PR #1293, opened 2020-01-30
Reorder README set-up instructions and make more detailed.

david-mears-dfe:
## Changes in this PR:

The set-up instructions had some lack of clarity for a new joiner.

**Reordering**: The https configuration needs to be done _before_ running `bin/rebuild`, because `bin/rebuild` runs `bin/dstart`.

**Adding**: Added instruction to copy from the secrets repo.

**Refactoring**: The instruction to start the application was redundant as `bin/rebuild` runs it, so I gave it its own section.

### Before
Screenshot 2020-01-30 at 18 16 27

### After
Screenshot 2020-01-30 at 18 17 48

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,22 +9,27 @@
 
 ### Setting up the project
 
-1. Copy the docker environment variables and fill in any missing secrets from the [secrets
-repository][secret-docker-compose]:
+1. Copy the docker environment variables:
 
 ```bash
-$ cp docker-compose.env.sample docker-compose.env
+cp docker-compose.env.sample docker-compose.env
 ```
 
-2. Build the docker container and set up the database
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,29 +9,36 @@
 
 ### Setting up the project
 
-1. Copy the docker environment variables and fill in any missing secrets from the [secrets
-repository][secret-docker-compose]:
+1. Copy the docker environment variables:
 
 ```bash
-$ cp docker-compose.env.sample docker-compose.env
+cp docker-compose.env.sample docker-compose.env
```
cesidio: Together with this, I believe we need to do also something like
```
cp config/database.example.yml config/database.yml
```
which I forgot to add when I removed that file. It might still work without it as it reads the config from `DATABASE_URL` environment variable, but for it to work outside docker it's required.

PR #1286, opened 2020-01-27
Add terraform variables for Cloud Storage.

Renny4Real:
This adds definitions for TVS's `Cloud Storage` and `Google Cloud Platform Project` related variables.

### Next steps:

- [ ] Terraform deployment is required for Staging (when merged into develop) and Production (when merged into master)

- [ ] New terraform configuration to be shared.

PR #1295, opened 2020-02-03
Production Release #142

dushan-madetech:
## Changes in this PR:

- Add ADR to reflect decision to use Big Query #1278
- Export DSI users to BigQuery #1281
- Bigquery initial checkin #1280
- Create ga-jobseeker-metrics-by-month.sql #1283
- Upload table data as CSV files into Cloud Storage #1282
- Fix memory limit in exporting to BigQuery #1285
- Add terraform variables for Cloud Storage. #1286
- Google drive api av spike #1279
- Import table csv files to big query #1288
- Add Export to big query job and rake task #1289
- Export to cloud storage and import to big query task #1290
- Set up feature flag for upload documents feature #1291
- Fix spelling mistake in ecs file #1292
- Add counties (Avon and West Midlands) #1294

## Next steps:


app/jobs/export_tables_to_cloud_storage_job.rb
```diff
@@ -0,0 +1,9 @@
+require 'export_tables_to_cloud_storage'
+
+class ExportTablesToCloudStorageJob < ApplicationJob
+  queue_as :export_tables
+
+  def perform
+    ExportTablesToCloudStorage.new.run!
```
Renny4Real: This is also defined in `def perform` in the `ImportCSVToBigQueryJob`. I think this was done so that the actions run in sequence.

I think this file is no longer needed.

app/jobs/import_csv_to_big_query_job.rb
```diff
@@ -0,0 +1,10 @@
+require 'import_csv_to_big_query'
+
```
Renny4Real: I think we need to add a `require 'export_tables_to_cloud_storage'` here.

cesidio: `require 'export_tables_to_cloud_storage'` is performed in `import_csv_to_big_query`.

PR #1300, opened 2020-02-04
Add page for users to select a file for upload (UD2)

Renny4Real:
TODO: 
- [x] Cherry-pick a4cd1309 to allow the router to recognise the feature flag.
- [x] Check integration with UD1 
- [x] Prevent users from accessing page by direct URL? (redirect to step one?)
- [x] Add relevant text to the translation file.
- [x] Review design with UX designer.
- [x] Check that code & tests cover the scope of [TEVA-407](https://dfedigital.atlassian.net/browse/TEVA-407).

image


app/frontend/styles/components/upload-documents.scss
```diff
@@ -0,0 +1,19 @@
...
+.create-title .govuk-back-link {
+  margin-bottom: 0;
+}
+
+.upload-group .govuk-button {
+  margin-bottom: 0;
+}
+
+.hidden_input {
+  display: none;
```
cpjmcquillan: Class names inconsistent (e.g. hidden-input)

Renny4Real: Thanks for spotting!

Renny4Real: Fixed in 6747a148

app/views/hiring_staff/vacancies/documents/index.haml
```diff
@@ -0,0 +1,17 @@
+.create-title
+  %h1.govuk-heading-m Create a job listing for Bexley Heath Academy
+  %h5.govuk-heading-m Step 2 of 3
+  %main.govuk-main-wrapper
+    .govuk-grid-row
+      .govuk-grid-column-two-thirds
+        %h2.govuk-heading-l Supporting documents
+        %form{:action => "#", :method => "post"}
```
tatyree: Can you break just the form out into a haml partial, please?  Chris' templates are already in `develop`, and we should use those as the source of truth to avoid merge conflicts.   You could just swap this as a partial for the dummy form he's got in there already. 

NOTE: They need some DRYing out, but they're all already there.

Renny4Real: Sorry for not seeing this. 
Chris templates are not in the folder I was expecting.

The controller `HiringStaff::Vacancies::DocumentsController` follows the same namespacing for where it's located in the app's directory. (Controllers -> Hiring Staff -> Vacancies -> DocumentsController)

We assumed the template would be located in a similar way. e.g. Views -> Hiring_Staff -> Vacancies -> Documents -> Index, which is why we created it there.

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -1,5 +1,14 @@
 class HiringStaff::Vacancies::DocumentsController < HiringStaff::Vacancies::ApplicationController
-  def index
-    # WIP
+  skip_before_action :verify_authenticity_token
+  before_action :school, :redirect_unless_vacancy_session_id, only: %i[new create]
+
+  def new
```
cesidio: I would keep this as `index` more than `new`, the reason is that it is actually an index of documents with a form to upload/edit documents.

spec/features/hiring_staff_can_upload_documents_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  before do
+    allow(UploadDocumentsFeature).to receive(:enabled?).and_return(feature_enabled?)
+    stub_hiring_staff_auth(urn: school.urn)
+
+    visit new_school_job_path
+    fill_in_job_specification_form_fields(vacancy)
+    click_on 'Save and continue'
+  end
+
+  context "without 'yes' previously selected for supporting documents" do
```
tatyree: This wording is confusing.  Maybe "if the user has selected 'no' in the previous step" or some such.

spec/features/hiring_staff_can_upload_documents_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    scenario 'redirects to step 2, supporting documents' do
+      visit documents_school_job_path
+      expect(page.current_path).to eq(supporting_documents_school_job_path)
+    end
+  end
+
+  context "with 'yes' previously selected for supporting documents" do
```
tatyree: See not on the context, above; This is similarly confusing.

spec/features/hiring_staff_can_upload_documents_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      visit supporting_documents_school_job_path
+      fill_in_supporting_documents_form_fields(vacancy)
+      click_on 'Save and continue'
+    end
+
+    scenario 'displays upload a file text' do
+      expect(page).to have_content('Upload a file')
+    end
+
+    scenario 'hiring staff can select a file for upload' do
```
tatyree: Run this spec in documentation mode and I think you'll find the overall output a bit confusing.

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -1,5 +1,14 @@
...
+    unless session[:vacancy_attributes]['supporting_documents'] == 'yes'
```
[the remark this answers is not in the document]
Renny4Real: - We've tried this and it doesn't access the value of the 'supporting_documents' key within the 'vacancy_attributes' hash.
- We had a look and found that using a method called `deep_symbolize_keys` on the sessions hash would convert all keys and nested keys within the sessions hash into symbols. But we don't think we should try to add this now.
- We also had a look in the codebase and have seen another place where this also happens:
`session[:vacancy_attributes]['id']` in the hiring_staff/vacancies/application_controller.rb:18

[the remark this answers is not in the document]
Renny4Real: Just spoke with @tatyree, the current implementation does not map secondary keys to symbols. Resolving this for now.

config/locales/en.yml
```diff
@@ -165,6 +165,9 @@ en:
     job_specification: 'Job specification'
     candidate_specification: 'Candidate specification'
     supporting_documents: 'Supporting documents'
+    upload_file: 'Upload a file'
```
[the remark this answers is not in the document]
cpjmcquillan: Confirmed with Michael

spec/features/hiring_staff_can_upload_documents_spec.rb
cesidio: I second this. All contexts should start either with `when` or `with`, in this case:
`when hiring stuff member did not select yes on supporting documents form`

app/frontend/src/uploadDocuments.js
```diff
@@ -0,0 +1,8 @@
+// This triggers a file input when the button is clicked
+
+document.addEventListener('DOMContentLoaded', function() {
+  document.getElementById('file-upload').addEventListener("click", function(e) {
+    e.preventDefault();
+    document.getElementById('upload').click();
+  });
+})
```
cesidio: Please fix the editor so that it automatically adds a new line at the end of the file so that we do not have end of line noise during reviews. GitHub also suggests it here

PR #1302, opened 2020-02-06
WIP: Upload documents to google drive (UD3~)

cpjmcquillan:
ACCEPTANCE ([TEVA-408](https://dfedigital.atlassian.net/browse/TEVA-408)):
- [X] Show users a list of documents they have uploaded 
- [X] While a file is uploading, I see a 'cancel' control 
- [X] The 'Save and continue' button is deactivated when files are uploading. It is reactivated when files have finished uploading
- [X] If the file upload is complete, show the user the file size 
- [X] Once a file has finished uploading, its status shows as 'Uploaded' 
- [x] If the file upload is not complete, show the user an upload percentage indicator. Collective indicator of all upload steps.
- [x] A happy path set of tests exist that ensures this interaction occurs
- [ ] The code has been peer reviewed and approved by at least two other developers
- [ ] A running example of this workflow has been signed off by Product, UX
- [ ] The code is released to live


app/models/document.rb
```diff
@@ -0,0 +1,3 @@
+class Document < ApplicationRecord
+  belongs_to :vacancy, optional: true
```
Renny4Real: On hindsight, I think we want the `optional: true` to be on the Vacancy model.
A vacancy could **have no documents attached** but a document **must have a vacancy it is attached to**.

spec/factories/documents.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :document do
+    name { 'Test.txt' }
```
Renny4Real: We should add the other fields here as well

spec/models/vacancy_spec.rb
```diff
@@ -499,6 +500,14 @@
     end
   end
 
+  describe 'when supporting documents are provided' do
+    it 'should return the document name' do
+      document = create(:document, name: 'Test.txt')
```
Renny4Real: We could improve this test by having two different documents attached to a vacancy and checking the names for both. Any thoughts?

cpjmcquillan: Yes definitely

Renny4Real:
This looks like the build is failing during the DB setup/migration step. @cpjmcquillan 
Error: `NoMethodError: undefined method `each' for #<Document:0x00005642dd17f400>`
We might need to rebuild the app locally and see if we can recreate this error.

cpjmcquillan:
> This looks like the build is failing during the DB setup/migration step. @cpjmcquillan
> Error: `NoMethodError: undefined method 'each' for #<Document:0x00005642dd17f400>`
> We might need to rebuild the app locally and see if we can recreate this error.

FactoryBot error. Fixed vacancies.rb to reflect has_many relationship with document.

**Before** 
`association :document`

**Now**
`
after :create do |vacancy|
  create_list :document, 3, vacancy: vacancy
end
`

Renny4Real:
Commit 54422e8
Screenshot 2020-02-13 at 21 49 39

PR #1303, opened 2020-02-06
Create schools_joined_with_metrics.sql

stevenleggdfe:
## Changes in this PR:
Joins the schools data from the database with GIAS data, and current values of signup/vacancy/MAT size metrics for efficient reference.

### DIFF
--- a/bigquery/schools_joined_with_metrics.sql
+++ b/bigquery/schools_joined_with_metrics.sql
@@ -1,0 +1,144 @@
+WITH
+  school_user_metrics AS ( #work out current user-related metric values for all individual schools for use in the main query later
+  SELECT
+    school_urn AS urn,
+    MIN(approval_datetime) AS dsi_signup_date,
+    #the earliest date that a user is recorded as authorised for TV access in DSI - for schools who signed up *after* we moved to DSI for authorisation in Nov 2019 this is the date the school first signed up
+    COUNT(*) AS number_of_users #the current number of users this school has authorised to access TV in DSI
+  FROM
+    `teacher-vacancy-service.production_dataset.users` AS users
+  GROUP BY
+    school_urn ),
+  school_vacancy_metrics AS ( #work out current vacancy-related metric values for all individual schools for use in the main query later
+  SELECT
+    school.urn AS urn,
+    COUNT(*) AS vacancies_published,
+    #the total number of vacancies this school published over all time
+    COUNTIF(CAST(publish_on AS DATE) > DATE_SUB(CURRENT_DATE(), INTERVAL 1 YEAR)) AS vacancies_published_in_the_last_year,
+    COUNTIF(CAST(publish_on AS DATE) > DATE_SUB(CURRENT_DATE(), INTERVAL 3 MONTH)) AS vacancies_published_in_the_last_quarter,
+    COUNTIF(status="published"
+      AND CAST(expiry_time AS DATE) > CURRENT_DATE()) AS vacancies_currently_live #count this as vacancies which have been published and have not yet expired
+  FROM
+    `teacher-vacancy-service.production_dataset.vacancies` AS vacancies
+  WHERE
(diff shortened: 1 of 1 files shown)


bigquery/schools_joined_with_metrics.sql
```diff
@@ -0,0 +1,144 @@
...
+    status != "trashed" #exclude deleted vacancies from the counts above
+    AND status != "draft" #exclude vacancies which have not (yet) been published from the counts above
+  GROUP BY
+    vacancies.school.urn ),
+  mat_metrics AS ( #make a table of academy trusts (MATs and SATs) with current values of trust related metrics for inclusion in main query later
+  SELECT
+    GIAS.Trusts__name_ AS trust_name,
+    COUNT(*) AS trust_size #count the total number of academies in the trust according to the latest GIAS download
+  FROM
+    `teacher-vacancy-service.production_dataset.STATIC_GIAS_Feb20` AS GIAS
```
tatyree: I'm guessing that `STATIC_GIAS_Feb20` is hardcoded for a reason. How frequently does it change? Is it possible to/worth declaring it in a variable at the top the statement?

bigquery/schools_joined_with_metrics.sql
```diff
@@ -0,0 +1,144 @@
...
+LEFT JOIN
+  `teacher-vacancy-service.production_dataset.detailed_school_type` AS detailed_school_type
+ON
+  detailed_school_type.id=school.detailed_school_type_id
+LEFT JOIN
+  school_user_metrics
+ON
+  school_user_metrics.urn=school.urn
+LEFT JOIN
+  `teacher-vacancy-service.production_dataset.STATIC_schools_historic_pre201119` AS historic_signups
```
tatyree: As per my comment, above.  I realise this one isn't likely to change, but is there anyway to flag that it is, in effect, a hard coded constant?  Again, assigning it to a variable (if possible) and moving it to the top is what I'd suggest.

bigquery/schools_joined_with_metrics.sql
```diff
@@ -0,0 +1,144 @@
...
+ON
+  historic_signups.URN=school.urn
+LEFT JOIN
+  school_vacancy_metrics
+ON
+  school_vacancy_metrics.urn=CAST(school.urn AS STRING)
+LEFT JOIN
+  `teacher-vacancy-service.production_dataset.STATIC_GIAS_Feb20` AS GIAS
```
tatyree: A quick glance at the BigQuery docs *suggests* `DECLARE` *might* work.  Feel free to disregard if you already know it won't or it adds too much work for now. 

https://cloud.google.com/bigquery/docs/reference/standard-sql/scripting#declare

bigquery/schools_joined_with_metrics.sql
tatyree: @stevenleggdfe I think I said this in a confusing way.  Apologies if so.  What I meant was: 

Is it possible to move this declaration: 

```
    `teacher-vacancy-service.production_dataset.STATIC_GIAS_Feb20` AS GIAS
```

To the top of the file, so that someone else looking at the file knows immediately that this is an external, hardcoded dependency?  

The reason I mentioned `DECLARE` is I wondered if you could do something like this: 

```
DECLARE GIAS_IMPORT_SHEET `teacher-vacancy-service.production_dataset.STATIC_GIAS_Feb20`

...

FROM 
    GIAS_IMPORT_SHEET AS GIAS
...
```

stevenleggdfe: @tatyree looks like it isn't possible to use a variable in a table name except if it's a wildcard at the end of the table name. But also wildcards aren't supported for 'external' tables like this one (which is a Google Sheet). So I don't think this approach is possible, sadly. However, I've removed the upload date from the filename so that this reference doesn't need to be maintained in future - we'll just need to remember to keep doing the manual updates to the GIAS download Google sheet until we put a better GIAS feed in place.

PR #1305, opened 2020-02-09
Removes end_of_day to show actual end time

HasnatH:
The current result on the `validThrough` attribute always returns end of day. The UI, however, uses the expires on flag. This change will ensure that the date/time returned matches the UI.

Alternative suggestion would be to add an additional variable to show the vacancy expiry time.


tatyree:
Thank you for your contribution @HasnatH.  I'm afraid that this is something that is already in our backlog and so we won't be merging this PR.  For future reference, at a minimum we require passing tests covering the changes before we can consider merging an external PR.

PR #1314, opened 2020-02-11
UD1 Redux: Change redirect and update tests, to conform with UX design

david-mears-dfe:
Previously this feature was implemented so that choosing 'No' and submit on supporting_documents led to the old version of Step 2. In fact, 'No' and submit should lead to Step 3.

There is an unrelated failing test on this commit, which was present on `develop` before I branched off. See discussion here: https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1581079193008500

## Jira ticket URL:

https://dfedigital.atlassian.net/browse/TEVA-406?atlOrigin=eyJpIjoiOGNiNDk4ZWY1YjRlNGQ3MGEyODY1MmU4ZGY1ZGMyYzIiLCJwIjoiaiJ9

### DIFF
--- a/app/controllers/hiring_staff/vacancies/supporting_documents_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/supporting_documents_controller.rb
@@ -29,6 +29,6 @@ class HiringStaff::Vacancies::SupportingDocumentsController < HiringStaff::Vacan
 
   def next_step
     @supporting_documents_form.supporting_documents == 'yes' ?
-      documents_school_job_path : candidate_specification_school_job_path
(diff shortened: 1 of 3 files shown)


cpjmcquillan:
Not sure if you've changed this yet, but can you check that there isn't a nested .wrapper class which adds more padding than necessary?

image

image

cesidio:
> Not sure if you've changed this yet, but can you check that there isn't a nested .wrapper class which adds more padding than necessary?

Good catch, we just need to remove line 8 in `supporting_documents/new.html.haml` partial
.govuk-main-wrapper
We can do it in this piece of work

PR #1316, opened 2020-02-12
Use GOVUKDesignSystemFormBuilder to upload files

cesidio:
By using the GOVUKDesignSystemFormBuilder (https://govuk-form-builder.netlify.com) we can reuse several useful helpers like the govuk_file_field and govuk_error_summary helpers. This functionality has been developed with progressive enhancements and works with javascript turned off too.

Screenshot 2020-02-13 at 12 36 03
Screenshot 2020-02-13 at 12 36 21


app/models/vacancy.rb
```diff
@@ -88,6 +88,8 @@ class Vacancy < ApplicationRecord
   extend FriendlyId
   extend ArrayEnum
 
+  attr_accessor :documents
```
cesidio: This is just temporary until we have the relation ready

app/frontend/src/uploadDocuments.js
```diff
@@ -1,8 +1,22 @@
-// This triggers a file input when the button is clicked
-
 document.addEventListener('DOMContentLoaded', function() {
-  document.getElementById('file-upload').addEventListener("click", function(e) {
+  inputFileUpload = document.getElementById('documents-form-documents-field')
```
tatyree: You don't need a ternary operator here and you should declare the variable (`const` is better as it's scoped and cannot be reassigned, but `var` is fine). 

```
const inputFileUpload = document.getElementById('documents-form-documents-field') || document.getElementById('documents-form-documents-field-error');
```

app/frontend/src/uploadDocuments.js
```diff
@@ -1,8 +1,22 @@
...
+    ? document.getElementById('documents-form-documents-field')
+    : document.getElementById('documents-form-documents-field-error')
+
+  selectFileButton = document.getElementsByClassName('govuk-button--secondary')[0];
```
tatyree: These two should also have explicit variable declarations, please. Also, both are going to raise exceptions if the element isn't in the dom as expected.

app/frontend/src/uploadDocuments.js
```diff
@@ -1,8 +1,22 @@
...
+  uploadFileButton = document.getElementsByClassName('govuk-button--secondary')[1];
+
+  selectFileButton.addEventListener('click', function(e) {
```
tatyree: Do we have on-page JS error monitoring with one of our alerting services?  If so, then this is all fine, we'll get alerts if the expected elements aren't in the dom.  If not, then we probably need to set something up (sentry.io is good) and plug it in before we release this.

app/frontend/src/uploadDocuments.js
tatyree: Er, `const` isn't supported on IE10 and below, which rules out three users in the last three months...

app/frontend/src/uploadDocuments.js
cesidio: We do have JS errors monitoring:

https://rollbar.com/dfe/teacher-vacancies

app/frontend/src/uploadDocuments.js
cesidio: `const` gets transpiled into `var` so it's fine.

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -1,14 +1,35 @@
...
-      redirect_to supporting_documents_school_job_path
-    end
+    @documents_form = DocumentsForm.new
   end
 
   def create
-    redirect_to documents_school_job_path
+    @documents_form = DocumentsForm.new(documents_form_params)
+
+    # @documents_form.errors.add(:base, 'One of the files contains a virus!')
```
tatyree: Can we kill unused code instead of comment it out, please?

app/frontend/src/uploadDocuments.js
tatyree: :+1:

PR #1318, opened 2020-02-13
Update schools table

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/browse/TEVA-28

## Changes in this PR:

- Perform migration to add columns to table schema
- Updated the update_schools_data.rb to set values of the new columns


db/migrate/20200212150214_add_key_info_to_schools.rb
```diff
@@ -0,0 +1,17 @@
+class AddKeyInfoToSchools < ActiveRecord::Migration[5.2]
+  def change
+    add_column :schools, :status, :string
```
Renny4Real: Wow. These are a lot of columns to add at once. 😄

lib/update_school_data.rb
```diff
@@ -29,6 +29,7 @@ def convert_to_school(row)
     school
   end
 
+  # rubocop:disable Metrics/AbcSize
```
Renny4Real: Not an issue but I think you can remove the need to disable this if you move all the new stuff into a separate method and call the method within set properties.

```
def set_properties(school, row)
    school.name = row['EstablishmentName']
    school.address = row['Street']
    school.locality = row['Locality'].presence
    school.address3 = row['Address3'].presence
    school.town = row['Town']
    school.county = row['County (name)'].presence
    school.postcode = row['Postcode']
    school.local_authority = row['LA (name)']
    school.minimum_age = row['StatutoryLowAge']
    school.maximum_age = row['StatutoryHighAge']
    school.easting = row['Easting']
    school.northing = row['Northing']
    school.url = valid_website(row['SchoolWebsite'])
    school.phase = row['PhaseOfEducation (code)'].to_i
    set_school_key_info(school, row)
  end

  def set_school_key_info(school, row)
    school.status = row['EstablishmentStatus (name)']
    school.trust_name = row['Trusts (name)']
    school.number_of_pupils = row['NumberOfPupils']
    school.head_title = row['HeadTitle (name)']
    school.head_first_name = row['HeadFirstName']
    school.head_last_name = row['HeadLastName']
    school.religious_character = row['ReligiousCharacter (name)']
    school.rsc_region = row['RSCRegion (name)']
    school.telephone = row['TelephoneNum']
    school.open_date = row['OpenDate']
    school.close_date = row['CloseDate']
    school.last_ofsted_inspection_date = row['OfstedLastInsp']
    school.oftsed_rating = row['OfstedRating (name)']
  end
```

spec/lib/update_school_data_spec.rb
```diff
@@ -119,6 +132,19 @@
         expect(school.northing).to eql('181201')
         expect(school.geolocation.x).to be_within(0.0000000000001).of(51.51396894535262)
         expect(school.geolocation.y).to be_within(0.0000000000001).of(-0.07751626505544208)
+        expect(school.status).to eql('Open')
```
Renny4Real: Could these expectancies be refactored into a method?

db/migrate/20200212150214_add_key_info_to_schools.rb
dushan-madetech: It's based on the columns needed for bigquery reporting and those needed for the key information tickets, wanted to avoid performing multiple migrations if possible 😄

lib/update_school_data.rb
dushan-madetech: Not all those columns are needed for the key info snippets, we decided to put it in one method based on `export_vacancy_records_to_big_query.rb` and just disable to metric in order to keep it all in one place

spec/lib/update_school_data_spec.rb
dushan-madetech: what kind of method did you have in mind?

Renny4Real: Not sure. Sorry for replying late @dushan-madetech 
These lines of expectations are in two different tests, I think. 
So I was wondering if there was a way to extract all the expectancies into a method and then call the method in the tests so that we could reduce the number of lines. 🤔

lib/update_school_data.rb
tatyree: These long lists of import attributes are difficult to read and maintain and fragile.

Can you please replace the hardcoded list of keys/attributes with an abstract method that maps keys of known format to symbols suitable for importing and have the migration follow the same schema.  Please make the attribute names follow the names used by the API (so `EstablishmentName` becomes `school.establishment_name` - you can alias a `school.name` public method to it to expose that for internal use).

PR #1323, opened 2020-02-17
Fix formatting of json import

tatyree:
Change representation of import data

Serializing the row results in a json array string being stored. What we want is a hash. ActiveRecord handles this automatically when you pass a ruby hash to a json or jsonb attribute.

Data from this hash can be accessed like `school.gias_data['URN']`.

### DIFF
--- a/lib/update_school_data.rb
+++ b/lib/update_school_data.rb
@@ -78,7 +78,10 @@ class UpdateSchoolData
   end
 
   def set_gias_data_as_json(school, row)
-    school.gias_data = row.to_json
+    scratch = {}
+    row.each { |element| scratch[element.first] = element.last }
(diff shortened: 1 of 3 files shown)


lib/update_school_data.rb
```diff
@@ -78,7 +78,10 @@ def set_simple_properties(school, row)
   end
 
   def set_gias_data_as_json(school, row)
-    school.gias_data = row.to_json
+    scratch = {}
```
david-mears-dfe: I'm not sure where the word 'scratch' came from?

david-mears-dfe:
The checks failed twice (we `retry build`ed once).

lib/update_school_data.rb
tatyree: 'scratch' as in 'temporary'.

PR #1324, opened 2020-02-17
Switch BigQuery queries from using old vacancies table to vacancy table

stevenleggdfe:
## Jira ticket URL:
https://dfedigital.atlassian.net/browse/TEVA-458

## Changes in this PR:
The production dataset in BigQuery used to contain a vacancies table which was written to via the BigQuery API. Now we export all tables, including 'vacancy', into Google Cloud Storage as CSVs and then import them into BigQuery that way. This PR moves SQL queries in BigQuery across from using the old vacancies table to using the new vacancy table in BigQuery to allow the code writing to the vacancies table to be removed in future.


tatyree:
I've just rebased this against `develop` and force-pushed the update to get the CI fixes in and the tests passing.

PR #1325, opened 2020-02-17
Fix brokent CI pipeline

tatyree:
Three days ago the postgres docker containers all got updated. It
appears that they now require a password to be set. That broke our
pipeline over the weekend. This fixes it.


dushan-madetech:
Curious to know where you looked in the AWS console to figure out the cause?

tatyree:
> 👍 Does this require any updates to the secrets?

Nope.  This is strictly local to the CI Pipeline and so can use a throwaway password.

tatyree:
> Curious to know where you looked in the AWS console to figure out the cause?

I inferred it when I went looking for the configuration.  I found that it was using `testspec.yml` and found that in the repo.  I then copied out the commands into a shell script and ran that.  It threw the same errors.  @adrianclay worked out the error was coming from the postgresql container so we took a look at their docker repo.  This revealed they had made a change on Friday.  From there, it was just a case of working out what to do. Fortunately, the logs explained what needed to be added...

PR #1326, opened 2020-02-17
Remove export vacancies to bigquery task

cesidio:
The task that export vacancies to big query is broken and we now use a different task for doing that.


lib/tasks/update_database_records_in_big_query.rake
```diff
@@ -2,7 +2,7 @@ namespace :database_records do
   desc 'Updates Big Query Datasets with latest database records'
   namespace :in_big_query do
     task update: :environment do
-      ExportVacancyRecordsToBigQueryJob.perform_later
+      # ExportVacancyRecordsToBigQueryJob.perform_later
```
tatyree: Can you please remove these rather than commenting them out?  They'll still be in the history if we need them.

spec/lib/tasks/update_database_records_in_big_query_rake_spec.rb
```diff
@@ -6,7 +6,7 @@
   end
 
   it 'queues the export vacancy records to big query job' do
-    expect { task.execute }.to have_enqueued_job(ExportVacancyRecordsToBigQueryJob)
+    # expect { task.execute }.to have_enqueued_job(ExportVacancyRecordsToBigQueryJob)
```
tatyree: Can you please remove these rather than commenting them out? They'll still be in the history if we need them.

PR #1327, opened 2020-02-17
Upload documents to google drive (UD3, UD6.1, UD6.2)

Renny4Real:
This PR is the rebased form of #1302 and will replace it.
(Pushed the rebased branch to a separate branch to check that the rebase worked without losing our original work)

**Cancel functionality has been broken out into a new ticket ([TEVA-504](https://dfedigital.atlassian.net/browse/TEVA-504))**

**Google drive issue has been documented in a separate ticket ([TEVA-505](https://dfedigital.atlassian.net/browse/TEVA-505))**

**Uploading progress indicator to be reviewed, amended and finalised ([TEVA-506](https://dfedigital.atlassian.net/browse/TEVA-506))** 

ACCEPTANCE ([TEVA-408](https://dfedigital.atlassian.net/browse/TEVA-408)):
- [X] Show users a list of documents they have uploaded 
- [X] While a file is uploading, I see a 'cancel' control 
- [X] The 'Save and continue' button is deactivated when files are uploading. It is reactivated when files have finished uploading
- [X] If the file upload is complete, show the user the file size 
- [X] Once a file has finished uploading, its status shows as 'Uploaded' 
- [x] If the file upload is not complete, show the user an upload percentage indicator. Collective indicator of all upload steps.
- [x] A happy path set of tests exist that ensures this interaction occurs
- [ ] The code has been peer-reviewed and approved by at least two other developers
- [ ] A running example of this workflow has been signed off by Product, UX
- [ ] The code is released to live


app/views/hiring_staff/vacancies/documents/new.html.haml
```diff
@@ -0,0 +1,42 @@
+%h1.govuk-heading-l
```
cesidio: This file should not be here

Renny4Real: Thanks for spotting this.

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
+         <th class='govuk-table__header'>Status</> \
+         <th class='govuk-table__header'>File size</> \
+         <th class='govuk-table__header'>Action</> \
+        </tr> \
+      </thead> \
+      <tbody id='table-body' class='govuk-table__body'> \
+      </tbody> \
+    </table> \
+  "
+  var p = document.getElementById('no-files');
```
dushan-madetech: what is this p meant to be?

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
+          "</td>  \
+          <td class='govuk-table__cell' scope='row'>" +
+            (filesList[i].size / 1024.0 / 1024.0).toFixed(2) + " MB" +
+          "</td>  \
+          <td class='govuk-table__cell' scope='row'>" +
+            "<a href='#' class='govuk-link govuk-link--no-visited-state'>Cancel</a>" +
+          "</td>  \
+        </tr> \
+      "
+      var b = document.getElementById('table-body');
```
dushan-madetech: And this b?

spec/factories/documents.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :document do
+    name { 'Test.png' }
+    size { 10000 }
```
dushan-madetech: is this defined in bytes?

cpjmcquillan: Yes

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -26,4 +30,68 @@ def redirect_if_no_supporting_documents
   def redirect_to_next_step_if_save_and_continue
     redirect_to application_details_school_job_path if params[:commit] == 'Save and continue'
   end
+
+  def add_documents(documents_attributes)
```
tatyree: Could you hoist this into the create action?  There's no real reason to have a four-line function that only gets called once.

app/frontend/src/uploadDocuments.js
tatyree: I agree with @dushan-madetech.  Let's avoid non-intention revealing variable names.  Also, let's go with a more modern, better scoped `const` or `let` variable declaration.  Finally, move variable declarations into the block they are associated with wherever possible.

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
+         <th class='govuk-table__header'>File size</> \
+         <th class='govuk-table__header'>Action</> \
+        </tr> \
+      </thead> \
+      <tbody id='table-body' class='govuk-table__body'> \
+      </tbody> \
+    </table> \
+  "
+  var p = document.getElementById('no-files');
+  var filesList = documentsInput.files;
```
tatyree: See note above on `const` or `let`.

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
+
+  if (filesList.length) {
```
tatyree: What happens if filesList is `undefined`?

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    if (p) {
+      p.insertAdjacentHTML("beforebegin", tableHTML);
+    }
+    for (var i = 0; i < filesList.length; i++) {
```
tatyree: Let's go with a more modern syntax on this, please: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

app/frontend/src/uploadDocuments.js
```diff
@@ -2,18 +2,68 @@ document.addEventListener('DOMContentLoaded', function() {
...
     inputFileUpload.form.submit();
   });
 
   inputFileUpload.classList.add('display-none');
   uploadFileButton.classList.add('display-none');
   selectFileButton.classList.remove('display-none');
 });
+
+injectDocumentsTable = function(documentsInput) {
+  var tableHTML = " \
```
tatyree: `const` or `let` please.

app/services/document_upload.rb
```diff
@@ -0,0 +1,52 @@
...
+    drive_service.create_permission(
+      uploaded.id,
+      Google::Apis::DriveV3::Permission.new(type: 'anyone', role: 'reader')
+    )
+  end
+
+  def google_drive_virus_check
+    download_path = "#{uploaded.id}"
+    begin
+      drive_service.get_file(
```
tatyree: So this is throwing an exception when there is a virus?  I didn't think it did based on my early (and admittedly light) testing.

app/controllers/hiring_staff/vacancies/documents_controller.rb
cpjmcquillan: I only did this because Rubocop checks were failing. Happy to refactor however makes most sense.

app/frontend/src/uploadDocuments.js
cpjmcquillan: Agreed, similar checks should also be carried out on the JS code already merged to avoid unnecessary Rollbar errors.

app/services/document_upload.rb
cpjmcquillan: Yes. The key parameter is acknowledge_abuse: false

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -26,4 +30,68 @@ def redirect_if_no_supporting_documents
...
+      document_upload.upload
+      unless document_upload.safe_download
+        virus_error(document_params.original_filename)
+      end
+      create_document_hash(document_params, document_upload)
+    end
+  end
+
+  def file_size_error(filename)
+    @errors = true
```
david-mears-dfe: I wonder if there's a risk, in using a separate variable to track whether there are errors, that it will become decoupled from the actual state of `@documents_form.errors` which is partly determined by ActiveModel.
```ruby
unless @errors 
```
could become (something like):
```ruby
unless @documents_form.errors
```
in order to have a single source of truth about whether there are errors. Maybe this is just a pattern I'm unaware of though.

app/frontend/styles/components/upload-documents.scss
```diff
@@ -1,7 +1,43 @@
...
+.upload-progress div:nth-child(1) {
+  animation-delay: -0.45s;
+}
+.upload-progress div:nth-child(2) {
+  animation-delay: -0.3s;
+}
+.upload-progress div:nth-child(3) {
+  animation-delay: -0.15s;
+}
+@keyframes lds-ring {
```
david-mears-dfe: What's lds-ring? (Also at line 24)

david-mears-dfe: Well, I know what a ring is. What's lds?

Renny4Real: @david-mears-dfe I think that's another way/short form of referring to the [loader/spinner.](https://github.com/loadingio/css-spinner/blob/master/dist/ring.html)
Not sure.

app/controllers/hiring_staff/vacancies/documents_controller.rb
Renny4Real: I'm not sure, but I think this is because we don't validate whether we have a virus in the form model.
We wouldn't know if there was a virus unless Google tells us so the method is invoked depending on the response from Google.

app/frontend/styles/components/upload-documents.scss
cpjmcquillan: It was just the name given to the animation (which was pulled from [here](https://loading.io/css/)). But this is going to change to a version of [MalcolmVonMoJ comment](https://github.com/alphagov/govuk-design-system-backlog/issues/28) as result of [TEVA-506](https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&projectKey=TEVA&modal=detail&selectedIssue=TEVA-506).

app/controllers/hiring_staff/vacancies/documents_controller.rb
david-mears-dfe: It looks like the error is created on line 84:
```ruby
@documents_form.errors.add(:base, t('jobs.file_virus_error_message', filename: filename))
```
Does that count as validating in the form model, for your purposes?

PR #1335, opened 2020-02-20
Key info for search snippet

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-129

## Changes in this PR:

This PR updates and re-styles the job search result snippets so that they display only the key info we decided to surface to jobseekers. Tests are updated.

The most common descriptions in GIAS data for schools without a religious character are 'None' and 'Does not apply'. If these are present, we do not display a religious character.

`FactoryBot::School` was updated in line with #1322 so that schools auto-populate themselves in the attribute `gias_data` (and so that some stray periods don't display), so you might want to re-seed your local database, if running this branch before merging, or running develop after merging.

Because `FactoryBot::School.gias_data` auto-populates the data inside a hash, the below type of school creation (any that specifies between 0 and all of the gias_data) will result in unspecified gias_data not being created. This isn't too much of a biggie, it just means they will be `nil` like the rest of the 110 or so columns I didn't mention in FactoryBot::School (all of these columns would have been `nil` anyway without these changes).

```ruby
school_one = FactoryBot.create(:school)
# school_one.gias_data['religious_character'] evaluates to e.g. "Church of England"
# school_one.gias_data['head_title'] evaluates to e.g. "Dr"
# school_one.gias_data['URN'] evalutes to nil
school = FactoryBot.create(:school, gias_data: { religious_character: "Pagan" })
# school.gias_data['religious_character'] evaluates to "Pagan"
# school_one.gias_data['head_title'] evaluates to nil
# school.gias_data['URN'] evalutes to nil
```

### Before

Screenshot 2020-02-20 at 10 51 57

### After

Screenshot 2020-02-20 at 10 51 00

## Next steps:

- [ ] Sign off by content and design


dushan-madetech:
I'd look at tidying up the commit history a little to remove the WIP commits, you can do this with `git rebase -i origin/develop' but be careful using this

app/views/vacancies/_vacancy.html.haml
```diff
@@ -1,23 +1,25 @@
-%li.vacancy
+%li.vacancy.bottom-border{ class: 'govuk-!-padding-0'}
```
dushan-madetech: Can you amend the spacing so that it's `{ class: 'govuk-!-padding-0' }` instead? I think that should be more consistent with other existing haml files in the project

app/views/vacancies/_vacancy.html.haml
```diff
@@ -1,23 +1,25 @@
...
-      = format_date(vacancy.publish_on)
-    %dt= t('jobs.expires_on')
-    - if vacancy.expiry_time.nil?
+    %dt= t('jobs.school_type')
+    - if vacancy.school.has_religious_character?
       %dd.double
-        = format_date(vacancy.expires_on)
-    - else 
+        = vacancy.school.school_type.label + ", " + vacancy.school.gias_data['religious_character']
+    - else
```
dushan-madetech: To save time refactoring this in the future I would create a presenter method that displays the school type and make it reference `vacancy.school.gias_data['TypeOfEstablishment (name)']` and `vacancy.school.gias_data['ReligiousCharacter (name)']`. Then you can remove the if display logic from here as a school will always have a type, so you only need to handle the displaying of the religious character in the presenter method

dushan-madetech: I say this because I think that the school type table will be removed sometime in the future due to the addition of the `gias_data`

spec/features/job_seekers_can_view_vacancies_spec.rb
```diff
@@ -233,6 +233,38 @@
...
+          school_type: unique_school_type,
+          gias_data: { religious_character: 'None' })
+        create(:vacancy, school: school_with_no_religion)
+        Vacancy.__elasticsearch__.client.indices.flush
+        visit jobs_path
+        expect(page).not_to have_content('Unique School Type' + linking_character)
+      end
+    end
+  end
+
```
dushan-madetech: If you create a presenter method for the school type you can then write unit tests for that method to test the same thing these tests would be testing. It might be better to do that but I'd probably check that with @cesidio @tatyree first 😄

david-mears-dfe: Good point, I was forgetting the
> _**testing pyramid**_
which I take to mean that feature tests are expensive.

config/locales/en.yml
```diff
@@ -104,8 +104,9 @@ en:
     inline_feedback_error: Please select an option.
```
tatyree: I appreciate that this is a large file with some potential sortability issues.  So, instead of trying to sort it as part of this PR, could you please put a ticket in the backlog to cover refactoring it?  Something like "increase the ease of maintenance of the `en.yml` file", or some such (Breaking it out into separate files would be a reasonable approach).

db/seeds.rb
```diff
@@ -33,28 +33,31 @@
...
-                                     urn: 137138,
-                                     address: 'Stockton Road',
-                                     phase: :secondary,
-                                     town: 'Middlesbrough',
-                                     postcode: 'TS5 4AG',
-                                     url: 'http://www.macmillan-academy.org.uk',
-                                     region: london,
-                                     geolocation: '(54.565770,-1.264489)')
+academy = FactoryBot.create(:school,
+                            address: 'Stockton Road',
```
tatyree: 👍

spec/factories/school_types.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :school_type do
-    label { Faker::Lorem.unique.sentence }
+    label { Faker::Lorem.unique.sentence.gsub('.', '') }
```
tatyree: Seems unnecessary given that using lorem for setting school type is going to give nonsense anyway.

spec/features/job_seekers_can_view_vacancies_spec.rb
tatyree: An even cheaper approach would be to write a view test to ensure the business logic in the view does what you're expecting.

spec/support/vacancy_helpers.rb
```diff
@@ -148,17 +148,15 @@ def verify_vacancy_show_page_details(vacancy)
 
   def verify_vacancy_list_page_details(vacancy)
     expect(page.find('.vacancy')).to have_content(vacancy.job_title)
```
tatyree: We should refactor these into something other than blackbox-full-of-lists-of-test tests.  These don't tell me much about the system and they are at least one level removed from the actual tests.  Also, what happens if the attribute being tested for is null?  Happy to leave it for this PR, but could you put a ticket in to look at it, please?

spec/factories/school_types.rb
david-mears-dfe: Was just annoyed by the period getting in the way of my nice '<school type>, <religious character>' formatting.

config/locales/en.yml
david-mears-dfe: Done

spec/support/vacancy_helpers.rb
david-mears-dfe: Done

app/models/school.rb
```diff
@@ -35,6 +35,11 @@ def northing=(northing)
     set_geolocation_from_easting_and_northing
   end
 
+  def has_religious_character?
+    return false if !self.respond_to?(:gias_data) || self.gias_data == nil
```
tatyree: No need to change, but this might be a bit clearer as 

`return false unless self.respond_to?(:gias_data) || self.gias_data == nil`

spec/models/school_spec.rb
```diff
@@ -4,6 +4,27 @@
   it { expect(subject.attributes).to include('gias_data') }
   it { expect(described_class.columns_hash['gias_data'].type).to eql(:json) }
 
+  describe '#has_religious_character?' do
+    it 'returns false when the school has no gias_data' do
+      subject.gias_data = nil
+      expect(subject.has_religious_character?).to be false
+    end
+
+    before do
```
tatyree: Move this at the top of the describe block - it's not positional, so it will apply to everything in the block.

spec/presenters/school_presenter_spec.rb
```diff
@@ -45,4 +45,20 @@
       end
     end
   end
+
+  describe '#school_type_with_religious_character', type: :view do
+    it 'links the school type with the religious character' do
+      school = SchoolPresenter.new(create(:school, gias_data: { religious_character: 'Roman Catholic' }))
```
tatyree: You probably don't need to persist the record for this test to work.  Prefer`build` to `create` wherever possible - it's much quicker as you don't have to round trip to the db nor clean it up afterwards. 

Don't spend a lot of time on it if `build` doesn't work - I'm expressing this as a general rule rather than as a result of having dug in to the model to see if would behave.

spec/presenters/school_presenter_spec.rb
```diff
@@ -45,4 +45,20 @@
...
+      allow(school).to receive(:has_religious_character?).and_return(true)
+      expect(school.school_type_with_religious_character).to eq("#{school.school_type.label}, Roman Catholic")
+    end
+
+    context 'when the school has no religious character' do
+      it 'returns only the school type' do
+        school = SchoolPresenter.new(create(:school))
```
tatyree: Same as above.

app/models/school.rb
david-mears-dfe: This changes the scope of the negative (in both English and Ruby).

david-mears-dfe: This changes the scope of the negative (in both Ruby and English).

david-mears-dfe:
I forgot to update the secrets repo with the new environment variable.

PR #1336, opened 2020-02-20
Show a list of uploaded documents on the summary page. (UD7)

Renny4Real:
**NB: This PR for ticket [UD7](https://dfedigital.atlassian.net/browse/TEVA-413) covers creating a section for users to see uploaded documents for a vacancy at the review step. It does not include enabling users to go back and change supporting documents from the review step. Ticket [UD8](https://dfedigital.atlassian.net/browse/TEVA-414) covers this, and this ticket will be picked up subsequently.**

**Screenshot (Full Screen):**
Screenshot 2020-02-21 at 10 36 53

**Screenshot (Supporting Documents Section):**
Screenshot 2020-02-21 at 10 37 00

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -252,6 +252,15 @@ class Vacancy < ApplicationRecord
     @skip_update_callbacks.present?
   end
 
+  def show_supporting_documents?
+    UploadDocumentsFeature.enabled? && !show_candidate_specification?
+  end
+
+  def show_candidate_specification?
+    # TODO: This method should change when the behaviour for changing candidate specification is defined
+    experience.present? && qualifications.present? && education.present?
+  end
+
   private
 
   def slug_candidates
--- a/app/views/hiring_staff/vacancies/review.html.haml
+++ b/app/views/hiring_staff/vacancies/review.html.haml
@@ -8,229 +8,28 @@
     %p.govuk-body-l
       - if @vacancy.publish_on.today?
         = t('jobs.review')
-      - else 
+      - else
         = t('jobs.review_future')
     = render 'hiring_staff/vacancies/error_messages', errors: @vacancy.errors
 
 .vacancy.govuk-grid-row
   .govuk-grid-column-two-thirds
+    = render 'hiring_staff/vacancies/review_vacancy_sections/review_job_specification'
+    - if UploadDocumentsFeature.enabled?
+      = render 'hiring_staff/vacancies/review_vacancy_sections/review_supporting_documents'
+    - else
+      = render 'hiring_staff/vacancies/review_vacancy_sections/review_candidate_specification'
 
-    %h2.govuk-heading-m.mb0
-      = t('jobs.job_specification')
+    = render 'hiring_staff/vacancies/review_vacancy_sections/review_application_details'
 
-    .govuk-body-s.mb1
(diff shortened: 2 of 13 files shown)


app/views/hiring_staff/vacancies/review_vacancy_sections/_review_supporting_documents.html.haml
```diff
@@ -0,0 +1,16 @@
+%h2.govuk-heading-m.mb0
+  = t('jobs.supporting_documents')
+
+.govuk-body-s.mb1
+-#   = link_to documents_school_job_path, class: 'govuk-link', 'aria-label': 'Change supporting documents' do
```
Renny4Real: NB: This will be added and tested as part of [UD8](https://dfedigital.atlassian.net/browse/TEVA-414) instead.

app/views/hiring_staff/vacancies/review.html.haml
```diff
@@ -8,229 +8,28 @@
...
         = t('jobs.review_future')
     = render 'hiring_staff/vacancies/error_messages', errors: @vacancy.errors
 
 .vacancy.govuk-grid-row
   .govuk-grid-column-two-thirds
+    = render 'hiring_staff/vacancies/review_vacancy_sections/review_job_specification'
+    - if UploadDocumentsFeature.enabled?
+      = render 'hiring_staff/vacancies/review_vacancy_sections/review_supporting_documents'
+    - else
+      = render 'hiring_staff/vacancies/review_vacancy_sections/review_candidate_specification'
```
dushan-madetech: When Upload Docs is enabled this wouldn't show the review candidate specification stuff and there would be existing live vacancies in the service with no documents uploaded to them

cpjmcquillan: I guess we should show both until the transition to the new journey is complete for all users and all vacancies

david-mears-dfe: @dushan-madetech, is the review page seen (a) only at the end of the process of creating a new vacancy, or (b) are there other times it's seen? If (a), new vacancies created after the feature flag is switched on will not have candidate spec', so this is fine, but if (b) then yes, we need to cater for legacy vacancies.

david-mears-dfe: Actually, even if (a), there may be some draft vacancies created before the feature flag is switched on, which have a candidate specification.

app/models/vacancy.rb
```diff
@@ -252,6 +252,14 @@ def skip_update_callbacks?
     @skip_update_callbacks.present?
   end
 
+  def show_supporting_documents?
+    UploadDocumentsFeature.enabled? && !show_candidate_specification?
+  end
+
+  def show_candidate_specification?
+    experience.present? && qualifications.present? && education.present?
```
david-mears-dfe: Is it possible for a pre-upload-docs vacancy to only have a subset of these attributes present?
Are these required fields in the candidate specification form?

david-mears-dfe: The answers are respectively 'no' and 'yes' so this is resolved.

cpjmcquillan: They are required fields. This method would change when the behaviour for changing candidate specification in an upload documents world is defined. Will add a TODO in the code!

PR #1338, opened 2020-02-21
Upload documents tweaks

cpjmcquillan:
Changes required following UX review.
- Remove cancel link from documents table
- Change content of error message
- Refactor errors to handle generic Google::Apis::Error when uploading

### DIFF
--- a/app/frontend/src/uploadDocuments.js
+++ b/app/frontend/src/uploadDocuments.js
@@ -38,11 +38,9 @@ injectDocumentsTable = function(documentsInput) {
             filesList[i].name +
           "</td>  \
           <td class='govuk-table__cell'>" +
-            "Uploading" + "<span class='upload-progress'><div class='upload-progress-spinner'></span></div>" +
+            "Uploading" + "<span class='upload-progress'><div class='upload-progress-spinner'></div></span>" +
           "</td>  \
-          <td class='govuk-table__cell' scope='row'>" +
-            (filesList[i].size / 1024.0 / 1024.0).toFixed(2) + " MB" +
-          "</td>  \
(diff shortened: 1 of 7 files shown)


app/frontend/src/uploadDocuments.js
```diff
@@ -55,9 +55,7 @@ injectDocumentsTable = function(documentsInput) {
             "Uploading" + "<span class='upload-progress'><div class='upload-progress-spinner'></span></div>" +
           "</td>  \
           <td class='govuk-table__cell'</td>  \
-          <td class='govuk-table__cell' scope='row'>" +
-            "<a href='#' class='govuk-link govuk-link--no-visited-state'>Cancel</a>" +
-          "</td>  \
+          <td class='govuk-table__cell'</td> \
```
tatyree: Your `td` isn't closed.

cesidio:
One thing, make sure there is only one commit, or if more than one, aim for meaningful changes grouped together

app/services/document_upload.rb
```diff
@@ -33,20 +40,18 @@ def set_public_permission_on_document
...
-    download_path = "#{uploaded.id}"
-    begin
-      drive_service.get_file(
-        uploaded.id,
-        acknowledge_abuse: false,
-        download_dest: download_path
-      )
-    rescue Google::Apis::ClientError
-      drive_service.delete_file(uploaded.id)
+    self.download_path = "#{uploaded.id}"
```
cesidio: If we just need to make sure to have an empty string instead of a `nil` value, can we go for `uploaded.id.to_s` instead of creating a new interpolated string?

That will be also faster and allocate less memory

PR #1342, opened 2020-02-24
Flatten and normalize records

tatyree:
While this has a lot of commits, the bulk of the changes are in lib/export_tables_to_big_query.rb

This is now the consolidated lib that handles all aspects of getting data out of postgres and in to BigQuery.


lib/export_tables_to_big_query.rb
```diff
@@ -0,0 +1,254 @@
...
+          fh.puts clean_record(row, table.to_s).to_json
+        end
+        logging_details[:records_processed] += batch.size
+      end
+      fh.close
+      Rails.logger.info(logging_details.to_json)
+    end
+  end
+
+  # Ensure that all records have a consistent map of the data keys. This is to prevnet BigQuery breaking when importing
```
cpjmcquillan: *prevent

lib/export_tables_to_big_query.rb
```diff
@@ -0,0 +1,254 @@
...
+          scratch[key] = record[key].to_s(:db)
+        end
+      elsif key.match?(/data/) # Flatten the data structure into one level
+        if record.data.respond_to?(:keys)
+          data_field_normalizer[table_name].each do |normalized_key|
+            value = record.data[normalized_key]
+
+            # Dates are the only values in gias data that can be cast without running in to trouble with variations in
+            # the data. BigQuery already does this intelligently with every other column type. Miscast values are
+            # *sometimes* raised by BigQuery as errors (see next paragraph) and can be ignored if there are staistically
```
cpjmcquillan: *statistically

lib/export_tables_to_big_query.rb
```diff
@@ -0,0 +1,254 @@
...
+          data_field_normalizer[table_name].each do |normalized_key|
+            scratch["data_#{normalized_key.chomp(')').gsub(/\W+/, '_')}"] = nil
+          end
+        end
+      elsif key == 'working_patterns'
+        # This attribute use the array_enum gem to map an array of integers to a text array. When the record is dumped
+        # using `as_json`, the gem does not do the conversion. Calling the non json version of the attribute ensures the
+        # data gets converted as expected.
+        scratch['working_patterns'] = record.working_patterns
+      elsif DROP_THESE_ATTRIBUTES.include?(key)
```
cpjmcquillan: Does it make sense to skip dropped attributes first?

tatyree: As per our conversation, we'll leave this for now.  Moving it is a lot of work and does little to mitigate the size or readability issues.

Gemfile
```diff
@@ -9,84 +9,77 @@ end
 
 gem 'rails', '~> 5.2.4'
 
-gem 'puma', '~> 4.3'
-gem 'pg', '~> 1.1'
+gem 'activerecord-session_store'
```
cesidio: We could just enable this in rubocop.yml, at the moment is disabled, and check it there are warnings.

```
 Bundler/OrderedGems:
    Enabled: true
```

Ideally put this piece of work into another PR

tatyree: I'll add it in, rebase and push.  Good suggestion.  I didn't realise rubocop offered this cop.

lib/export_tables_to_big_query.rb
```diff
@@ -0,0 +1,254 @@
+require 'fileutils'
```
cesidio: I preferred to have 2 services instead of just 1. One that exports to CloudStorage  and one that imports on BigQuery. 250 lines of code in one file is a lot, plus I do not see the `spec/lib.export_tables_to_big_query_spec` file.

tatyree: It performs the single task of loading files from our system to BigQuery.  The two phases required to do that are an implementation detail, subject to change, and one is entirely dependent on the other. It does not make sense to separate them into different files for the sole purpose of reducing class size.  It may be useful from a testing standpoint.  

Given that this is almost three sprints behind schedule, and the now-missing tests had very light and specific (read 'fragile') coverage, it did not make sense to try and adapt them; rewriting them to be more robust and provide higher coverage would have taken at least another week. 

I've put a ticket in JIRA to review the overall approach and tech debt and decide what our next steps need to be: https://dfedigital.atlassian.net/browse/TEVA-527

Are you willing to accept this as a compromise?

cesidio: Yes willing to accept the compromise, thanks

PR #1346, opened 2020-02-25
UD8: Add change upload documents link to summary page

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-414&search=ud8

## Changes in this PR:

Draft PR for review.

Edited to add: This draft PR is intended to review only the commits by author david-mears-dfe, since it branches off Renny4Real's branch `show-uploaded-documents-on-summary-page` which will be reviewed in a different PR for UD7.

In the two commits so far, a test was written to ensure we have the correct redirects in place. We did not (after checking with UX), so the redirect logic was changed, using `application_controller.rb`'s `redirect_to_next_step` method.

## Next steps:

- [ ] Test for the other possible paths, both happy and off the beaten track. Unsure if we need tests where the user does already have a document uploaded (time-expensive test?). We can certainly test cases where the user has selected 'Yes' on supporting_documents_path before review (but doesn't upload).


app/models/concerns/vacancy_candidate_specification_validations.rb
```diff
@@ -12,6 +12,10 @@ module VacancyCandidateSpecificationValidations
                                if: proc { |model| model.qualifications.present? }
   end
 
+  def upload_feature_enabled?
+    UploadDocumentsFeature.enabled?
+  end
+
```
dushan-madetech: Curious as to why you introduced a method instead of just using `UploadDocumentsFeature.enabled?` directly where needed?

app/views/hiring_staff/vacancies/review.html.haml
```diff
@@ -8,229 +8,28 @@
...
+      - else
         = t('jobs.review_future')
     = render 'hiring_staff/vacancies/error_messages', errors: @vacancy.errors
 
 .vacancy.govuk-grid-row
   .govuk-grid-column-two-thirds
+    = render 'hiring_staff/vacancies/review_vacancy_sections/review_job_specification'
+    - if UploadDocumentsFeature.enabled?
+      = render 'hiring_staff/vacancies/review_vacancy_sections/review_supporting_documents'
+    - else
```
cpjmcquillan: After the feature flag is on, there may still be jobs with candidate specification data that haven't yet been published - correct?

app/models/concerns/vacancy_candidate_specification_validations.rb
david-mears-dfe: Apologies, this is out of scope for the PR - rebased and re-pushed to reflect this.

app/views/hiring_staff/vacancies/review.html.haml
david-mears-dfe: Apologies, this is out of scope for the PR - rebased and re-pushed to reflect this.

PR #1351, opened 2020-02-26
UD8: Add change upload documents link to summary page

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-414&search=ud8

## Changes in this PR:

This PR adds the link from the Review page section for supporting documents to the documents index page, and ensures that clicking 'Save and continue' from that index page returns the user to the Review page. This navigation sequence was checked with Chris (UX).

This used `application_controller.rb`'s `redirect_to_next_step` method.

Question for discussion: is it OK to not include adding or removing a document in the tests, but instead just test the pathway (page sequence)? By not including uploading in the tests, the tests are simpler and quicker. But we don't have the security of knowing that a change in supporting documents results in a change in the review page view, except insofar as existing tests for uploading cover this. My feeling is it's fine not to include uploading in the tests, since the uploading is performed by the same route (documents index page) in both the existing tests (for vacancy creation) and in the case of editing via the review link introduced by this PR.

## Before:

Screenshot 2020-02-26 at 12 21 46

## After:

Screenshot 2020-02-26 at 12 20 14
Screenshot 2020-02-26 at 12 20 51

### DIFF
--- a/app/controllers/hiring_staff/vacancies/documents_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/documents_controller.rb
@@ -55,8 +55,12 @@ class HiringStaff::Vacancies::DocumentsController < HiringStaff::Vacancies::Appl
     redirect_to supporting_documents_school_job_path unless supporting_documents
   end
 
+  def next_step
+    application_details_school_job_path
+  end
+
   def redirect_to_next_step_if_save_and_continue
-    redirect_to application_details_school_job_path if params[:commit] == 'Save and continue'
+    redirect_to_next_step(vacancy) if params[:commit] == 'Save and continue'
   end
 
   def process_documents
--- a/app/controllers/hiring_staff/vacancies/documents_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/documents_controller.rb
@@ -52,7 +52,7 @@ class HiringStaff::Vacancies::DocumentsController < HiringStaff::Vacancies::Appl
 
   def redirect_if_no_supporting_documents
     supporting_documents = session[:vacancy_attributes]['supporting_documents']
-    redirect_to supporting_documents_school_job_path unless supporting_documents == 'yes'
+    redirect_to supporting_documents_school_job_path unless supporting_documents
   end
 
   def redirect_to_next_step_if_save_and_continue
--- a/spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
+++ b/spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
@@ -655,15 +655,6 @@ RSpec.feature 'Creating a vacancy' do
 
       context 'editing the supporting_documents' do
         let(:feature_enabled?) { true }
-        let(:document_upload) { double('document_upload') }
-
-        before do
-          allow(DocumentUpload).to receive(:new).and_return(document_upload)
-          allow(document_upload).to receive(:upload)
-          allow(document_upload).to receive_message_chain(:uploaded, :web_content_link).and_return('test_url')
-          allow(document_upload).to receive_message_chain(:uploaded, :id).and_return('test_id')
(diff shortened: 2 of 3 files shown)


app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -52,11 +52,16 @@ def documents_form_params
...
+  end
+
+  def next_step
+    application_details_school_job_path
   end
 
   def redirect_to_next_step_if_save_and_continue
-    redirect_to application_details_school_job_path if params[:commit] == 'Save and continue'
+    vacancy = Vacancy.find(session[:vacancy_attributes]['id'])
+    redirect_to_next_step(vacancy) if params[:commit] == 'Save and continue'
```
cpjmcquillan: The documents controller does not have a next_step method currently, which next step does the application controller user in this case?

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -653,6 +653,47 @@
...
+        let(:feature_enabled?) { true }
+        let(:document_upload) { double('document_upload') }
+
+        before do
+          allow(DocumentUpload).to receive(:new).and_return(document_upload)
+          allow(document_upload).to receive(:upload)
+          allow(document_upload).to receive_message_chain(:uploaded, :web_content_link).and_return('test_url')
+          allow(document_upload).to receive_message_chain(:uploaded, :id).and_return('test_id')
+          allow(document_upload).to receive(:safe_download).and_return(true)
+        end
```
cpjmcquillan: Do we need this block if the document upload is not re-tested?

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -653,6 +653,47 @@
...
+          expect(page).to have_content('Review this job')
+
+          within '#change-supporting-documents' do
+            click_on 'Change'
+          end
+
+          expect(page).to have_content('Step 2 of 3')
+          expect(page.current_path).to eq(documents_school_job_path)
+
+          # Omit the step of uploading a document as this is tested elsewhere
```
cpjmcquillan: Remove comment when above document_upload `before` block removed

app/controllers/hiring_staff/vacancies/documents_controller.rb
david-mears-dfe: In `develop`, the application controller never gets asked to call `next_step` by this documents controller. The next step was managed by deleted line 59 (no involvement of application controller).
With this change, `next_step` is added, so involving application controller becomes possible.

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
david-mears-dfe: Good catch!

david-mears-dfe:
Thanks for the comments @cpjmcquillan

app/controllers/hiring_staff/vacancies/documents_controller.rb
```diff
@@ -52,11 +52,16 @@ def documents_form_params
...
+    redirect_to supporting_documents_school_job_path unless supporting_documents
+  end
+
+  def next_step
+    application_details_school_job_path
   end
 
   def redirect_to_next_step_if_save_and_continue
-    redirect_to application_details_school_job_path if params[:commit] == 'Save and continue'
+    vacancy = Vacancy.find(session[:vacancy_attributes]['id'])
```
cesidio: This line is redundant as we have a method called `vacancy` that does that

david-mears-dfe: Got rid of the line

PR #1356, opened 2020-02-27
Add supporting documents section to vacancy view

cpjmcquillan:
[UD11 (TEVA-420)](https://dfedigital.atlassian.net/browse/TEVA-420)

Given that this view will change drastically with the key information epic, I have not refactored the view or eliminated the code duplication yet.

**Vacancy before feature flag on**
image

**Vacancy after feature flag on**
image

**TODO**
- [x] Review with Content, UX


david-mears-dfe:
Does this PR include the ability to download a document, as on the ticket?

cpjmcquillan:
> Does this PR include the ability to download a document, as on the ticket? Is there a reason that isn't tested?

The ability to download relies on Google Drive. The user only clicks a link. I don't feel like we should be testing that html elements are rendered correctly, but maybe I'm wrong?

david-mears-dfe:
> > Does this PR include the ability to download a document, as on the ticket? Is there a reason that isn't tested?
> 
> The ability to download relies on Google Drive. The user only clicks a link. I don't feel like we should be testing that html elements are rendered correctly, but maybe I'm wrong?

I didn't realise the download came just from a link - good point. I wonder if there's an argument for adding a ticket to test that `download_url` generates the correct link - depends how closely it's based on code we wrote vs code from Google.

cpjmcquillan:
> > > Does this PR include the ability to download a document, as on the ticket? Is there a reason that isn't tested?
> > 
> > 
> > The ability to download relies on Google Drive. The user only clicks a link. I don't feel like we should be testing that html elements are rendered correctly, but maybe I'm wrong?
> 
> I didn't realise the download came just from a link - good point. I wonder if there's an argument for adding a ticket to test that `download_url` generates the correct link - depends how closely it's based on code we wrote vs code from Google.

`download_url` is assigned directly from the `web_content_link` field returned by the `create_file` method on `Google::Apis::DriveV3::DriveService`

spec/features/job_seekers_can_view_a_vacancy_spec.rb
```diff
@@ -91,6 +97,90 @@
       expect(page).to have_content(30)
     end
 
+    context 'when the upload documents feature flag is OFF' do
+      before do
+        vacancy = create(:vacancy, :published)
+        visit job_path(vacancy)
```
david-mears-dfe: This `before` block, and the 3 other new `before` blocks, are probably time-expensive when running the test suite, as they run before each example. You might want to refactor each of these 4 `context`s to only have one `scenario` instead of 2, so that each `context` only has to run the `before` stuff once. I recognise that reduces the tidiness of the code, though.

app/frontend/styles/components/vacancy.scss
```diff
@@ -37,4 +37,37 @@
...
+
+          svg {
+            float: left;
+            margin-right: 0.5rem;
+            margin-top: 0.3rem;
+          }
+        }
+      }
+    }
+}
```
cesidio: not aligned

PR #1359, opened 2020-02-28
Delete docs on vacancy deletion

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&modal=detail&selectedIssue=TEVA-424

## Changes in this PR:

With this PR, a request to delete a vacancy leads to a DocumentDelete object deleting each attached document. I don't test this all the way down the full stack as it were, because there is no change in the view and I trust the existing tests on the DocumentDelete; so I test that Vacancy#delete_documents is called, and that Vacancy#delete_documents calls the appropriate methods on DocumentDelete.

### DIFF
--- a/app/controllers/hiring_staff/vacancies_controller.rb
+++ b/app/controllers/hiring_staff/vacancies_controller.rb
@@ -2,7 +2,7 @@ class HiringStaff::VacanciesController < HiringStaff::Vacancies::ApplicationCont
   before_action :set_vacancy, only: %i[review]
 
   def show
-    vacancy = current_school.vacancies.active.find(id)
+    vacancy = find_active_vacancy_by_id
     unless vacancy.published?
       return redirect_to school_job_review_path(vacancy.id),
                          alert: I18n.t('messages.jobs.view.only_published')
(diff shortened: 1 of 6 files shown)


david-mears-dfe:
Does this PR include the ability to download, as in the ticket?

cesidio:
I think that the last commit should be a PR on its own. Big PRs are difficult to review

spec/factories/vacancies.rb
```diff
@@ -10,42 +10,43 @@
       create_list :document, 3, vacancy: vacancy
     end
 
-    job_title { Faker::Lorem.sentence[1...30].strip }
-    job_description { Faker::Lorem.paragraph(sentence_count: 4) }
+    application_link { Faker::Internet.url }
```
cpjmcquillan: Is alphabetical more useful than in order of hiring staff journey? I'm not sure, but agree we should have an approach!

david-mears-dfe:
> I think that the last commit should be a PR on its own. Big PRs are difficult to review

Reverted.

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -86,4 +83,8 @@ def clear_cache_and_step
   def set_vacancy
     @vacancy = current_school.vacancies.active.find(vacancy_id)
```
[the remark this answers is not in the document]
david-mears-dfe: Now that I have reverted the part of this PR about combining the `job_id` and `id` params together, this will have to wait until we reduce the variety of param names down from 2 to 1.

spec/factories/vacancies.rb
david-mears-dfe: I think our approach is alphabetization for long lists

PR #1360, opened 2020-02-28
Combine two param types into one for hiring_staff/vacancies_controller

david-mears-dfe:
This PR builds on #1359 (now merged). It's written in the spirit of the agreement to do refactoring as we come across things that need refactoring.

I made a judgement call about whether there should be some inconsistency in the `hiring_staff/vacancies_controller` file or in the `config/routes.rb` file. I traded off consistency in the params used in the former against regularity in the latter. Previously, #show, #edit and #delete used `id` as the param key, while #review and #summary used `job_id`. This had caused some duplication of methods in the controller. Open to disagreement on this.

A couple of path aliases are different but the urls are the same as before.


david-mears-dfe:
It occurs to me that changing the name of the params for some routes might impact any external users of our API, if they are specifying param values in their requests. How would I check this?

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -16,7 +16,7 @@ def new
   end
 
   def edit
-    vacancy = current_school.vacancies.find(id)
+    vacancy = current_school.vacancies.find(vacancy_id)
```
david-mears-dfe: I could change `vacancy_id` across this file into `job_id` to match the param's name instead of matching the ruby object's name. This would be a trade-off between two different sorts of consistency. Which is preferred?

tatyree: Personally, I'd rather it matched the Ruby object's name. However, @stevenleggdfe is likely to be the person most affected by changes in parameters.

david-mears-dfe:
Replaced by #1428

PR #1363, opened 2020-02-28
Disable the export to BigQuery job

tatyree:
Quick hack to turn off the exporter and prevent it from eating all the disk space.


tatyree:
Closing this for now as there are better workarounds.

PR #1366, opened 2020-03-02
Copy/edit supporting documents for existing vacancies

cpjmcquillan:
[UD14: TEVA-440](https://dfedigital.atlassian.net/browse/TEVA-440)

**In this PR**
- Copy supporting documents from an existing vacancy (published/scheduled/expired)
- Edit supporting documents for an existing vacancy (published/scheduled/draft)
- Refactor documents controller and routing to simplify code and remove duplication. (Now passes job_id as parameter and does not use the session)
- Include notification label on review/edit page to highlight supporting documents section

**TODO**
- [x] Code review by 2+ devs
- [ ] UX review (view on staging)
- [ ] Product review (view on staging)


app/views/hiring_staff/vacancies/documents/_title.html.haml
```diff
@@ -0,0 +1,12 @@
+- if @vacancy.published?
+  - content_for :page_title_prefix, "Edit the Supporting documents"
```
tatyree: Shouldn't Supporting be lower case or documents be title case?

cpjmcquillan: I agree. I just followed convention for all the other page titles. Happy to change them all 👍🏼

cesidio: It would be nice to I18n these strings

tatyree: If you're following convention then don't sweat it.  Let's get it shipped and refactor for casing/I18N later.

PR #1379, opened 2020-03-03
Revert "Key info for search snippet"

david-mears-dfe:
Reverts DFE-Digital/teacher-vacancy-service#1335

Misunderstood a couple of things: our workflow (shouldn't merge to develop unless ready for production) and use of Figma (should not match styling of Figma unless requested).

Reverting so I can undo styling.


david-mears-dfe:
Closed so I can instead revert the specific file that is not wanted

PR #1386, opened 2020-03-05
Add missing cloud_storage_bucket env variables

cesidio:
## Jira ticket URL:
https://dfedigital.atlassian.net/secure/RapidBoard.jspa?rapidView=21&projectKey=TEVA&modal=detail&selectedIssue=TEVA-571


cesidio:
> Weird that it was working before 🤔

It was working in the worker container but tasks are run in the web container, that was missing that environment variable

PR #1390, opened 2020-03-06
Information banner when editing/reviewing legacy listings

cpjmcquillan:
[UD14.2 (TEVA-576)](https://dfedigital.atlassian.net/browse/TEVA-576)

image

Adds an information banner that alerts hiring staff users to the supporting documents feature when they edit/review a vacancy created before the feature existed.

### DIFF
--- a/app/views/feedback/_outlined_flash_message.html.haml
+++ b/app/views/feedback/_outlined_flash_message.html.haml
@@ -2,4 +2,13 @@
   .govuk-body.mv0{ class: "govuk-#{style}-summary__body" }
     %a.govuk-link.govuk-link--no-visited-state.govuk-success-summary__dismiss-link.js-dismissable__link{ href: '#' }
       = t('buttons.dismiss')
+    - if defined?(heading)
+      %h3.govuk-heading-s.mb0
+        = heading
     = message
+    <br>
+    - if defined?(links)
+      - links.each do |section|
+        %a.govuk-link.govuk-link--no-visited-state{ href: "##{section}"}
+          = t("jobs.#{section}")
+        <br>
--- a/app/frontend/styles/application.scss
+++ b/app/frontend/styles/application.scss
@@ -25,6 +25,7 @@ $govuk-image-url-function: frontend-image-url;
 @import 'components/feedback_form';
 @import 'components/filter_vacancies';
 @import 'components/flashes';
+@import 'components/information_summary.scss';
 @import 'components/links';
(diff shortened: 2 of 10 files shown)


app/models/vacancy.rb
```diff
@@ -268,6 +268,16 @@ def delete_documents
     self.documents.each { |document| DocumentDelete.new(document).delete }
   end
 
+  def new_information
```
cpjmcquillan: Maybe there is a cleaner way of doing this. We want to add a link for every section that has new information.

cesidio: I would use a rails helper for this

app/views/feedback/_outlined_flash_message.html.haml
```diff
@@ -2,4 +2,13 @@
   .govuk-body.mv0{ class: "govuk-#{style}-summary__body" }
     %a.govuk-link.govuk-link--no-visited-state.govuk-success-summary__dismiss-link.js-dismissable__link{ href: '#' }
       = t('buttons.dismiss')
+    - if defined?(heading)
```
cesidio: I would just check the action here, you already have `@vacancy` defined and accessible, so no need to pass heading and links local variables

app/models/vacancy.rb
david-mears-dfe: How about calling this 'new_sections'? It reveals that it returns a list of sections and is more specific than 'information'

david-mears-dfe: This would change en.yml too

app/models/vacancy.rb
```diff
@@ -268,6 +268,16 @@ def delete_documents
...
+    sections = []
+    sections << 'supporting_documents' if supporting_documents_new
+    sections
+  end
+
+  def supporting_documents_new
```
david-mears-dfe: I like the Ruby convention where a method that returns a Boolean ends with a question mark

app/models/vacancy.rb
david-mears-dfe: What kind of rails helper would help with this?

cesidio: Have a look at the helper methods defined in `app/helpers/vacancies_helper.rb`

app/views/feedback/_outlined_flash_message.html.haml
cpjmcquillan: Heading and links are passed because they might not always be relevant when using this partial

PR #1393, opened 2020-03-09
Key info feature base rebase

tatyree:
I didn't want to overwrite history in @david-mears-dfe's branch, so I branched again from the branch I   used to rebase `develop` on to `key-info-feature-base` and will merge this instead.  This way the original branch's history is retained in case something goes wrong with this merge. 

This has already been PR'd as `key-info-feature-base`, so this branch is not for PR but only for merging back in to `develop`.


app/frontend/styles/application.scss
```diff
@@ -18,32 +18,43 @@ $govuk-image-url-function: frontend-image-url;
 
 @import 'components/awaiting_feedback_vacancies';
 @import 'components/banners';
+@import 'components/banners';
```
cesidio: Duplicate file imports

app/frontend/styles/components/vacancies-show.scss
```diff
@@ -0,0 +1,22 @@
+
```
cesidio: Would be nice not to have empty lines at the beginning of files

tatyree:
And I accidentally force-pushed to the branch I branched this from, so we've lost that history now, unfortunately.

PR #1402, opened 2020-03-10
Ensure tmp files get removed. 

tatyree:
Prior to this change any exception would prevent the temp files being removed. This fixes that and should temporarily stop breakages that arise when a container's drive gets full.

The best long term solution is to stop storing AuditData in our database and store it somewhere better suited to unbounded data-sets, like DynamoDB or Firestore.
1
The second best solution is to increase the disk size of the ec2 instances hosting the containers.


tatyree:
Merge to wrong branch. Closing.

PR #1415, opened 2020-03-11
Teva 579 bigquery try to reduce loading

tatyree:
It appears that ECS (or terraform) is killing the container running the export job for reasons of excessive usage.  I *think* it is around the auto-scaling setting for max CPU usage, but am not sure and experimentation takes at least 40 minutes per run. 

After discussing with Steven, we decided we can turn off `AuditData` for now.

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -30,7 +30,7 @@ class ExportTablesToBigQuery
     activities
     ar_internal_metadata
     audit_data
-    friendly_id_slugs
+    friendlyuid_slugs
     schema_migrations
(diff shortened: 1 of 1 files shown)


lib/export_tables_to_big_query.rb
```diff
@@ -29,7 +29,8 @@ class ExportTablesToBigQuery
   EXCLUDE_TABLES = %w[
     activities
     ar_internal_metadata
-    friendly_id_slugs
+    audit_data
+    friendlyuid_slugs
```
cesidio: This should be 
`friendly_id_slugs`

PR #1422, opened 2020-03-16
Upload directly to BigQuery

tatyree:
This is quicker and generates lower loads.


david-mears-dfe:
Rubocop offenses, all fixable if someone runs `rubocop -a`

tatyree:
> Rubocop offenses, all fixable if someone runs `rubocop -a`

Fixed.  I keep forgetting to run the blasted thing, so I'm going to set up a hook.

PR #1425, opened 2020-03-16
Turn on upload documents feature

cpjmcquillan:
Ensure environment variable feature flag is updated in all places


cpjmcquillan:
> I've already done a TF apply to production having set it `production.tfvars`, so it's live. Will PR that on the secrets repo as well.

I've updated the secrets already and it has been merged (I was doing this while you were but I started at the opposite end of the process)

PR #1428, opened 2020-03-16
Combine two params into one (into master)

david-mears-dfe:
This PR replaces https://github.com/DFE-Digital/teacher-vacancy-service/pull/1360/ which was a request to merge into *develop*. There is a small amount of conversation there. @stevenleggdfe confirmed that this change does not impact him.

Original PR description:

This PR builds on #1359 (now merged). It's written in the spirit of the agreement to do refactoring as we come across things that need refactoring.

I made a judgement call about whether there should be some inconsistency in the `hiring_staff/vacancies_controller` file or in the `config/routes.rb` file. I traded off consistency in the params used in the former against regularity in the latter. Previously, #show, #edit and #delete used `id` as the param key, while #review and #summary used `job_id`. This had caused some duplication of methods in the controller. Open to disagreement on this.

A couple of path aliases are different but the urls are the same as before.


david-mears-dfe:
I'm confused about where the interloping 5 commits (3 from @tatyree and 2 from @cpjmcquillan ) came from.

tatyree:
This one has been open for a while and has accidentally picked up external commits.  Can we go ahead and just close and delete it for now?

PR #1432, opened 2020-03-17
GOVUK PaaS

cesidio:


app/controllers/application_controller.rb
```diff
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
 
-  before_action :check_staging_auth, except: :check
+  # before_action :check_staging_auth, except: :check
```
cesidio: We disable staging HTTP basic auth on staging for now

tatyree: :+1: Let's not leave it in as comment lint for too long, but I'm ok with it while we're still polishing things.

config/schedule.yml
```diff
@@ -0,0 +1,19 @@
+daily_alerts:
```
tatyree: A massive improvement on those awful pipeline jobs. 👍

config/initializers/00_cf_services.rb
```diff
@@ -0,0 +1,7 @@
+if ENV['VCAP_SERVICES']
+  VCAP_SERVICES = JSON.parse(ENV['VCAP_SERVICES'])
```
tatyree: Is it worth `.fetch`ing these so we it raises an exception here if the variable isn't set?  I sometimes find it hard to track down bugs caused by unset variables due to the fact that `[]` assigns `null` rather than failing. 

In other words `ENV.fetch('VCAP_SERVICES')`.

lib/tasks/cloudfoundry.rake
```diff
@@ -0,0 +1,7 @@
+namespace :cf do
+  desc 'Only run on the first application instance'
+  task :on_first_instance do # rubocop:disable Rails/RakeEnvironment
+    instance_index = JSON.parse(ENV['VCAP_APPLICATION'])['instance_index'] rescue nil
```
tatyree: See comment in the initializer, above.

spec/requests/staging_auth_spec.rb
```diff
@@ -11,7 +11,8 @@
 
   context 'when in staging' do
     before(:each) { stub_global_auth(return_value: true) }
-    it_behaves_like 'basic auth'
+    # it_behaves_like 'basic auth'
```
tatyree: See note on comment lint, above.  Actually, if we kill this completely, then we should probably just remove this test altogether.

config/initializers/00_cf_services.rb
cesidio: These services are only set on CloudFoundry, when we run the server on development machine we do not have them, so we do not want to raise an exception in that case

cesidio:
> Your call as to whether or not to change to `.fetch` (see comments). I'm happy with it either way. Good work. :shipit:

Thanks!

PR #1433, opened 2020-03-18
Shared partials for vacancy form sections

cpjmcquillan:
Throwing this out there as an initial step towards simplifying the vacancy forms. 

**In this PR**
- Move the title and submit code to partials
- Introduce some abstraction around steps, but this logic needs refining (I've left this to whoever picks up the sidebar ticket). 

Feel free to use these partials on forms wherever suitable, but as we create new pages in the vacancy form this may become more useful


tatyree:
This one also gets done in #1437 doesn't it?

PR #1435, opened 2020-03-20
Show subject row in job details only when a subject is present

cpjmcquillan:
Quick bug fix

### DIFF
--- a/app/presenters/vacancy_presenter.rb
+++ b/app/presenters/vacancy_presenter.rb
@@ -106,11 +106,7 @@ class VacancyPresenter < BasePresenter
   end
 
   def subject_count
-    count = 0
-    count += 1 if main_subject.present?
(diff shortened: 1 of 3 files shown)


app/presenters/vacancy_presenter.rb
```diff
@@ -101,6 +101,18 @@ def supporting_subjects
     "#{first_supporting_subject}, #{second_supporting_subject}"
   end
 
+  def any_subjects?
+    main_subject.present? || other_subjects.present?
+  end
+
+  def subject_count
+    count = 0
```
cesidio: This could be rewritten in:

```
[main_subject, first_supporting_subject, second_supporting_subject].count(&:present?)
```

cpjmcquillan: I love it

PR #1444, opened 2020-03-25
Move salary conversion migration into a rake task to be run manually

cpjmcquillan:
Timeout limit was reached when executing migration due to large number of records.
This rake task can be run after deployment, and columns can be removed once the rake task has been completed.


lib/tasks/convert_salary_ranges_to_salary_string.rake
```diff
@@ -0,0 +1,24 @@
+namespace :data do
+  desc 'Convert salary ranges to string'
+  namespace :convert_salary_ranges do
+    task vacancies: :environment do
+      Vacancy.in_batches.each_record do |vacancy|
+        min_pay_scale = vacancy.min_pay_scale_id.present? ? PayScale.find(vacancy.min_pay_scale_id) : nil
```
tatyree: So, first off, I'd get rid of these two db calls - they're comparatively expensive and you've already got the data in `PAYSCALE_DATA` in `seeds.rb`.   Just move it to it's own initializer and use that everywhere you need it.

lib/tasks/convert_salary_ranges_to_salary_string.rake
```diff
@@ -0,0 +1,24 @@
...
+        maximum_salary = vacancy.maximum_salary.present? ?
+          "£#{vacancy.maximum_salary.gsub(/\d(?=(...)+$)/, '\0,')}" : nil
+        salary_range = [minimum_salary, maximum_salary].reject(&:blank?).join(' to ')
+        salary = [pay_scales, salary_range].reject(&:blank?).join(', ')
+        if vacancy.pro_rata_salary?
+          salary += ' per year pro rata'
+        elsif vacancy.flexible_working?
+          salary += ' per year (full-time equivalent)'
+        end
+        vacancy.update(salary: salary)
```
tatyree: I haven't done this in a while so I'm a bit fuzzy on the details, but you could persist the changes into an Array of Arrays while you're reading/altering the data then iterate over that that using `.in_batches` to do batch updates as well.  That will drop something like 30-40k individual writes to 30 or 40, which should be significants quicker.

PR #1445, opened 2020-03-25
Fix salary migrations

cpjmcquillan:
**In this PR**
- Add salary column
- Add rake task to convert salary fields to new string field

**Todo**
- [ ] Once deployed to production, execute rake task
- [ ] Merge salary input changes to master
- [ ] Execute rake task again to catch any jobs created while between previous rake task and deploy
- [ ] Create a PR with the migrations to remove old columns


db/schema.rb
```diff
@@ -97,6 +97,13 @@
     t.index ["title"], name: "index_leaderships_on_title", unique: true
   end
 
+  create_table "pay_scales", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "label", null: false
```
tatyree: For now, just remove `null: false` and we can use a validation.  It'll stop it failing.

tatyree: To expedite, though, don't bother with the validation unless it's already there.  Put in a ticket to add it later.

tatyree:
I'm merging with @cpjmcquillan's blessing to get it moving.

PR #1447, opened 2020-03-25
Return nil when pay scale is missing

cpjmcquillan:
This rake task is running nicely on PaaS using production data. 
Hopefully after this we can move the rest of the changes forward 🤞

### DIFF
--- a/lib/tasks/convert_salary_ranges_to_string.rake
+++ b/lib/tasks/convert_salary_ranges_to_string.rake
@@ -2,6 +2,7 @@ namespace :data do
   desc 'Convert salary ranges to strings for vacancies without the new salary field'
   namespace :convert_salary_ranges do
     task vacancies: :environment do
+      Rails.logger.info('Conversion of salary ranges to strings has been started')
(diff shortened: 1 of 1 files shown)


lib/tasks/convert_salary_ranges_to_string.rake
```diff
@@ -2,6 +2,7 @@ namespace :data do
   desc 'Convert salary ranges to strings for vacancies without the new salary field'
   namespace :convert_salary_ranges do
     task vacancies: :environment do
+      Rails.logger.info('Conversion of salary ranges to strings has been started')
```
cesidio: This line is duplicated

lib/tasks/convert_salary_ranges_to_string.rake
```diff
@@ -25,10 +25,12 @@ namespace :data do
         elsif vacancy.flexible_working?
           salary += ' per year (full-time equivalent)'
         end
-
+
```
cesidio: Empty spaces here

PR #1449, opened 2020-03-26
Remove validations when updating salary field

cpjmcquillan:
- Required because we are updating all vacancies (including those created before various validations were added on the model)
- Tested on PaaS using production dataset


lib/tasks/convert_salary_ranges_to_string.rake
```diff
@@ -26,12 +33,15 @@ namespace :data do
           salary += ' per year (full-time equivalent)'
         end
 
-        vacancy.update(salary: salary)
+        vacancy.salary = salary
+        # Some vacancies being updated will have been created prior to certain validations
+        vacancy.save(validate: false)
```
cesidio: There is a better way to just update the columns without running validations and updating the timestamps
https://apidock.com/rails/ActiveRecord/Persistence/update_columns

PR #1451, opened 2020-03-26
Salary input changes without migrations

cpjmcquillan:
Re-merge of the salary input changes, but without database migrations


app/controllers/hiring_staff/vacancies/pay_package_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def pay_package_form_params
+    (params[:pay_package_form] || params).permit(:salary, :benefits)
+  end
+
+  def next_step
+    UploadDocumentsFeature.enabled? ? supporting_documents_school_job_path : candidate_specification_school_job_path
```
cesidio: You could remove the feature flag here too while you are at it

app/form_models/pay_package_form.rb
```diff
@@ -0,0 +1,7 @@
+class PayPackageForm < VacancyForm
+  include VacancyPayPackageValidations
+
+  def completed?
```
cesidio: This does not make much sense, you can just invoke `valid?`

app/controllers/hiring_staff/vacancies/pay_package_controller.rb
cpjmcquillan: Can I remove the feature flag in a separate PR? It's in a lot of places

app/models/concerns/vacancy_pay_package_validations.rb
```diff
@@ -0,0 +1,17 @@
...
+  extend ActiveSupport::Concern
+  include ApplicationHelper
+
+  included do
+    validates :salary, presence: true
+    validates :salary, length: { minimum: 1, maximum: 256 }, if: :salary?
+  end
+
+  def salary=(value)
+    super(sanitize(value, tags: []))
```
cesidio: Why are you passing an array with no tags here?

app/views/hiring_staff/vacancies/pay_package/show.html.haml
```diff
@@ -0,0 +1,20 @@
+= render 'hiring_staff/vacancies/vacancy_form_partials/title', form_section_title: 'Pay package', form_section_step: 2, form_section_total_steps: 4
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_for @pay_package_form, url: school_job_pay_package_path(@vacancy.id), html: { method: "patch" }, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f|
```
cesidio: Very happy to see the GOVUKDesignSystemFormBuilder 🎊

spec/models/vacancy_spec.rb
```diff
@@ -566,6 +425,14 @@
       expect(vacancy.job_title).to eq(sanitized_title)
     end
 
+    it '#salary' do
```
cesidio: it does not read that well, I would use a context `#salary` and inside would put something like `it strips the tags`

app/models/concerns/vacancy_pay_package_validations.rb
cpjmcquillan: Because no tags are allowed

PR #1452, opened 2020-03-26
Remove upload documents feature flag logic

cpjmcquillan:
**In this PR**
- Remove logic across the application for the feature being on/off

**Todo**
- Remove all candidate specification code at a later date


app/models/concerns/vacancy_candidate_specification_validations.rb
```diff
@@ -13,7 +13,7 @@ module VacancyCandidateSpecificationValidations
   end
 
   def upload_feature_enabled?
```
cesidio: This method should not exist anymore

cpjmcquillan: I agree, making a PR for the candidate spec stuff to be removed following product approval

PR #1454, opened 2020-03-27
Remove candidate specification code except jobseeker view

cpjmcquillan:
Following user research, jobseekers were allowed to upload documents in place of entering information into candidate specification fields. This PR removes redundant candidate specification code.

**In this PR**
- Remove candidate specification controller and view code

**Todo**
- Remove jobseeker view code once no live vacancies have candidate specification fields 
- Remove columns from database once no live vacancies have candidate specification fields


tatyree:
Right now I think we all know why this is being done, but it seems unlikely that someone looking through the history in a year's time will remember why.  Can you put a couple of lines of context in the PR description, please?   Otherwise, a good call.

PR #1457, opened 2020-03-30
Fix schools' religious character

david-mears-dfe:
## Jira ticket URL:

https://dfedigital.atlassian.net/browse/TEVA-605?atlOrigin=eyJpIjoiOGY3NTA4YWMzZTI3NDc5ZTg1NmE0YzhlMjI2MmQ5ZGQiLCJwIjoiaiJ9

## Changes in this PR:

This PR addresses half of the concerns in the ticket: those around the missing religious character. The fault was a misspelling in the codebase and the tests (which were therefore not fit for purpose).

I'm waiting for instructions on whether we should use `detailed_school_type` or `school_type` data, which correspond to different GIAS data columns.

## Next steps:

- [ ] If necessary, update `school_type` to `detailed_school_type` in the SchoolPresenter method `school_type_with_religious_character`

- [ ] Check that this fixes the problem on staging, using real school data


david-mears-dfe:
Alice said that we don't in fact need to change the school type from `school_type` to `detailed_school_type`.

PR #1438, opened 2020-03-23
Combine leadership and NQT to job role in hiring staff journey

david-mears-dfe:
This PR adds the new job role field to the hiring staff journey.

## In this PR
- Add job role string array
- Add NQT to job role for vacancies where NQT true
- When NQT selected in job role, set NQT boolean to true
- When NQT not selected in job role, set NQT boolean to false
- Don't change search
- Don't drop NQT column yet

## Jira ticket URL
[TEVA-534](https://dfedigital.atlassian.net/browse/TEVA-534)

## Next steps
- Update jobseeker view to display job role info
- Remove leadership from code and database
- Update search to filter roles based on NQT being a job role

PR #1464, opened 2020-04-01
Completed step on db

cpjmcquillan:
This PR addresses inconsistencies in the application of the visited class to the sidebar circles which are arising as a result of using the session to store the completed step.

**In this PR**
- Create a completed step field on the vacancy model as a more robust method of determining whether or not a section in the hiring staff journey has been completed

**Next steps** ([TEVA-647](https://dfedigital.atlassian.net/browse/TEVA-647))
- Only use session where strictly necessary
- Refactor job spec and application details forms to use GOVUKDesignSystemFormBuilder
- Implement simpler form logic
    - Currently edit/new/create/update and review/edit involves a lot of unnecessary complexity and duplication in the code
    - A vacancy has been created in the model as soon as the first section of the hiring staff journey has been completed (everything after this is an update of the vacancy)
    - See Pay Package section as an example

### DIFF
--- a/app/controllers/hiring_staff/vacancies/application_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_controller.rb
@@ -2,6 +2,8 @@ require 'auditor'
 require 'indexing'
 
 class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseController
+  before_action :set_vacancy
+
   def school_id
     params.permit![:school_id]
   end
@@ -15,7 +17,11 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
   end
 
   def set_vacancy
-    @vacancy = params[:job_id] ? current_school.vacancies.find(params[:job_id]) : nil
+    if params[:job_id]
+      @vacancy = current_school.vacancies.find(params[:job_id])
+    elsif session_vacancy_id
+      @vacancy = current_school.vacancies.find(session_vacancy_id)
+    end
   end
 
   def current_step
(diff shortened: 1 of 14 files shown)


app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -15,7 +17,12 @@ def session_vacancy_id
   end
 
   def set_vacancy
-    @vacancy = params[:job_id] ? current_school.vacancies.find(params[:job_id]) : nil
+    if params[:job_id]
+      @vacancy = current_school.vacancies.find(params[:job_id])
+    elsif session_vacancy_id
+      @vacancy = current_school.vacancies.find(session_vacancy_id)
+    end
+    @vacancy
```
cesidio, line 25: no need to return the `@vacancy`

db/schema.rb
```diff
@@ -225,6 +225,8 @@
     t.datetime "expiry_time"
     t.string "supporting_documents"
     t.string "salary"
+    t.string "job_roles", array: true
```
david-mears-dfe, line 228: This should probably wait until https://github.com/DFE-Digital/teacher-vacancy-service/pull/1438 is merged

spec/helpers/hiring_staff/job_creation_helper_spec.rb
```diff
@@ -136,6 +136,8 @@
   end
 
   describe '#set_visited_step_class' do
+    let(:vacancy) { double('vacancy') }
```
david-mears-dfe, line 139: Does `instance_double` work here? If so, it's slightly better practice as it will give you warning if the test calls a method that doesn't exist for real `Vacancy`s.

db/schema.rb
cpjmcquillan: Good spot, I will add the correct schema 👍

PR #1467, opened 2020-04-02
Fix bug on production where job_roles.any? throws error

david-mears-dfe:
https://rollbar.com/dfe/teacher-vacancies/items/928/
ActionView::Template::Error: undefined method 'any?' for nil:NilClass

I have tested this change (and the unchanged version) locally, with vacancies that have job_roles = nil.


david-mears-dfe:
Apologies for non-informative PR description. This is to fix a bug on production where jobs which have no job_roles attribute can't be displayed, for example:
https://teaching-vacancies.service.gov.uk/jobs/lead-practitioner-in-design-and-technology-garth-hill-college-bracknell-berkshire

PR #1468, opened 2020-04-02
Fix possible nil value for job_roles

cesidio:


david-mears-dfe:
While you are doing this can you change `if @vacancy.benefits.present?` to if `@vacancy.benefits&.any?` ?
For consistency within the same file.

PR #1473, opened 2020-04-05
Job summary page

cpjmcquillan:
This PR adds a new step (Job summary) to the hiring staff 'Create a job' journey

**TODO**
- [x] Pre-populate about_school field in job summary form with About School data
- [x] Refactor .yml files to wrap after 120 characters
- [x] Refactor user input sanitization
- [x] Product/UX review
- [x] Code review

## Jira ticket URL
[TEVA-542](https://dfedigital.atlassian.net/browse/TEVA-542)
[Figma for design](https://www.figma.com/file/n2wLtKVQ2imMbjdPSHBRKW/Sprint-50?node-id=0%3A1)
## Changes in this PR
- Rename job_description column to job_summary on the vacancies table and allow it to be NULL
- Add about_school column to the vacancies table
- Add Job Summary step to the 'Create a job' journey
## Screenshots of UI changes
image
## Next steps
- Show about_school field on the vacancy in the school overview section that jobseekers see

### DIFF
--- a/app/controllers/hiring_staff/vacancies/job_summary_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/job_summary_controller.rb
@@ -1,0 +1,34 @@
+class HiringStaff::Vacancies::JobSummaryController < HiringStaff::Vacancies::ApplicationController
+  before_action :redirect_unless_vacancy
+
+  def show
+    @job_summary_form = JobSummaryForm.new(@vacancy.attributes)
+  end
+
+  def update
+    @job_summary_form = JobSummaryForm.new(job_summary_form_params)
+
+    if @job_summary_form.valid?
+      update_vacancy(job_summary_form_params, @vacancy)
+      update_google_index(@vacancy) if @vacancy.listed?
+      return redirect_to_next_step_if_save_and_continue
+    end
+
+    render :show
+  end
+
+  private
+
+  def job_summary_form_params
+    (params[:job_summary_form] || params).permit(:job_summary, :about_school)
+                                         .merge(completed_step: current_step)
+  end
+
+  def redirect_to_next_step_if_save_and_continue
+    if params[:commit] == I18n.t('buttons.save_and_continue')
+      redirect_to school_job_review_path(@vacancy.id)
+    elsif params[:commit] == I18n.t('buttons.update_job')
+      redirect_to edit_school_job_path(@vacancy.id), success: I18n.t('messages.jobs.updated')
+    end
+  end
+end
--- a/app/models/concerns/vacancy_job_summary_validations.rb
+++ b/app/models/concerns/vacancy_job_summary_validations.rb
@@ -6,12 +6,4 @@ module VacancyJobSummaryValidations
     validates :job_summary, presence: true
     validates :about_school, presence: true
   end
-
-  def job_summary=(value)
-    super(sanitize(value))
-  end
-
-  def about_school=(value)
-    super(sanitize(value))
-  end
 end
--- a/app/models/concerns/vacancy_job_summary_validations.rb
+++ b/app/models/concerns/vacancy_job_summary_validations.rb
@@ -1,6 +1,5 @@
 module VacancyJobSummaryValidations
   extend ActiveSupport::Concern
-  include ApplicationHelper
 
   included do
     validates :job_summary, presence: true
--- a/app/models/concerns/vacancy_job_summary_validations.rb
+++ b/app/models/concerns/vacancy_job_summary_validations.rb
@@ -1,0 +1,17 @@
+module VacancyJobSummaryValidations
+  extend ActiveSupport::Concern
+  include ApplicationHelper
+
+  included do
+    validates :job_summary, presence: true
+    validates :about_school, presence: true
+  end
+
+  def job_summary=(value)
+    super(sanitize(value))
+  end
+
+  def about_school=(value)
+    super(sanitize(value))
+  end
+end
--- a/app/views/hiring_staff/vacancies/edit_vacancy_sections/_edit_job_summary.html.haml
+++ b/app/views/hiring_staff/vacancies/edit_vacancy_sections/_edit_job_summary.html.haml
@@ -1,0 +1,16 @@
+%h2.govuk-heading-m.app-task-list__section#job_summary_heading
+  %span.app-task-list__section-number 5.
+  = t('jobs.job_summary')
+  = link_to t('buttons.change'), school_job_job_summary_path(@vacancy.id), class: 'govuk-link change-button', 'aria-label': t('jobs.aria_labels.change_job_summary')
+
+%dl.app-check-your-answers.app-check-your-answers--short
+
+  .app-check-your-answers__contents
+    %dt.app-check-your-answers__question.first-question#job_summary
+      %h4.govuk-heading-s= t('jobs.job_summary')
+    %dd.app-check-your-answers__answer.first-question= @vacancy.job_summary
+
+  .app-check-your-answers__contents
+    %dt.app-check-your-answers__question#about_school
+      %h4.govuk-heading-s= t('jobs.about_school', school: current_school.name)
+    %dd.app-check-your-answers__answer= @vacancy.about_school
--- a/app/views/hiring_staff/vacancies/job_summary/show.html.haml
+++ b/app/views/hiring_staff/vacancies/job_summary/show.html.haml
@@ -22,7 +22,7 @@
       = f.govuk_text_area :about_school,
         label: { text: t('helpers.fieldset.job_summary_form.about_school_html', school: current_school.name), size: 's' },
         hint_text: t('helpers.hint.job_summary_form.about_school'),
-        value: @vacancy.about_school.presence || (@vacancy.school.description.presence || ''),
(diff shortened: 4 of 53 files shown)


app/views/hiring_staff/vacancies/edit_vacancy_sections/_edit_job_summary.html.haml
```diff
@@ -0,0 +1,16 @@
+%h2.govuk-heading-m.app-task-list__section#job_summary_heading
+  %span.app-task-list__section-number 5.
+  = t('jobs.job_summary')
+  = link_to t('buttons.change'), school_job_job_summary_path(@vacancy.id), class: 'govuk-link change-button', 'aria-label': t('jobs.aria_labels.change_job_summary')
+
+%dl.app-check-your-answers.app-check-your-answers--short
+
+  .app-check-your-answers__contents
```
tatyree, line 8: In another PR @cesidio observed these are good candidates for abstracting.  I'm adding this note to forestall other change requests - my preferred path would be to start to replace some of this template code with a completely different workflow. 

This is sufficient for now-a partial or helper method would have to deal with at least two different logic branches (`.first-question`, etc.) and there are higher value things to be cleaning up right now.

lib/tasks/vacancies_to_update.yaml
```diff
@@ -26,7 +26,7 @@ vacancies:
     -
       slug: 'teacher-of-english-norham-high-school'
       job_title: 'Teacher of English'
-      job_description: "<p>Norham High school is seeking to appoint a Teacher of English who will join and strengthen the Department on a maternity cover contract. We are looking to appoint a dynamic and enthusiastic teacher with high expectations, who will be passionate about their subject, able to inspire and motivate young people, they must be committed to the highest standards of teaching and learning, through lesson planning, and high quality assessment.</p>
+      job_summary: "<p>Norham High school is seeking to appoint a Teacher of English who will join and strengthen the Department on a maternity cover contract. We are looking to appoint a dynamic and enthusiastic teacher with high expectations, who will be passionate about their subject, able to inspire and motivate young people, they must be committed to the highest standards of teaching and learning, through lesson planning, and high quality assessment.</p>
```
tatyree, line 29: Wrap these at 120 lines, please: 

https://yaml-multiline.info/

app/models/concerns/vacancy_job_summary_validations.rb
```diff
@@ -0,0 +1,17 @@
+module VacancyJobSummaryValidations
+  extend ActiveSupport::Concern
+  include ApplicationHelper
+
+  included do
+    validates :job_summary, presence: true
+    validates :about_school, presence: true
+  end
+
+  def job_summary=(value)
```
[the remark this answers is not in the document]
cpjmcquillan: > Can you refactor these please? There are two specific problems:
> 
> First, setters shouldn't have side effects. If we want to sanitize user input we should do it explicitly in a `before_save` hook on the model. This is too much magic.
> 
> Second, it's a concern containing validations. Setters shouldn't be in here.

Sure, I was just following previous convention for job_description. Shall I refactor everywhere this happens?

app/controllers/hiring_staff/vacancies/job_summary_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    (params[:job_summary_form] || params).permit(:job_summary, :about_school)
+                                         .merge(completed_step: current_step)
+  end
+
+  def redirect_to_next_step_if_save_and_continue
+    if params[:commit] == I18n.t('buttons.save_and_continue')
+      redirect_to school_job_review_path(@vacancy.id)
+    elsif params[:commit] == I18n.t('buttons.update_job')
+      redirect_to edit_school_job_path(@vacancy.id), success: I18n.t('messages.jobs.updated')
+    end
```
alexbowen, line 32: is this safe if neither condition is met?

app/models/concerns/vacancy_job_summary_validations.rb
```diff
@@ -0,0 +1,9 @@
+module VacancyJobSummaryValidations
```
tatyree, line 1: Overall I'm dubious about adding an additional class just to have two validations in two different models (one of which,`JobSummaryForm`, isn't really necessary either). Leave it for now, but let's start to think about getting rid of some of this layered-model-class complexity.

david-mears-dfe:
Is the reason we are adding a new column about_school (rather than use vacancy.school.description) so that it can vary by vacancy?

db/schema.rb
```diff
@@ -184,7 +184,7 @@
   create_table "vacancies", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "job_title", null: false
     t.string "slug", null: false
-    t.text "job_summary", null: false
+    t.text "job_summary"
```
david-mears-dfe, line 187: In what scenario will job_summary be null?

cpjmcquillan: When a user completes step 1 of the create a job journey (e.g every time a vacancy is created)

cpjmcquillan:
> Is the reason we are adding a new column about_school (rather than use vacancy.school.description) so that it can vary by vacancy?

Yes

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -33,7 +33,7 @@
       fill_in_pay_package_form_fields(draft_vacancy)
       click_on I18n.t('buttons.save_and_continue')
 
-      expect(page).to have_content(I18n.t('jobs.current_step', step: 3, total: 5))
+      expect(page).to have_content(I18n.t('jobs.current_step', step: 3, total: 6))
```
david-mears-dfe, line 36: Can we store `6` somewhere central so it doesn't have to be ceremonially changed in six places every day?

david-mears-dfe: Or dynamically get it somehow?

app/controllers/hiring_staff/vacancies/job_summary_controller.rb
cpjmcquillan: If we changed the way forms are submitted this would have to change, but unless that happens I think it's safe to assume one of these will be true

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -83,9 +83,36 @@
...
-      expect(page).to have_content(I18n.t('jobs.current_step', step: 4, total: 5))
+      expect(page).to have_content(I18n.t('jobs.current_step', step: 4, total: 6))
       expect(page).to have_content(I18n.t('jobs.application_details'))
     end
+
+    scenario 'incomplete job summary step' do
+      visit edit_school_job_path(id: draft_vacancy.id)
+
+      draft_vacancy.salary = 'Pay scale 1 to Pay scale 2'
+      draft_vacancy.benefits = 'Gym, health insurance'
```
david-mears-dfe, line 94: This couple of lines occurs about 5 times in this file. Can they be DRYed away?

alexbowen:
massive pull request, maybe unavoidable dunno, but maybe a code smell that tasks could be broken down more? tests look comprehensive , good work

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
david-mears-dfe: Same comment for `spec/features/hiring_staff_can_publish_a_vacancy_spec.rb`

david-mears-dfe: If it's a constant that can be accessed by the view too, what's the best place for that? I don't know.

alexbowen:
not specifically about this PR, but for discussion, do we have a team GIT methodolgy/workflow/approach - was going to bring this up in retro. for instance in this PR and git history ive seen overall, a bit of interactive rebasing to merge or fixup some commits might be a good idea as its quite verbose

app/views/hiring_staff/vacancies/job_summary/show.html.haml
```diff
@@ -0,0 +1,32 @@
...
+      = f.govuk_text_area :job_summary,
+        label: { text: t('helpers.fieldset.job_summary_form.job_summary_html'), size: 's' },
+        hint_text: t('helpers.hint.job_summary_form.job_summary'),
+        rows: 10,
+        required: true
+
+      = f.govuk_text_area :about_school,
+        label: { text: t('helpers.fieldset.job_summary_form.about_school_html', school: current_school.name), size: 's' },
+        hint_text: t('helpers.hint.job_summary_form.about_school'),
+        value: @vacancy.about_school.presence || (@vacancy.school.description.presence || ''),
```
tatyree, line 25: Couldn't `(@vacancy.school.description.presence || '')` be `@vacancy.school.description.presence.to_s`?

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
david-mears-dfe: Or maybe _tests_ shouldn't use constants, so that if the constant is wrong we get to find out by running the test. Sorry for all the thinking out loud here.

PR #1477, opened 2020-04-06
Remove code to write to Google Spreadsheets

cesidio:
This PR removes all the code, specs, tasks involved in putting data into Google Sheets. We now use Google BigQuery for that.


david-mears-dfe:
Where did the PR description go?

PR #1484, opened 2020-04-08
Redirect to DfE Sign-in when signing out

cesidio:
With this fix we redirect to DfE Sign-in for signing out, effectively destroying the session and logging out the user.

### DIFF
--- a/app/controllers/hiring_staff/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sessions_controller.rb
@@ -5,11 +5,16 @@ class HiringStaff::SessionsController < HiringStaff::BaseController
   skip_before_action :check_terms_and_conditions, only: %i[destroy]
 
   def destroy
-    session.destroy
-    redirect_to root_path, notice: I18n.t('messages.access.signed_out')
+    redirect_to dsi_logout_url
   end
 
   private
+
+  def dsi_logout_url
+    url = URI.parse("#{ENV['DFE_SIGN_IN_ISSUER']}/session/end")
+    url.query = { post_logout_redirect_uri: auth_dfe_signout_url, id_token_hint: session[:id_token] }.to_query
+    url.to_s
+  end
 
   def redirect_to_dfe_sign_in
     # This is defined by the class name of our Omniauth strategy
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -2,8 +2,8 @@ class HiringStaff::SignIn::Dfe::SessionsController < HiringStaff::BaseController
   include SignInAuditConcerns
 
   skip_before_action :check_session, only: %i[create new]
-  skip_before_action :check_terms_and_conditions, only: %i[create new]
-  skip_before_action :verify_authenticity_token, only: %i[create new]
+  skip_before_action :check_terms_and_conditions, only: %i[create new destroy]
+  skip_before_action :verify_authenticity_token, only: %i[create new destroy]
 
   def new
     # This is defined by the class name of our Omniauth strategy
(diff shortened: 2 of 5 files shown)


spec/features/hiring_staff_can_sign_out_spec.rb
```diff
@@ -8,6 +8,7 @@
     visit root_path
 
     click_on(I18n.t('nav.sign_out'))
+    visit auth_dfe_signout_path
```
david-mears-dfe, line 11: In both specs, the `click_on` should cause the `visit`, right? We don't want the user to need to type in the `auth_dfe_signout_path` themselves.

david-mears-dfe: I've now made this change - see next commit.

app/controllers/hiring_staff/sessions_controller.rb
```diff
@@ -5,12 +5,20 @@ class HiringStaff::SessionsController < HiringStaff::BaseController
...
   end
 
   private
 
+  def logout_url
+    url = URI.parse(ENV['DFE_SIGN_IN_ISSUER'] + '/session/end')
+    url.query = {
+      post_logout_redirect_uri: auth_dfe_signout_url,
+      id_token_hint: session[:id_token]
+    }.to_query
```
david-mears-dfe, line 18: Could we move the `session.destroy` into this method? We want to be sure the session is destroyed, but we need to use some of the information in the session first to generate logout_url.

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -17,6 +17,11 @@ def create
     perform_dfe_sign_in_authorisation
   end
 
+  def destroy
```
david-mears-dfe, line 20: If you put a `binding.pry` in you will see that this method is never run, correct?

app/controllers/hiring_staff/sessions_controller.rb
david-mears-dfe: I tested this idea. If you do this, then the method `destroy` on the other controller app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb never gets run, and signing out redirects you to the wrong page: /identifications/new.

So some part of the session must be necessary to do one or both of those. Session keys:
`["_csrf_token", "omniauth.nonce", "session_id", "urn", "multiple_schools", "id_token"]`

david-mears-dfe:
Bug:

If you sign out and are redirected to the homepage, and then click 'List a teaching job', the method below is run:

```ruby
def redirect_signed_in_users
    return redirect_to school_path if session.key?(:urn)
end
```

david-mears-dfe:
Just to record this somewhere public, DfE Sign In let us know that "these are the configured logout redirect urls for teaching vacancies":

```
https://localhost:3000/auth/dfe/signout
https://teaching-vacancies-staging.london.cloudapps.digital/auth/dfe/signout
https://tvs.edge.dxw.net/auth/dfe/signout
https://tvs.edge.dxw.net:443/auth/dfe/signout
https://tvs.staging.dxw.net/auth/dfe/signout
https://tvs.staging.dxw.net:443/auth/dfe/signout
https://tvs.testing.dxw.net/auth/dfe/signout
https://tvs.testing.dxw.net:443/auth/dfe/signout
```

david-mears-dfe:
Currently looking at a bug where sometimes a signed in user can only see `Page not found` pages.

david-mears-dfe:
> Currently looking at a bug where sometimes a signed in user can only see `Page not found` pages.

I've been unable to recreate this error, so I assume it's to do with my own set-up (outside of docker). Approving now.

david-mears-dfe:
@cesidio says:

> Before merging, we need to make sure that DSI has whitelisted our new sign out URL for production environment, otherwise users will see a broken link back.

PR #1485, opened 2020-04-09
Bugfix/sign in flow

alexbowen:
no jira ticket - small fix to sign in template as workflow and links were incorrect


alexbowen:
needs rebase - on it

PR #1492, opened 2020-04-14
Fix bug that would raise error on text inputs including an ampersand

cpjmcquillan:
Currently if hiring staff enter an ampersand '&' in the job_title or salary text input fields, an incorrect error is returned stating that the field contains html inputs. 

image

**In this PR**
- Fix the bug in the field validation causing this
- Add a test to ensure hiring staff can use '&'

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,6 +2,10 @@ module ApplicationHelper
   include ActionView::Helpers::NumberHelper
   include ActionView::Helpers::SanitizeHelper
 
+  def sanitize(text, options = {})
+    super(text, options)&.gsub('&amp;', '&')
+  end
+
   def body_class
     auth_class = authenticated? ? 'hiring-staff' : ''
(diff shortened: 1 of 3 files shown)


alexbowen:
if im understanding this correctly, encodeURIComponent in javascript might be better option?

tatyree:
> if im understanding this correctly, encodeURIComponent in javascript might be better option?

I'd be curious to see what an up-to-date html validator makes of 'ambiguous' ampersands.  This feels like a 2012 problem to me.  If it turns out they still are, then I'm inclined to agree with @alexbowen.  However, we this fix is, in essence, just putting something back that came out in a previous commit, so let's run with that as path of least resistance.

alexbowen:
i was missing some context here and just spoke to connor to address that. Alarm bells were going off that unsanitized user input was going directly to DB, i see rails provides protection out of the box, but client and server side validation is best for user input. Use of formbuilder will address this

PR #1503, opened 2020-04-17
GOVUK Form Builder job specification refactor

cpjmcquillan:
This PR replaces simple_form with GOVUKDesignSystemFormBuilder for the job specification section of the create a job process

## Changes in this PR:
- Refactor the view to reduce code duplication
- Simplify the controller actions
- Implement GOVUKDesignSystemFormBuilder
- Update the controller to handle multipart date assignment and blank checkbox values introduced by GOVUKDesignSystemFormBuilder


spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
```diff
@@ -494,21 +494,6 @@
         verify_all_vacancy_details(vacancy)
       end
 
-      scenario 'enables the user to resolve cross-form errors' do
```
david-mears-dfe, line 497: Can't we keep this spec?

app/helpers/vacancies_helper.rb
```diff
@@ -72,7 +72,8 @@ def radius_filter_options
```
david-mears-dfe, line 72: What's going on in this commit?
`Update the job role options and subject options for GOVUKDesignSystemFormBuilder`
Your commit messages could afford to be more explicit about the 'why' sometimes - makes it easier to follow along at home / understand changes months and years in the future.

david-mears-dfe:
Thanks for doing this, it will make life a lot easier!

app/helpers/vacancies_helper.rb
cpjmcquillan: GOVUKDesignSystemFormBuilder doesn't implement any `include_blank` type functionality (partly I guess because selects are actively discouraged, so I added a blank subject to the list of options. 

OpenStructs because that's what the documentation recommended, but I've since found another way of solving the same issue, so will readdress this.

Also, noted re commits - will make a conscious effort to be more explicit

spec/features/hiring_staff_can_publish_a_vacancy_spec.rb
cpjmcquillan: I removed it because it's testing a scenario that wouldn't ever arise, and the feature specs are long enough already

PR #1505, opened 2020-04-20
Release notes

cpjmcquillan:
This PR implements an updates page on the teaching vacancies site.

Product will create an update file `_{title_separated_by_underscores}_{yyyy_mm_dd}.html.erb` that contains a short paragraph of text in `app/views/updates/update_files` and create a pull request via the Github web interface. 

**TODO**
- [ ] Sign off with Product
- [ ] Update content of updates

## Jira ticket
[TEVA-706](https://dfedigital.atlassian.net/browse/TEVA-706)

## Changes in this PR
- Add release notes page that dynamically creates a list of all release notes

## Screenshots
image


app/controllers/updates_controller.rb
```diff
@@ -0,0 +1,26 @@
+class UpdatesController < ApplicationController
+  def show
```
cesidio, line 2: This should be an index action, it is effectively showing an index of updates, it is more RESTful this way

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   root 'pages#home'
 
+  get 'updates' => 'updates#show'
```
cesidio, line 4: `resources :updates, only: %i[index]`

app/controllers/updates_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    update_paths = Dir.glob(Rails.root.join('app/views/updates/update_files/*').to_s)
+    @updates = update_file_paths_to_hash(update_paths).sort.reverse.to_h
+  end
+
+  def update_file_paths_to_hash(update_paths)
+    updates_by_date = {}
+    update_paths.each do |update_path|
+        path, name, date_array = process_update_path(update_path)
```
cesidio, line 10: indentation here is 4 spaces, I guess we need to set up rubocop to check these too

app/controllers/updates_controller.rb
```diff
@@ -0,0 +1,26 @@
+class UpdatesController < ApplicationController
+  def show
+    update_paths = Dir.glob(Rails.root.join('app/views/updates/update_files/*').to_s)
+    @updates = update_file_paths_to_hash(update_paths).sort.reverse.to_h
+  end
+
+  def update_file_paths_to_hash(update_paths)
```
cesidio, line 7: There is too much logic here in the controller, controllers are only responsible for checking params and invoking models/services for preparing the view. We can pair together to refactor it if you want.

app/controllers/updates_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    updates_by_date
+  end
+
+  private
+
+  def process_update_path(update_path)
+    path = update_path.split('/').last.split('.html').first[1..-1]
+    name = update_path.split('/').last.split('.html').first.split('_')[1..-4].join(' ').humanize
+    date_array = update_path.split('/').last.split('.html').first.split('_').last(3).map(&:to_i)
+    return path, name, date_array
```
cesidio, line 24: no need to return the last statement in ruby

app/controllers/updates_controller.rb
cpjmcquillan: Rubocop automatically did this somehow... but yes I agree

PR #1516, opened 2020-04-24
Create a job tweaks

cpjmcquillan:
This PR implements small tweaks in the create a job journey to reduce code complexity and bring all steps of the process more in line with one another 

## Changes in this PR:
- Fixed bug that meant the update action in the job specification was only used for published vacancies
- Refactored the supporting documents controller
- Refactored redirect function to remove code duplication
- Add a page title helper to remove code duplication

### DIFF
--- a/app/controllers/hiring_staff/vacancies/pay_package_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/pay_package_controller.rb
@@ -11,7 +11,7 @@ class HiringStaff::Vacancies::PayPackageController < HiringStaff::Vacancies::App
     if @pay_package_form.valid?
       update_vacancy(pay_package_form_params, @vacancy)
       update_google_index(@vacancy) if @vacancy.listed?
-      return redirect_to_next_step_if_save_and_continue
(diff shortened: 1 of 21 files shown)


app/controllers/hiring_staff/vacancies/pay_package_controller.rb
```diff
@@ -20,18 +20,10 @@ def update
   private
 
   def pay_package_form_params
-    (params[:pay_package_form] || params).permit(:salary, :benefits).merge(completed_step: current_step)
+    params.require(:pay_package_form).permit(:salary, :benefits).merge(completed_step: current_step)
```
cesidio, line 23: much better!

app/helpers/vacancies_helper.rb
```diff
@@ -113,4 +113,14 @@ def new_sections(vacancy)
     sections << 'supporting_documents' unless vacancy.supporting_documents
     sections
   end
+
+  def page_title_prefix(vacancy, form_object, page_heading)
```
cesidio, line 117: 👍 nice one!

PR #1529, opened 2020-04-30
Remove id from supporting subject attributes in index

david-mears-dfe:
We were seeing long meaningless id strings in records in the Algolia index, for supporting subjects. We can get rid of these and make them consistent with the 'subject' attribute at no cost.

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -33,11 +33,16 @@ class Vacancy < ApplicationRecord
 
   include AlgoliaSearch
 
+  # rubocop:disable Metrics/BlockLength
   algoliasearch disable_indexing: !Rails.env.production? do
-    attributes :first_supporting_subject, :job_roles, :job_title, :salary, :second_supporting_subject, :working_patterns
+    attributes :job_roles, :job_title, :salary, :working_patterns
 
     attribute :expiry_date do
       convert_date_to_unix_time(self.expires_on)
+    end
+
+    attribute :first_supporting_subject do
+      self.first_supporting_subject&.name
     end
 
     attribute :last_updated_at do
@@ -72,6 +77,10 @@ class Vacancy < ApplicationRecord
         town: self.school.town }
     end
 
+    attribute :second_supporting_subject do
+      self.second_supporting_subject&.name
+    end
+
     attribute :start_date do
       convert_date_to_unix_time(self.starts_on)
     end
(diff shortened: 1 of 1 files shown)


app/models/vacancy.rb
```diff
@@ -33,13 +33,18 @@ class Vacancy < ApplicationRecord
 
   include AlgoliaSearch
 
+  # rubocop:disable Metrics/BlockLength
   algoliasearch disable_indexing: !Rails.env.production? do
-    attributes :first_supporting_subject, :job_roles, :job_title, :salary, :second_supporting_subject, :working_patterns
```
tatyree, line 37: I don't actually see `objectID` anywhere in here?

tatyree: Screenshot 2020-04-30 at 13 42 41
Are you talking about this?

tatyree: https://www.algolia.com/doc/api-client/methods/indexing/#the-object-id

david-mears-dfe: That is a good objectID to have. I want to get rid of extraneous objectIDs.

We currently get an objectID for each Vacancy record and that's fine and should stay that way. Algolia seems to be setting it to `Vacancy.id`.

Example record:

```
{
      "first_supporting_subject": {
        "id": "45eb4f2d-67fd-49d2-b039-8f9b746b75ff",
        "name": "Maths"
      },
      "job_roles": [
        "Suitable for NQTs"
      ],
      "job_title": "Mathematics Teacher ",
      "salary": "Main pay range 1 to Upper pay range 3, £25,543 to £41,635",
      "second_supporting_subject": null,
      "working_patterns": [
        "full_time",
        "part_time"
      ],
      "expiry_date": 1587726000,
      "last_updated_at": 1586224842,
      "listing_status": "published",
      "newly_qualified_teacher_status": true,
      "permalink": "mathematics-teacher-becket-keys-church-of-england-free-school",
      "publication_date": 1585911600,
      "school": {
        "name": "Becket Keys Church of England Free School",
        "address": "Sawyers Hall Lane",
        "county": "Essex",
        "local_authority": "Essex",
        "phase": "secondary",
        "postcode": "CM15 9DA",
        "region": "East of England",
        "town": "Brentwood"
      },
      "start_date": 1598958000,
      "subject": "Maths",
      "_geoloc": {
        "lat": 51.6271124337987,
        "lng": 0.302248497586538
      },
      "objectID": "fff7b992-3066-4818-87d1-9e03fadda4e0"
```

I want to make the supporting subject attributes consistent with `subject` in that they should be a string, not a json containing a name and an objectID.

david-mears-dfe: Just realised my confusion: I have been calling `id`s `ObjectID`s in my description and comment.

PR #1532, opened 2020-05-01
Index expiry_time for sorting

david-mears-dfe:
## Changes in this PR:

Add expiry_time to the Algolia index so that we can sort Algolia-side.

These changes were tested by indexing a Vacancy with nil expiry_time and a Vacancy with a real expiry_time.

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -40,9 +40,12 @@ class Vacancy < ApplicationRecord
       convert_date_to_unix_time(self.expires_on)
     end
 
+    attribute :expiry_time do
+      convert_time_to_unix_time(self.expiry_time)
+    end
+
     attribute :last_updated_at do
-      # Convert from ActiveSupport::TimeWithZone object to Unix time
-      self.updated_at.to_i
+      convert_time_to_unix_time(self.updated_at)
     end
 
     attribute :listing_status do
@@ -319,6 +322,13 @@ class Vacancy < ApplicationRecord
     return nil if date == nil
     # Convert to unix time via DateTime object in order to use correct time zone
     Time.zone.at(date.to_time).to_datetime.midday.to_i
+  end
+
+  def convert_time_to_unix_time(time)
+    # nil.to_i returns 0 in unix time (midnight 1970), so if time is nil we should return nil.
+    return nil if time == nil
+    # Convert from ActiveSupport::TimeWithZone object to Unix time
+    time.to_i
   end
 
   def slug_candidates


app/models/vacancy.rb
```diff
@@ -321,6 +324,13 @@ def convert_date_to_unix_time(date)
     Time.zone.at(date.to_time).to_datetime.midday.to_i
   end
 
+  def convert_time_to_unix_time(time)
+    # nil.to_i returns 0 in unix time (midnight 1970), so if time is nil we should return nil.
+    return nil if time == nil
```
ddippolito, line 329: Advice here: 

It's always recommended not to return `nil` directly. If you're returning `nil` there's probably something else higher up that we're missing.

In this case, we want to somehow say that vacancies might not have an expiry time.
In that case, I'd set a default value on the attribute `expiry_time` to be midnight of the expiry date.

app/models/vacancy.rb
```diff
@@ -40,9 +40,12 @@ class Vacancy < ApplicationRecord
       convert_date_to_unix_time(self.expires_on)
     end
 
+    attribute :expiry_time do
+      convert_time_to_unix_time(self.expiry_time)
```
tatyree, line 44: You don't need a separate method.  Just do:

```
time.nil? ? time : time.to_i
```

app/models/vacancy.rb
cpjmcquillan: Vacancies might not have an expiry date either (a vacancy is created in the database, and could be indexed, before all the required data to list a vacancy has been entered by a user)

tatyree: I believe that's a DateTime object, @ddippolito so combining our two comments it could just be:

```
time.nil? ? DateTime.now.to_i : time.to_i
```

That said, Alogolia needs in seconds or milliseconds since the epoch for sorting purposes, so zero would be fine under the circumstances.

david-mears-dfe: @cpjmcquillan is right, but we could still account for the case where we do have expiry_date but no expiry_time.

tatyree: Might be worth renaming that attribute, @david-mears-dfe to something like `expiry_time_epoch` or suchlike and sending the 'real' attribute as a string. That way the record can be sorted by time and the client doesn't need to do anything to covert the sorting timestamp.

app/models/vacancy.rb
david-mears-dfe: How about:

```ruby
self.expiry_time.nil? ? self.expiry_time : convert_date_to_unix_time(self.expires_on)
```

- accounting for the case where we don't know `expiry_time` but we do have `expires_on` (date).

app/models/vacancy.rb
cpjmcquillan: @tatyree I thought sorting was achieved through replicas of the index?

tatyree: @cpjmcquillan I think we're talking at cross purposes here. 

"Note that this attribute needs to be a numeric value for Algolia to be able to sort on it."

From

https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/sort-an-index-by-date/#configuring-the-index

david-mears-dfe:
Replaced by #1541

PR #1542, opened 2020-05-11
Update PULL_REQUEST_TEMPLATE

david-mears-dfe:
Adds a time-saving template url for Jira link and a prompt to request specific feedback.

### DIFF
--- a/PULL_REQUEST_TEMPLATE
+++ b/PULL_REQUEST_TEMPLATE
@@ -3,9 +3,15 @@ Pull Request template
 - Fill in or remove sections as necessary
 - Delete this section before submitting your PR
 
-## Jira ticket URL:
+## Jira ticket URL
+
+- Just add the ticket number to the end:
+
+https://dfedigital.atlassian.net/browse/TEVA-
 
 ## Changes in this PR:
+
+- Is there anything specific you want feedback on?
 
 ## Screenshots of UI changes:
 


ddippolito:
That's great 👏 

Can I propose another section about guidance for reviewers?
It'd be nice for chunky pull requests to include some guidance in how to review. 
Something like:

`How should you review this PR `

Examples might include:

* Start from this file and move to ....
* Commit by commit
* Start with the test and then move to code ...

Thoughts?

david-mears-dfe:
> 🍰

Thank you for this lovely slice of cake.

tatyree:
> That's great 👏
> 
> Can I propose another section about guidance for reviewers?
> It'd be nice for chunky pull requests to include some guidance in how to review.
> Something like:
> 
> `How should you review this PR `
> 
> Examples might include:
> 
> * Start from this file and move to ....
> * Commit by commit
> * Start with the test and then move to code ...
> 
> Thoughts?

I've got mixed feelings about this. It sounds like a good idea, but I suspect that the complexity of the current codebase was frustrate efforts to give sensible guidance. I'm also reluctant to allow the requestor to frame the review at any meaningful level. I've seen too many examples of things getting missed exactly because the review was framed in that way. That said, these have all been of the 'let me talk you through it' variety. 

What does everyone else think?

PR #1544, opened 2020-05-12
Update vacancy index for sorting

cpjmcquillan:
- Add replicas for sorting vacancies
- Add attributes for faceting for legacy alerts filtering

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -102,6 +102,24 @@ class Vacancy < ApplicationRecord
     end
 
     geoloc :lat, :lng
+
+    attributesForFaceting [:job_roles, :working_patterns, :school]
+
+    add_replica 'Vacancy_publish_on_desc', inherit: true do
+      ranking ['desc(publication_date_timestamp)']
+    end
+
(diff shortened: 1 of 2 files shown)


app/models/vacancy.rb
```diff
@@ -102,6 +102,24 @@ class Vacancy < ApplicationRecord
     end
 
     geoloc :lat, :lng
+
+    attributesForFaceting [:job_roles, :working_patterns, :school]
```
david-mears-dfe, line 106: We can configure these in the dashboard. Listing the attributes there has the advantage of being more easily configurable.

https://www.algolia.com/apps/QM2YE0HRBW/explorer/configuration/Vacancy/facets

Screenshot 2020-05-12 at 10 40 43

cpjmcquillan: Part of the thinking here was that we don't want someone to un-configure this by mistake

david-mears-dfe: Is it going to need to change very often (more than every couple of months)? If not, seems fine to have it here.

PR #1549, opened 2020-05-14
Remove algolia fall-back environment variables

david-mears-dfe:
The fall-back variables were introduced because when we used AWS as our
hosting solution, some instances did not load all the environment variables.

The problem seems to have gone away since GovUK PaaS migration.

We need all instances to have access to the real environment variables,
so that they can communicate with Algolia.

### DIFF
--- a/config/initializers/algoliasearch.rb
+++ b/config/initializers/algoliasearch.rb
@@ -1,18 +1,4 @@
-def application_id
-  ENV.fetch('ALGOLIA_APP_ID')
-rescue KeyError => e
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
Interestingly, the problem has perhaps not gone away. From the test run:

```
 Set up test database7s
/home/runner/work/teacher-vacancy-service/teacher-vacancy-service/vendor/bundle/gems/railties-5.2.4/lib/rails/initializable.rb:50:in `tsort_each_child'
Run bin/rails db:create db:schema:load
rails aborted!
KeyError: key not found: "ALGOLIA_APP_ID"
```

PR #1550, opened 2020-05-14
Remove Algolia fall-back environment variables

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-713

## Changes in this PR:

The fall-back variables were introduced because when we used AWS as our
hosting solution, some instances did not load all the environment variables.

We need all instances to have access to the real environment variables,
so that they can communicate with Algolia.

I am attempting to have the github workflow know about the environment variables so that the test run can pass.

### DIFF
--- a/config/initializers/algoliasearch.rb
+++ b/config/initializers/algoliasearch.rb
@@ -1,18 +1,4 @@
-def application_id
-  ENV.fetch('ALGOLIA_APP_ID')
-rescue KeyError => e
-  Rails.logger.error("Could not find environment variable. Defaulting to 'fake_algolia_app_id'. #{e.message}")
-  'fake_algolia_app_id'
-end
-
-def api_key
-  ENV.fetch('ALGOLIA_WRITE_API_KEY')
-rescue KeyError => e
-  Rails.logger.error("Could not find environment variable. Defaulting to 'fake_algolia_write_api_key'. #{e.message}")
-  'fake_algolia_write_api_key'
-end
-
 AlgoliaSearch.configuration = {
-  application_id: application_id,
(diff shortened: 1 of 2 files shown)


config/initializers/algoliasearch.rb
```diff
@@ -1,18 +1,4 @@
...
-  ENV.fetch('ALGOLIA_WRITE_API_KEY')
-rescue KeyError => e
-  Rails.logger.error("Could not find environment variable. Defaulting to 'fake_algolia_write_api_key'. #{e.message}")
-  'fake_algolia_write_api_key'
-end
-
 AlgoliaSearch.configuration = {
-  application_id: application_id,
-  api_key: api_key
+  application_id: ENV.fetch('ALGOLIA_APP_ID'),
```
cpjmcquillan, line 2: Might it instead be useful to only use the configuration when the variables are set? I don't think we don't need them to be there for any particular reason, do we?

david-mears-dfe: We shouldn't need them, but I suspect that the Vacancy model will throw a fit if we give it no config and then try to `include AlgoliaSearch`; and I'm uncomfortable with including fall-back variables in the config in case a production environment tries to use them, and we miss out on indexing some records.

I will try it.

david-mears-dfe:
Attempt number 3 worked, yay!

config/initializers/algoliasearch.rb
david-mears-dfe: It's as I expected:

```
Algolia::AlgoliaError:
        API not initialized
```

fails about a third of the specs.

PR #1555, opened 2020-05-14
Restrict documents that can be uploaded

cpjmcquillan:
Restrict content-types accepted by the file input and implement server side check on content-type.

Error message has been signed off by Content.

## Jira ticket URL
[TEVA-697](https://dfedigital.atlassian.net/browse/TEVA-697)

## Changes in this PR:
- Accept only .doc, .xls, .ppt, .pdf, .jpeg and .mp4 file types in the file input so users can't select other file types
- Restrict this server side as well in case an invalid file type was able to be submitted

## Screenshots of UI changes:
image

### DIFF
--- a/app/views/hiring_staff/vacancies/documents/show.html.haml
+++ b/app/views/hiring_staff/vacancies/documents/show.html.haml
@@ -13,6 +13,7 @@
 
       = f.govuk_file_field :documents,
         label: { text: t('jobs.upload_file'), size: 's' },
+        accept: '.doc, .xls, .ppt, .pdf, image/jpeg, image/png, video/mp4',
         multiple: true,
         enctype: 'multipart/form-data' do
 
--- a/app/views/hiring_staff/vacancies/documents/show.html.haml
+++ b/app/views/hiring_staff/vacancies/documents/show.html.haml
@@ -13,7 +13,7 @@
 
       = f.govuk_file_field :documents,
         label: { text: t('jobs.upload_file'), size: 's' },
-        accept: '.doc, .xls, .ppt, .pdf, image/jpeg, image/png, video/mp4',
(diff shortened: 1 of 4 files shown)


app/views/hiring_staff/vacancies/documents/show.html.haml
```diff
@@ -13,6 +13,7 @@
 
       = f.govuk_file_field :documents,
         label: { text: t('jobs.upload_file'), size: 's' },
+        accept: '.doc, .xls, .ppt, .pdf, image/jpeg, image/png, video/mp4',
```
tatyree, line 16: Shouldn't this have `.docx` and `.xlsx` as well (is there a `.pptx`?).

config/locales/en.yml
```diff
@@ -265,6 +265,7 @@ en:
     file_google_error_message: '%{filename} could not be uploaded - try again'
     file_input_error_message: 'The selected file(s) could not be uploaded'
     file_size_error_message: '%{filename} must be smaller than %{size_limit}'
+    file_type_error_message: '%{filename} must be a valid file type'
```
david-mears-dfe, line 268: Could we list the valid filetypes here?

app/views/hiring_staff/vacancies/documents/show.html.haml
cpjmcquillan: There is, and I asked the question but no response. Will add now 👍

config/locales/en.yml
cpjmcquillan: It could be a long list...

cpjmcquillan: Also, should add that an end user would probably never see this message.

PR #1558, opened 2020-05-14
Sign user out after (30) minutes of inactivity

david-mears-dfe:
# Draft

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-698

## Changes in this PR:

Clicks 'sign out' link after 30 minutes of no mouse movement, key presses, or clicks.

This JS is adapted from something I found online, and I am not a JS person (YET (GROWTH MINDSET)), so let me know what is bad JS style.

We don't have any JS tests at the moment, so is it OK for this code to be untested, or should we reintroduce JS tests, or should this be totally reimplemented in Ruby?

I tested this by reducing the inactivity timeout to 6 seconds. When there is no sign_out link visible, there is an error message in the console of not finding the link; but the user is already not signed in, if that is the case.

### DIFF
--- a/app/frontend/packs/application.js
+++ b/app/frontend/packs/application.js
@@ -16,6 +16,7 @@ import 'src/googleOptimise';
 import 'src/googleTagManagerUrlSnippet';
 import 'src/removeCommaFromNumber';
 import 'src/shareUrl';
+import 'src/signOutAfterInactivity.js';
 import 'src/sortJobList';
 import 'src/submitFeedback';
 import 'src/uploadDocuments';
--- a/app/frontend/src/signOutAfterInactivity.js
+++ b/app/frontend/src/signOutAfterInactivity.js
@@ -1,0 +1,28 @@
+var timeoutInMinutes = 30
+
+var timeoutId; 
+
+function resetTimer() { 
+    window.clearTimeout(timeoutId)
+    startTimer();
+}
+
+function clickSignOut() {
+    document.getElementById('sign_out').click();
+}
+
+function startTimer() { 
+    var timeoutInMilliseconds = timeoutInMinutes * 60000;
+    timeoutId = window.setTimeout(clickSignOut, timeoutInMilliseconds)
+}
+ 
+function setupTimers () {
+    document.addEventListener("mousemove", resetTimer, false);
+    document.addEventListener("mousedown", resetTimer, false);
+    document.addEventListener("keypress", resetTimer, false);
+    document.addEventListener("touchmove", resetTimer, false);
+     
+    startTimer();
+}
+
+setupTimers();
--- a/app/views/shared/_navigation.haml
+++ b/app/views/shared/_navigation.haml
@@ -3,7 +3,7 @@
     - if authenticated? && !ReadOnlyFeature.enabled?
       %li{ class: active_link_class(school_path) }= link_to t('nav.school_page_link'), school_path, class: 'govuk-header__link'
       %li{ class: active_link_class(root_path) }= link_to t('nav.jobseekers_index_link'), root_path, class: 'govuk-header__link'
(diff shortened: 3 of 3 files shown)


app/frontend/src/signOutAfterInactivity.js
```diff
@@ -0,0 +1,28 @@
+var timeoutInMinutes = 30
```
alexbowen, line 1: can we use let/const instead of var

app/frontend/src/signOutAfterInactivity.js
```diff
@@ -0,0 +1,28 @@
...
+
+var timeoutId; 
+
+function resetTimer() {
```
alexbowen, line 5: can we use arrow functions? e.g function blah() {} becomes const blah = () => {}

app/frontend/src/signOutAfterInactivity.js
```diff
@@ -0,0 +1,28 @@
...
+    window.clearTimeout(timeoutId)
+    startTimer();
+}
+
+function clickSignOut() {
+    document.getElementById('sign_out').click();
+}
+
+function startTimer() { 
+    var timeoutInMilliseconds = timeoutInMinutes * 60000;
```
alexbowen, line 15: whilst testing is limited you can do with this, pass `timeoutInMinutes` into this function as an argument and will make testing this possible, other wise the test will take 30mins to run and we would test on a much shorter timeout. i suggest we pair on a few tests here once algolia is merged (and test suite and linting is available). but for now id be happy that this isnt tested

app/frontend/packs/application.js
```diff
@@ -16,6 +16,7 @@ import 'src/googleOptimise';
 import 'src/googleTagManagerUrlSnippet';
 import 'src/removeCommaFromNumber';
 import 'src/shareUrl';
+import 'src/signOutAfterInactivity.js';
```
alexbowen, line 19: remove the .js for consistency

app/frontend/src/signOutAfterInactivity.js
alexbowen: just discussed with @tatyree - dont worry about this newer JS syntax for now

tatyree: @david-mears-dfe, just talked it over with @alexbowen and we're not 100% sure we're transpiling arrow functions (they don't work in IE and need to be changed by Babel), so just leave it as is for now.  Once we're sure, we'll have a clear up of them.

app/frontend/src/signOutAfterInactivity.js
alexbowen: just discussed with @tatyree - dont worry about this newer JS syntax for now

tatyree:
Good point, @cesidio.  As some things won't be coming back to the server anymore (client side searches), we may want to adapt this to ping an API endpoint on the server to update `last_activity_at` periodically.

app/frontend/src/signOutAfterInactivity.js
```diff
@@ -0,0 +1,28 @@
...
+
+var timeoutId; 
+
+function resetTimer() { 
+    window.clearTimeout(timeoutId)
+    startTimer();
+}
+
+function clickSignOut() {
+    document.getElementById('sign_out').click();
```
tatyree, line 11: Following on from @cesidio observation, I'd advise against simulating user interaction. Instead find where the session information is stored and delete that.  It is most-likely stored in a cookie, but it might also be in Local or Session storage.

alexbowen: i agree, yes good point @cesidio

cesidio:
> Good point, @cesidio. As some things won't be coming back to the server anymore (client side searches), we may want to adapt this to ping an API endpoint on the server to update `last_activity_at` periodically.

@tatyree search is not performed by authenticated user, so we do not actually need that. We only need to keep track of the activity that happens in the hiring staff area.

PR #1566, opened 2020-05-18
Fetch vacancy state and push to the dataLayer

cpjmcquillan:
This PR adds a vacancy_state attribute to the model and updates it appropriate. This state is pushed to the dataLayer for google tag manager every time the DOM is loaded.

**Todo**
- [x] Sign off with Performance (Steven)

## Jira ticket URL
[TEVA-787](https://dfedigital.atlassian.net/browse/TEVA-787)

## Changes in this PR:
- Add state to the vacancy model 
- Update state server-side when hiring staff create/review/edit vacancies
- Set the vacancy state in the dataLayer

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,7 @@ jobs:
         run: bundle exec rspec
 
   frontend-tests:
-    name: Run yarn test
+    name: Run jest
 
     runs-on: ubuntu-latest
 
--- a/app/frontend/src/addVacancyStateToDataLayer.js
+++ b/app/frontend/src/addVacancyStateToDataLayer.js
@@ -1,0 +1,5 @@
+document.addEventListener("DOMContentLoaded", function(e) {
+  const element = document.querySelector('.new_copy_vacancy_form') || document.body || {};
+  const dataset = element.dataset || {};
+  dataLayer.push({ vacancy_state: dataset.vacancyState });
+});
--- a/spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
+++ b/spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
@@ -11,153 +11,172 @@ RSpec.feature 'Hiring staff can edit a draft vacancy' do
 
   before do
     stub_hiring_staff_auth(urn: school.urn)
-    visit new_school_job_path
-    fill_in_job_specification_form_fields(vacancy)
-    click_on I18n.t('buttons.save_and_continue')
   end
 
-  context '#redirects_to' do
-    scenario 'incomplete pay package step' do
-      visit edit_school_job_path(id: draft_vacancy.id)
-
-      expect(page).to have_content(I18n.t('jobs.current_step', step: 2, total: 6))
-      within('h2.govuk-heading-l') do
-        expect(page).to have_content(I18n.t('jobs.pay_package'))
-      end
-    end
-
-    scenario 'incomplete supporting documents step' do
-      visit edit_school_job_path(id: draft_vacancy.id)
-
-      draft_vacancy.salary = 'Pay scale 1 to Pay scale 2'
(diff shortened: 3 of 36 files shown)


tatyree:
As per our conversation.  I think most of this can be simplified down to a script tag on the vacancy page that looks something like this in the Rails template: 

```ruby
<script>
document.addEventListener("DOMContentLoaded", function(e) {
  const vacancyStatus = '#{@vacancy.status}";
  if (vacancyStatus) datalayer.push({ vacancyStatus });
});
</script>
```

The tests don't need to be complicated, just make sure it doesn't wipe other entries in data layer when it pushes and that it pushes something.  look at @alexbowen's Jest tests for some examples.

app/frontend/src/addVacancyStateToDataLayer.js
```diff
@@ -0,0 +1,37 @@
...
+
+  getBodyVacancyDataset() {
+    const bodyVacancyDataset = document.body.dataset;
+    return bodyVacancyDataset
+  },
+
+  getBodyVacancyState() {
+    return this.getBodyVacancyDataset() && this.getBodyVacancyDataset().vacancyState;
+  },
+
```
alexbowen, line 22: 20 lines of code could be 6 as these should all be in the same function as pass in the selector as argument

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -11,153 +11,172 @@
...
+  context 'editing a complete draft vacancy' do
+    let(:vacancy) { create(:vacancy, :draft, school: school) }
 
-      fill_in 'application_details_form[application_link]', with: 'https://example.com'
-      click_on I18n.t('buttons.update_job')
+    scenario 'vacancy state is edit' do
+      visit school_job_review_path(vacancy.id, edit_draft: true)
 
-      expect(page).to have_content(I18n.t('messages.jobs.updated'))
+      expect(Vacancy.last.state).to eql('edit')
```
david-mears-dfe, line 175: Why `Vacancy.last` rather than `vacancy`?

.github/workflows/test.yml
```diff
@@ -107,5 +107,5 @@ jobs:
       - name: Install node.js dependencies
         run: yarn install
 
-      # - name: Run yarn test
-      #   run: yarn test
+      - name: Run jest
+        run: npm run js:test
```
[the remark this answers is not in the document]
tatyree: Let's leave this for now, I'll stick a ticket in to fix it later.

app/frontend/src/addVacancyStateToDataLayer.js
```diff
@@ -0,0 +1,15 @@
...
+    const element = document.querySelector('.new_copy_vacancy_form') || document.body
+    return element.dataset;
+  },
+
+  getObjectForDataLayer() {
+    const vacancyDataset = this.getVacancyDataSet();
+    return { vacancy_state: vacancyDataset.vacancyState };
+  }
+}
+
```
tatyree, line 12: This is a very ruby way to do it. A more JS way would be:

```
export const default = document.addEventListener("DOMContentLoaded", function(e) {
  const element = document.querySelector('.new_copy_vacancy_form') || {};
  const dataset = element.dataset || {};
  const vacancy_state = dataset.vacancyState;
  dataLayer.push({ vacancy_state });
});
```
I'm pretty sure you don't need the `export const default = ` either, but I'm showing it in case you get console errors.

tatyree: I'm OK with it being rubish, but the extra function calls add unnecessary overhead and you don't need the `export` as all that does is expose the exported function(s) to the importing lib.  All the work gets done the moment the script gets hit.  I'll leave it up to you if you want to adapt this.  If you stay with this way, let me know and I'll put a snagging ticket in for later.

PR #1568, opened 2020-05-20
Sync staging from production database

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-773


.github/workflows/test.yml
```diff
@@ -1,6 +1,9 @@
 name: Test
 
-on: [push]
+on:
```
cesidio, line 3: This was to temporarily disable tests, you can revert this file back

PR #1571, opened 2020-05-21
Clean up Elasticsearch

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-668

## Changes in this PR:

Removed Elasticsearch from codebase.

I have no way to know whether this regresses the _job alerts_, because staging didn’t send me an email when I tried to set up a job alert _before_ I pushed my changes, let alone afterwards. Therefore, for that bit of certainty, we are relying on the specs to catch any regression. But since Elasticsearch was barely present in the codebase at all, it seems unlikely that this PR would regress job alerts.

Tested by pushing to staging and seeing whether it broke anything in an obvious way. Staging's search functionality already had several issues with it before I pushed.

Removed elasticsearch services on staging and dev CF_SPACEs.

The acceptance criteria mention removing 'libraries for frontend search'. I presume these are just the libraries that I removed (the ones with elasticsearch in the name)?


db/migrate/20180905101200_reindex_vacancies_after_changing_max_salary_type.rb
```diff
@@ -1,5 +0,0 @@
-class ReindexVacanciesAfterChangingMaxSalaryType < ActiveRecord::Migration[5.2]
```
cpjmcquillan, line 1: Should we be removing migrations?

david-mears-dfe: It's a fake migration which does not touch the database, but just runs a task. But your question still stands.
https://github.com/DFE-Digital/teacher-vacancy-service/commit/825e85c2ecffe8596734f7b09f6768bae7b78827

cpjmcquillan: Ok cool, I just wasn't sure how migrations work in ActiveRecord

alexbowen:
looks pretty thorough

db/migrate/20180905101200_reindex_vacancies_after_changing_max_salary_type.rb
tatyree: Given the way Rails tracks migrations, and the fact that they can be fiddly and difficult to maintain, a better solution would be to leave the migration in place but this:

```ruby
  def change
     #noop
     # <brief explaination>
  end
```

david-mears-dfe:
Hang on, I'm going to implement @tatyree 's migration comment

db/migrate/20180905101200_reindex_vacancies_after_changing_max_salary_type.rb
tatyree: Otherwise, migration versioning numbers may well get out of sync which can cause problems maintaining `schema.rb` and make it difficult to reconcile versions later.

PR #1574, opened 2020-05-22
Add subjects array to vacancy

cpjmcquillan:
## Jira ticket URL
[TEVA-631](https://dfedigital.atlassian.net/browse/TEVA-631)

## Changes in this PR:
- Add a postgresql string array column for subjects to the vacancy table
- Add a rake task to convert existing subject fields to the new subjects field
- On vacancy save, update the new subjects field
- Index the subjects field in Algolia

## Next steps:
- Add fancy checkbox input and filter
- Update review page
- Update jobseeker page

### DIFF
--- a/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
@@ -55,12 +55,13 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
   def job_specification_form_params
     persist_nqt_job_role_to_nqt_attribute(:job_specification_form)
     strip_empty_checkboxes(:job_specification_form, [:working_patterns, :job_roles])
+    convert_subject_ids_to_subjects_array
     params.require(:job_specification_form)
           .permit(:state, :job_title,
                   :subject_id, :first_supporting_subject_id, :second_supporting_subject_id,
                   :starts_on, :ends_on,
                   :newly_qualified_teacher,
-                  working_patterns: [], job_roles: [])
+                  job_roles: [], working_patterns: [], subjects: [])
           .merge(completed_step: current_step)
   end
 
@@ -76,5 +77,13 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
 
   def next_step
     school_job_pay_package_path(@vacancy&.id.present? ? @vacancy.id : session_vacancy_id)
+  end
+
+  def convert_subject_ids_to_subjects_array
+    subjects = params.require(:job_specification_form)
+                     .values_at(:subject_id, :first_supporting_subject_id, :second_supporting_subject_id)
+                     .reject(&:blank?)
+                     .map { |subject_id| Subject.find(subject_id).name }
+    params[:job_specification_form][:subjects] = subjects
   end
 end
--- a/lib/tasks/convert_subject_fields_to_subject_array.rake
+++ b/lib/tasks/convert_subject_fields_to_subject_array.rake
@@ -1,0 +1,39 @@
+namespace :data do
+  desc 'Convert subject fields to string array for vacancies without the new subject field'
+  namespace :convert_subject_fields do
+    task vacancies: :environment do
+      updated_count = 0
+      should_be_updated_count = Vacancy.where("subjects is NULL or subjects = '{}'").count
+
+      Rails.logger.info(
+        "Conversion of subject fields to string array has been started for #{should_be_updated_count} vacancies"
+      )
+      Rollbar.log(
+        :info,
+        "Conversion of subject fields to string array has been started for #{should_be_updated_count} vacancies"
+      )
+
+      Vacancy.where("subjects is NULL or subjects = '{}'").in_batches(of: 100).each_record do |vacancy|
+        subject_name = vacancy.subject.present? ? vacancy.subject.name : nil
+        first_subject_name = vacancy.first_supporting_subject.present? ? vacancy.first_supporting_subject.name : nil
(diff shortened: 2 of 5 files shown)


app/controllers/hiring_staff/vacancies/job_specification_controller.rb
```diff
@@ -55,12 +55,13 @@ def set_up_job_specification_form
   def job_specification_form_params
     persist_nqt_job_role_to_nqt_attribute(:job_specification_form)
     strip_empty_checkboxes(:job_specification_form, [:working_patterns, :job_roles])
+    convert_subject_ids_to_subjects_array
     params.require(:job_specification_form)
           .permit(:job_title,
                   :subject_id, :first_supporting_subject_id, :second_supporting_subject_id,
```
david-mears-dfe, line 61: Does `convert_subject_ids_to_subjects_array` make this line redundant?

cesidio:
Aren't we going to lose the fact that before we had main subject and supporting subjects, and if I search for a subject, the fact that is main subject, instead of first or second supporting subject should have more weight in the results. With this all subjects will be the same. Is it intended?

david-mears-dfe:
> Aren't we going to lose the fact that before we had main subject and supporting subjects, and if I search for a subject, the fact that is main subject, instead of first or second supporting subject should have more weight in the results. With this all subjects will be the same. Is it intended?

I have a dim recollection of a setting in Algolia that gives more weight to results where the search query is earlier in the field. Will look for it.

david-mears-dfe:
> > Aren't we going to lose the fact that before we had main subject and supporting subjects, and if I search for a subject, the fact that is main subject, instead of first or second supporting subject should have more weight in the results. With this all subjects will be the same. Is it intended?
> 
> I have a dim recollection of a setting in Algolia that gives more weight to results where the search query is earlier in the field. Will look for it.

"[There is also an importance to the order of the matches within the attribute itself. By default, records whose matched words are closer to the beginning of a given attribute will be ranked higher. For example, words in position 2 of an attribute are ranked higher than words in position 5. Otherwise, the position of the word is not taken into account.](https://www.algolia.com/doc/guides/managing-results/relevance-overview/in-depth/ranking-criteria/)"

app/controllers/hiring_staff/vacancies/job_specification_controller.rb
cpjmcquillan: No, this PR is simply to implement the field on the DB and ensure the field gets populated. Nothing is exposed in any view to either hiring staff or jobseekers. As outlined in the next steps, subsequent PRs will deal with implementing a new subjects field in the form and displaying this to users.

tatyree:
It is also worth noting that you cannot use Arrays for custom rankings in Algolia.  It only accepts primitive types.  My suggestion is to consider having an explicit ranking expressed in the select like (`subjects[0] == subject`, `subjects[1] == first_supporting_subject`, and so on).  Then, for custom ranking purposes iterate them into strings before indexing so you get `subject0`, `subject1`, etc. in the rank-able attributes.

lib/tasks/convert_subject_fields_to_subject_array.rake
```diff
@@ -0,0 +1,39 @@
...
+
+      Vacancy.where("subjects is NULL or subjects = '{}'").in_batches(of: 100).each_record do |vacancy|
+        subject_name = vacancy.subject.present? ? vacancy.subject.name : nil
+        first_subject_name = vacancy.first_supporting_subject.present? ? vacancy.first_supporting_subject.name : nil
+        second_subject_name = vacancy.second_supporting_subject.present? ? vacancy.second_supporting_subject.name : nil
+
+        subjects = [subject_name, first_subject_name, second_subject_name].reject(&:blank?)
+
+        # rubocop:disable Rails/SkipsModelValidations
+        vacancy.update_columns(subjects: subjects)
```
david-mears-dfe, line 24: Rubocop said `Avoid using update_columns because it skips validations.` If you want to skip validations you can use `.update`.

cpjmcquillan:
> Could you either include a change to `models/vacancy.rb` with this PR or create a ticket for us to make changes in that file to index the new subjects attribute?

This PR doesn't close the ticket I'm working on, so I don't see the need to make a new ticket or do this in this PR

lib/tasks/convert_subject_fields_to_subject_array.rake
cpjmcquillan: `.update_columns` doesn't update the timestamps

lib/tasks/convert_subject_fields_to_subject_array.rake
```diff
@@ -0,0 +1,39 @@
...
+        "Conversion of subject fields to string array has been started for #{should_be_updated_count} vacancies"
+      )
+      Rollbar.log(
+        :info,
+        "Conversion of subject fields to string array has been started for #{should_be_updated_count} vacancies"
+      )
+
+      Vacancy.where("subjects is NULL or subjects = '{}'").in_batches(of: 100).each_record do |vacancy|
+        subject_name = vacancy.subject.present? ? vacancy.subject.name : nil
+        first_subject_name = vacancy.first_supporting_subject.present? ? vacancy.first_supporting_subject.name : nil
```
cesidio, line 18: Here you can use the `presence` method, is more concise:
https://andycroll.com/ruby/use-the-presence-method

cpjmcquillan: Would this return the name or the object?

david-mears-dfe: `vacancy.subject.name.presence`

PR #1583, opened 2020-05-26
Recaptcha

tatyree:
This implements a [v3 recaptcha](https://developers.google.com/recaptcha/docs/v3) on the feedback form, which does not challenge users, but only scores them. 

https://dfedigital.atlassian.net/browse/TEVA-812

## TODO:

 - [ ] Chris to review on staging or dev

Will do the job alerts after this is merged.


config/initializers/recaptcha.rb
```diff
@@ -0,0 +1,4 @@
+Recaptcha.configure do |config|
+  config.site_key = ENV['RECAPTCHA_SITE_KEY']
```
cesidio, line 2: It is useful to add these ENV variables to the file `.env.example` so we know we need them.

spec/features/general_feedback_spec.rb
```diff
@@ -1,63 +0,0 @@
-require 'rails_helper'
```
cesidio, line 1: It is good that we have controller specs to test permutations faster and only in one place, but I would keep just the happy path of this feature spec as they are valuable.

db/schema.rb
```diff
@@ -224,9 +225,9 @@
     t.uuid "publisher_user_id"
     t.datetime "expiry_time"
     t.string "supporting_documents"
-    t.string "job_roles", array: true
```
cpjmcquillan, line 227: These sort of things keep springing up in schema changes

tatyree: It's one of the long-lived problems with Rails schemas unfortunately. `job_roles` needs to stay in, doesn't it?

cpjmcquillan: Yes definitely, just the order of fields in the schema always seems to jump around when we do migrations

david-mears-dfe:
:shipit:

PR #1586, opened 2020-05-27
New subjects field

cpjmcquillan:
This PR implements the frontend changes to add the new subjects input. 

**TODO**
- [x] Update css
- [x] Make JS filter functional in IE
- [x] Filter hint text as well as subject name
- [x] Handle subjects that aren't in the subject options
- [x] Move pre-checked subjects to top of the list
- [x] Run rake task on production to convert subject fields
- [x] Product/UX sign off

## Jira ticket URL
[TEVA-631](https://dfedigital.atlassian.net/browse/TEVA-631)

## Changes in this PR:
- Add new subjects input
- Add subject filter JS functionality 
- Update views

### Before
image

### After
image

## Next steps:
- Remove old subject fields

### DIFF
--- a/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
@@ -56,11 +56,9 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
 
   def job_specification_form_params
     persist_nqt_job_role_to_nqt_attribute(:job_specification_form)
-    strip_empty_checkboxes(:job_specification_form, [:working_patterns, :job_roles])
-    convert_subject_ids_to_subjects_array
+    strip_empty_checkboxes(:job_specification_form, [:working_patterns, :job_roles, :subjects])
     params.require(:job_specification_form)
           .permit(:state, :job_title,
-                  :subject_id, :first_supporting_subject_id, :second_supporting_subject_id,
                   :starts_on, :ends_on,
                   :newly_qualified_teacher,
                   job_roles: [], working_patterns: [], subjects: [])
@@ -79,15 +77,5 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
 
   def next_step
     school_job_pay_package_path(@vacancy&.id.present? ? @vacancy.id : session_vacancy_id)
-  end
-
-  def convert_subject_ids_to_subjects_array
-    subjects = params.require(:job_specification_form)
-                     .values_at(:subject_id, :first_supporting_subject_id, :second_supporting_subject_id)
-                     .reject(&:blank?)
-                     .map { |subject_id| get_subject_name(Subject.find(subject_id)) }
-                     .uniq
-                     .reject(&:blank?)
-    params[:job_specification_form][:subjects] = subjects
   end
 end
--- a/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/job_specification_controller.rb
@@ -32,6 +32,7 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
 
   def update
     if @job_specification_form.complete_and_valid?
+      remove_subject_fields(@vacancy) unless @vacancy.subjects.nil?
       update_vacancy(job_specification_form_params, @vacancy)
       update_google_index(@vacancy) if @vacancy.listed?
       return redirect_to_next_step_if_save_and_continue(@vacancy.id)
@@ -63,6 +64,12 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
                   :newly_qualified_teacher,
                   job_roles: [], working_patterns: [], subjects: [])
           .merge(completed_step: current_step)
+  end
+
+  def remove_subject_fields(vacancy)
+    @vacancy.subject = nil
+    @vacancy.first_supporting_subject = nil
+    @vacancy.second_supporting_subject = nil
   end
 
   def save_vacancy_without_validation
--- a/app/frontend/src/subjectsFilter.js
+++ b/app/frontend/src/subjectsFilter.js
@@ -1,0 +1,80 @@
+document.addEventListener('DOMContentLoaded', function() {
+  var subjectFilter, filteredSubjectNames, filteredSubjectHints;
+  var checkedSubjectElements = [];
+  var duplicateSubjectElements = [];
+
+  const subjectSearchInput = document.getElementById('job-specification-form-subject-search');
+  const subjectElements = document.querySelector('.checkboxes-with-scroll.subjects-options');
+
+  if (subjectElements) {
+    const subjectArray = [...subjectElements.children];
+    subjectArray.shift();
+    const subjectNames = subjectArray.map(ele => ele.children[1].textContent.toLowerCase());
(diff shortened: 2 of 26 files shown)


app/frontend/src/subjectsFilter.js
```diff
@@ -0,0 +1,62 @@
...
+        this.checked = false;
+        this.parentElement.style.display = 'none';
+      }
+    });
+  });
+
+  subjectSearchInput.addEventListener('input', function(v) {
+    subjectFilter = subjectSearchInput.value.toLowerCase();
+    filteredSubjectNames = subjectNames.filter(function(a) {
+      return a.includes(subjectFilter);
```
alexbowen, line 54: includes not supported in IE

app/frontend/src/subjectsFilter.js
```diff
@@ -0,0 +1,62 @@
...
+        duplicateElement.addEventListener('change', function() {
+          if (!this.checked) {
+            var originalElement = checkedSubjectElements.find(function(ele) {
+              return ele.children[1].textContent == duplicateElement.children[1].textContent;
+            });
+            originalElement.style.display = 'block';
+            this.remove();
+          }
+        });
+        subjectElements.prepend(duplicateElement);
```
alexbowen, line 44: again no IE support for prepend

app/frontend/src/subjectsFilter.js
```diff
@@ -0,0 +1,62 @@
...
+    filteredSubjectNames = subjectNames.filter(function(a) {
+      return a.includes(subjectFilter);
+    });
+
+    for (var i = 0; i < subjectArray.length; i++) {
+      var element = subjectArray[i];
+      displayElement(element) ? element.style.display = 'block' : element.style.display = 'none';
+    }
+  });
+});
```
alexbowen, line 62: in general I would be tempted to use jquery for DOM selection and because of IE support for some stuff. im 95% that we will carry on using a thing called sizzle which is the DOM selection library used in jquery, so this should mean little to no refactoring needed when jquery is removed

app/frontend/styles/components/checkboxes.scss
```diff
@@ -4,3 +4,41 @@
...
+
+  .govuk-checkboxes__item label{
+    font-weight: 600;
+  }
+}
+
+.checkboxes-with-scroll__filter {
+  position: relative;
+  background: #ffffff;
+  border-left: 5px solid #dee0e2;
```
alexbowen, line 25: should use a GDS color or its nearest equivalent here

cpjmcquillan: Yeah these need to be updated, all css is just copy and pasted from the prototype at the minute

app/presenters/vacancy_presenter.rb
```diff
@@ -163,4 +117,8 @@ def job_title_and_school
   def show_job_roles
     model.job_roles&.join(', ')
   end
+
+  def show_subjects
+    model.subjects&.join(', ')
+  end
 end
```
david-mears-dfe, line 124: Clearing out 10 methods all at once. Love it.

spec/factories/vacancies.rb
```diff
@@ -121,13 +123,5 @@
       expires_on { Faker::Time.between(from: Time.zone.today + 10.days, to: Time.zone.today + 20.days) }
       expiry_time { nil }
     end
-
-    trait :first_supporting_subject do
```
david-mears-dfe, line 125: Maybe we ought to keep these attributes in the factory for now, if we've still got these attributes on some records?

cpjmcquillan: Vacancies won't be created with these attributes anymore, so I don't think it makes sense to add them by default in the factory

app/helpers/vacancies_helper.rb
```diff
@@ -45,7 +45,7 @@ def uncapitalize_words(location_name)
 
   def new_sections(vacancy)
     sections = []
-    sections << 'job_role' unless vacancy.job_roles&.any?
```
david-mears-dfe, line 51: I wonder, do we have an agreed principle for when we stop showing the new-section notification? Is it after some time period or is it when the user hasn't seen the section before?

Am I misremembering, or is job_role actually a newer section than supporting_documents is?

app/helpers/vacancies_helper.rb
```diff
@@ -59,4 +59,11 @@ def page_title_prefix(vacancy, form_object, page_heading)
         'Error: ' : ''}#{page_heading} — #{t('jobs.create_a_job', school: current_school.name)}"
     end
   end
+
+  def missing_subjects?(vacancy)
+    legacy_subjects = [vacancy.subject,
+                       vacancy.first_supporting_subject,
+                       vacancy.second_supporting_subject].reject(&:blank?)
```
david-mears-dfe, line 66: It's much of a muchness, but I discovered the ruby method `.compact` yesterday, which rejects nil elements from an array. (`.blank?` captures a few more things that are unlikely to crop up here, e.g. `false`, and empty arrays.)

spec/helpers/vacancies_helper_spec.rb
```diff
@@ -39,7 +39,12 @@
 
     it 'should include job_role for legacy listings' do
```
david-mears-dfe, line 40: Update description of test to match change (job_role -> job_details)

spec/helpers/vacancies_helper_spec.rb
```diff
@@ -39,7 +39,12 @@
...
       allow(vacancy).to receive_message_chain(:job_roles, :any?).and_return(false)
-      expect(helper.new_sections(vacancy)).to include('job_role')
+      expect(helper.new_sections(vacancy)).to include('job_details')
+    end
+
+    it 'should include subjects for legacy listings' do
+      allow(helper).to receive(:missing_subjects?).with(vacancy).and_return(true)
+      expect(helper.new_sections(vacancy)).to include('job_details')
```
david-mears-dfe, line 47: This expectation doesn't seem to match the description, which mentions subjects?

app/controllers/hiring_staff/vacancies/job_specification_controller.rb
```diff
@@ -32,6 +32,7 @@ def create
 
   def update
     if @job_specification_form.complete_and_valid?
+      remove_subject_fields(@vacancy)
```
david-mears-dfe, line 35: This is a nice place for this method. I'm not sure how we know that the rake task for converting legacy subject fields to `subjects` array is run before this method is run, and so how we know that we aren't losing the legacy subject fields.

app/helpers/vacancies_helper.rb
cpjmcquillan: Indefinitely. There is a ticket for refining how this works I believe

app/views/vacancies/_job_details_section.html.haml
```diff
@@ -9,17 +9,12 @@
             = t('jobs.job_roles')
           %td.govuk-table__cell
             = @vacancy.show_job_roles
-      - if @vacancy.any_subjects?
+      - if @vacancy.subjects&.any?
         %tr.govuk-table__row
           %th.govuk-table__cell{ class: "govuk-!-font-weight-bold" }
-            = t('jobs.subject', count: @vacancy.subject_count)
+            = t('jobs.subject', count: @vacancy.subjects.count)
```
cesidio, line 15: maybe you want to use again the ruby safe operator like you did on line 12 `@vacancy.subjects&.count`?

PR #1588, opened 2020-05-27
Convert subject fields

cpjmcquillan:
This PR handles vacancies with subjects that are not in the new subject list.

**TODO**
- [x] Confirm subject synonyms for matching
- [x] Confirm new subjects list

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-631

## Changes in this PR
- Add get subject name service that matches legacy subjects to new subjects
- Update rake task
- Populate subjects field when vacancies are created
- Populate subjects field when vacancies are copied

## Next steps
- Update subjects field on all vacancies to nil
- Run rake task on all vacancies to convert subjects to the new subject field


app/models/vacancy.rb
```diff
@@ -27,6 +28,8 @@ class Vacancy < ApplicationRecord
     [I18n.t('jobs.sort_by.expiry_time.ascending'), 'expiry_time_asc']
   ]
 
+  SUBJECT_OPTIONS = Subjects::SUBJECT_OPTIONS.freeze
```
david-mears-dfe, line 31: `include Subjects` is a little terser and ensures that `SUBJECT_OPTIONS` is always referencing the same object.

app/models/subjects.rb
```diff
@@ -0,0 +1,51 @@
+module Subjects
+  SUBJECT_OPTIONS = [
```
cesidio, line 2: Maybe have these subjects in YAML format? Kind of what we have done with boroughs, cities, counties and regions

app/services/get_subject_name.rb
```diff
@@ -0,0 +1,25 @@
...
+    'Media Studies': 'Media studies',
+    'Physical Education': 'Physical education',
+    'Religious Studies': 'Religious education',
+    'General Science': 'Science'
+  }.stringify_keys
+
+  def get_subject_name(subject)
+    return nil if subject.blank?
+
+    subject_option_names = SUBJECT_OPTIONS.map { |subject_option| subject_option.first }
```
cesidio, line 19: Here you can use:

`SUBJECT_OPTIONS.map(&:first)`

https://borgs.cybrilla.com/tils/symbol-to-proc-in-ruby/

PR #1591, opened 2020-05-29
Change all 1-element arrays to strings

tatyree:
Change all 1-element arrays to strings

    We've recently change the the way we handle subjects in the vacancy
    table. This has resulted in a group of records with single-element
    arrays, which BQ does not like. To prevent this from happening again,
    I've added the code in this commit.

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -92,7 +92,8 @@ class ExportTablesToBigQuery
       # being nullable.
       data = '' if c.name == 'hired_status' && data.nil?
       # This prevents an error whereby BigQuery rejects arrays of exactly one element.
-      data = data.to_s if c.is_a?(Array) && c.length == 1
+      data = data.to_s if c.name == 'job_roles'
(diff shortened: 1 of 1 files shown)


lib/export_tables_to_big_query.rb
```diff
@@ -92,7 +92,7 @@ def bigquery_data(record, table)
       # being nullable.
       data = '' if c.name == 'hired_status' && data.nil?
       # This prevents an error whereby BigQuery rejects arrays of exactly one element.
-      data = data.to_s if c.name == 'job_roles'
+      data = data.to_s if c.is_a?(Array) && c.length == 1
       data = data.to_s if c.name == 'working_patterns'
```
cpjmcquillan, line 96: Can we lose this line as well?

tatyree: We can, indeed.

tatyree: Ah, wait, just realised why it wasn't done this way the first time. Once it's cast as a string, you can't then push an array to it and vice-versa.  Correcting.

PR #1595, opened 2020-06-01
Minor updates to README and to env.example

david-mears-dfe:
- Clarify README
- Add needed variable names to env.example


.env.example
```diff
@@ -4,6 +4,9 @@ CF_ORG=dfe-teacher-services
 CF_SPACE=teaching-vacancies-dev
 CF_USERNAME=
 CF_PASSWORD=
+CF_POSTGRES_SERVICE_ORIGIN=
+CF_SPACE_ORIGIN=
+CF_POSTGRES_SERVICE_TARGET=
```
cesidio, line 9: You need `CF_SPACE_TARGET` too

cesidio: but only if you need to take the production dump and load it to `dev` or `staging`

PR #1597, opened 2020-06-01
fix capitalization and spacing of autocomplete refinement

alexbowen:
production style fixes can be viewed/verified https://teaching-vacancies-dev.london.cloudapps.digital/

use `lou` as autocomplete test, previously it was displaying `g Loucestershire` for example


david-mears-dfe:
Is this already on dev?

alexbowen:
@david-mears-dfe yes it just finished deploying the first 2, the 3rd commit is deploying now

PR #1606, opened 2020-06-02
locations that are not postcodes and do not appear in predefined list…

alexbowen:
to fix production issue


cpjmcquillan:
If I type part of a location, and select the autocomplete suggestion, the search is carried out with the location I have typed so far rather than the full location (as far as I can tell)

PR #1598, opened 2020-06-01
Use presenter methods for indexing

david-mears-dfe:
## Changes in this PR:

- Route indexed vacancy attributes via presenter methods. We want a single source of truth for the presentation of these attributes.
- Visible changes in UI are the inclusion of 'pm/am' in the expiry time, and the capitalization and punctuation of working patterns.

### Before

([current production](https://teaching-vacancies.service.gov.uk/jobs?utf8=%E2%9C%93&keyword=teacher&location=&commit=Search#vacancy-results))

Screenshot 2020-06-01 at 19 40 28

### After 

(on staging)

Screenshot 2020-06-01 at 19 35 25

## Next steps:

- [ ] Run `Vacancy.reindex` (without bang) on production immediately after merging.
- [ ] Reindexing will probably knock out the synonyms. The latest JSON of synonyms will be found [here](https://drive.google.com/drive/u/0/folders/1RcwUjcIlpt7WXN9qPmuZKUgXvyvP9EMW) and can be directly uploaded on the Algolia dashboard.

PR #1607, opened 2020-06-03
Update index in advance of front end changes

david-mears-dfe:
This isolates the index changes from #1598 so that we can reindex before making the front end changes. #1598 is closed until we merge this and run `Vacancy.reindex`.

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -29,7 +29,7 @@ class Vacancy < ApplicationRecord
 
   include ApplicationHelper
   include Auditor::Model
-
+  include DateHelper
   include VacancyJobSpecificationValidations
   include VacancyPayPackageValidations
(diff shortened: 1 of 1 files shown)


alexbowen:
do we need to do this for the school address display format too? that isnt in the index currently but this probably has a presenter or some logic in the template that prob shouldnt be there

cpjmcquillan:
> do we need to do this for the school address display format too? that isnt in the index currently but this probably has a presenter or some logic in the template that prob shouldnt be there

I think that's the location presenter. @david-mears-dfe can confirm

david-mears-dfe:
> do we need to do this for the school address display format too? that isnt in the index currently but this probably has a presenter or some logic in the template that prob shouldnt be there

As Connor says, #1598 covers this, using the `location` attribute. It delegates to a presenter method.

PR #1610, opened 2020-06-04
Teva 844 recaptcha job alerts

tatyree:
This PR sets up [recaptcha v3](https://www.google.com/recaptcha/admin/site/352086690) score collection for new job alerts.  It does *nothing* beyond save the score as an attribute in the `Subscription` record.  We can decide what to do with bad scores (the lower the score, the more recaptcha thinks the interaction if from a bot) once we have some collected. 

The only change to the page is shown in this screenshot: 

Screenshot 2020-06-04 at 14 38 54

UX has already approved this change on the feedback page, which has been live for over a week now. UX has also agreed that it does not need to be approved again.


app/controllers/subscriptions_controller.rb
```diff
@@ -12,7 +12,9 @@ def new
   def create
     subscription = Subscription.new(daily_subscription_params)
     @subscription = SubscriptionPresenter.new(subscription)
+    recaptcha_valid = verify_recaptcha(model: subscription, action: 'subscription')
```
david-mears-dfe, line 15: If this is Boolean, can we denote this with a `?`?

tatyree: Good catch.  We can and will do.

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -28,6 +28,55 @@
   end
 
   describe '#create' do
+    let(:subscription) do
```
david-mears-dfe, line 31: Do not these new `let` blocks belong inside the scope of the new `context` block, as they aren't used outside it?

tatyree: Ah, yes, I was going to add another block of tests, but they aren't necessary.  I'll move them inside.

PR #1616, opened 2020-06-05
TEVA-804 Search results heading & job alerts

alexbowen:
- ensure server and client heading is formatted as per UX requirements
- ensure server and client headings are consistent (this will remove any flashing in heading on page load)
- relocate the job alert cta to directly under heading
- add new icon and wording

https://dfedigital.atlassian.net/browse/TEVA-804

figma - https://www.figma.com/file/KnFYJyTGiyshGDGr61uEvi/Sprint-53?node-id=14%3A147


david-mears-dfe:
My comment seemed to disappear. Oh well, here it is again!

Screenshot 2020-06-08 at 11 45 41

Re your last commit, I think the figma and prototype show that we want to include quote marks around the location and keyword.

david-mears-dfe:
Have you checked whether the keyword and location should be capitalized?

david-mears-dfe:
Is it in the scope of this PR/ticket to update the wording of `' Receive a job alert whenever a job matching this search is published'` to `'Subscribe to a job alert. We’ll email you whenever a job matching this search is published.'` ?

david-mears-dfe:
Getting very picky: can you make the bell a tiny bit larger and give it a bit more space? The Figma looks a bit cleaner as a result.

Screenshot 2020-06-08 at 12 02 10

Screenshot 2020-06-08 at 12 02 18

alexbowen:
> My comment seemed to disappear. Oh well, here it is again!
> 
> Screenshot 2020-06-08 at 11 45 41
> 
> Re your last commit, I think the figma and prototype show that we want to include quote marks around the location and keyword.

yes chris picked up the lack of quotes, its opened up a can of worms!

alexbowen:
> Have you checked whether the keyword and location should be capitalized?

yes it should be

app/frontend/styles/base/_utilities.scss
```diff
@@ -72,6 +72,10 @@ b {
   font-weight: bold;
 }
 
+.text-capitalize {
+  text-transform: capitalize;
```
cpjmcquillan, line 76: Do we want to capitalize the first letter of every word in the phrase, or just the first letter of the phrase? (thinking of the search results heading here)

alexbowen: im waiting to confirm 100% with chris. i know using this css doesnt match the figma 100% as figma suggests should be first word only capitalised

alexbowen:
> Getting very picky: can you make the bell a tiny bit larger and give it a bit more space? The Figma looks a bit cleaner as a result.
> 
> Figma:
> Screenshot 2020-06-08 at 12 02 10
> 
> Currently:
> Screenshot 2020-06-08 at 12 02 18

gave it a 5px right margin as per figma

app/frontend/styles/base/_utilities.scss
cpjmcquillan: Can you apply `uppercase` to `selector::first-letter`? I've never used it before

alexbowen: agreed with chris - keyword in heading should be exactly as user types it, location should be capitalised whatever

alexbowen:
the stylistic comments above have been addressed apostophes/bell spacing/capitalisation

app/frontend/src/search/hits.js
```diff
@@ -52,16 +52,28 @@ export const getJobAlertLinkParam = (key, value) => {
...
 
-export const createHeadingMarkup = (numberHits, keyword = null, location = null) => {
-  const prefix = keyword || location ? ` ${numberHits > 1 ? 'match' : 'matches'} ` : ' listed';
-  const postfix = `${prefix}${createHeadingHTMLForSearchTerm('', keyword)} ${createHeadingHTMLForSearchTerm('near', location)}`;
-  const hits = keyword || location ? `<span class="govuk-!-font-weight-bold">${numberHits}</span>` : `There ${numberHits > 1 ? 'are' : 'is'} <span class="govuk-!-font-weight-bold">${numberHits}</span>`;
+export const createHeadingMarkup = (options) => {
+  const { count, keyword = '', location = '' } = options;
 
-  return `${hits} ${numberHits > 1 ? 'jobs' : 'job'} ${postfix}`;
+  const postfix = `${getSearchTermsPrefix(location, keyword, count)}${createHeadingHTMLForSearchTerm('', keyword, false)} ${createHeadingHTMLForSearchTerm('near', location, true)}`;
```
david-mears-dfe, line 58: I'm a bit confused about the use of 'postfix' and 'prefix' here and in the method getSearchTermsPrefix. Are they both referring to the same thing?

app/presenters/vacancies_presenter.rb
```diff
@@ -26,30 +26,59 @@ def any?
     decorated_collection.count.nonzero?
   end
 
-  def total_count_message(keyword)
-    if total_count == 1
-      return I18n.t('jobs.job_count_without_search', count: total_count) unless @searched
+  def search_heading(keyword: '', location: '')
+      case true
+      when (keyword && keyword.length > 0) && (location && location.length > 0)
```
david-mears-dfe, line 31: Alternatively, you can use `&`: 
```ruby
pry(main)> keyword = nil
=> nil
pry(main)> keyword&.length
=> nil
pry(main)> if keyword&.length; puts 'hi'; end
=> nil
pry(main)> unless keyword&.length; puts 'hi'; end
hi
=> nil
```

david-mears-dfe, line 31: Alternatively, `when (keyword.length > 0)` (and same for location).
If keyword is nil, `>` throws an error, but keyword is never not a string.

app/presenters/vacancies_presenter.rb
```diff
@@ -26,30 +26,59 @@ def any?
...
 
-  def total_count_message_with_location(location, keyword)
+  def total_count_message_with_keyword_location(keyword, location)
     return I18n.t(
-      'jobs.job_count_with_location_category',
-      count: number_with_delimiter(total_count),
-      location: location,
-      keyword: keyword
+      'jobs.job_count_with_location_keyword',
+      count: total_count,
```
david-mears-dfe, line 51: Just checking: deliberate or accidental getting rid of `number_with_delimiter`?

app/frontend/src/search/hits.js
alexbowen: yeah i might have got these a bit topsy turvy due to refactoring, ill get it a bit more understanbable

app/presenters/vacancies_presenter.rb
alexbowen: yes this was the main bit of ruby i thought was a bit inelegant, i am used to `def search_heading(keyword: '', location: '')` type declarations meaning that keyword or location are always empty string even if not passed as arguments. and i think that correct as we are dealing with string manipulation/construction here. but i dont see why you would need this part if that was the case `when (keyword && `

alexbowen: as in keyword has to exist

alexbowen: @cesidio ?

app/presenters/vacancies_presenter.rb
alexbowen: yes that was deliberate as there is no point delimiting the number 1?

alexbowen: although i made assumption that `number_with_delimiter` turn 1000 into 1,000, am i wrong?

app/presenters/vacancies_presenter.rb
cesidio: In ruby you can use the shorter syntax:

keyword.present? && location.present?`

same for the ones below

app/presenters/vacancies_presenter.rb
```diff
@@ -26,22 +26,57 @@ def any?
...
-    end
+  def total_count_message_with_keyword_location(keyword, location)
+    return I18n.t(
+      'jobs.job_count_with_location_keyword',
+      count: total_count,
+      keyword: keyword,
+      location: location
+    ) if total_count == 1
+
+    I18n.t('jobs.job_count_plural_with_location_keyword', \
```
cesidio, line 56: No need to use `\` to indicate that the line extends on the following line

app/presenters/vacancies_presenter.rb
```diff
@@ -26,22 +26,57 @@ def any?
...
+  def search_heading(keyword: '', location: '')
+      case true
+      when (keyword && keyword.length > 0) && (location && location.length > 0)
+        total_count_message_with_keyword_location(keyword, location)
+      when keyword && keyword.length > 0
+        total_count_message_with_keyword(keyword)
+      when location && location.length > 0
+        total_count_message_with_location(location)
+      else
+        total_count_message()
```
cesidio, line 38: in ruby we omit the () if the method does not take any arguments

config/locales/en.yml
```diff
@@ -214,12 +214,14 @@ en:
...
-    job_count_plural: '%{count} jobs match your search'
-    job_count_with_location_category: '%{count} job match near %{location}'
-    job_count_plural_with_location_category: '%{count} jobs match near %{location}'
+    job_count_without_search: There is <span class='govuk-!-font-weight-bold'>%{count}</span> job listed
+    job_count_plural_without_search: "There are <span class='govuk-!-font-weight-bold'>%{count}</span> jobs listed"
+    job_count_with_keyword: "<span class='govuk-!-font-weight-bold'>%{count}</span> job matches <span class='govuk-!-font-weight-bold text-wrap-apostrophe'>%{keyword}</span>"
+    job_count_plural_with_keyword: "<span class='govuk-!-font-weight-bold'>%{count}</span> jobs match <span class='govuk-!-font-weight-bold text-wrap-apostrophe'>%{keyword}</span>"
+    job_count_with_location: "<span class='govuk-!-font-weight-bold'>%{count}</span> job found near <span class='govuk-!-font-weight-bold text-capitalize text-wrap-apostrophe'>%{location}</span>"
+    job_count_plural_with_location: <span class="govuk-!-font-weight-bold">%{count}</span> jobs found near <span class="govuk-!-font-weight-bold text-capitalize text-wrap-apostrophe">%{location}</span>
```
cesidio, line 222: have a look at how to use
https://thoughtbot.com/blog/pluralizing-rails-i18n-translations
here you can do:
```ruby
job_count_with_location:
  one: ...
  other: ...
```

config/locales/en.yml
```diff
@@ -214,12 +214,14 @@ en:
...
       other: 'Subjects'
     leadership_level: 'Leadership level'
-    job_count_without_search: 'There is %{count} job listed'
-    job_count_plural_without_search: 'There are %{count} jobs listed'
-    job_count: '%{count} job matches your search'
-    job_count_plural: '%{count} jobs match your search'
-    job_count_with_location_category: '%{count} job match near %{location}'
-    job_count_plural_with_location_category: '%{count} jobs match near %{location}'
+    job_count_without_search: There is <span class='govuk-!-font-weight-bold'>%{count}</span> job listed
+    job_count_plural_without_search: "There are <span class='govuk-!-font-weight-bold'>%{count}</span> jobs listed"
```
cesidio, line 218: because these keys contain HTML, they should end in `_html` like:

`job_count_plural_without_search_html` so that you do not need to use the method `raw` in the views

alexbowen: 👍

alexbowen: the raw thing caught me out for an hour or two :D

alexbowen: it seems that you still need the raw method in the views, but i have changed all these to _html anyway as it makes it clear

config/locales/en.yml
alexbowen: i had a look at this and it doesnt seem to work as expected, especially with `_html` postfix. so i have structured the en.yml a bit more in line with this structurally and it does make it more readable

app/presenters/vacancies_presenter.rb
alexbowen: works like a charm and a lot neater/readable

app/frontend/src/search/hits.js
alexbowen: have renamed variables in this function to be a bit intuitive i hope

app/views/vacancies/index.html.haml
```diff
@@ -1,19 +1,19 @@
...
   = content_for :page_description,  t('jobs.location_description', location: format_location_name(@vacancies_search.location_category))
 - else
   = content_for :page_title_prefix, t('jobs.heading')
 = render partial: 'head_links'
 
-%h1.govuk-heading-xl#job-count{'role': 'heading', 'aria-level': '1'}
-  = @vacancies_search.location_category_search? ? @vacancies.total_count_message_with_location(format_location_name(@vacancies_search.location_category)) : @vacancies.total_count_message
-.govuk-grid-row
+%h1#job-count{ class: 'govuk-!-margin-bottom-2 govuk-heading-xl govuk-!-font-weight-regular', role: 'heading', 'aria-level': '1' }
+  = raw @vacancies.search_heading(location: @vacancies_search.location, keyword: @vacancies_search.keyword)
```
cpjmcquillan, line 9: Is there any associated security risk using `raw` on user input here?

PR #1621, opened 2020-06-08
Changes to copy part 1

cpjmcquillan:
This PR implements the first part of content changes and UX refinements to the copy process.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-766

## Changes in this PR:
- Updated content to the copy page
- Updated review page for copying vacancies
- Cancel copy link that cancels the copy process

### DIFF
--- a/app/controllers/hiring_staff/vacancies_controller.rb
+++ b/app/controllers/hiring_staff/vacancies_controller.rb
@@ -32,7 +32,8 @@ class HiringStaff::VacanciesController < HiringStaff::Vacancies::ApplicationCont
     unless @vacancy.valid?
       redirect_to_incomplete_step
     else
-      state = params[:edit_draft] == 'true' ? 'edit' : 'review'
+      state = %w(copy copy_review).include?(@vacancy.state) ? 'copy_review' :
+        params[:edit_draft] == 'true' ? 'edit' : 'review'
       @vacancy.update(state: state) unless @vacancy&.state == state
       set_completed_step
     end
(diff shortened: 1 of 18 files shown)


app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -32,7 +32,7 @@ def review
     unless @vacancy.valid?
       redirect_to_incomplete_step
     else
-      state = params[:edit_draft] == 'true' ? 'edit' : 'review'
+      state = params[:edit_draft] == 'true' ? 'edit' :  @vacancy&.state == 'copy' ? 'copy' : 'review'
```
cesidio, line 35: It's not a good practice to have a ternary inside a ternary. Maybe you can move that to a method

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -32,7 +32,7 @@ def review
...
       @vacancy.update(state: state) unless @vacancy&.state == state
```
cesidio, line 36: I think that if you try to call `update` and you set the same value, nothing happens anyway so you do not need to check that.

app/views/hiring_staff/vacancies/vacancy_form_partials/_hidden_state_input.html.haml
```diff
@@ -1 +1,2 @@
-= f.hidden_field :state, value: @vacancy&.published? ? 'edit_published' : %w(review edit).include?(@vacancy&.state) ? @vacancy.state : 'create'
+- copy = false if local_assigns[:copy].nil?
+= f.hidden_field :state, value: copy ? 'copy' : @vacancy&.published? ? 'edit_published' : %w(copy review edit).include?(@vacancy&.state) ? @vacancy.state : 'create'
```
cesidio, line 2: Maybe a helper for all this logic, or in the presenter

app/controllers/hiring_staff/vacancies_controller.rb
cpjmcquillan: Ah ok nice, thanks

PR #1628, opened 2020-06-10
Improve specs folder structure

david-mears-dfe:
Restructuring the folders to be more legible was suggested by @vigneshmsft 

## Changes in this PR:

- Unit tests and feature/integration being the main dichotomy
- Also, move some root-level support files to support folder
- Does this seem sensible (as a low-hanging fruit change rather than a deep-dive)?


david-mears-dfe:
Closing because it requires more investment of time to do. 

We rely on `config.infer_spec_type_from_file_location!`, which forces a conventional file structure.

PR #1634, opened 2020-06-11
Custom Cloudfront caching for 404 end 500 errors

saliceti:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-889

## Changes in this PR:
By default it is 5 min. This has created a small downtime on 9/06.
We reduce it to avoid waiting for longer than expected.

## How to test
- Query the app, it should return 200
- Stop the app
- It should return 404
- Start it and wait 10s
- It should return 200

### DIFF
--- a/terraform/modules/cloudfront/cloudfront.tf
+++ b/terraform/modules/cloudfront/cloudfront.tf
@@ -22,6 +22,16 @@ resource "aws_cloudfront_distribution" "default" {
   }
 
   custom_error_response {
+    error_code            = "404"
+    error_caching_min_ttl = "10"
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
> 60s for error 500 as there may be an internal problem with the app so we can
use cloudfront to reduce the load

I don't quite understand what's being said here. You want to cache 500 for longer than 404, because why? Call me if it's easier than replying on GitHub.

david-mears-dfe:
Colin updated the commit message to have more explanation. He also kindly gave me a demo!

PR #1635, opened 2020-06-11
Bug/teva 894 geocoding fallback

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-894

### DIFF
--- a/app/frontend/src/search/index.js
+++ b/app/frontend/src/search/index.js
@@ -44,17 +44,19 @@ if (document.querySelector('#vacancies-hits')) {
             return new Promise(resolve => {
                 if (stringMatchesPostcode(query) || (query.length && locations.indexOf(query.toLowerCase()) === -1)) {
                     getGeolocatedCoordinates(query).then(coords => {
-                        coords.success ? enableRadiusSelect() : disableRadiusSelect();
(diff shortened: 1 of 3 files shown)


app/frontend/src/search/index.js
```diff
@@ -44,17 +44,19 @@ if (document.querySelector('#vacancies-hits')) {
...
-                        setDataAttribute(document.querySelector('#radius'), 'radius', document.querySelector('#radius').value || 10);
-                        resolve();
+                        if (coords.success) {
+                            enableRadiusSelect();
+                            setDataAttribute(document.querySelector('#location'), 'coordinates', `${coords.lat}, ${coords.lng}`);
+                            setDataAttribute(document.querySelector('#radius'), 'radius', document.querySelector('#radius').value || 10);
+                            resolve();
+                        }
                     });
-                } else {
```
cpjmcquillan, line 52: Does the removal of this else statement always disable the radius select input, or does `resolve` on line 51 return from the Promise?

alexbowen: yes resolve on 51 stops execution of the promise, returns it in a sense (although you `resolve(something);` if you want a return value so its not quite the same)

PR #1646, opened 2020-06-15
Implement session timeout after 60 mins inactivity

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-840

### DIFF
--- a/app/controllers/hiring_staff/base_controller.rb
+++ b/app/controllers/hiring_staff/base_controller.rb
@@ -1,7 +1,11 @@
 class HiringStaff::BaseController < ApplicationController
-  before_action :redirect_to_root_if_read_only
-  before_action :check_session
-  before_action :check_terms_and_conditions
+  TIMEOUT_PERIOD = 60.minutes
+
+  before_action :check_user_last_activity_at,
+    :update_user_last_activity_at,
+    :redirect_to_root_if_read_only,
+    :check_session,
+    :check_terms_and_conditions
 
   helper_method :current_school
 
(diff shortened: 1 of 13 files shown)


app/controllers/hiring_staff/base_controller.rb
```diff
@@ -28,4 +32,23 @@ def current_user
 
     @current_user ||= User.find_or_create_by(oid: current_session_id)
   end
+
+  def check_user_last_activity_at
+    return redirect_to dsi_logout_url if current_user&.last_activity_at.blank?
+
+    if Time.zone.now > (current_user.last_activity_at + TIMEOUT_PERIOD)
+      session[:signing_out_for_inactivity] = true
+      redirect_to dsi_logout_url
```
david-mears-dfe, line 41: I think perhaps this should be `return redirect_to`, in order to make the action `check_user_last_activity_at` closed and thus avoid the possibility of too many redirects.

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -18,8 +19,13 @@ def create
...
 
   def destroy
+    if session[:signing_out_for_inactivity]
+      flash_message = { notice: I18n.t('messages.access.signed_out_for_inactivity') }
+    else
+      flash_message = { success: I18n.t('messages.access.signed_out') }
+    end
     session.destroy
-    redirect_to root_path, notice: I18n.t('messages.access.signed_out')
+    redirect_to new_identifications_path, flash_message
```
david-mears-dfe, line 28: I think perhaps this should be `return redirect_to`, in order to make the action `destroy` closed and thus avoid the possibility of too many redirects.

cesidio: No need to return on the last line of the method

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -18,8 +19,13 @@ def create
   end
 
   def destroy
+    if session[:signing_out_for_inactivity]
+      flash_message = { notice: I18n.t('messages.access.signed_out_for_inactivity') }
+    else
+      flash_message = { success: I18n.t('messages.access.signed_out') }
```
cpjmcquillan, line 25: Check with UX, but I think this should be `notice` for both

config/locales/en.yml
```diff
@@ -648,6 +648,7 @@ en:
         only_published: You can only view published jobs
     access:
       signed_out: You've been signed out.
+      signed_out_for_inactivity: You've been signed out due to inactivity.
```
cpjmcquillan, line 651: You have been and no full stops

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
david-mears-dfe: Correct https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1589793769222100?thread_ts=1589793583.221500&cid=CG9A8H1HP

PR #1652, opened 2020-06-15
Bring preview message in line with Figma

david-mears-dfe:
I spotted a difference between production and the Figma for sprint 50, which was a 5-minute job so I just did it.

## Figma

See 1.8. https://www.figma.com/file/n2wLtKVQ2imMbjdPSHBRKW/Sprint-50?node-id=0%3A1

### Before

Screenshot 2020-06-15 at 17 48 02

### After

Screenshot 2020-06-15 at 18 36 57

## Next steps:

- [x] UX sign-off

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -327,7 +327,7 @@ en:
       button: 'Preview job listing'
       page_title: 'Preview job listing for %{school}'
       summary:
-        heading: 'Preview of %{title}'
+        heading_html: 'Preview of &#8216;%{title}&#8217; job listing'
         body_html:
           zero: >-
             This is how your job listing will appear to jobseekers visiting Teaching Vacancies.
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -327,7 +327,7 @@ en:
       button: 'Preview job listing'
       page_title: 'Preview job listing for %{school}'
       summary:
-        heading: 'Preview of %{title}'
+        heading_html: 'Preview of &#8216;%{title}&#8217; job listing'
```
cpjmcquillan, line 330: `heading: Preview of '%{title}' job listing` if you prefer. We don't use fancy apostrophes anywhere else on the site yet as far as I know

david-mears-dfe: I thought the straight apostrophes looked ugly.

david-mears-dfe: Screenshot 2020-06-15 at 18 19 59

For your consideration

cpjmcquillan: Tomayto tomahto, I'm happy either way if UX/Content are happy

PR #1654, opened 2020-06-15
Display notification when user 'saves' a draft job from the review page

cpjmcquillan:
This PR adds an endpoint that redirects the user to the appropriate tab on the schools page, with a success notification.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-865

## Changes in this PR
- Add endpoint and update link on review page

### DIFF
--- a/app/views/hiring_staff/vacancies/review.html.haml
+++ b/app/views/hiring_staff/vacancies/review.html.haml
@@ -30,7 +30,7 @@
     = link_to t('jobs.submit_listing.button'), school_job_publish_path(@vacancy.id), method: :post, class: 'govuk-button govuk-button--secondary submit-listing-without-preview-gtm', id: 'vacancy-review-submit'
 
     %br
-    = render 'hiring_staff/vacancies/back_to_manage_jobs'
+    = link_to t('buttons.back_to_manage_jobs'), jobs_with_type_school_path('draft', from_review: @vacancy.id), class: 'govuk-link govuk-link--no-visited-state'
 
   .govuk-grid-column-one-third
     - if %w(create review).include?(@vacancy.state)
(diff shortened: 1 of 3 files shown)


app/controllers/hiring_staff/base_controller.rb
```diff
@@ -51,4 +52,10 @@ def dsi_logout_url
     url.query = { post_logout_redirect_uri: auth_dfe_signout_url, id_token_hint: session[:id_token] }.to_query
     url.to_s
   end
+
+  def render_link_to_flash_message
```
cesidio, line 56: It might be useful also for the jobseekers area, so this could go to the generic ApplicationController

cpjmcquillan: I didn't think we would want to, that's why I put it there. But maybe it's nicer to move it there just in case we do want to 👍

app/views/hiring_staff/vacancies/review.html.haml
```diff
@@ -30,7 +30,7 @@
     = link_to t('jobs.submit_listing.button'), school_job_publish_path(@vacancy.id), method: :post, class: 'govuk-button govuk-button--secondary submit-listing-without-preview-gtm', id: 'vacancy-review-submit'
 
     %br
-    = render 'hiring_staff/vacancies/back_to_manage_jobs'
+    = link_to t('buttons.back_to_manage_jobs'), jobs_with_type_school_path('draft', flash_success: t('messages.jobs.draft_saved_html', job_title: @vacancy.job_title)), class: 'govuk-link govuk-link--no-visited-state'
```
cesidio, line 33: Passing the content of the message could introduce a security problem here. I think you can pass just a parameter that flags that specific message to that specific view. We do not need to make it general

PR #1659, opened 2020-06-17
introduce airbnb eslint rules

alexbowen:
JS linting was added a little while back but not many rules added.

this PR introduces https://www.npmjs.com/package/eslint-config-airbnb-base as this widely acknowledged to be a comprehensive modern ruleset.

obvious and inconsequential rules (style/spacing) have been fixed
some lines have been disabled from linting as to fix these i feel has highlighted the need for testing and so these will be addressed in a separate PR


.eslintrc.json
```diff
@@ -21,6 +21,7 @@
     },
     "rules": {
         "semi": ["warn", "always"],
+        "no-use-before-define": ["warn"],
```
david-mears-dfe, line 24: I think we should downgrade this rule even further, into non-existence. Robert C. Martin says in Clean Code that for readability "If one function calls another, they should be vertically close, and the caller should be above the callee, if at all possible." I tend to agree.

david-mears-dfe: If you get rid of it, is it worth reverting the changes in this commit? If we revert, we get closer to the Robert C. Martin ideal, but if not, we would get consistency.

alexbowen: whilst i havent read this book, i am in agreement, its readability that becomes the primary concern if ordering of functions wont *break* anything. id rather see associated methods near one another rather than one of them being 100 lines away just because it uses a utility function or something like this. agreed, will disable this rule entirely

app/frontend/src/search/ui/alert.js
```diff
@@ -1,43 +1,49 @@
+/* eslint-disable max-len */
```
david-mears-dfe, line 1: I'm uncomfortable with disabling rules at all, let alone for entire files.
Applies to 4 files.

david-mears-dfe: Sorry, should have read the PR description!

alexbowen: personally i disagree, it depends on the rule. in this case it is indicating that the code might be hard to read because of the length of the line. but when you are talking about strings being on one line refactoring them into smaller lines makes them harder to read and in the case of tests can break your tests. if the rule was === rule for instance, i would never disable this as this indicates your code more than likely has a problem. ill meet you halfway and disable the lines explicitly that it applies to rather than the whole file

alexbowen: but you are absolutely right picking up anything lint wise that says disable, its more often than not a code smell if it hasn't been thought about

PR #1660, opened 2020-06-18
Teva 884 revised

tatyree:
This is in draft so everyone can see where I'm heading with it.  

Read the commit messages to understand my thinking about it.


spec/jobs/send_daily_alert_email_job_spec.rb
```diff
@@ -1,139 +0,0 @@
...
-    it 'limits the number of vacancies' do
-      relation = Vacancy.none
-
-      allow(subscription).to receive(:vacancies_for_range) { relation }
-      expect(relation).to receive(:limit).with(500)
-
-      job.vacancies_for_subscription(subscription)
-    end
-
-    context 'with algolia search', algolia: true do
```
cpjmcquillan, line 74: You can just get rid of this bit

app/models/vacancy.rb
```diff
@@ -229,7 +229,7 @@ def coordinates
   end
 
   def listed?
-    published? && !publish_on.future? && expires_on.future?
+    published? && !publish_on&.future? && expires_on&.future?
```
cpjmcquillan, line 232: `expires_on` doesn't contain the time

app/models/vacancy.rb
```diff
@@ -119,15 +119,15 @@ class Vacancy < ApplicationRecord
...
       ranking ['desc(publication_date_timestamp)']
     end
 
-    add_replica Rails.env.test? ? "Vacancy_test#{ENV.fetch('GITHUB_RUN_ID', '')}_expiry_time_desc" : 'Vacancy_expiry_time_desc', inherit: true do
+    add_replica 'Vacancy_expiry_time_desc', inherit: true do
       ranking ['desc(expires_at_timestamp)']
     end
 
-    add_replica Rails.env.test? ? "Vacancy_test#{ENV.fetch('GITHUB_RUN_ID', '')}_expiry_time_asc" : 'Vacancy_expiry_time_asc', inherit: true do
+    add_replica 'Vacancy_expiry_time_asc', inherit: true do
```
cpjmcquillan, line 130: The test logic isn't needed if the integration tests are removed. 
The `add_replica` blocks are there because the algoliarails client could never find the replica indices in the app settings correctly - despite them existing on the dashboard.

spec/jobs/send_daily_alert_email_job_spec.rb
cpjmcquillan: The whole spec file doesn't need to be deleted, does it?

app/models/vacancy.rb
```diff
@@ -45,7 +45,7 @@ class Vacancy < ApplicationRecord
   # rubocop:disable Metrics/BlockLength
   # rubocop:disable Metrics/LineLength
   # There must be a better way to pass these settings to the block, but everything seems to break
-  algoliasearch index_name: Rails.env.test? ? "Vacancy_test#{ENV.fetch('GITHUB_RUN_ID', '')}" : 'Vacancy', auto_index: Rails.env.production?, auto_remove: Rails.env.production?, synchronous: Rails.env.test?, disable_indexing: !(Rails.env.production? || Rails.env.test?) do
+  algoliasearch if: :listed? do
```
david-mears-dfe: Can we retain `disable_indexing` for the staging environment, etc? Otherwise I don't know what is stopping staging data appearing in our index. Perhaps `auto_index` and `auto_remove` are load-bearing too, I'm not sure.

david-mears-dfe: I see the next commit handles this. Is it worth having any extra layer of security by using `disable_indexing: Rails.env.staging?` or something?

tatyree: Read the comment in the README about how it handles attempts by non-prod environments attempting to start using the prod index.

spec/jobs/send_daily_alert_email_job_spec.rb
tatyree: Oops.  Didn't mean to pull the whole thing.  I'll put the rest back now. Thanks!

app/models/vacancy.rb
tatyree: Also, you seem to have commented on an outdated version of the PR, @david-mears-dfe...

spec/jobs/send_daily_alert_email_job_spec.rb
tatyree: @cpjmcquillan fixed.

app/models/vacancy.rb
```diff
@@ -37,15 +37,59 @@ class Vacancy < ApplicationRecord
...
 
-  # For guidance on sanity-checking an indexing change, read documentation/algolia_sanity_check.md
+  # NOTE: the `if: :listed?` filter in the `algoliasearch` stanza below *only* excludes records from being *added* to
+  # the index. It DOES NOT prevent the ruby client from checking that the record exists in the Algolia index in the
+  # first place. Even if the record should not be in the index (unpublished or expired records), the client still
+  # consumes an Algolia operation to try and look it up if it appears in canonical list of records returned by the
+  # model.
+  #
+  # To illustrate: if you run the unmodified `Vacancy.reindex!` on a recent (2020-06-18) production dataset you will
+  # consume more than 30,000 operations on the Alogolia app. This occurs because it looks up each of the 30,000+
```
alexbowen, line 50: minor typo alogolia -> algolia

app/models/vacancy.rb
```diff
@@ -181,28 +225,29 @@ def lng
...
-                          published.where('expiry_time IS NULL AND publish_on <= ? AND expires_on >= ?',
-                                          Time.zone.today, Time.zone.today)
-                        })
-  scope :awaiting_feedback, (-> { expired.where(listed_elsewhere: nil, hired_status: nil) })
+    published.where('expiry_time IS NULL AND publish_on <= ? AND expires_on >= ?',
+                    Time.zone.today, Time.zone.today)
+  })
+  scope :pending, (-> { published.where('publish_on > ?', Time.zone.today) })
+  scope :published_on_count, (->(date) { published.where(publish_on: date.all_day).count })
+  scope :unindexed, (-> { live.where(initially_indexed: false) })
```
alexbowen, line 250: 👍  cross referenced this with filtering rules and exactly the same i think

spec/spec_helper.rb
```diff
@@ -19,7 +19,6 @@
 
   config.shared_context_metadata_behavior = :apply_to_host_groups
   config.filter_run_when_matching :focus
-  config.filter_run_excluding algolia: true
```
alexbowen, line 22: i thought this was a bit odd

app/models/vacancy.rb
```diff
@@ -37,15 +37,59 @@ class Vacancy < ApplicationRecord
...
+  def self.remove_vacancies_that_expired_yesterday!
+    expired_records = where('expiry_time BETWEEN ? AND ?', Time.zone.yesterday.midnight, Time.zone.today.midnight)
+    index.delete_objects(expired_records.map(&:id)) if expired_records.present?
+  end
 
   # rubocop:disable Metrics/BlockLength
   # rubocop:disable Metrics/LineLength
-  # There must be a better way to pass these settings to the block, but everything seems to break
-  algoliasearch index_name: Rails.env.test? ? "Vacancy_test#{ENV.fetch('GITHUB_RUN_ID', '')}" : 'Vacancy', auto_index: Rails.env.production?, auto_remove: Rails.env.production?, synchronous: Rails.env.test?, disable_indexing: !(Rails.env.production? || Rails.env.test?) do
+  algoliasearch auto_index: true, auto_remove: true, if: :listed? do
```
cpjmcquillan, line 92: I think `auto_index` and `auto_remove` are true by default

tatyree: Noted.  Doesn't hurt to make them explicit :)

spec/spec_helper.rb
tatyree: It was only because the integration tests made indexes and records and so should have been used sparingly.

spec/jobs/send_daily_alert_email_job_spec.rb
```diff
@@ -71,10 +71,8 @@
       job.vacancies_for_subscription(subscription)
     end
 
-    context 'with algolia search', algolia: true do
+    context 'subscription created before algolia' do
```
cpjmcquillan, line 74: Lines 74-137 inclusive (on the original file) need to go I think

spec/spec_helper.rb
cpjmcquillan: @alexbowen it was needed so that expensive specs with `algolia: true` weren't executed every time, but instead only executed if `rspec --tag algolia` was explicitly called

config/initializers/algoliasearch.rb
```diff
@@ -1,3 +1,19 @@
+ALGOLIA_PRODUCTION_APP_ID = 'QM2YE0HRBW'
```
cesidio, line 1: Shouldn't this be an env variable?

app/models/vacancy.rb
tatyree: Fixed.

config/initializers/algoliasearch.rb
tatyree: Given that the intent is to prevent us from accidentally reindexing the production app index from another environment, that that key is safe to publicly expose, and that we're unlikely to change the production app without advanced planning (would have to change the billing details) I felt like hardcoding it was the safer, more intention-revealing way to go with this.  Plus, if we put it in an ENV variable, then we've got yet another to manage across multiple environments :(

spec/jobs/send_daily_alert_email_job_spec.rb
tatyree: I left those in because I wasn't sure how many (any?) old-school subscriptions we might still have running.  If they are all gone, I'm happy to zap those lines.

tatyree: I just went ahead and killed them.

PR #1669, opened 2020-06-19
Update .gitignore to include other certificate from keybase

david-mears-dfe:
The teacher-vacancy-service isn't where this file lives anyway, but still, better safe than sound.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ config/localhost/https/*.key
 config/localhost/https/*.crt
 localhost.key
 localhost.crt
+rootCA.pem
 
 # GCP development credentails
 serviceAccount.json


cesidio:
No need to have that file in that folder:

https://github.com/DFE-Digital/teacher-vacancy-service/blob/master/config/localhost/https/README.md

david-mears-dfe:
> No need to have that file in our repo:
> 
> https://github.com/DFE-Digital/teacher-vacancy-service/blob/master/config/localhost/https/README.md

You're right, but it's easy to accidentally put it in our repo. Do we lose anything by having this extra security?

alexbowen:
lol, i made exactly the same PR in my first week on the project. but i guess it doesnt do any harm as you say @david-mears-dfe

PR #1673, opened 2020-06-22
Document secrets

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-920

## Changes in this PR:
This is a follow-up to the security incident on 16/06. We want to be
able to regenerate the secrets as quickly as possible in case such
incident occurs again.

### DIFF
--- a/documentation/secrets.md
+++ b/documentation/secrets.md
@@ -1,0 +1,94 @@
+# Application secrets
+
+The application requires configuration to run. Most of the values are stored in the application github repository but some of them are secrets which can't be exposed in a public repository.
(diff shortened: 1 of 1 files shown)


saliceti:
> None of the lists increment

Have you tried in markdown? It should automatically increment. See:
https://github.com/DFE-Digital/teacher-vacancy-service/blob/71f31272d9f869012e5ba2f663dab21efcf30da9/documentation/secrets.md

tatyree:
@saliceti I went ahead and put these in the README with #1706 so I'm going to close and delete this one.

PR #1674, opened 2020-06-23
Client side search bugs

alexbowen:
## Changes in this PR:

- more unit test coverage (most components at 90% plus coverage apart from autocomplete which we are going to replace with alphagov version).
- decouple autocomplete entirely from algolia client.
- initialise search client when user starts to interact with form. otherwise if on page load (as before) the client does not work nicely with the results that the homepage produces.
- make sure coordinates (if geocoded) from homepage search are available to the front end.


app/frontend/src/search/index.js
```diff
@@ -3,7 +3,8 @@ import 'regenerator-runtime/runtime';
 import '../polyfill/classlist.polyfill';
 
 import {
-  connectSearchBox, connectAutocomplete, connectHits, connectSortBy, connectMenu, connectStats, connectPagination,
+  connectSearchBox, connectHits, connectSortBy, connectMenu, connectStats, connectPagination,
+  connectSearchBox, connectHits, connectSortBy, connectMenu, connectStats,
```
cpjmcquillan: duplicated imports duplicated imports

alexbowen: i did have a rebase conflict here and ended up with 2 but i removed and that is case in file chnages? ill resolve this

app/frontend/src/locationAutocomplete.js
```diff
@@ -4,13 +4,11 @@ import { locations } from './search/data/locations';
 $(document).ready(function () {
```
cpjmcquillan, line 4: Do we not want to move away from a jQuery dependency?

app/views/shared/_location_search_tag.html.haml
```diff
@@ -1,7 +1,7 @@
 .govuk-form-group.js-location-finder#location-search
   = label_tag 'location', t('jobs.filters.location'), class: "govuk-label #{extra_label_class}"
   .js-location-finder__input-container#js-location-finder__input-container
-    = search_field_tag :location, @vacancies_search&.location, class: 'govuk-input js-location-finder__input app-site-search__input app-site-search__input--default', placeholder: t('jobs.filters.location_hint'), autocomplete: 'off',  "aria-expanded": "true", "aria-owns": "location__listbox", "aria-autocomplete": "list", role: "combobox"
+    = search_field_tag :location, @vacancies_search&.location, class: 'govuk-input js-location-finder__input app-site-search__input app-site-search__input--default', placeholder: t('jobs.filters.location_hint'), autocomplete: 'off',  "aria-expanded": "true", "aria-owns": "location__listbox", "aria-autocomplete": "list", role: "combobox", 'data-coordinates': @vacancies_search&.coordinates
```
cpjmcquillan, line 4: `data: { coordinates: @vacancies_search&.coordinates }` should do the job too if you prefer

app/frontend/src/locationAutocomplete.js
alexbowen: yes this was a bit lazy, im going to leave it for the moment however as this is a part of modernizing the rest if the app JS

PR #1678, opened 2020-06-24
DSI contingency documentation

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-815

## Changes in this PR:

You can see the documentation [here](https://github.com/DFE-Digital/teacher-vacancy-service/blob/dsi-contingency-doc/documentation/dsi-contingency-fallback.md)

- Does it cover everything you would want to know?


documentation/cron-jobs.md
```diff
@@ -0,0 +1,9 @@
+# Cron jobs
```
cpjmcquillan, line 1: This documentation is in `govuk-paas.md` I think

david-mears-dfe: Yep, I'll remove the file.

PR #1679, opened 2020-06-24
Update kaminari templates for new pagination design

cpjmcquillan:
This PR updates kaminari templates so that server and client side pagination follow the same design exactly.

## Changes in this PR:
- Remove first and last links
- Change window to 2
- Update styling


david-mears-dfe:
If you are dropping the translations, should they be removed from the en.yml too? Although I'd rather we used the en.yml.

cpjmcquillan:
> If you are dropping the translations, should they be removed from the en.yml too? Although I'd rather we used the en.yml.

Those translations don't exist anyway

PR #1691, opened 2020-06-25
Correct 'sort by' is selected dependent on search scenario

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-898

## Changes in this PR:

- With this PR we will introduce a 'most_relevant' value to the job sort drop-down, and map it to the base search index, 'Vacancy'.

Scenario A: If the user searches by keyword & location, we sort vacancies by most relevant

Scenario B: If the user searches by location only, we sort by newest job listing

Scenario C: If the user clicks search without entering search terms, we show them all vacancies, with the heading 'There are x jobs listed' and we sort by newest job listing

### DIFF
--- a/app/frontend/src/search/ui/sort.js
+++ b/app/frontend/src/search/ui/sort.js
@@ -8,7 +8,7 @@ export const renderSortSelectInput = (renderOptions, isFirstRender) => {
 
     if (document.querySelector('#jobs_sort_select')) {
       document.querySelector('#jobs_sort_select').addEventListener('change', (event) => {
-        refine(getSearchReplicaName(event.target.value));
+        refine(getSearchIndexName(event.target.value));
       });
     }
   }
@@ -22,4 +22,10 @@ export const renderSortSelectInput = (renderOptions, isFirstRender) => {
   }
 };
 
-export const getSearchReplicaName = (selected) => (selected ? `Vacancy_${selected}` : 'Vacancy');
+export const getSearchIndexName = (selected) => {
+  const defaultSearchIndexName = 'Vacancy_publish_on_desc';
+  if (selected === 'most_relevant') {
+    return 'Vacancy';
+  }
+  return selected ? `Vacancy_${selected}` : defaultSearchIndexName;
+};
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -20,7 +20,7 @@ class Vacancy < ApplicationRecord
   }.merge(FLEXIBLE_WORKING_PATTERN_OPTIONS).freeze
 
   JOB_SORTING_OPTIONS = [
-    [I18n.t('jobs.sort_by.most_relevant'), ''],
+    [I18n.t('jobs.sort_by.most_relevant'), 'most_relevant'],
     [I18n.t('jobs.sort_by.publish_on.descending'), 'publish_on_desc'],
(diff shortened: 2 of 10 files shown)


spec/features/job_seekers_can_search_for_vacancies_with_algolia_spec.rb
```diff
@@ -234,8 +234,20 @@
   end
```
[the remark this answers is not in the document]
david-mears-dfe: As I discovered to my chagrin

app/models/vacancy.rb
```diff
@@ -20,7 +20,7 @@ class Vacancy < ApplicationRecord
   }.merge(FLEXIBLE_WORKING_PATTERN_OPTIONS).freeze
 
   JOB_SORTING_OPTIONS = [
-    [I18n.t('jobs.sort_by.most_relevant'), ''],
+    [I18n.t('jobs.sort_by.most_relevant'), 'most_relevant'],
```
cpjmcquillan, line 23: I think this adds unnecessary complexity

app/frontend/src/search/ui/sort.js
```diff
@@ -22,4 +22,10 @@ export const renderSortSelectInput = (renderOptions, isFirstRender) => {
   }
 };
 
-export const getSearchReplicaName = (selected) => (selected ? `Vacancy_${selected}` : 'Vacancy');
+export const getSearchIndexName = (selected) => {
```
cpjmcquillan, line 25: You don't need to change this function at all. You do need to add a function that sets the appropriate default sort based on user input.

spec/support/search_helper.rb
```diff
@@ -1,9 +1,11 @@
 module SearchHelper
+  DEFAULT_INDEX = "Vacancy_test#{ENV.fetch('GITHUB_RUN_ID', '')}_publish_on_desc"
```
david-mears-dfe, line 2: This line - and indeed this file - is obsolete as of https://github.com/DFE-Digital/teacher-vacancy-service/commit/923c78946291cb99899d78ed09ff5f2d97987749. But we expect to reinstate it in the future, and when we do, the changes to this file will save someone some confusion.

app/services/vacancy_algolia_search_builder.rb
```diff
@@ -118,8 +119,10 @@ def build_location_filter
   end
 
   def build_search_replica
-    return nil if sort_by.blank?
-    self.search_replica = ['Vacancy', sort_by].reject(&:blank?).join('_')
+    # Returning nil means the search should use the main index, not a replica
```
cpjmcquillan, line 122: You shouldn't need to change this. I would maybe add a public method `sort_option` which returns the correct default `sort_by` value or the `sort_by` that was selected by the user, and change the value in the `select_tag` in _sorting_options.html.haml

app/models/vacancy.rb
david-mears-dfe: You could be right... the only reason to have 'most_relevant' is so that a user can override the default and get the 'most relevant' results for a null query, which doesn't make too much sense.

david-mears-dfe:
Reimplemented by https://github.com/DFE-Digital/teacher-vacancy-service/pull/1698

PR #1696, opened 2020-06-25
[TEVA-858] Deploy prometheus ecosystem

saliceti:
Depends on https://github.com/DFE-Digital/bat-platform-building-blocks/pull/47

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-858

## Changes in this PR:

- Monitoring terraform module
- Makefile commands
- run-in-env script to fetch secrets from AWS
- Moved app terraform code to its own directory

## How to review
- You need an AWS account with permission to read the secrets
- Run `make monitoring-plan` (The components are already deployed)
- Access the apps (space teaching-vacancies-monitoring)


bin/run-in-env
```diff
@@ -0,0 +1,133 @@
...
+    with_decryption: true,
+  })
+
+  YAML.load(response.parameter.value)
+end
+
+def make_tf_vars_map(config_map)
+  @log.debug 'Transforming variables into terraform TF_VAR_* variables'
+
+  config_map.map { |k, v|
```
cesidio, line 53: In ruby there is a convention that blocks spanning on one line we you use `{}`

`config_map.map { |k, v| ["TF_VAR_#{k}", v] }.to_h`

or if you want to go on multiple lines you do:

```
config_map.map do |k, v| 
  ["TF_VAR_#{k}", v]
end.to_h
```
which is not that nice considering that at the end of it you have to do `.to_h`

saliceti: done, thanks

PR #1697, opened 2020-06-25
Import trust data

cpjmcquillan:
This PR creates a daily job and task for importing SchoolGroup data into the database. 

The task downloads data from [two .csv files](https://get-information-schools.service.gov.uk/Downloads). `All group records` and `All group links records`. Currently the data retrieved is insufficient (due to a bug), however this has been raised with the GIAS team and should be resolved. 

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-684

## Changes in this PR
- Daily job that runs a task to parse .csv files from GIAS in order to create SchoolGroups and create SchoolGroupMembership relationships between Schools and SchoolGroups

## Next steps
- Implement methods/tasks to parse the GIAS data json for MATs and populate useful fields on SchoolGroup records

### DIFF
--- a/lib/import_school_group_data.rb
+++ b/lib/import_school_group_data.rb
@@ -1,28 +1,59 @@
 require 'csv'
 require 'httparty'
 
-CSV_URL = 'https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/allgroupsdata.csv'
-TEMP_CSV_FILE_LOCATION = './tmp/school-groups-data.csv'
+SCHOOL_GROUP_URL = 'https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/allgroupsdata.csv'
+SCHOOL_GROUP_TEMP_LOCATION = './tmp/school-group-data.csv'
+
+SCHOOL_GROUP_MEMBERSHIP_URL = 'https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/alllinksdata.csv'
+SCHOOL_GROUP_MEMBERSHIP_TEMP_LOCATION = './tmp/school-group-membership-data.csv'
 
 class ImportSchoolGroupData
   def run!
-    save_csv_file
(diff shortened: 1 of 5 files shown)


lib/import_school_group_data.rb
```diff
@@ -0,0 +1,69 @@
...
+      File.write(location, request.body, mode: 'wb')
+    elsif request.code == 404
+      raise HTTParty::ResponseError, 'CSV file not found.'
+    else
+      raise HTTParty::ResponseError, 'Unexpected problem downloading CSV file.'
+    end
+  end
+
+  def convert_to_school_group(row)
+    school_group = SchoolGroup.find_or_initialize_by(uid: row['Group UID'])
```
david-mears-dfe, line 58: `SchoolGroup`s don't have a `.uid` method yet, correct?

cpjmcquillan: All SchoolGroup records have a `uid` field by definition

lib/import_school_group_data.rb
```diff
@@ -0,0 +1,69 @@
+require 'csv'
+require 'httparty'
+
+SCHOOL_GROUP_URL = 'https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/allgroupsdata.csv'
```
alexbowen, line 4: do these URLs need to be env vars to point at diff environments fo things like integration testing? (cause we havent got enough environment variables)

cesidio: We want the full list of data on all environments. Why would you want a different source of data for staging?

alexbowen: i dont why you would tbh, just posing the question 👍

alexbowen: i guess i was thinking along the line of if the data source was versioned. you might want to test against a new version, was the sort of thing i was thinking, but if thats not the case then no bother

PR #1698, opened 2020-06-25
Set the correct default sorting strategy per scenario

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-898

## Changes in this PR:

> Scenario A: If the user searches by keyword & location, we sort vacancies by most relevant
> 
> Scenario B: If the user searches by location only, we sort by newest job listing
> 
> Scenario C: If the user clicks search without entering search terms, we show them all vacancies, with the heading 'There are x jobs listed' and we sort by newest job listing

### DIFF
--- a/spec/services/vacancy_algolia_search_builder_spec.rb
+++ b/spec/services/vacancy_algolia_search_builder_spec.rb
@@ -1,5 +1,17 @@
 require 'rails_helper'
 
+RSpec.shared_examples 'a search in the base Vacancy index' do
+  it 'does not use any search replica' do
+    expect(subject.search_replica).to be_nil
+  end
+end
+
+RSpec.shared_examples 'a search in the default search replica' do
+  it 'uses the default search replica' do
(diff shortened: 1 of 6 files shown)


spec/helpers/vacancies_helper_spec.rb
```diff
@@ -26,6 +28,18 @@
         ]
       )
     end
+
+    it "does not include 'most relevant' when there is not keyword" do
```
cpjmcquillan, line 32: `no keyword` perhaps? Or `when keyword is empty`

spec/services/vacancy_algolia_search_builder_spec.rb
```diff
@@ -1,5 +1,17 @@
 require 'rails_helper'
 
+RSpec.shared_examples 'a search in the base Vacancy index' do
```
cpjmcquillan, line 3: 👍

PR #1699, opened 2020-06-26
Remove unused gems

cesidio:


alexbowen:
as an aside, i have wondered if any other ruby gems can be removed, there seems to be a lot of them

PR #1701, opened 2020-06-26
Location suggestion spike

cpjmcquillan:
This PR is an outcome of the spike, and implements an API endpoint that returns location autocomplete suggestions using the Google Places Autocomplete API.

The matched terms are returned as well as the location description, since the matched term will probably be useful when carrying out location search on the user selection. 

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-924

## Changes in this PR
- Add location suggestion service
- Add location suggestion endpoint

## Next steps
- Set up a production API key
- Use this API in the frontend to render suggestions and subsequently pass them to search

### DIFF
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -1,5 +1,6 @@
 class LocationSuggestion
   MINIMUM_LOCATION_INPUT_LENGTH = 3
+  NUMBER_OF_SUGGESTIONS = 5
 
   class InsufficientLocationInput < StandardError; end
   class MissingLocationInput < StandardError; end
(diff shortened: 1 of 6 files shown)


app/services/location_suggestion.rb
```diff
@@ -0,0 +1,49 @@
...
+  def initialize(location_input)
+    raise MissingLocationInput if location_input.nil?
+    raise InsufficientLocationInput if location_input.length < MINIMUM_LOCATION_INPUT_LENGTH
+    self.location_input = location_input
+  end
+
+  def suggest_locations
+    predictions = get_suggestions_from_google['predictions'].take(5)
+    suggestions = predictions.map { |prediction| prediction['description'] }
+    matched_terms = predictions.map {
```
cesidio, line 19: use do end when block spans on multiple lines

app/services/location_suggestion.rb
```diff
@@ -0,0 +1,50 @@
...
+    self.location_input = location_input
+  end
+
+  def suggest_locations
+    predictions = get_suggestions_from_google['predictions'].take(NUMBER_OF_SUGGESTIONS)
+    suggestions = predictions.map { |prediction| prediction['description'] }
+    matched_terms = predictions.map do
+      |prediction| prediction['terms'].select { |term| term['offset'] == 0 }.map { |hit| hit['value'] }
+    end
+    return suggestions, matched_terms
```
cesidio, line 23: no need to return on the last line of the method

cpjmcquillan: I thought when you want to return multiple values you do need to

cesidio: [suggestions, matched_terms]

david-mears-dfe:
> Set up a production API key

Done.

PR #1702, opened 2020-06-26
current-location

alexbowen:
ensure functionality isnt initialised on pages that dont need it

### DIFF
--- a/app/frontend/src/lib/currentLocation.js
+++ b/app/frontend/src/lib/currentLocation.js
@@ -96,9 +96,9 @@ const currentLocation = {
 
 export default currentLocation;
 
-if (navigator.geolocation) {
-  window.addEventListener('DOMContentLoaded', (event) => {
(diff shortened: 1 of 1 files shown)


app/frontend/src/lib/currentLocation.js
```diff
@@ -96,8 +96,8 @@ const currentLocation = {
 
 export default currentLocation;
 
-if (navigator.geolocation) {
-  window.addEventListener('DOMContentLoaded', (event) => {
+if (navigator.geolocation && document.querySelector('.js-location-finder')) {
+  window.addEventListener('DOMContentLoaded', () => {
```
cpjmcquillan, line 100: Should the DOMContentLoaded event listener go outside the if statement?

alexbowen: im not sure it would affect it too much, but it is better logic anyway so have updated

PR #1714, opened 2020-06-30
Run app without algolia env variables

saliceti:
## Changes in this PR:
Rake tasks may not require access to algolia so make the environment variables optional

## How to review
Run docker build

### DIFF
--- a/config/initializers/algoliasearch.rb
+++ b/config/initializers/algoliasearch.rb
@@ -10,12 +10,14 @@ See main README.md for instructions how to create new dev indexes.
   end
 end
 
-if !Rails.env.production? && ENV.fetch('ALGOLIA_APP_ID') == ALGOLIA_PRODUCTION_APP_ID
+if !Rails.env.production? && ENV['ALGOLIA_APP_ID'] == ALGOLIA_PRODUCTION_APP_ID
   raise AlgoliaEnvironmentError.new
 end
 
-AlgoliaSearch.configuration = {
-  application_id: ENV.fetch('ALGOLIA_APP_ID'),
-  api_key: ENV.fetch('ALGOLIA_WRITE_API_KEY'),
-  pagination_backend: :kaminari
-}
+if ENV['ALGOLIA_APP_ID']
+  AlgoliaSearch.configuration = {
+    application_id: ENV['ALGOLIA_APP_ID'],
+    api_key: ENV['ALGOLIA_WRITE_API_KEY'],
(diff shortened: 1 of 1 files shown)


config/initializers/algoliasearch.rb
```diff
@@ -10,12 +10,14 @@ def initialize
   end
 end
 
-if !Rails.env.production? && ENV.fetch('ALGOLIA_APP_ID') == ALGOLIA_PRODUCTION_APP_ID
+if !Rails.env.production? && ENV['ALGOLIA_APP_ID'] == ALGOLIA_PRODUCTION_APP_ID
```
cpjmcquillan, line 13: Could we use this instead?
```ruby
ENV.fetch(‘VALUE’, ‘’)
```

david-mears-dfe:
I have just now got a failing deploy, before this PR being merged:

```
rake aborted!
KeyError: key not found: "ALGOLIA_APP_ID"
/teacher-vacancy/config/initializers/algoliasearch.rb:13:in `fetch'
```

So if you get the same, it might **not** be caused by this PR.

config/initializers/algoliasearch.rb
saliceti: It returns nil if empty so it's useful on line 17 to do `if ENV['ALGOLIA_APP_ID']`
If you prefer I don't mind, I'll change it to `if ENV['ALGOLIA_APP_ID'] == ''`

config/initializers/algoliasearch.rb
```diff
@@ -10,12 +10,14 @@ def initialize
...
   raise AlgoliaEnvironmentError.new
 end
 
-AlgoliaSearch.configuration = {
-  application_id: ENV.fetch('ALGOLIA_APP_ID'),
-  api_key: ENV.fetch('ALGOLIA_WRITE_API_KEY'),
-  pagination_backend: :kaminari
-}
+if ENV['ALGOLIA_APP_ID']
```
david-mears-dfe, line 17: This line makes me worried that we will deploy an app which won't connect to Algolia. It could pass all the tests (because the tests have the env), and be deployed with no Algolia initialization without throwing an error.

david-mears-dfe: This would be caught by the smoke test, but that seems a little too late.

david-mears-dfe: Maybe it's an acceptable risk

PR #1717, opened 2020-07-01
Remove unused environment variable

david-mears-dfe:
We use `ALGOLIA_WRITE_API_KEY` everywhere. It has the same permissions as `ALGOLIA_SEARCH_API_KEY` and then some.


terraform/app/container-definitions/web_container_definition.json
```diff
@@ -209,10 +209,6 @@
       {
         "name": "ALGOLIA_WRITE_API_KEY",
         "value": "${algolia_write_api_key}"
-      },
```
cesidio, line 212: We probably do not need this file anymore @saliceti

saliceti: @david-mears-dfe can you remove the whole `terraform/container-definitions` directory please?

cpjmcquillan:
Will #1721 necessitate this API key still exists? Client search needs far fewer permissions and the search API key can be exposed in the client safely

david-mears-dfe:
> Will #1721 necessitate this API key still exists? Client search needs far fewer permissions and the search API key can be exposed in the client safely

Yes.

Issue #1723, opened 2020-07-01
Trust user selects which school the vacancy is for


tatyree:
[TEVA-683](https://dfedigital.atlassian.net/browse/TEVA-683)

## User story

**As a**
hiring staff user working for a trust

**I need**
to select which school the vacancy is at when I create a listing (and that includes the trust as an option)

**So that**
I can list a vacancy for any school I manage

## Details
Scenario 1: User logs in with URN = Skip step
Scenario 2: User logs in with UID = Show step
Scenario 1 and 2 cannot happen at the same time because DSI send us only one organisation ID when user logs in.

## Design/IA

Refer to the following Figma Screens for anything relating to design or IA.

[Figma Screen 1.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A739)
[Figma Screen 1.2](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A1039)

If you have any questions, clarify with Chris Taylor before proceeding. In the event of any corrections, please ensure the Figmas are updated and notes of conversations are recorded in this issue for future reference. 

### An additional step has been added to the **Create a job** journey: 
- [x] This user journey is feature-flagged and this entry point to the user journey does not appear if the feature flag is set to false
- [x] The page shown in [Figma Screen 1.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A739) exists and matches the design, including the copy. 
- [x] A sidebar element exists for this page as shown in [Figma Screen 1.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A739). 
- [x] If they select 'At the trust head office' radio button they are taken to Step 2
- [x] If they select 'At one school in the trust' radio button Step 1b, [Figma Screen 1.2](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A1039), is shown instead.
- [x] The user's selection is persisted to the data model and, if they come back to these pages, they are shown what they chose. 
- [x] The updated sidebar is shown on [Figma Screen 1.2](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A1039) and for all subsequent steps that show the sidebar. 
[...]

PR #1732, opened 2020-07-02
Allow Vacancy to belong to either a School or a SchoolGroup

### LINKED ISSUE #1731: Change vacancy so it can belong to a school or a school_group (aka trust)
## Background

An initial requirement of Multi Academy Trust (MAT) access to TVS is that trusts can list vacancies for jobs at the trust, proper, or any number of schools in the trust.  Our current vacancy data model does not fully enable this behaviour, so we need to change it such that a vacancy can belong to a trust or a school.  

Eventually, a vacancy will need to belong to a trust and/or any number of schools in that trust, but this is out of scope for this issue.

## Done when

 - [x] A vacancy can belong to a school **or** a trust (a school association is currently required)
 - [x] The owning organisation of the school is returned by a single method call (`school_or_trust`)
 - [ ] All occurrences of `Vacancy#school` have been changed to use this new method call
 - [x] The code has been PR'd and approved
 - [ ] The code has been merged to master 
 - [ ] The code has been released to production

## Notes

The current `school` association is declared here: 
https://github.com/DFE-Digital/teacher-vacancy-service/blob/master/app/models/vacancy.rb#L214

This code will only change back end implementation details.  Once the new method call has been implemented across the codebase it is safe to release to production as is regardless of the implementation status of the other MAT tickets.

cpjmcquillan:
This PR changes the data structure to allow a vacancy to belong to either a School or a SchoolGroup. There is no constraint on this at a table level, but validation should be added as part of #1723.

This PR does **not** update all instances of `vacancy.school` to use `vacancy.school_or_school_group`. These methods should be updated as and when the feature continues to be implemented, so that the relevant methods and properties can be added to the SchoolGroup model.

## Issues
Closes #1731 

## Changes in this PR
- Allow school to be null for a vacancy
- Allow a vacancy to belong to a school or a school group (there is no validation on the model, this should be implemented on the form object)
- Add a school_or_school_group method
- Update delegated school methods
- Add a temporary name method to the SchoolGroup model


david-mears-dfe:
I wonder how we will validate that vacancies have exactly one of the `school_id` and `school_group_id` attributes set.

We can do custom validation by [class](https://edgeguides.rubyonrails.org/active_record_validations.html#performing-custom-validations) or by [method](https://edgeguides.rubyonrails.org/active_record_validations.html#custom-methods) or even by  [conditional validation](https://edgeguides.rubyonrails.org/active_record_validations.html#conditional-validation).

david-mears-dfe:
> I wonder how we will validate that vacancies have exactly one of the `school_id` and `school_group_id` attributes set.
> 
> We can do custom validation by [class](https://edgeguides.rubyonrails.org/active_record_validations.html#performing-custom-validations) or by [method](https://edgeguides.rubyonrails.org/active_record_validations.html#custom-methods) or even by [conditional validation](https://edgeguides.rubyonrails.org/active_record_validations.html#conditional-validation).

Ignore this, I realise now that this is only going to be a temporary situation.

cpjmcquillan:
> > I wonder how we will validate that vacancies have exactly one of the `school_id` and `school_group_id` attributes set.
> > We can do custom validation by [class](https://edgeguides.rubyonrails.org/active_record_validations.html#performing-custom-validations) or by [method](https://edgeguides.rubyonrails.org/active_record_validations.html#custom-methods) or even by [conditional validation](https://edgeguides.rubyonrails.org/active_record_validations.html#conditional-validation).
> 
> Ignore this, I realise now that this is only going to be a temporary situation.

```ruby
school.present? ^ school_group.present?
```

app/models/vacancy.rb
```diff
@@ -144,7 +144,12 @@ def self.remove_vacancies_that_expired_yesterday!
         religious_character: school.gias_data['ReligiousCharacter (name)'],
         region: school.region&.name,
         school_type: school.school_type&.label&.singularize,
-        town: school.town }
+        town: school.town } if school.present?
+    end
+
+    attribute :school_group do
+      school_group = self.school_group
+      { uid: school_group&.uid }
```
david-mears-dfe, line 152: is defining the method `uid` in scope?

spec/models/vacancy_spec.rb
```diff
@@ -382,12 +383,23 @@
     end
   end
 
-  describe 'delegate school_name' do
-    it 'should return the school name for the vacancy' do
-      school = create(:school, name: 'St James School')
-      vacancy = create(:vacancy, school: school)
+  describe '#school_or_school_group_name' do
```
david-mears-dfe, line 386: can we have the same for the method `#school_or_school_group_geolocation` which you've used?

app/models/vacancy.rb
cpjmcquillan: We don't need a `uid` method?

spec/models/vacancy_spec.rb
cpjmcquillan: You could, but it would always be nil at the minute since there is no `geolocation` method on `SchoolGroup`

app/models/vacancy.rb
david-mears-dfe: Nonetheless, it looks like you are using it

spec/models/vacancy_spec.rb
cpjmcquillan: Also, there was no test for the delegate geolocation in the first place

app/models/school_group.rb
```diff
@@ -1,4 +1,11 @@
 class SchoolGroup < ApplicationRecord
   has_many :school_group_memberships
   has_many :schools, through: :school_group_memberships
+
+  has_many :vacancies
+
+  # TODO: This should ideally be a stored as column in the database table
```
tatyree, line 7: Not sure I agree with this as it's just copying source-of-truth for no real benefit.

app/views/vacancies/_school_overview_section.html.haml
```diff
@@ -61,7 +61,7 @@
   - if @vacancy.school.geolocation
     %div{ id: 'map_zoom', role: 'presentation', aria: { hidden: 'true', label: t('schools.map_aria_label') },  }
     = render partial: '/vacancies/school', formats: [:js], locals: { name: @vacancy.school.name,
-                                                          lat: @vacancy.school_geolocation.x,
-                                                          lng: @vacancy.school_geolocation.y }
+                                                          lat: @vacancy.school_or_school_group_geolocation.x,
```
david-mears-dfe, line 64: I'm confused about if you are using methods that don't exist.

app/models/school_group.rb
cpjmcquillan: The thinking behind the TODO is that it would make it easier to lookup `SchoolGroups` by name if you ever wanted to do that

app/models/vacancy.rb
tatyree: I know `&.` is the uber concise way to do this, but it does make the code harder to read.  Can we please go with `school_group.try(:uid)` instead?  It looks less like perl-ish line noise.

tatyree: Also, I'm not entirely clear on what this block is supposed to be doing anymore.  Is it assigning the school_group object to the Algolia index, or is only yielding `{ uid: <some uid> }` presumably as JSON?

app/models/vacancy.rb
```diff
@@ -179,11 +184,11 @@ def self.remove_vacancies_that_expired_yesterday!
   # rubocop:enable Metrics/BlockLength
 
   def lat
-    self.school.geolocation&.x&.to_f
+    self.school.geolocation&.x&.to_f if school.present?
   end
 
   def lng
-    self.school.geolocation&.y&.to_f
+    self.school.geolocation&.y&.to_f if school.present?
```
tatyree, line 191: See my comment on `&.` vs `.try(...)`, above.

app/views/vacancies/_school_overview_section.html.haml
cpjmcquillan: Getting the locations for `SchoolGroups` isn't in the scope of this ticket as I read it?

app/models/vacancy.rb
```diff
@@ -179,11 +184,11 @@ def self.remove_vacancies_that_expired_yesterday!
   # rubocop:enable Metrics/BlockLength
 
   def lat
-    self.school.geolocation&.x&.to_f
+    self.school.geolocation&.x&.to_f if school.present?
```
tatyree, line 187: See my comment on `&.` vs `.try(...)`, above.

app/views/vacancies/_school_overview_section.html.haml
tatyree: I am, too.

tatyree: Is this a misfiring replaceall?

tatyree: It's coming from a delegation on `Vacancy` line 231.  Personally, I'm not a fan of this much metaprogramming relational magic, but I'm happy that it addresses the need right now.

cpjmcquillan:
@tatyree I can live without the delegated geolocation method completely for now, especially since the corresponding method doesn't exist yet for `SchoolGroup`. The algolia block is a placeholder to highlight the fact that `school_group` will likely need to be indexed in a similar way to schools are now - again this can be removed for now.

app/models/vacancy.rb
cpjmcquillan: @tatyree the algolia block here was a placeholder to indicate that the school_group will likely need to be indexed similarly to the school. It's not in the scope of this ticket, so I will remove it

cpjmcquillan: @david-mears-dfe It's a field on the record, not a method

app/views/vacancies/_school_overview_section.html.haml
david-mears-dfe: Can you copy (or permalink) me the line on Vacancy please @tatyree as I can't see it and would like to understand what is creating this method.

PR #1734, opened 2020-07-02
NodeJS engine version altered to min version

alexbowen:
node and npm version altered to min version


david-mears-dfe:
Worth mentioning under 'dependencies' in README?

PR #1743, opened 2020-07-03
[TEVA-903] enable radius select when input contains a number

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-903

By default, the dropdown 'Within radius of' is hidden when the location box is empty or when it only contains letters.

'Within radius of' appears INSTANTLY when user inputs a number in the location field.

'Within radius of' appears INSTANTLY if user clicks on 'Use my current location'

'Within radius of' disappears INSTANTLY if user deletes the number from the location box.

## Changes in this PR:

- add custom event handler to index of search bundle (I do think the handle is better in a seperate file to bundle differently as it shouldnt just be bundles just with the search functionality but i have an idea of what to do that warrant another PR. the acceptance criteria for this immediate ticket are met however

- add utility function to detect for number and use this to decide if radius is shown when event is fired

- associated JS unit tests

- over rode the `keyword-field` and reverted to keyword for consistency

### DIFF
--- a/app/frontend/src/lib/utils.test.js
+++ b/app/frontend/src/lib/utils.test.js
@@ -1,5 +1,5 @@
 import {
-  getUnixTimestampForDayStart, constructNewUrlWithParam, stringMatchesPostcode, convertMilesToMetres, convertEpochToUnixTimestamp, extractQueryParams,
+  getUnixTimestampForDayStart, constructNewUrlWithParam, stringMatchesPostcode, convertMilesToMetres, convertEpochToUnixTimestamp, extractQueryParams, stringContainsNumber
 } from './utils';
 
 describe('constructNewUrlWithParams', () => {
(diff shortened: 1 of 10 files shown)


app/frontend/src/search/index.js
```diff
@@ -128,3 +128,7 @@ renderAutocomplete({
     updateUrlQueryParams('location', value, window.location.href);
   },
 });
+
+document.getElementById('location').addEventListener('input', (e) => {
```
cpjmcquillan, line 132: What happens if this element isn't there? It probably should be, but this could _potentially_ be executed before the DOM has loaded couldn't it?

app/frontend/src/lib/utils.test.js
```diff
@@ -55,6 +55,13 @@ describe('stringMatchesPostcode', () => {
   });
 });
 
+describe('stringContainsNumber', () => {
+  test('is true when supplied string contains a number otherwise false', () => {
+    expect(stringContainsNumber('alex8')).toBe(true);
+    expect(stringContainsNumber('alex')).toBe(false);
```
cpjmcquillan, line 61: This could be 2 separate tests I think

alexbowen: i had excatly the same thought, yes you are right but at first i couldnt be arsed then i thought its petty. but ill split it will take 2mins

app/frontend/src/search/index.js
alexbowen: doesnt do any harm, whilst im not sure totally neccesary as it use betElementById, im sure ie9 or something might say 'why didnt you just do it'?

alexbowen: actually this does work

PR #1746, opened 2020-07-06
Trust user step 1

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-683

## GitHub issue
https://github.com/DFE-Digital/teacher-vacancy-service/issues/1723

## Changes in this PR

Adds step 1a and 1b for SchoolGroup-level hiring staff: job location.

I have not added feature tests for creating and publishing a vacancy, because updating the 'job summary' and 'review' and 'preview' steps (required to publish) is out of scope for this ticket. But the new step 1 is covered by feature tests for editing a draft vacancy.

## Demo video

https://drive.google.com/file/d/1bhcYseLYr8OsZWkT6qHlSbc5eAIgOZh1/view?usp=sharing

### An additional step has been added to the **Create a job** journey: 
- [x] This user journey is feature-flagged and this entry point to the user journey does not appear if the feature flag is set to false
- [x] The page shown in [Figma Screen 1.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A739) exists and matches the design, including the copy. 
- [x] A sidebar element exists for this page as shown in [Figma Screen 1.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A739). 
- [x] If they select 'At the trust head office' radio button they are taken to Step 2
- [x] If they select 'At one school in the trust' radio button Step 1b, [Figma Screen 1.2](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A1039), is shown instead.
- [x] The user's selection is persisted to the data model and, if they come back to these pages, they are shown what they chose. 
- [x] The updated sidebar is shown on [Figma Screen 1.2](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=229%3A1039) and for all subsequent steps that show the sidebar. 
- [x] Happy path tests exist to cover this interaction (unit tests are sufficient, but this additional tests are left to assignee's discretion)
 - [x] An error is triggered if the user tries to submit either page without first selecting an option. 
     - [x] Content Design has provided an error message for this step.  
- [x] The rendered pages **exactly match** the listed figmas.
 - [ ] The code has been signed off on staging (or other suitable test environment) by: 
    - [ ] Design/IA
    - [ ] Product
    - [ ] Content
  - [ ] A suitable PR has been raised and approved
  - [ ] The code has been merged to master and deployed to production.  **NOTE** the code **does not** need to be switched on for this issue to be considered complete.


david-mears-dfe:
I would appreciate a review of this work in progress PR, before it gets too big to manage. At the moment it adds the page you can see on the Figma ([Step 1a](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=0%3A1 )).

Tomorrow I'll work on Step 1b, which should be less complex.

There is complexity in the first page because it's the first form, for only some user journeys, and because it makes the step-sidebar numbers dynamic (depending on whether the first step is used).

Also, now that we are using SchoolGroups sometimes instead of Schools, we might (later) need to rethink our RESTful routes, which currently are like `/school/jobs/123/job_specification`. We could just get rid of the `school` part.

(FYI Joe, the SchoolGroup model is to represent things like Multi-Academy Trusts.)

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -28,12 +28,20 @@ def current_school
...
+  end
+
   def current_user
     return if current_session_id.blank?
 
     @current_user ||= User.find_or_create_by(oid: current_session_id)
   end
 
+  def school_group_user?
+    session[:uid] != '' && session[:uid] != nil
```
cesidio, line 42: you can use `session[:uid].present?`

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -28,12 +28,20 @@ def current_school
     @current_school ||= School.find_by!(urn: session[:urn])
   end
 
+  def current_school_group
```
cpjmcquillan, line 31: This and the above method are refactored out to avoid repetition in #1749

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -18,14 +20,18 @@ def session_vacancy_id
 
   def set_vacancy
     if params[:job_id]
-      @vacancy = current_school.vacancies.find(params[:job_id])
+      @vacancy = Vacancy.find(params[:job_id])
```
cpjmcquillan, line 23: `current_organisation.vacancies` ensures only vacancies for the current School or SchoolGroup can be found

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -34,8 +40,7 @@ def store_vacancy_attributes(attributes)
   end
 
   def update_vacancy(attributes, vacancy = nil)
-    vacancy ||= current_school.vacancies.find(session_vacancy_id)
-
+    vacancy ||= Vacancy.find(session_vacancy_id)
```
cpjmcquillan, line 43: `current_organisation.vacancies`

app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+  private
+
+  def set_up_url
+    @job_location_url_method = @vacancy&.id.present? ? 'patch' : 'post'
+    @job_location_url = @vacancy&.id.present? ?
+      school_job_job_location_path(@vacancy.id) : job_location_school_job_path(school_group_id: current_school_group.id)
+  end
+
+  def set_up_job_location_options
+    @job_location_options = [
```
cpjmcquillan, line 51: Probably better to move this away from the controller

app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def redirect_to_school_selection_or_next_step
+    if session[:current_step].eql?(:review) && @job_location_form.job_location == 'at_one_school'
+      redirect_to school_job_school_path(@vacancy.id)
+    else
+      redirect_to_next_step_if_save_and_continue(@vacancy.id, @vacancy.job_title)
+    end
+  end
+
+  def save_vacancy_without_validation
```
cpjmcquillan, line 84: If these methods are significantly similar to the job_specification methods, can we refactor and reuse?

app/helpers/hiring_staff/job_creation_helper.rb
```diff
@@ -1,14 +1,27 @@
 module HiringStaff::JobCreationHelper
   def current_step
-    params[:create_step]
+    step = params[:create_step]
```
cpjmcquillan, line 3: Can we reuse this helper in the application_controller to avoid repeating ourselves?

app/helpers/hiring_staff/job_creation_helper.rb
```diff
@@ -27,4 +40,8 @@ def set_visited_step_class(step_number, vacancy)
     return 'app-step-nav__step--visited' if
       step_number != current_step && step_number <= completed_step
   end
+
+  def school_group_user?
```
cpjmcquillan, line 44: As above

app/models/vacancy.rb
```diff
@@ -358,6 +358,7 @@ def slug_candidates
       :job_title,
       %i[job_title school_or_school_group_name],
       %i[job_title location],
+      :id
```
cpjmcquillan, line 361: Is this because you need to set the slug without a job title being specified? I think you can wait to set the slug until the job_specification page

app/views/hiring_staff/vacancies/vacancy_form_partials/_submit.html.haml
```diff
@@ -2,4 +2,5 @@
   = f.govuk_submit t('buttons.update_job'), classes: 'update-listing-gtm'
 - else
   = f.govuk_submit classes: 'govuk-!-margin-bottom-5 save-listing-gtm'
-  = f.govuk_submit t('buttons.save_and_return_later'), secondary: true, classes: 'button-link save-and-return-listing-gtm'
+  - unless @job_location_form.present? # || the other form
```
cpjmcquillan, line 5: I'd prefer to the keep the partial as simple as possible, and make the distinction in the view for the job_location_form instead - if that makes sense?

config/initializers/flags.rb
```diff
@@ -6,8 +6,8 @@
 FEATURE_IMPORT_VACANCIES = ENV['FEATURE_IMPORT_VACANCIES']
 FEATURE_SCHOOL_GROUP_JOBS = ENV['FEATURE_SCHOOL_GROUP_JOBS']
 
-AuthenticationFallback = FeatureFlag.new('authentication_fallback', is_feature: false)
-ReadOnlyFeature = FeatureFlag.new('read_only')
-EmailAlertsFeature = FeatureFlag.new('email_alerts')
-ImportVacanciesFeature = FeatureFlag.new('import_vacancies')
-SchoolGroupJobsFeature = FeatureFlag.new('school_group_jobs')
+AuthenticationFallback = Flag.new('authentication_fallback', is_feature: false)
```
cpjmcquillan, line 9: 👍

config/locales/en.yml
```diff
@@ -289,12 +294,14 @@ en:
     per_year_pro_rata: 'per year pro rata'
     per_year_fte: 'per year (full-time equivalent)'
     create_a_job_title: 'Create a job listing for %{school}'
```
cpjmcquillan, line 296: This could end up as `Create a job listing for %{organisation}`

db/schema.rb
```diff
@@ -10,13 +10,14 @@
...
 
-ActiveRecord::Schema.define(version: 2020_07_02_125917) do
+ActiveRecord::Schema.define(version: 2020_07_06_172428) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "fuzzystrmatch"
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
+  enable_extension "postgis"
```
cpjmcquillan, line 20: Not sure how/why this and the table below have crept in?

app/models/vacancy.rb
david-mears-dfe: It is because of that.

db/schema.rb
david-mears-dfe: Good catch, it was stuff that entered my local database after I put a production dump into it. But I don't know why the production database has this stuff in it. @cesidio do you know why?

app/controllers/hiring_staff/vacancies/application_controller.rb
david-mears-dfe: This was almost very bad security

app/controllers/hiring_staff/base_controller.rb
david-mears-dfe: Given that, I'll get rid of the method.

app/controllers/hiring_staff/vacancies/application_controller.rb
david-mears-dfe: I'll change this when #1749 is merged to master.

app/models/vacancy.rb
cpjmcquillan: Setting the slug as the id doesn't make much sense to me, we may as well not set a slug at that point

david-mears-dfe: It's because the slug can't be null, at the moment. The slug is refreshed whenever the vacancy is updated, so as soon as we can have a meaningful text, it will use that. Worth making `slug` nullable?

cpjmcquillan: Yeh I think that makes more sense

spec/features/hiring_staff_can_edit_a_draft_vacancy_spec.rb
```diff
@@ -7,223 +7,285 @@
...
+            )
+          end
+        end
       end
 
-      def edit_a_published_vacancy
-        published_vacancy = create(:vacancy, :published, school: school)
-        visit edit_organisation_job_path(published_vacancy.id)
-        click_header_link(I18n.t('jobs.application_details'))
+      context 'as a SchoolGroup-level user' do
```
david-mears-dfe, line 232: GitHub hid this file for being a large diff. That was mainly from indentation changes.

Here I added meaningful diff.

david-mears-dfe:
The last reviews were I believe on the code as it stood before the commit called `♻️ Refactor`.

db/schema.rb
cesidio: Not sure why, as the extensions we ask for are only these
`"enable_extensions": ["pgcrypto", "fuzzystrmatch", "plpgsql"]`
maybe it's a default on PaaS

app/controllers/concerns/first_step_form_concerns.rb
```diff
@@ -0,0 +1,20 @@
+# Methods shared by controllers which can serve Step 1
+# HiringStaff::Vacancies::JobSpecificationController is school-level users' Step 1
+module FirstStepFormConcerns
+  def set_up_form(form_class)
+    @form = form_class.new(form_params)
+  end
+
+  def save_form_params_on_vacancy_without_validation
+    @form.vacancy.status = :draft
+    Auditor::Audit.new(@form.vacancy, 'vacancy.create', current_session_id).log do
```
cpjmcquillan, line 10: I think it's worth talking with @stevenleggdfe about whether we can remove these sorts of things from the codebase, and rely on GA

app/controllers/concerns/first_step_form_concerns.rb
```diff
@@ -0,0 +1,20 @@
+# Methods shared by controllers which can serve Step 1
+# HiringStaff::Vacancies::JobSpecificationController is school-level users' Step 1
+module FirstStepFormConcerns
```
cpjmcquillan, line 3: This module isn't a concern, should it be named differently (and moved somewhere else)?

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -139,4 +132,8 @@ def add_errors_to_form(errors, form_object)
       form_object.errors.add(field, error)
     end
   end
+
+  def redirect_unless_school_group_user_flag_on
+    redirect_to job_specification_organisation_job_path(request.parameters) unless SchoolGroupJobsFeature.enabled?
```
cpjmcquillan, line 137: Could we be explicit about what we're passing here (i.e `school_id: id` or `school_group_id: id`)? It feels wrong to blindly pass through all request parameters

david-mears-dfe: I'm not sure what parameters will be used. My intention was to redirect any requests to a path blocked by the flag, and redirect the parameters whatever they may be. Should I explicitly define what parameters are possible?

app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      @form = JobLocationForm.new(session[:vacancy_attributes])
+    else
+      @form = JobLocationForm.new(school_group_id: current_school_group.id)
+    end
+  end
+
+  def create
+    store_vacancy_attributes(@form.vacancy.attributes)
+
+    if @form.complete_and_valid?
```
cpjmcquillan, line 23: This method is only necessary at the minute because of the way we handle dates. So both here and in `JobSpecificationController` we should be able to use @form.valid? It might be worth adding a comment to the `complete_and_valid?` method explaining that (or you can leave that to me)

app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      session_vacancy_id ? update_vacancy(form_params) : save_vacancy_without_validation
+      store_vacancy_attributes(@form.vacancy.attributes)
+      return redirect_to_next_step_if_continue(@vacancy&.id.present? ? @vacancy.id : session_vacancy_id)
```
cpjmcquillan, line 26: I think we're moving to a standard approach of `if... render, else... render` so let's change when we see this more pythonic 🐍  style

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -80,13 +86,20 @@ def step_valid?(step_form)
...
 
   def redirect_to_incomplete_step
-    return redirect_to organisation_job_job_specification_path(@vacancy.id) unless step_valid?(JobSpecificationForm)
-    return redirect_to organisation_job_pay_package_path(@vacancy.id) unless step_valid?(PayPackageForm)
-    return redirect_to organisation_job_important_dates_path(@vacancy.id) unless step_valid?(ImportantDatesForm)
-    return redirect_to organisation_job_supporting_documents_path(@vacancy.id) unless
-      step_valid?(SupportingDocumentsForm)
-    return redirect_to organisation_job_application_details_path(@vacancy.id) unless step_valid?(ApplicationDetailsForm)
-    return redirect_to organisation_job_job_summary_path(@vacancy.id) unless step_valid?(JobSummaryForm)
+    computer_readable_step_names.each do |step|
```
cpjmcquillan, line 89: This is very nice, but does it make the code better? I'd argue it obfuscates the intention of the code and makes it less readable. If the pros > cons let's do it, but I'd like to hear what others think about it 👍

app/views/hiring_staff/vacancies/application_details/show.html.haml
```diff
@@ -1,9 +1,9 @@
 - content_for :page_title_prefix, page_title_prefix(@vacancy, @application_details_form, t('jobs.application_details'))
-= render 'hiring_staff/vacancies/vacancy_form_partials/title'
+= render 'hiring_staff/vacancies/vacancy_form_partials/heading'
 
 - if @vacancy.present? && @vacancy&.state != 'create'
   = link_to t('buttons.cancel_and_return'), edit_organisation_job_path(@vacancy.id), class: 'govuk-back-link'
-- else
+- elsif @vacancy.present?
```
cpjmcquillan, line 6: What's the reason for this change?

app/views/hiring_staff/vacancies/application_details/show.html.haml
```diff
@@ -35,7 +35,8 @@
         label: { text: t('helpers.fieldset.application_details_form.application_link'), size: 's' },
         hint_text: t('helpers.hint.application_details_form.application_link')
 
-      = render 'hiring_staff/vacancies/vacancy_form_partials/submit', f: f
+      = render 'hiring_staff/vacancies/vacancy_form_partials/continue_or_update_submit', f: f
```
cpjmcquillan, line 38: I prefer the submit buttons being a single partial, but that might just be me

config/initializers/govuk_design_system_formbuilder.rb
```diff
@@ -1,5 +1,5 @@
 GOVUKDesignSystemFormBuilder.configure do |conf|
-  conf.default_error_summary_title  = 'Please correct the following errors'
+  conf.default_error_summary_title  = 'There is a problem'
```
cpjmcquillan, line 2: I think `There is a problem` is a placeholder in the designs. I'd be hesitant changing this and want to be absolutely confident this is what Content wants

db/schema.rb
```diff
@@ -13,11 +13,9 @@
 ActiveRecord::Schema.define(version: 2020_07_10_080945) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "citext"
```
cpjmcquillan, line 16: What's the reason behind these being removed?

spec/support/auth_helpers.rb
```diff
@@ -16,8 +16,8 @@ def stub_env_for_staging_auth(env_field_for_username: :http_user,
   end
 
   def stub_hiring_staff_auth(urn: nil, uid: nil, session_id: 'session_id', email: nil)
-    page.set_rack_session(urn: urn) if urn.present?
-    page.set_rack_session(uid: uid) if uid.present?
+    urn.present? ? page.set_rack_session(urn: urn) : page.set_rack_session(urn: '')
```
cpjmcquillan, line 19: Maybe this could be
```ruby
if urn.present?
  page.set_rack_session(urn: urn)
elsif uid.present?
  page.set_rack_session(uid: uid)
end
```

config/initializers/govuk_design_system_formbuilder.rb
cpjmcquillan: And even if this is something that UX/Content want, can we raise a separate PR for it? Scope creep = PR bloat = hard to review PRs

app/views/hiring_staff/vacancies/application_details/show.html.haml
david-mears-dfe: I put a note in the commit message for `Prevent nil errors in views`: 
```
    I noticed that although in each case the first
    conditional checked for the presence of @vacancy, the else-case
    did not do likewise.
    This commit makes the two parts of the conditional consistent
    by having them both check for @vacancy.present? before using
    @vacancy.
```

david-mears-dfe: We could in theory get rid of @vacancy.present? in both parts of the conditional, as line 1 will throw a nil error in any case.

app/controllers/hiring_staff/vacancies_controller.rb
david-mears-dfe: I probably got carried away here. Not sure if it makes code better - prevents a very small need for upkeep if we add more steps to hiring staff journey.

app/controllers/concerns/first_step_form_concerns.rb
david-mears-dfe: I wonder where to. @cesidio resident Rubyist, got any opinions?

config/initializers/govuk_design_system_formbuilder.rb
david-mears-dfe: Confirmed with content. I can do a separate PR.
For my own reference: `f0be6f1509aa9d80ae0ab44cc85ef65264959cf9` is the main commit.

db/schema.rb
david-mears-dfe:
```
commit 5209cd7703b503f128b8d1330530618444a21d66
Author: David Mears <david.mears@digital.education.gov.uk>
Date:   Wed Jul 8 12:28:56 2020 +0100

    Remove database extensions

    These were added 13 days ago by tatyree.
    If I drop my database and recreate and migrate, these lines
    no longer exist.
    They most likely were created by tatyree importing a
    production database dump into his local database.
```

spec/support/auth_helpers.rb
david-mears-dfe: We need to explicitly set `session[:urn/:uid]` to `''`, to avoid the impossible state of having both present. I will use `if` and `elsif` to eliminate the unnecessarily checking both uid and urn.

README.md
```diff
@@ -301,6 +301,14 @@ Populate your environment with real school data. This is taken from
 rake data:schools:import
 ```
 
+To get trust data:
```
cesidio, line 304: There should be a rake task for this as well, the same as we do for importing schools

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -4,9 +4,7 @@
 class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseController
   before_action :set_vacancy
 
-  def school_id
-    params.permit![:school_id]
-  end
+  include HiringStaff::JobCreationHelper
```
cesidio, line 7: include statements should always come first after class declaration

app/controllers/hiring_staff/vacancies/application_controller.rb
cesidio: I agree with Connor here

db/schema.rb
cesidio: if your migration is not involved in that, leave it like it is. We have different extensions on our local machine and PaaS postgres service, that's why you end up with changes in that file. Please only commit the changes that are the effect of your migrations. Do not commit changes that are a result of the different databases

app/views/hiring_staff/vacancies/application_details/show.html.haml
david-mears-dfe: Not possible because some pages have the two buttons and some have just the one.

README.md
david-mears-dfe: Yes, there should be

app/controllers/hiring_staff/vacancies/application_controller.rb
david-mears-dfe: Where should this redirect, if the flag is on? I don't know how the person would find themselves on this page if the flag is on. Maybe just redirect to homepage?

cpjmcquillan: We need to agree an approach, but the most sensible place to redirect - from my perspective - would be the dashboard

README.md
cpjmcquillan: Indeed, there is

app/controllers/concerns/first_step_form_concerns.rb
david-mears-dfe: I'm putting the methods in application_controller
(I couldn't sleep)

app/controllers/hiring_staff/vacancies/job_location_controller.rb
david-mears-dfe: Added comment `# This method is only necessary for forms with specific error messages for date inputs.` Does that sound right?
(Couldn't sleep)

app/controllers/hiring_staff/vacancies_controller.rb
david-mears-dfe: Reverted

app/views/hiring_staff/vacancies/application_details/show.html.haml
david-mears-dfe: > We could in theory get rid of @vacancy.present? in both parts of the conditional, as line 1 will throw a nil error in any case.

Tried this, didn't work.

config/initializers/govuk_design_system_formbuilder.rb
david-mears-dfe: https://github.com/DFE-Digital/teacher-vacancy-service/pull/1771

README.md
david-mears-dfe: I can't see it. Reverting commit.

david-mears-dfe:
As far as I can see, GitHub has deleted a lot of @cesidio and @cpjmcquillan's comments, and my replies. I had clicked resolved on all of them.

app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      @form = JobLocationForm.new(session[:vacancy_attributes])
+    else
+      @form = JobLocationForm.new(school_group_id: current_school_group.id)
+    end
+  end
+
+  def create
+    store_vacancy_attributes(@form.vacancy.attributes)
+
+    if @form.valid?
```
cesidio, line 21: Nice, thanks for using this standard rails if...redirect...else...render

app/controllers/hiring_staff/vacancies/job_specification_controller.rb
```diff
@@ -4,40 +4,42 @@ class HiringStaff::Vacancies::JobSpecificationController < HiringStaff::Vacancie
...
+      store_vacancy_attributes(@form.vacancy.attributes)
       return redirect_to_next_step_if_continue(@vacancy&.id.present? ? @vacancy.id : session_vacancy_id)
     end
 
     render :show
   end
 
   def update
-    if @job_specification_form.valid?
+    if @form.valid?
```
cesidio, line 40: Can we have here too the standard if...redirect...else...render?

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -30,9 +34,11 @@ def review
     reset_session_vacancy!
     store_vacancy_attributes(@vacancy.attributes)
 
+
     unless @vacancy.valid?
```
cesidio, line 38: Maybe this could be a one-liner, the same as on line 24, to keep it consistent

david-mears-dfe: There are two lines in the `else` part, so I'm not sure how it could be condensed!

david-mears-dfe: (Which is clearer when I get rid of the extra newline in the middle)

Issue #1733, opened 2020-07-02
Trust user sees all schools in the trust when they first log in


tatyree:
[TEVA-680](https://dfedigital.atlassian.net/browse/TEVA-680)

## Background

Trust users managing vacancies for the trust need to be able to see all the schools in their trust when they first log in. They also need to be able to select the subset of these school they would like to manage. 

## Workflows
1. User logs in with URN = show them vacancies for that URN.
1. User logs in with UID = show them vacancies for all URNs linked to that UID.

**NOTE:** These cannot happen at the same time because DSI send us only one organisation ID when user logs in.

The user journey is feature-flagged and this entry point to the user journey does not appear if the feature flag is set to false

When user linked to a UID logs in for the first time, they are presented with a interstitial screen.

## Design/IA

Refer to the following Figma Screens for anything relating to design or IA.

[Figma Screen 6.0](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=233%3A0)
[Figma Screen 6.0.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=100%3A0)
[Figma Screen 6.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=233%3A116)

If you have any questions, clarify with Chris Taylor before proceeding.  In the event of any corrections, please ensure the Figmas are updated and notes of conversations are recorded in this issue for future reference. 

## Done when 
- [x] This user journey is feature-flagged and this entry point to the user journey does not appear if the feature flag is set to false
    - [x] This user journey uses the same feature flag as the other user journeys in the MAT Access epics.
- [x] A page exists based on [Figma Screen 6.0](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=233%3A0).
- [x] A user linked to a UID sees this page the first time they log in
- [x] That user can click 'Skip this step',  and they are taken to [Figma Screen 6.1](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=233%3A116), which is the dashboard. 
    - [ ] That user then sees all the schools in the trust
- [x] That user can select options from[Figma Screen 6.0](https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=233%3A0) and click 'Save and continue'  
[...]

PR #1749, opened 2020-07-07
Allow SchoolGroup users to sign in

cpjmcquillan:
This PR allows SchoolGroup users to sign in with DfE sign in, as well as the AuthenticationFallback. Unless the SchoolGroupJobsFeature is enabled, this will have no effect on the application. If the feature is enabled, a SchoolGroup user can sign in and is redirected to a placeholder page - but no other functionality will work yet.

This PR is part of the work required to complete #1733.

## Changes in this PR
- Allow SchoolGroup users to sign in with DfE sign in
- Allow SchoolGroup users to sign in with Authentication fallback

## Next steps
- Incrementally remove the dependency on vacancies and users being associated to a School, so that the feature can be enabled

### DIFF
--- a/app/controllers/concerns/authentication_concerns.rb
+++ b/app/controllers/concerns/authentication_concerns.rb
@@ -21,10 +21,10 @@ module AuthenticationConcerns
   end
 
   def current_school
-    @current_school ||= School.find_by!(urn: session[:urn])
+    @current_school ||= School.find_by!(urn: session[:urn]) if session[:urn].present?
   end
 
   def current_school_group
-    @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid])
+    @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid]) if session[:uid].present?
   end
 end
--- a/app/controllers/concerns/authentication_concerns.rb
+++ b/app/controllers/concerns/authentication_concerns.rb
@@ -13,11 +13,7 @@ module AuthenticationConcerns
   end
 
   def current_organisation
-    if SchoolGroupJobsFeature.enabled?
-      current_school_group || current_school
-    else
-      current_school
-    end
+    current_school || current_school_group
   end
 
   def current_school
@@ -25,6 +21,8 @@ module AuthenticationConcerns
   end
 
   def current_school_group
-    @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid]) if session[:uid].present?
+    if SchoolGroupJobsFeature.enabled?
+      @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid]) if session[:uid].present?
+    end
   end
 end
--- a/app/controllers/concerns/authentication_concerns.rb
+++ b/app/controllers/concerns/authentication_concerns.rb
@@ -3,14 +3,28 @@ module AuthenticationConcerns
 
   included do
     helper_method :authenticated?
+    helper_method :current_organisation
     helper_method :current_school
+    helper_method :current_school_group
   end
 
   def authenticated?
     session[:session_id].present?
   end
 
+  def current_organisation
+    if SchoolGroupJobsFeature.enabled?
+      current_school_group || current_school
+    else
+      current_school
+    end
+  end
+
   def current_school
-    School.find_by!(urn: session[:urn])
+    @current_school ||= School.find_by!(urn: session[:urn])
+  end
+
+  def current_school_group
+    @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid])
   end
 end
--- a/app/controllers/hiring_staff/base_controller.rb
+++ b/app/controllers/hiring_staff/base_controller.rb
@@ -49,6 +49,9 @@ class HiringStaff::BaseController < ApplicationController
   end
 
   def redirect_signed_in_users
+    if SchoolGroupJobsFeature.enabled?
+      return redirect_to school_group_temporary_path if session.key?(:uid)
+    end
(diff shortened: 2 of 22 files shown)


app/controllers/hiring_staff/base_controller.rb
```diff
@@ -24,10 +22,6 @@ def check_terms_and_conditions
     redirect_to terms_and_conditions_path unless current_user&.accepted_terms_and_conditions?
   end
 
-  def current_school
```
david-mears-dfe, line 27: Good catch

app/controllers/hiring_staff/sign_in/email/sessions_controller.rb
```diff
@@ -35,12 +35,14 @@ def change_organisation
...
-    # This flow needs to be adapted when we have implemented school groups/trusts/LAs
-    # It is likely that we will not sign in Trust users as a school.
     information = GetInformationFromLoginKey.new(get_key)
-    @reason_for_failing_sign_in, @schools = information.reason_for_failing_sign_in, information.schools
+    @reason_for_failing_sign_in = information.reason_for_failing_sign_in
+    @schools, @school_groups = information.schools, information.school_groups
     update_session_without_urn(information.details_to_update_in_session)
-    redirect_to auth_email_create_session_path(urn: @schools.first.urn) if @schools&.one?
+    # .to_i used to convert nil to 0
+    only_one_organisation = @schools&.count.to_i + @school_groups&.count.to_i == 1
```
david-mears-dfe, line 43: A minor thing, but I like the Ruby convention where names for boolean variables end with a `?` - makes it easy to see what it's for.

app/controllers/hiring_staff/sign_in/email/sessions_controller.rb
```diff
@@ -54,6 +56,10 @@ def update_session_without_urn(options)
     Rails.logger.warn("Hiring staff signed in via fallback authentication: #{options[:oid]}")
```
david-mears-dfe, line 56: While we're adding UIDs to this controller, let's rename `update_session_without_urn` to e.g. `update_session_without_urn_and_uid` / `without_org_id`

david-mears-dfe:
Can you make sure that your temporary placeholder route is blocked by the feature flag, as an extra layer of certainty? We might instantiate SchoolGroup objects (for some reason) before the feature is switched on. And it's possible that we will mistakenly associate users with a Trust on DSI before the feature is switched on.

```ruby
# console
> FactoryBot.create(:school_group, uid: 4505)
> ENV['FEATURE_SCHOOL_GROUP_JOBS']
=> "false"
```

Screenshot 2020-07-07 at 10 45 02

Screenshot 2020-07-07 at 10 48 50

app/controllers/concerns/authentication_concerns.rb
```diff
@@ -3,14 +3,26 @@ module AuthenticationConcerns
...
+  end
+
   def current_school
-    School.find_by!(urn: session[:urn])
+    @current_school ||= School.find_by!(urn: session[:urn]) if session[:urn].present?
+  end
+
+  def current_school_group
+    if SchoolGroupJobsFeature.enabled?
+      @current_school_group ||= SchoolGroup.find_by!(uid: session[:uid]) if session[:uid].present?
```
david-mears-dfe, line 25: Is it possible for these instance variables to outlive the deletion of the session?

cpjmcquillan: If they could, that would be no different to what is currently the case. But if the session is deleted then you're not signed in

cpjmcquillan:
> Can you make sure that your temporary placeholder route is blocked by the feature flag, as an extra layer of certainty? We might instantiate SchoolGroup objects before the feature is switched on. And it's possible that we will mistakenly associate users with a Trust on DSI before the feature is switched on. If the latter happens but not the former, then they will be directed to a `Page not found` when they log in with a UID.
> 
> ```ruby
> # console
> > FactoryBot.create(:school_group, uid: 4505)
> # the feature flag is false
> > ENV['FEATURE_SCHOOL_GROUP_JOBS']
> => "false"
> ```
> 
> Screenshot 2020-07-07 at 10 45 02
> 
> Screenshot 2020-07-07 at 10 48 50

You can't sign in with a SchoolGroup now unless the feature is enabled, so I don't think it's necessary to protect a placeholder route that has no real purpose and won't exist when the feature is enabled.

david-mears-dfe:
> You can't sign in with a SchoolGroup now unless the feature is enabled, so I don't think it's necessary to protect a placeholder route that has no real purpose and won't exist when the feature is enabled.

Fair enough.

app/controllers/hiring_staff/sign_in/email/sessions_controller.rb
```diff
@@ -35,17 +35,17 @@ def change_organisation
...
 
   def choose_organisation
-    # This flow needs to be adapted when we have implemented school groups/trusts/LAs
-    # It is likely that we will not sign in Trust users as a school.
     information = GetInformationFromLoginKey.new(get_key)
-    @reason_for_failing_sign_in, @schools = information.reason_for_failing_sign_in, information.schools
-    update_session_without_urn(information.details_to_update_in_session)
-    redirect_to auth_email_create_session_path(urn: @schools.first.urn) if @schools&.one?
+    @reason_for_failing_sign_in = information.reason_for_failing_sign_in
+    @schools, @school_groups = information.schools, information.school_groups
```
cesidio, line 40: minor thing, better to have one assignment per line

PR #1752, opened 2020-07-07
allow algolia api keys to come from .env value using erb template

alexbowen:
Allow the api keys for algolia to be dynamically inserted into an .erb template (so they match with current environment). the client to be initialised can then be imported from front end javascript.

So this backend and frontend are both using the correct and consistent algolia api key env vars so client side search can be used in any (staging/dev/prod) environment

https://dfedigital.atlassian.net/browse/TEVA-981

## Changes in this PR:

- was a minor bug i noticed and fixed around job alerts markup mixed in this pr

### DIFF
--- a/app/views/vacancies/index.html.haml
+++ b/app/views/vacancies/index.html.haml
@@ -23,19 +23,20 @@
             = render partial: 'vacancy', locals: { vacancy: vacancy }
 
       - elsif @vacancies.user_search?
-        .divider-bottom
-          %p.govuk-heading-m
-            = t('jobs.no_jobs.intro')
+        #server-no-results
+          .divider-bottom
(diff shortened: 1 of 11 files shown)


app/views/vacancies/index.html.haml
```diff
@@ -23,19 +23,20 @@
             = render partial: 'vacancy', locals: { vacancy: vacancy }
 
       - elsif @vacancies.user_search?
-        .divider-bottom
-          %p.govuk-heading-m
-            = t('jobs.no_jobs.intro')
+        #server-no-results
```
cpjmcquillan, line 26: You can just add this id to the other element rather than nesting it in another div element - unless that was the intention?

PR #1753, opened 2020-07-08
Update rails to 5.2.4.3

cesidio:


david-mears-dfe:
So this update forces us to use `session.to_h` etc instead of `session[:whatever]`?

cpjmcquillan:
Similar to @david-mears-dfe question, do we need to change every instance of `session[:some_key]` to `session.to_h['some_key']`?

cesidio:
It is only for the key :session_id, you access all the other keys in the session as you would normally do with []. It might have to do with `append_info_to_payload` method.

david-mears-dfe:
Thank you for addressing this vulnerability by the way!

PR #1756, opened 2020-07-08
[TEVA 1032] MAT dashboard page HTML preparation

alexbowen:
*** these screenshots show a demo, it hasnt been added to page in this PR ***
Screenshot 2020-07-08 at 12 24 32
Screenshot 2020-07-08 at 12 22 33

tweeks to dashboard page markup so it matches design and will accept new filter controls

https://dfedigital.atlassian.net/browse/TEVA-1032

### DIFF
--- a/app/views/hiring_staff/organisations/show.html.haml
+++ b/app/views/hiring_staff/organisations/show.html.haml
@@ -2,28 +2,25 @@
 
 .school.govuk-grid-row
   .govuk-grid-column-full
-    %h1.govuk-heading-l.mb0
+    %h1.govuk-heading-l{ class: 'govuk-!-margin-bottom-2' }
       = t('schools.jobs.index', organisation: @organisation.name)
     - if @multiple_organisations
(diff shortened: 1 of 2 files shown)


app/views/hiring_staff/organisations/show.html.haml
```diff
@@ -2,28 +2,25 @@
 
 .school.govuk-grid-row
   .govuk-grid-column-full
-    %h1.govuk-heading-l.mb0
+    %h1.govuk-heading-l.mb0{ class: 'govuk-!-margin-bottom-2' }
```
cpjmcquillan, line 5: we can lose the `.mb0`

cpjmcquillan:
Can you add a screenshot of what these changes look like please?

alexbowen:
as above, but the table will be full width as there are no filters in place yet

alexbowen:
pointless having PR for this, will be in another

PR #1760, opened 2020-07-10
put keys in template as data attributes

alexbowen:
add keys to template as data attributes for JS to consume

https://dfedigital.atlassian.net/browse/TEVA-981

### DIFF
--- a/app/frontend/src/search/client.js
+++ b/app/frontend/src/search/client.js
@@ -6,9 +6,14 @@ import { getKeyword } from './ui/input/keyword';
 import { getCoords } from './ui/input/location';
 import { getRadius } from './ui/input/radius';
 
+export const clientElement = document.getElementsByClassName('filter-form')[0];
+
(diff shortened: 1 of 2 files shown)


app/frontend/src/search/client.js
```diff
@@ -8,7 +8,10 @@ import { getRadius } from './ui/input/radius';
 
 // This is the public API key which can be safely used in your frontend code.
 // This key is usable for search queries and list the indices you've got access to.
-export const search = algoliasearch('QM2YE0HRBW', '4082ba44346a92023eac1f794d739dd1');
+export const search = algoliasearch(
+  document.getElementById('new_').getAttribute('data-app'),
```
cpjmcquillan, line 12: The id being `new_` is a bug. (Possibly I think due the form builder not knowing that name of what it's been passed), so we should probably sort this

alexbowen: i havent fixed the source problem but this code is no longer dependant on the erroneous id

PR #1758, opened 2020-07-09
Sign up to NQT job alerts

cpjmcquillan:
This PR creates a new page for NQTs to sign up for job alerts. It piggybacks onto the way we currently create and manage subscriptions in the application.

Closes #1740 #1739 

**Todo**
- [x] Confirm all content
- [x] UX sign off
- [x] Content sign off
- [x] Product sign off
- [x] Wider sign off

This PR should only be merged once all the above have been completed.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1010

## Changes in this PR
- NQT specific job alert sign up page
- NQT specific job alert confirmation page
- Create a subscription through NQT specific job alert form

## Designs
https://www.figma.com/file/uRge4QPS2rBADfR9vLlkTO/Sprint-61?node-id=1%3A20

PR #1763, opened 2020-07-13
filter and accordian group components for MAT/search refinements

alexbowen:
This screenshot represents the component(s) being used but they are NOT added to any page in this PR

Screenshot 2020-07-13 at 15 44 20

- https://dfedigital.atlassian.net/browse/TEVA-1025
- https://dfedigital.atlassian.net/browse/TEVA-1026
- https://dfedigital.atlassian.net/browse/TEVA-1027
- https://dfedigital.atlassian.net/browse/TEVA-1032

## Changes in this PR:

- a component that can be used for GDS accordian groups
- a component that can be used for Formbuilder checkbox groups
- this also changes the JS folder structure to align with GDS and be a little more intuitive

https://www.figma.com/file/8BS7fuTGkUxx3ReqKdkufm/TVS-Design-System?node-id=110%3A11


app/views/layouts/application.html.haml
```diff
@@ -17,6 +17,7 @@
     %link{ href: asset_pack_path('media/images/govuk-apple-touch-icon-167x167.png'), rel: "apple-touch-icon", sizes: "167x167" }/
     %link{ href: asset_pack_path('media/images/govuk-apple-touch-icon-152x152.png'), rel: "apple-touch-icon", sizes: "152x152" }/
     %link{ href: asset_pack_path('media/images/govuk-apple-touch-icon.png'), rel: "apple-touch-icon" }/
+    %link{ href: "https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css", rel: "stylesheet" }/
```
david-mears-dfe, line 20: Just curious what this is being used for?

alexbowen: i just commented about it on connors wip PR https://github.com/DFE-Digital/teacher-vacancy-service/pull/1758

alexbowen: its amongst other things an svg icon library that is quite well used across the web

app/views/shared/_checkbox_group.html.haml
```diff
@@ -0,0 +1,12 @@
...
+    %input.govuk-input.search-input__input{ placeholder: "Search" }
+    %button{type: "submit"}
+      %i.fa.fa-search
+.govuk-form-group
+  .govuk-checkboxes.govuk-checkboxes--small
+    = form_for '', url: 'url', html: { method: 'post' } do |f|
+      = f.govuk_collection_check_boxes :content,
+        options,
+        :first,
+        :first
```
david-mears-dfe, line 12: If we were using VacanciesHelper's `working_pattern_options`, for example, then this would need to be
```ruby
options,
:second,
:first
```
where `:second` gets the computer-readable name and `:first` gets the human-readable name.

Is there a way to allow this type of options to be used by the checkbox group?

I *think* if instead of `:second` you use `:last`, then it should work for both types, because in any element of a one-dimensional array its last and first elements are the same element, could be wrong about how it works though.

Alternatively, if we can't easily abstract this further, you could add a comment somewhere, e.g. top of `app/views/shared/_filter_group.html.haml`, to say what format the list of options needs to take (a one-dimensional array). This is less preferred (by me) because it means we can't separate human-readable and computer-readable names.

david-mears-dfe: Correct me if I am misunderstanding something!

alexbowen: no you are not misunderstanding. im not sure about those params cause i was struggling to find docs that werent just examples and tbh i used the bare minimum for the example data i was using from the vacancy presenter. although working patterns and subject data works fine without change. i think this is something to standardise when the components are being used. but in the scope of this PR i am unsure. i think its best to document this (with a few other bits in the PR) so ill do that.

alexbowen: simplified the arguments need for the checkbox list and put a link to the proper docs page

app/frontend/src/lib/ui/components/searchCheckbox.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  input.addEventListener('change', (e) => {
+    filterCheckboxes(checkboxes, e.target);
+  });
+}
+
+export const filterCheckboxes = (checkboxes, input) => Array.from(checkboxes).map(checkbox => checkboxDisplay(checkbox, input));
+
+export const stringExistsInString = (original, input) => original.toUpperCase().indexOf(input.toUpperCase()) > -1;
```
cpjmcquillan, line 18: Maybe `substringExistsInString`?

app/frontend/src/lib/ui/components/searchCheckbox.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const checkboxDisplay = (checkbox, input) => {
```
cpjmcquillan, line 20: Is the expectation from UX that this will behave how the subjects filter currently does? (i.e when selected the element goes to the top of the input). Not saying this needs to be done right now, but worth thinking about!

app/frontend/src/lib/ui/components/searchCheckbox.test.js
```diff
@@ -0,0 +1,28 @@
+import { init, stringExistsInString } from './searchCheckbox';
+
+describe('search group of checkboxes', () => {
```
cpjmcquillan, line 3: maybe `'search govuk checkbox collection'` or similar?

app/frontend/styles/components/accordion-group.scss
```diff
@@ -0,0 +1,78 @@
+.js-enabled .govuk-accordion {
+    border: none;
```
cpjmcquillan, line 2: Is this file 4 space indented?

app/frontend/styles/components/search_input.scss
```diff
@@ -0,0 +1,25 @@
+.search-input {
+  background: white;
+  position: relative;
+  width: 100%;
+  margin-bottom: 10px;
+
+  input {
+    width: 100%;
+    padding: 20px 60px 20px 40px;
```
cpjmcquillan, line 9: not crucial, but would be good to use `govuk-spacing(X)` where we can

app/frontend/styles/components/search_input.scss
```diff
@@ -0,0 +1,25 @@
...
+    display: block;
+  }
+
+  button {
+    background: transparent;
+    border: none;
+    cursor: pointer;
+    display: inline-block;
+    font-size: 20px;
```
cpjmcquillan, line 18: Same with font-sizes

app/views/shared/_checkbox_group.html.haml
cpjmcquillan: `:last` and `:first` makes sense to me 👍

app/frontend/src/lib/ui/components/searchCheckbox.js
alexbowen: good point, i hadnt accounted for this

alexbowen: thinking about this, yes lets leave this for now. i think this behaviour is bad user experience and id like to discuss with chris first

PR #1768, opened 2020-07-14
Link scraped vacancies to TV vacancies if possible

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-935

## Changes in this PR:

Previously we just matched scraped vacancies to schools in our database.
This matches them to vacancies those schools posted if:
1. The vacancy was posted by a school AND
2. The vacancy was posted within a fortnight either side of the matched vacancy AND
3. No more than 1 in 5 words in one vacancy title is not included in the matched vacancy title (ignoring a list of small words which don't add meaning) [this last part boosts the match rate by ~50% over just comparing the two strings directly]

### DIFF
--- a/bigquery/link-scraped-vacancies-to-schools-and-vacancies.sql
+++ b/bigquery/link-scraped-vacancies-to-schools-and-vacancies.sql
@@ -1,6 +1,11 @@
   # Transforms the latest versions of scraped vacancy NoSQL documents from Firestore in scraped_vacancies_raw_latest into SQL-queriable rows
   # Also matches vacancies to schools where possible to allow this to be joined to the schools table, and if this is possible matches vacancies to vacancies on Teaching Vacancies.
 WITH
+  stop_words AS (
+  SELECT
+    *
+  FROM
+    UNNEST(['at','of','to','and','or','for','with','but','may','be','as','x','from','including','per','in']) ),
   TV_vacancy AS (
   SELECT
     id,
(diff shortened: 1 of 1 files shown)


bigquery/link-scraped-vacancies-to-schools-and-vacancies.sql
```diff
@@ -102,31 +108,91 @@ WITH
...
+        SELECT
+          SAFE_DIVIDE(COUNTIF(scraped_words IS NULL
+          OR TV_words IS NULL),COUNTIF(scraped_words IS NOT NULL
+          AND TV_words IS NOT NULL)) #number of words that are in one title but not in the other as a proportion of the number of words that are in both titles
+        FROM (
+          SELECT
+            *
+          FROM
+            UNNEST(SPLIT(LOWER(REGEXP_REPLACE(TV_vacancy.job_title,r'[^a-zA-Z0-9]', '')),' ')) AS words
+          WHERE words NOT IN ('at','of','to','and','or','for','with','but','may','be','as','x','from','including','per','in') #don't count these words when working out the proportion
```
david-mears-dfe, line 164: What happens if we also exclude 'teacher'? My thinking is that 'teacher' is a common but generally meaningless word in this context.

david-mears-dfe: For avoidance of duplication, it would be a nice-to-have for the list of stop-words to have a single source of truth. Not sure how easy that is in SQL.

david-mears-dfe:
I don't feel qualified to give an approval as I don't know very much SQL. This is on my list of things to know.

bigquery/link-scraped-vacancies-to-schools-and-vacancies.sql
stevenleggdfe: Thanks @david-mears-dfe - 'teacher' is actually quite important hear, because we need to avoid matching teacher roles to teaching assistant roles or headteacher roles, for example. Many primary jobs are in fact just listed as 'Teacher'.

I'll look into the list of stop words. Something might be possible, although it might be clunky!

stevenleggdfe: @david-mears-dfe - sorted the list of stop words.

PR #1770, opened 2020-07-15
correct layout for dashboard header

alexbowen:
this corrects the layout for dashboard header to be exactly as per https://www.figma.com/file/jcScod22AfWwcMCKCmv0Br/Sprint-55?node-id=222%3A450

Screenshot 2020-07-15 at 15 04 36

### DIFF
--- a/app/views/hiring_staff/organisations/show.html.haml
+++ b/app/views/hiring_staff/organisations/show.html.haml
@@ -13,7 +13,7 @@
   .govuk-grid-column-full{ class: 'govuk-!-margin-top-5' }
     .govuk-form-group
       - if @organisation.vacancies.active.any?
-        .create-a-job
+        .dashboard-header
           = link_to t('buttons.create_job'), new_organisation_job_path, class: 'govuk-button govuk-!-margin-bottom-0 govuk-!-margin-top-0'
         .overflow-scroll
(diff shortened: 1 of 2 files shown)


app/views/hiring_staff/organisations/show.html.haml
```diff
@@ -2,18 +2,19 @@
...
       - else
         = link_to t('sign_in.organisation.change'), auth_dfe_callback_path, class: 'govuk-link'
 
-  .govuk-grid-column-full{ class: 'govuk-!-margin-top-9' }
-    .govuk-form-group.create-a-job
+  .govuk-grid-column-full{ class: 'govuk-!-margin-top-5' }
+    .govuk-form-group
       - if @organisation.vacancies.active.any?
-        = link_to t('buttons.create_job'), new_organisation_job_path, class: 'govuk-button govuk-!-margin-bottom-0 govuk-!-margin-top-0'
+        .create-a-job
```
cpjmcquillan, line 16: Rather than a div around a link, you can add the class to the link_to helper directly

alexbowen: i decided to keep it as a container div as this will be the container for other buttons that will live in there to hide/show filters. i think ill rename the container to be a bit more descriptive

PR #1775, opened 2020-07-16
Changed about_school to about_organisation

josephhull676:
Added migration file to change about_school field in vacancies table to about_organisation. Replaced all references to about_school to about_organisation

## Changes in this PR:

- Created a migration file to change about_school field to about_organisation
- Replaced all references to about_school to about_organisation

### DIFF
--- a/config/locales/activerecord.yml
+++ b/config/locales/activerecord.yml
@@ -45,7 +45,7 @@ en:
   job_summary_errors: &job_summary_errors
     job_summary:
       blank: Enter a job summary
-    about_school:
+    about_organisation:
       blank: Enter a description of the school
   hiring_staff_user_preference_errors:
     managed_organisations:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -169,7 +169,7 @@ en:
         application_link: Link to online application
       job_summary_form:
         job_summary_html: "Job summary (<span class='text-red'>Required</span>)"
-        about_school_html: "About %{school} (<span class='text-red'>Required</span>)"
+        about_organisation_html: "About %{school} (<span class='text-red'>Required</span>)"
       organisation_form:
         description: Description of school or trust
     hint:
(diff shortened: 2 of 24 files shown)


config/locales/activerecord.yml
```diff
@@ -45,7 +45,7 @@ en:
   job_summary_errors: &job_summary_errors
     job_summary:
       blank: Enter a job summary
-    about_school:
+    about_organisation:
       blank: Enter a description of the school
```
cesidio, line 49: Probably we need to change this line too with school/trust

config/locales/en.yml
```diff
@@ -169,7 +169,7 @@ en:
         application_link: Link to online application
       job_summary_form:
         job_summary_html: "Job summary (<span class='text-red'>Required</span>)"
-        about_school_html: "About %{school} (<span class='text-red'>Required</span>)"
+        about_organisation_html: "About %{school} (<span class='text-red'>Required</span>)"
```
cesidio, line 172: you may want to change the argument you pass from %{school} to %{organisation}

config/locales/en.yml
```diff
@@ -212,7 +212,7 @@ en:
         job_summary: >-
           Give a brief summary of the job that job seekers can quickly scan.
           This will be the first thing they see on the listing or in search engine results
-        about_school: "This will appear next to the job details under 'School overview'. Feel free to adapt this text"
+        about_organisation: "This will appear next to the job details under 'School overview'. Feel free to adapt this text"
```
cesidio, line 215: `This will appear next to the job details under 'School overview'` this part may need to change as well to consider the case of a trust

config/locales/en.yml
```diff
@@ -371,7 +371,7 @@ en:
     time_at: ' at '
     publication_date: 'Date job will be listed'
     job_summary: 'Job summary'
-    about_school: 'About %{school}'
+    about_organisation: 'About %{school}'
```
cesidio, line 374: you may want to change the name of the parameter here too

config/locales/en.yml
cpjmcquillan: This and the error one you can have `about_school` and `about_trust` and use a different one, or something more complicated would be passing `count: current_organisation.is_a?(School) ? 0 : 1` and adding `zero` and `one` translations under `about_organisation`

config/locales/activerecord.yml
josephhull676: Sure. So you want two separate translations (is that the correct term?)  - about_school and about_organisation?

david-mears-dfe: Will be picked up by https://dfedigital.atlassian.net/browse/TEVA-1117

PR #1776, opened 2020-07-16
Created rake task for importing school group data

josephhull676:
## Changes in this PR:

- Rake task for school group data importation created in data.rake file in lib/tasks


cesidio:
I would also add something to the README.md file the same is done for the `rake data:school:import` task

PR #1781, opened 2020-07-19
Remove client side algolia related code/functionality/tests

alexbowen:
this will also re-enable autocomplete on home/search results and show radius if user enters number

https://dfedigital.atlassian.net/browse/TEVA-1084


alexbowen:
this is pretty much there, just want to verify the ticket i created is agreed and maybe fine tune this PR a little bit

cpjmcquillan:
> this will also re-enable autocomplete on home/search results and hide/show radius if user enters number
> 
> https://dfedigital.atlassian.net/browse/TEVA-1084

Hide/show radius if the user enters a number might be acceptable on the home page, but on the job results page, the radius input should be hidden/shown depending on the type of search carried out. At the minute you can determine this based on whether a `location_category` parameter is present. If it's just based on a number, users will never be able to select a radius even if it's suitable for their search (e.g 'Timbuktu')

alexbowen:
@cpjmcquillan javascript is only responsible for the behaviour in this ticket https://dfedigital.atlassian.net/browse/TEVA-903, wether the radius is shown based on any backend/url params means boing back down the path of should/shouldnt geocode in JS and i don't think we want to do that

PR #1796, opened 2020-07-22
Sidebar view component

cpjmcquillan:
This PR implements a ViewComponent for the Sidebar in the Create a job listing journey.

## Changes in this PR
- Remove redundant sidebar view and spec
- Implement sidebar ViewComponent
- Fix bug related to conditional sidebar rendering
- Fix bug that meant sidebar could never have completed steps until the vacancy had been created


Gemfile
```diff
@@ -51,6 +51,7 @@ gem 'sidekiq'
 gem 'sidekiq-cron', '~> 1.1'
 gem 'skylight'
 gem 'validate_url', '~> 1.0.8'
+gem 'view_component'
```
alexbowen, line 54: this gem has already been added in the school group overview PR from last week?

cpjmcquillan: That PR won the race to merge

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      result.css('.app-step-nav__step--active .app-step-nav__circle-background').to_html, current_step.to_s
+    )
+  end
+
+  context 'when a step is completed' do
+    let(:completed_step) { 1 }
+    let(:current_step) { 2 }
+
+    it 'renders visited class on completed steps' do
+      result = render_inline(described_class.new(vacancy: vacancy))
```
cesidio, line 33: It's nice to have this into a before and ideally have one assertion per spec

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  it 'renders the sidebar' do
+    render_inline(described_class.new(vacancy: vacancy))
+
+    expect(rendered_component).to include('Create a job listing steps')
+  end
+
+  it 'renders active class on current_step' do
+    result = render_inline(described_class.new(vacancy: vacancy))
+
+    assert_includes(
```
david-mears-dfe, line 23: Is there a reason you sometimes use Rspec's `expect(fridge).to include(milk)` and sometimes Minitest's `assert_includes(fridge, milk)`?

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        result.css('.app-step-nav__step--visited .app-step-nav__circle-background').to_html, completed_step.to_s
+      )
+    end
+  end
+
+  context 'when a School user creates a job' do
+    let(:school_group_user?) { false }
+
+    it 'renders the correct steps' do
+      render_inline(described_class.new(vacancy: vacancy))
```
cesidio, line 45: It's nice to have this into a before and ideally have one assertion per spec

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      expect(rendered_component).not_to include(I18n.t('jobs.job_location'))
+      expect(rendered_component).to include(I18n.t('jobs.job_details'))
+      expect(rendered_component).to include(I18n.t('jobs.pay_package'))
+      expect(rendered_component).to include(I18n.t('jobs.important_dates'))
+      expect(rendered_component).to include(I18n.t('jobs.supporting_documents'))
+      expect(rendered_component).to include(I18n.t('jobs.application_details'))
+      expect(rendered_component).to include(I18n.t('jobs.job_summary'))
+      expect(rendered_component).to include(I18n.t('jobs.review_heading'))
```
david-mears-dfe, line 54: There might be a case for using a regex match assertion, if you want to test the order of these inclusions. There is already some order-checking in `job_creation_helper_spec.rb` though.

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+RSpec.describe CreateAJobSidebarComponent, type: :component do
+  let(:vacancy) { create(:vacancy, completed_step: completed_step) }
+  let(:completed_step) { 0 }
+  let(:current_step) { 1 }
+  let(:school_group_user?) { true }
+
+  before do
+    allow_any_instance_of(HiringStaff::JobCreationHelper).to receive(:current_step).and_return(current_step)
+    allow_any_instance_of(HiringStaff::JobCreationHelper).to receive(:school_group_user?).and_return(school_group_user?)
+    @result = render_inline(described_class.new(vacancy: vacancy))
```
cesidio, line 12: This should be

`let(:inline_result) { render_inline(described_class.new(vacancy: vacancy)) }`

spec/components/create_a_job_sidebar_component_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      @result.css('.app-step-nav__step--active .app-step-nav__circle-background').to_html, current_step.to_s
+    )
+  end
+
+  context 'when a step is completed' do
+    let(:completed_step) { 1 }
+    let(:current_step) { 2 }
+
+    it 'renders visited class on completed steps' do
+      assert_includes(
```
cesidio, line 72: probably we have some capybara matchers for these too, so that we do not need to `require 'view_component/test_helpers'`

spec/components/create_a_job_sidebar_component_spec.rb
cpjmcquillan: Just that I was following the docs

app/components/create_a_job_sidebar_component/create_a_job_sidebar_component.html.haml
```diff
@@ -0,0 +1,18 @@
+.app-step-nav-related.app-step-nav-related--singular
```
cesidio, line 1: Why is this file in a folder? I would keep it consistent with the rest for now and maybe refactor all of them at some point if it makes more sense to have views in folders

cpjmcquillan: Ah I didn't mean to leave this there. I was messing around with having the html and css in a folder for the component

spec/components/create_a_job_sidebar_component_spec.rb
cpjmcquillan: We need it to be rendered before each block, but after lines 10/11. Not sure how we can achieve that other than this?

cesidio: ah right, cannot think of a better way for now

Issue #1725, opened 2020-07-01
Capture human-readable user identification when a hiring staff user logs in


tatyree:
## Background 

For the MAT user journey we need to be able to show which users created which vacancies.  The user model is already associated with the vacancy model, so we need to ensure that the users' names are stored in the user record so we can show it in the MAT vacancy table. 

## Done When 
 - [x] The user's name is stored on the user record (can be when the record is created, or when they login - developer's discretion) 
 - [x] This name is available to be displayed on vacancy pages-particularly for MAT access vacancy lists.
 - [ ] A suitable PR has been raised and approved
 - [ ] The code has been merged to master and deployed to production. **NOTE** the code **does not** need to be switched on for this issue to be considered complete.

bencmitchell:
Jira ticket https://dfedigital.atlassian.net/browse/TEVA-1004 created.

david-mears-dfe:
Here is an example DSI response for reference.

{"approvedAt"=>"2020-01-31T14:12:32.822Z",
 "updatedAt"=>"2020-01-31T14:12:32.822Z",
 "organisation"=>
  {"id"=>"blah blah",
   "name"=>"Bexleyheath Academy",
   "Category"=>"001",
   "Type"=>"28",
   "URN"=>"137138",
   "UID"=>nil,
   "UKPRN"=>"10034964",
   "EstablishmentNumber"=>"4002",
   "Status"=>1,
   "ClosedOn"=>nil,
   "Address"=>"Woolwich Road, Bexleyheath, Kent, DA6 7DA",
   "phaseOfEducation"=>4,
   "statutoryLowAge"=>11,
   "statutoryHighAge"=>18,
   "telephone"=>"02083035696",
   "regionCode"=>"H",
   "legacyId"=>"1005605",
   "companyRegistrationNumber"=>nil,
   "createdAt"=>"2019-04-05T08:59:16.883Z",
   "updatedAt"=>"2019-04-05T09:00:00.873Z"},
 "roleName"=>"End user",
 "userId"=>"blah blah",
 "email"=>"david.mears@digital.education.gov.uk",
 "familyName"=>"Mears",
 "givenName"=>"David"}

PR #1808, opened 2020-07-27
Capture users' names

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-997

## GitHub Issue

Closes https://github.com/DFE-Digital/teacher-vacancy-service/issues/1725

## Changes in this PR:

- Store user's first and last names when the User record is created, using DSI data
- Create UserPresenter to concatenate first and last names together. This falls back to email when either name is unavailable.

### DIFF
--- a/db/migrate/20200727095226_add_names_to_user.rb
+++ b/db/migrate/20200727095226_add_names_to_user.rb
@@ -1,0 +1,6 @@
+class AddNamesToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :family_name, :string
+    add_column :users, :given_name, :string
+  end
+end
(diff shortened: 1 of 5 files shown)


app/presenters/user_presenter.rb
```diff
@@ -0,0 +1,9 @@
+class UserPresenter < BasePresenter
```
cesidio, line 1: No need to add a UserPresenter here as it is not used anywhere yet. I am also trying to get rid of presenters and just use ViewComponent

david-mears-dfe: Removed commit.

PR #1803, opened 2020-07-24
Change vacancy form field labels and create a job title depending on vacancy organisation

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-756

## Changes in this PR:

Create a job title does not include the school or trust name before the location of the vacancy is selected

Create a job title includes the school name when the vacancy is at a school or the trust name when at the trust's central office

School visits and school visit instructions (on step is 6 and the review page) change to trust visits and trust visit instructions when the vacancy is at the trust's central office

Hint text under about [trust]/about [school] change depending on vacancy location

Tests changed to reflect changes above

## Screenshots of UI changes:

Create a job title does not include the school / trust name before the location of the vacancy is chosen

Screenshot 2020-07-24 at 10 23 03

Create a job title includes the trust or school name depending on the vacancy location:

Trust:

Screenshot 2020-07-24 at 10 25 31

School:

Screenshot 2020-07-24 at 10 27 28

School visits changes to Trust visits when the vacancy is at the trust's central office:

Application details section:

Screenshot 2020-07-24 at 10 30 11

Review section:

Screenshot 2020-07-24 at 10 32 36

Hint text:

Trust: 

Screenshot 2020-07-27 at 12 02 52

School:

Screenshot 2020-07-27 at 12 04 13

PR #1811, opened 2020-07-28
Create organisation helper to remove heavy logic from views

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-756

## Changes in this PR:

These changes are a result of the changes suggested on https://github.com/DFE-Digital/teacher-vacancy-service/pull/1803 after it was merged.

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -17,6 +17,14 @@ module OrganisationHelper
     end
   end
 
+  def organisation_type_basic(organisation)
+    if organisation.is_a?(School)
+      'school'
+    else
+      'trust'
+    end
+  end
+
   def school_size(school)
     if school.gias_data.present?
       return number_of_pupils(school) if school.gias_data['NumberOfPupils'].present?
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -71,12 +71,10 @@ module VacanciesHelper
   end
 
   def page_title(vacancy)
-    organisation = vacancy.school_or_school_group.present? ? vacancy.school_or_school_group.name : ''
-
     if vacancy.state == 'copy'
       I18n.t('jobs.copy_job_title', job_title: vacancy.job_title)
     elsif %w(create review).include?(vacancy.state)
-      I18n.t('jobs.create_a_job_title', organisation: "#{organisation}")
+      I18n.t('jobs.create_a_job_title', organisation: vacancy.school_or_school_group&.name.presence || '')
     else
       I18n.t('jobs.edit_job_title', job_title: vacancy.job_title)
     end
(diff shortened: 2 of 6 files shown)


app/helpers/vacancies_helper.rb
```diff
@@ -71,12 +71,12 @@ def review_heading(vacancy)
   end
 
   def page_title(vacancy)
-    organisation = vacancy.school_or_school_group.present? ? vacancy.school_or_school_group.name : ''
+    organisation_name = vacancy.school_or_school_group.present? ? vacancy.school_or_school_group.name : ''
```
cpjmcquillan, line 74: You might be able to do `vacancy.school_or_school_group&.name.presence || ''`

app/views/hiring_staff/vacancies/edit_vacancy_sections/_edit_application_details.html.haml
```diff
@@ -18,7 +18,7 @@
 
   .app-check-your-answers__contents
     %dt.app-check-your-answers__question#school_visits
-      %h4.govuk-heading-s= t("jobs.#{organisation}")
+      %h4.govuk-heading-s= t("jobs.#{organisation_type_basic(@vacancy.school_or_school_group)}_visits")
```
cpjmcquillan, line 21: Doesn't this depend on where the job is located as opposed to whether the organisation is a Trust or not?

config/locales/en.yml
```diff
@@ -173,7 +173,7 @@ en:
         contact_email_html: "Contact email (<span class='text-red'>Required</span>)"
         school_visits: School visits
         how_to_apply: How to apply
-        trust_visits: Trust visits
+        schoolgroup_visits: Trust visits
```
cpjmcquillan, line 176: I'd prefer to keep these all as `trust_visits`

app/helpers/organisation_helper.rb
```diff
@@ -17,6 +17,14 @@ def organisation_type(organisation)
     end
   end
 
+  def organisation_type_basic(organisation)
+    if organisation.is_a?(SchoolGroup)
+      'trust'
```
david-mears-dfe, line 22: I think, if you are introducing this helper to return `'trust'`, then you don't need to (or, you need to not) rename the translations from `trust_whatever` to `schoolgroup_whatever`. I may have misunderstood though.

app/views/hiring_staff/vacancies/edit_vacancy_sections/_edit_application_details.html.haml
josephhull676: Yeah, I believe that is why the organisation provided to organisation_type_basic comes from the vacancy. If the vacancy is at the school, @vacancy.school_or_school_group will return the school, if not, it will return the school group... I think?

cpjmcquillan: Yep, my bad sorry!

app/helpers/organisation_helper.rb
josephhull676: Yup, changed it back to trust_whatever lol

PR #1813, opened 2020-07-28
Refactor School overview section in the view to a ViewComponent

cpjmcquillan:
This PR refactors some of the view logic in the organisation dashboard template. This is part of the work required to complete [TEVA-682](https://dfedigital.atlassian.net/browse/TEVA-682)

## Changes in this PR
- Add ViewComponent for the school overview in the organisation dashboard


spec/components/organisation/school_overview_component_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+RSpec.describe Organisation::SchoolOverviewComponent, type: :component do
+  let!(:inline_component) { render_inline(described_class.new(organisation: organisation)) }
+
+  context 'when organisation is a SchoolGroup' do
+    let(:organisation) { build(:school_group) }
+
+    it 'does not render the school overview component' do
+      expect(rendered_component).to eql('')
```
cesidio, line 10: This reads better with:

`expect(rendered_component).to be_blank`

cpjmcquillan: Indeed it does, I'd tried `to be nil` but it returns the empty string. Thanks 👍

app/components/organisation/school_overview_component.html.haml
```diff
@@ -0,0 +1,43 @@
...
+        %td.govuk-table__cell
+          %strong.strong= t('schools.type')
+        %td.govuk-table__cell
+          = @organisation.school_type.label
+        %td.govuk-table__cell
+      %tr.govuk-table__row
+        %td.govuk-table__cell
+          %strong.strong Website
+        %td.govuk-table__cell
+          = link_to("#{@organisation.name} website (opens in a new window)", @organisation.url, class: 'govuk-link wordwrap', target: '_blank')
```
cesidio, line 42: While you are at it, I would I18n this link text

app/components/organisation/school_overview_component.html.haml
```diff
@@ -0,0 +1,43 @@
...
+        %td.govuk-table__cell
+      %tr.govuk-table__row
+        %td.govuk-table__cell
+          %strong.strong= t('schools.type')
+        %td.govuk-table__cell
+          = @organisation.school_type.label
+        %td.govuk-table__cell
+      %tr.govuk-table__row
+        %td.govuk-table__cell
+          %strong.strong Website
```
cesidio, line 40: and this one I18n too

app/components/organisation/school_overview_component.rb
```diff
@@ -0,0 +1,11 @@
+class Organisation::SchoolOverviewComponent < ViewComponent::Base
```
cesidio, line 1: As a convention, module names or scopes in general should be plural, so I would call it Organisations::SchoolOverviewComponent. Not sure why you are putting this under `organisation` though

cpjmcquillan: 1. This is the first in a set of components that are all to do with the organisations dashboard. I envisage them all being under `organisation`
2. There is already a component called `SchoolOverviewComponent

PR #1814, opened 2020-07-28
Create LocationPolygon model and location_polygons migration

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-941

## Changes in this PR:

- Created LocationPolygon model and migration to create location_polygons table.

- There was one less attribute in the schema spike when compared to the attributes listed on the ticket (is_active? was missing from the spike). It has been added.

> "Another field that might be useful is a string field that indicates the resolution of the boundary data. This could just add unnecessary complexity, however."

Feel I'm unable to give a judgement on the above

### DIFF
--- a/app/models/location_polygon.rb
+++ b/app/models/location_polygon.rb
@@ -1,0 +1,6 @@
+class LocationPolygon < ApplicationRecord
+  scope :cities, -> { where(location_type: 'cities') }
+  scope :counties, -> { where(location_type: 'counties') }
+  scope :local_authorities, -> { where(location_type: 'local_authorities') }
+  scope :regions, -> { where(location_type: 'regions') }
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_27_095226) do
+ActiveRecord::Schema.define(version: 2020_07_28_145709) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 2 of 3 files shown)


app/models/location_polygon.rb
```diff
@@ -0,0 +1,6 @@
+class LocationPolygon < ApplicationRecord
+    scope :cities, -> { where(location_type: 'cities') }
+    scope :counties, -> { where(location_type: 'counties') }
+    scope :local_authorities, -> { where(location_type: 'local_authorities') }
+    scope :regions, -> { where(location_type: 'regions') }
+end
```
cpjmcquillan, line 6: Newline. You can configure your IDE/editor to do this automatically

db/schema.rb
```diff
@@ -10,14 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_27_095226) do
+ActiveRecord::Schema.define(version: 2020_07_28_145709) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "citext"
```
cpjmcquillan, line 16: Try to make sure you only commit the changes that your migration introduced (which should just be the version change and the create_table block)

josephhull676: Will make sure to do this in the future, but have re-added the lines that were removed

app/models/location_polygon.rb
```diff
@@ -0,0 +1,6 @@
+class LocationPolygon < ApplicationRecord
+    scope :cities, -> { where(location_type: 'cities') }
```
cesidio, line 2: 4 space indentation, we use only 2 in ruby

josephhull676: Dunno what happened there. Apologies!

josephhull676: Seems like vscode lost my defaults as I had it set to 2

PR #1820, opened 2020-07-29
Make job alerts typo-intolerant

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-955

## Changes in this PR:

Searches should be typo-tolerant: 'physics' can return 'physical education'
Alerts should not be: 'physics' can NOT return 'physical education'

This commit also removes the use of a magic number by reusing a
pre-existing constant.

## How I tested this

I think we don't have tests for our search results.

I tested this change by firstly creating a job alert job in the console with keyword 'physics' and seeing that it returned only a physics vacancy, and secondly I tested that the regular search was still returning both 'physics' and 'physical education' jobs for that search-term.

### DIFF
--- a/app/services/vacancy_algolia_alert_builder.rb
+++ b/app/services/vacancy_algolia_alert_builder.rb
@@ -28,7 +28,7 @@ class VacancyAlgoliaAlertBuilder < VacancyAlgoliaSearchBuilder
       replica: search_replica,
       hitsPerPage: MAXIMUM_SUBSCRIPTION_RESULTS,
       filters: search_filter,
-      typoTolerance: false,
+      typoTolerance: false
     )
     Rails.logger.info(
       "#{vacancies.count} vacancies found for job alert with criteria: #{subscription_hash}, "\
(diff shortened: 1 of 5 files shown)


app/services/vacancy_algolia_alert_builder.rb
```diff
@@ -27,7 +27,8 @@ def call
       aroundRadius: location_filter[:radius],
       replica: search_replica,
       hitsPerPage: MAXIMUM_SUBSCRIPTION_RESULTS,
-      filters: search_filter
+      filters: search_filter,
+      typoTolerance: false,
```
cpjmcquillan, line 31: trailing comma

spec/support/search_helper.rb
```diff
@@ -1,16 +1,31 @@
...
+  def get_arguments_to_algolia(algolia_hash)
+    {
       aroundLatLng: algolia_hash[:aroundLatLng] || nil,
       aroundRadius: algolia_hash[:aroundRadius] || nil,
       replica: algolia_hash[:replica] || nil,
       hitsPerPage: algolia_hash[:hitsPerPage] || 10,
       filters: algolia_hash[:filters] ||
         "publication_timestamp <= #{Time.zone.today.to_datetime.to_i} AND "\
-        "expires_at_timestamp > #{Time.zone.today.to_datetime.to_i}"
+        "expires_at_timestamp > #{Time.zone.today.to_datetime.to_i}",
```
cpjmcquillan, line 23: trailing comma

PR #1828, opened 2020-07-30
Add @job_location_options to organisation vacancy view_component

david-mears-dfe:
We will need to pass in an array of options to the filter UI. This method creates the array.

## Jira ticket URL

This is the first small part of
https://dfedigital.atlassian.net/browse/TEVA-755

### DIFF
--- a/app/components/hiring_staff/vacancies_component.rb
+++ b/app/components/hiring_staff/vacancies_component.rb
@@ -32,16 +32,16 @@ class HiringStaff::VacanciesComponent < ViewComponent::Base
   private
 
   def job_location_options
-    locations = Set.new
-    locations.add(I18n.t('hiring_staff.organisations.school_groups.readable_job_location'))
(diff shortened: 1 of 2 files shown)


david-mears-dfe:
I've realised this needs some tweaking - these options should be something like `[[text, value], [text, value]]` (or the other way round), not just `[text, text, text, text]`.

_filter_group.html.haml will need to be changed to pass different `value_method`s and `text_method`s on the last line.

david-mears-dfe:
> I've realised this needs some tweaking - these options should be something like `[[text, value], [text, value]]` (or the other way round), not just `[text, text, text, text]`.
> 
> _filter_group.html.haml will need to be changed to pass different `value_method`s and `text_method`s on the last line.

Fixed this in latest commit

david-mears-dfe:
This is done in #1838

PR #1829, opened 2020-07-31
Create import polygons task

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-945

## Changes in this PR:

- Added LOCATION_POLYGON_SETTINGS to location_data initializer
- Created task to import boundaries from ONS
- Changes to lists of cities and counties in respective location category yml files


spec/fixtures/files/ons_local_authorities.json
```diff
@@ -0,0 +1,59 @@
...
+    "name": "ctyua19nm",
+    "type": "esriFieldTypeString",
+    "alias": "ctyua19nm",
+    "length": 36
+   }
+  ],
+  "features": [
+   {
+    "attributes": {
+     "ctyua19nm": "Tower Hamlets"
```
cesidio, line 22: nice choice 👍

josephhull676: ;)

app/services/import_polygons.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      geometry_rings = region_response.dig('geometry', 'rings')
+
+      # The first ring tends to contain far more points than subsequent rings.
+      # Boundary should be visualised to check how it should be used.
+      # If algolia searches by polygon are slow, these boundaries could be downsampled significantly.
+      points = []
+      geometry_rings[0].each do |point|
+        # API returns coords in an unconventional lng,lat order
+        points.push(*point.reverse.map { |coord| coord.round(13) })
```
cpjmcquillan, line 24: Might be worth adding a comment about why we're rounding the float and why 13 decimal places rather than any other number

app/services/import_polygons.rb
```diff
@@ -0,0 +1,40 @@
...
+    response = HTTParty.get(LOCATION_POLYGON_SETTINGS[location_type][:api])
+
+    response.fetch('features', []).each do |region_response|
+      region_name = region_response.dig('attributes', LOCATION_POLYGON_SETTINGS[location_type][:name_key]).downcase
+
+      next unless location_categories_include?(region_name)
+
+      geometry_rings = region_response.dig('geometry', 'rings')
+
+      # The first ring tends to contain far more points than subsequent rings.
```
cpjmcquillan, line 18: I know I wrote this initially, but my understanding/assumption now is that the first ring is the outer boundary, and that subsequent rings are bodies of water (essentially exclusion rings) that we don't need to care about. So it might be helpful to update the comment to reflect that

app/services/import_polygons.rb
```diff
@@ -0,0 +1,40 @@
...
+                     .update(boundary: points)
+    end
+  end
+
+  private
+
+  def location_categories_include?(region_name)
+    location_type == :regions && DOWNCASE_REGIONS.include?(region_name) ||
+      location_type == :counties && DOWNCASE_COUNTIES.include?(region_name) ||
+      location_type == :local_authorities && DOWNCASE_BOROUGHS.include?(region_name) ||
```
cpjmcquillan, line 37: We could have the location type as `:london_boroughs` (or something else) if that is more revealing. Let's not have a mismatch between the `location_type` and what the data is referred to in the rest of the code base 👍

josephhull676: I did think this too. I'll speak to Cesi and Ben about it.

PR #1830, opened 2020-07-31
add region:uk param to google query

alexbowen:
Use built in google API parameter to define region as Uk rather than appending it to each suggestion


david-mears-dfe:
Can we have a PR description of what change this makes?

PR #1832, opened 2020-08-03
Add filtering functionality to dashboard

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-755


app/services/vacancy_filter.rb
```diff
@@ -0,0 +1,31 @@
+class VacancyFilter
```
cpjmcquillan, line 1: Could we scope this to HiringStaff, given that we will need jobseeker filters too, which are different. Or we could do that for jobseeker filters?

app/services/vacancy_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    @managed_school_urns = @user_preference.managed_school_urns
+    @managed_organisations = @user_preference.managed_organisations
+  end
+
+  def update(params)
+    @managed_school_urns = params[:managed_school_urns]&.reject(&:blank?)
+    @managed_organisations = params[:managed_organisations]
+
+    if managed_organisations.include?('all') || managed_organisations.blank? && managed_school_urns.none?
```
cpjmcquillan, line 15: Is it more readable and explicit like this? I'm happy either way, but this saves anyone having to think too much about precedence when following the logic
```ruby
if managed_organisations.include?('all') || (managed_organisations.blank? && managed_school_urns.none?)
```

app/components/hiring_staff/vacancies_component.rb
```diff
@@ -49,4 +52,13 @@ def published
   def awaiting_feedback
     @organisation.vacancies.awaiting_feedback.order(@sort.column => @sort.order)
   end
+
+  def set_filters(vacancies, filters)
```
cesidio, line 56: this should be done with something like:

```  def set_filters(vacancies, filters)
    return vacancies if filters.none? || filters[:managed_organisations] == 'all'

    return vacancies.in_school_urns(filters[:managed_school_urns]) if filters[:managed_school_urns]&.any? && filters[:managed_organisations] != 'school_group'

    return vacancies.in_central_office if filters[:managed_organisations] == 'school_group' && filters[:managed_school_urns]&.none?

    vacancies.in_school_urns(filters[:managed_school_urns]).or(vacancies.joins(:school).in_central_office)
  end
```

cesidio: the or query causes problems because of the joins

cpjmcquillan: Both sides of the or query need to be the same structure (except for the where). We can't have a join for schools on the `in_central_office` scope because these vacancies don't have a school_id. We need to join on the `in_school_urns` scope because urn isn't a column on the database table. 

Approving the PR and merging for now, and this method will be re-implemented in due course. One potential option could be to manage organisation preferences by school_id instead of by urn, this way we don't need do a join at all.

PR #1837, opened 2020-08-04
Trust or School overview should be conditional on vacancy.job_location

david-mears-dfe:
A job belonging to a school group which is located at a single school should display the school overview tab not the trust overview tab.

### DIFF
--- a/app/views/shared/vacancy/_jobseeker_view.html.haml
+++ b/app/views/shared/vacancy/_jobseeker_view.html.haml
@@ -30,11 +30,8 @@
               = t('schools.school_overview')
 
       = render(Jobseekers::VacancyDetailsComponent.new(vacancy: @vacancy))
-
-      - if @vacancy.school_group
-        = render(Jobseekers::SchoolGroupOverviewComponent.new(vacancy: @vacancy))
-      - else
-        = render(Jobseekers::SchoolOverviewComponent.new(vacancy: @vacancy))
+      = render(Jobseekers::SchoolGroupOverviewComponent.new(vacancy: @vacancy))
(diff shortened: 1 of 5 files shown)


app/views/shared/vacancy/_jobseeker_view.html.haml
```diff
@@ -24,14 +24,14 @@
...
-            - if @vacancy.school_group
+            - if @vacancy.job_location == 'central_office'
               = t('school_groups.trust_overview')
             - else
               = t('schools.school_overview')
 
       = render(Jobseekers::VacancyDetailsComponent.new(vacancy: @vacancy))
 
-      - if @vacancy.school_group
+      - if @vacancy.job_location == 'central_office'
```
cpjmcquillan, line 34: Since you're here..... and you've opened the can of worms...

May I request we remove the conditional logic from the template, and utilise the `render?` method in the respective ViewComponents (and add relevant tests to each component) please?

app/presenters/vacancy_presenter.rb
```diff
@@ -108,4 +108,8 @@ def show_job_roles
   def show_subjects
     model.subjects&.join(', ')
   end
+
+  def immediate_parent_organisation_type
```
cpjmcquillan, line 112: My concern with adding this code is two-fold:

1. We're moving away from the use of presenters
2. This tabs design is being replaced imminently with a pills design

My proposal would be either:

- Leave the small piece of logic for the tab heading as it was, or
- Use a small inline view_component 

I would lean towards the first option since this will be refactored soon anyway, and this isn't being done as part of any ticket

PR #1840, opened 2020-08-05
Server-side polygon search

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-950

## Changes in this PR:

- Use LocationPolygons to construct Algolia search query for location categories.
- Rename coordinates to point coordinates within search builder, to distinguish it from polygon coordinates.
- Move import polygons task to lib/
- To test locally, you should import LocationPolygons with:

```ruby
require 'import_polygons'
ImportPolygons.new(location_type: :regions).call; ImportPolygons.new(location_type: :counties).call; ImportPolygons.new(location_type: :cities).call; ImportPolygons.new(location_type: :london_boroughs).call
...
checking_all_location_types = [LocationPolygon.regions.size, LocationPolygon.counties.size, LocationPolygon.cities.size, LocationPolygon.london_boroughs.size]
=> [9, 68, 106, 33]
```

I have not yet checked if it's definitely having the desired effect on front end - that Algolia is using the polygons appropriately, and we are no longer getting the results a text search would give. I can see that the API requests are coming through with the correct parameters.

Screenshot 2020-08-04 at 16 53 34

## Next steps:

- [ ] Product signs off the search results on pre-production site.

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -9,7 +9,7 @@ class VacanciesController < ApplicationController
       @vacancies_search.vacancies,
       searched: @vacancies_search.any?,
       total_count: @vacancies_search.vacancies.raw_answer['nbHits'],
-      coordinates: @vacancies_search.coordinates
+      coordinates: @vacancies_search.point_coordinates
     )
     AuditSearchEventJob.perform_later(audit_row) if valid_search?
     expires_in 5.minutes, public: true
--- a/app/views/pages/home/_search.html.haml
+++ b/app/views/pages/home/_search.html.haml
@@ -17,7 +17,7 @@
     'aria-autocomplete': 'list',
     'aria-describedby': 'js-location-finder__error',
     'role': 'combobox',
-    'data-coordinates': @vacancies_search&.coordinates
+    'data-coordinates': @vacancies_search&.point_coordinates
 
   = f.govuk_collection_select :radius,
     radius_filter_options,
--- a/app/views/vacancies/_filters.html.haml
+++ b/app/views/vacancies/_filters.html.haml
@@ -19,7 +19,7 @@
       'aria-autocomplete': 'list',
       'aria-describedby': 'js-location-finder__error',
       'role': 'combobox',
-      'data-coordinates': @vacancies_search&.coordinates
+      'data-coordinates': @vacancies_search&.point_coordinates
 
     .js-location-finder__link.govuk-body-s
       or
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -86,7 +86,7 @@ Metrics/BlockLength:
     Exclude:
       - 'spec/**/*'
       - 'config/routes.rb'
-      - 'lib/tasks/**'
+      - 'lib/tasks/data.rake'
 
 Layout/EmptyLinesAroundAccessModifier:
     Enabled: false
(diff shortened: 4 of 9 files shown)


spec/components/jobseekers/school_group_overview_component_spec.rb
```diff
@@ -11,10 +11,6 @@
     expect(rendered_component).to include(organisation_type(vacancy.school_group))
   end
 
-  it 'renders the trust website label' do
```
cpjmcquillan, line 14: This shouldn't be here

spec/lib/import_polygons_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'import_polygons'
```
cpjmcquillan, line 2: I don't understand why this is needed?

david-mears-dfe: Most of our lib specs have a require line or they don't load the class they're testing. I guess this is because lib/ is not in app/. Could do something like the existing `Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }` which is in the rails_helper.rb.

spec/components/jobseekers/school_group_overview_component_spec.rb
david-mears-dfe: True

spec/lib/import_polygons_spec.rb
david-mears-dfe: I'll make that a separate tech debt ticket

david-mears-dfe: https://github.com/DFE-Digital/teacher-vacancy-service/pull/1844/files

app/controllers/vacancies_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
       @vacancies_search.vacancies,
       searched: @vacancies_search.any?,
       total_count: @vacancies_search.vacancies.raw_answer['nbHits'],
-      coordinates: @vacancies_search.coordinates
+      coordinates: @vacancies_search.point_coordinates
```
cpjmcquillan, line 11: I think we can remove this line now

app/views/pages/home/_search.html.haml
```diff
@@ -17,7 +17,7 @@
     'aria-autocomplete': 'list',
     'aria-describedby': 'js-location-finder__error',
     'role': 'combobox',
-    'data-coordinates': @vacancies_search&.coordinates
+    'data-coordinates': @vacancies_search&.point_coordinates
```
cpjmcquillan, line 20: We can remove this line

app/views/vacancies/_filters.html.haml
```diff
@@ -19,7 +19,7 @@
       'aria-autocomplete': 'list',
       'aria-describedby': 'js-location-finder__error',
       'role': 'combobox',
-      'data-coordinates': @vacancies_search&.coordinates
+      'data-coordinates': @vacancies_search&.point_coordinates
```
cpjmcquillan, line 22: And we can remove this line

app/controllers/vacancies_controller.rb
cpjmcquillan: Correction - we can't remove this (front end code this shows/hides radius input uses this data attribute)

app/views/pages/home/_search.html.haml
cpjmcquillan: See above

app/views/vacancies/_filters.html.haml
cpjmcquillan: See above

PR #1847, opened 2020-08-06
1117 about school blank form validation error

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1117

## Changes in this PR:

- Changed error messages when about school is not valid. Now depends on the location of the vacancy.
- Added tests for the above.

- Manually added @vacancy.job_location from the job_summary controller to the job_summary_form object so that the validation method knows the location of the vacancy and can change depending on its value.


spec/form_models/job_summary_form_spec.rb
```diff
@@ -19,17 +19,37 @@
...
+          let(:job_location) { nil }
+
+          it 'requests the user to complete the about school field' do
+            expect(job_summary_form.valid?).to be false
+            expect(job_summary_form.errors.messages[:about_school].first).to match(/school/)
+          end
+        end
+
+        context 'for a vacancy at a school group' do
+          let(:vacancy) { create(:vacancy, :with_school_group) }
```
david-mears-dfe, line 46: I know I created most of these but I think we aren't using any of these `let(:vacancy)`s or `let(:school)`s: they are never referenced in the test examples.

app/models/concerns/vacancy_job_summary_validations.rb
```diff
@@ -3,6 +3,18 @@ module VacancyJobSummaryValidations
 
   included do
     validates :job_summary, presence: true
-    validates :about_school, presence: true
+    validate :about_school_must_not_be_blank
+  end
+
+  def about_school_must_not_be_blank
+    if defined?(vacancy) && vacancy.job_location == 'central_office'
```
david-mears-dfe, line 10: For reviewers: `if defined?(vacancy)` is here because `vacancy` is initialized by `VacancyForm#initialize`, and creating a vacancy (e.g. with FactoryBot) runs the validations separately.

spec/form_models/job_summary_form_spec.rb
```diff
@@ -19,17 +19,34 @@
...
 
-        it 'requests an entry in the field' do
-          expect(job_summary_form.valid?).to be false
-          expect(job_summary_form.errors.messages[:about_school])
-            .to include(
-              I18n.t('activemodel.errors.models.job_summary_form.attributes.about_school.blank')
-            )
+          it 'requests the user to complete the about school field' do
+            expect(job_summary_form.valid?).to be false
+            expect(job_summary_form.errors.messages[:about_school].first).to match(/school/)
```
cpjmcquillan, line 30: I'd prefer to equal the message exactly

app/models/concerns/vacancy_job_summary_validations.rb
cesidio: I prefer:
```
if vacancy&.job_location == 'central_office'
```

app/models/concerns/vacancy_job_summary_validations.rb
```diff
@@ -3,6 +3,18 @@ module VacancyJobSummaryValidations
...
+      organisation = 'trust'
+    else
+      organisation = 'school'
+    end
+    errors.add(:about_school,
+      I18n.t('job_summary_errors.about_school.blank',
+      organisation: organisation)
+    ) if about_school.blank?
```
cesidio, line 18: I would do:
```
def about_school_must_not_be_blank
  return unless about_school.blank?
  ...
end
```

app/models/concerns/vacancy_job_summary_validations.rb
david-mears-dfe: Unfortunately, this throws an error because vacancy is not defined. I might try to pass in an argument to this method. What do you think about that tactic?

david-mears-dfe: I ended up using `begin` `rescue NameError`

PR #1852, opened 2020-08-10
Refactor search inputs into partials and remove aria-describedby

cpjmcquillan:
This PR addresses TEVA-1141 in the sense that the automated accessibility score will be higher, but I don't think that means it is necessarily a good thing to do.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1141

## Changes in this PR
- Refactor search form inputs into shared partials
- Remove aria-describedby attribute because the error element is added by JS and not present in the template

### DIFF
--- a/app/views/pages/home/_search.html.haml
+++ b/app/views/pages/home/_search.html.haml
@@ -2,33 +2,9 @@
 
 = form_for @jobs_search_form, as: 'jobs_search_form', url: jobs_path(anchor: 'vacancy-results'), html: { method: 'get' } do |f|
 
-  = f.govuk_text_field :keyword,
-    label: { text: t('jobs.filters.keyword'), size: 's' },
-    placeholder: t('jobs.filters.keyword_hint')
-
-  = f.govuk_text_field :location,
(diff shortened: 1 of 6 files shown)


app/views/pages/home/_search.html.haml
```diff
@@ -2,33 +2,9 @@
...
-    label: { text: t('jobs.filters.within_radius'), size: 's' },
-    html_options: { class: 'govuk-!-width-full' },
-    form_group_classes: 'location-radius-select location-radius-select-inline-block'
-
-  .js-location-finder__link.govuk-body-s
-    or
-    = link_to t('jobs.filters.use_current_location'), '#', id: 'current-location', class: 'govuk-link govuk-link--no-visited-state'
+  = render 'shared/search/keyword', f: f
+  = render 'shared/search/location', f: f
+  = render 'shared/search/radius', f: f, filters: false
```
alexbowen, line 7: is filters false needed?

cpjmcquillan: Yes, otherwise you get an undefined error if it's not there (which is annoying)

PR #1860, opened 2020-08-11
Refactor job role options to enum

cpjmcquillan:
What it says on the proverbial tin...

## Why this, why now
This is being done in the interests of consistency (see working_patterns). It will also help us avoid nasty issues if we change how the data is presented at a later date. 

It will also help us filter hiring staff jobs in their dashboard easily and more performantly if we need to:
```ruby
Vacancy.with_job_roles(:nqt_suitable)
```

## Next steps
Next on the menu is the job_location field, which should undergo a similar operation before we start implementing `multiple_schools` as an option


cpjmcquillan:
The migration to convert job roles isn't working as expected. Please do not merge this PR while I troubleshoot

PR #1864, opened 2020-08-11
Show age range in search snippet for vacancies at single schools

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1088

## Changes in this PR:

- Display age range alongside school type and religious character, but only in search snippets. For a singular school in a trust, school type is displayed with religious character and age range (i.e. it here behaves like non-trust vacancies, as desired).
- Removed redundant test examples for vacancy summary view component to save time running test suite
- Refactor: changed out `School#has_religious_character?` for `School#religious_character`

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -11,15 +11,9 @@ module OrganisationHelper
 
   def organisation_type(organisation: organisation, with_age_range: false)
     return organisation.group_type unless organisation.is_a?(School)
-    school_type = school_type_and_religious_character(organisation)
-    school_type.push age_range_or_nil(organisation) if with_age_range
-    school_type.reject(&:blank?).join(', ')
-  end
-
-  def school_type_and_religious_character(school)
-    school_type = [school.school_type.label.singularize]
-    school_type.push school.gias_data['ReligiousCharacter (name)'] if school.has_religious_character?
-    school_type
+    school_type_details = [organisation.school_type.label.singularize, organisation.religious_character]
+    school_type_details.push age_range(organisation) if with_age_range
+    school_type_details.reject(&:blank?).reject { |str| str == I18n.t('schools.not_given') }.join(', ')
   end
 
   def organisation_type_basic(organisation)
@@ -41,11 +35,8 @@ module OrganisationHelper
   end
 
   def age_range(school)
-    age_range_or_nil(school) || I18n.t('schools.not_given')
-  end
-
-  def age_range_or_nil(school)
-    "#{school.minimum_age} to #{school.maximum_age}" if school.minimum_age? && school.maximum_age?
+    return I18n.t('schools.not_given') unless school.minimum_age? && school.maximum_age?
+    "#{school.minimum_age} to #{school.maximum_age}"
   end
 
   def edit_vacancy_section_number(section, organisation)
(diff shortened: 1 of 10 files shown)


spec/components/jobseekers/school_overview_component_spec.rb
```diff
@@ -28,7 +28,10 @@
   end
 
   it 'renders the school type' do
-    expect(rendered_component).to include(organisation_type(vacancy.school))
+    expect(rendered_component).to include(organisation_type(organisation: vacancy.school,
```
cesidio, line 31: test should make only one assertion

spec/models/school_spec.rb
```diff
@@ -7,24 +7,24 @@
...
 
-  describe '#has_religious_character?' do
+  describe '#religious_character' do
     before do
       subject.gias_data = {}.to_json
     end
 
-    it 'returns false when the school has no gias_data' do
+    it 'returns nil when the school has no gias_data' do
       subject.gias_data = nil
```
cesidio, line 16: this should be set outside in a `let`

spec/models/school_spec.rb
```diff
@@ -7,24 +7,24 @@
...
-      expect(subject.has_religious_character?).to be false
+      expect(subject.religious_character).to eq nil
     end
 
-    it 'returns false when the school has no religious_character' do
+    it 'returns nil when the school has no religious_character' do
       allow(subject.gias_data).to receive(:[]).with('ReligiousCharacter (name)').and_return 'Does not apply'
```
cesidio, line 21: `gias_data` should be also set in a `let` and the content defined outside

app/helpers/organisation_helper.rb
```diff
@@ -40,11 +35,11 @@ def ofsted_report(school)
   end
 
   def age_range(school)
-    if school.minimum_age && school.maximum_age?
-      "#{school.minimum_age} to #{school.maximum_age}"
-    else
-      I18n.t('schools.not_given')
-    end
+    age_range_or_nil(school) || I18n.t('schools.not_given')
```
cesidio, line 38: I would rewrite this method like this:

```
def age_range(school)
  return I18n.t('schools.not_given') unless school.minimum_age? && school.maximum_age?
  "#{school.minimum_age} to #{school.maximum_age}"
end
```

david-mears-dfe: I was trying to avoid duplicating `school.minimum_age? && school.maximum_age?`, but maybe it's OK to duplicate something so small.

david-mears-dfe: I found a way to do this without using a method age_range_or_nil - I reject `I18n.t('schools.not_given')` in `organisation_type`

spec/models/school_spec.rb
```diff
@@ -7,24 +7,30 @@
...
 
-  describe '#has_religious_character?' do
-    before do
-      subject.gias_data = {}.to_json
-    end
+  describe '#religious_character' do
+    let(:religious_character) { 'Roman Catholic' }
+    let(:gias_data) { { 'ReligiousCharacter (name)' => religious_character } }
+
+    before { allow(subject).to receive(:gias_data).and_return gias_data }
```
cesidio, line 14: I think this needs some refactoring. I don't understand why a normal field like `gias_data` is managed this way. we should just have something like

```
    subject { build(:school, gias_data: gias_data) }

    let(gias_data) { {...} }
```

PR #1865, opened 2020-08-12
Add hardcoded meta keywords tag to meta partial

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1106

## Changes in this PR:

- Hardcoded meta keyword tags have been added to the meta partial.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,4 +24,9 @@ module ApplicationHelper
       t('app.description')
     end
   end
+
+  def meta_keywords
+    "teacher,maths,primary,history,english,science,geography,\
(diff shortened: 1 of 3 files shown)


alexbowen:
i think this needs to be added to `app/views/shared/_open_graph.html.haml` also?

alexbowen:
i think a simple test would be good features/application_meta_spec.rb  or something - doesnt need to be anything more than they are present for now imo

alexbowen:
> i think this needs to be added to `app/views/shared/_open_graph.html.haml` also?

actually keywords are not needed for open graph i just checked, so dont worry about this

PR #1869, opened 2020-08-12
Fix school/trust overview heading

david-mears-dfe:
By moving logic from views to vacancy model.

## Jira ticket URL

None - this seemed small so I just went for it.

## Changes in this PR:

- Fix bug. See screenshots. We want to show a school overview for vacancies at 1 school in a trust. This was inconsistently implemented.
- Move the logic to a method on the vacancy model. Where should this method live? I avoided using a helper or presenter as I know we want to eventually get rid of these. The method is used in three places, not in any particular view component.

### Before
Screenshot 2020-08-12 at 12 28 15

### After
Screenshot 2020-08-12 at 12 33 39

### DIFF
--- a/spec/models/vacancy_spec.rb
+++ b/spec/models/vacancy_spec.rb
@@ -478,4 +478,32 @@ RSpec.describe Vacancy, type: :model do
       end
     end
   end
+
+  describe '#overview_heading' do
+    context 'a vacancy at a trust head office' do
+      subject { build(:vacancy, :with_school_group) }
+
+      it 'enables translation to generate a TRUST overview heading' do
+        expect(subject.overview_heading).to eq('Trust overview')
+      end
+    end
+
+    context 'a vacancy at a single school' do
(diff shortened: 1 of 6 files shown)


cpjmcquillan:
I have flirted with the idea of InlineViewComponents precisely for usages like this. 

I don't think it should live on the model, but I don't have any strong suggestions about exactly where it should live. @cesidio what do you think? We probably need to have a sit down (virtual or otherwise) and chat about our longer term strategy for things like this!

spec/models/vacancy_spec.rb
```diff
@@ -478,4 +478,32 @@
       end
     end
   end
+
+  describe '#overview_heading' do
+    context 'a vacancy at a trust head office' do
```
cesidio, line 483: context should always start with `when` or `with`

david-mears-dfe:
Obsoleted by an other PR fixing this issue

PR #1873, opened 2020-08-13
Update organisations to use STI

cesidio:
This make School and SchoolGroup become subclasses of Organisation model.

### DIFF
--- a/db/migrate/20200813105100_remove_foreign_keys.rb
+++ b/db/migrate/20200813105100_remove_foreign_keys.rb
@@ -1,0 +1,6 @@
+class RemoveForeignKeys < ActiveRecord::Migration[5.2]
+  def change
+    remove_foreign_key :user_preferences, :school_groups
+    remove_foreign_key :vacancies, :school_groups
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/tasks/create_organisations.rake
```diff
@@ -11,5 +11,11 @@ namespace :data do
         Organisation.create(school_group.attributes)
       end
     end
+
+    desc 'Update organisations data'
```
alexbowen, line 15: indentation looks slightly out here?

alexbowen:
whilst code looks pretty straightfoward change, could you add description as i dont what its supposed to acheive?

db/migrate/20200813105100_remove_foreign_keys.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveForeignKeys < ActiveRecord::Migration[5.2]
+  def change
+    remove_foreign_key :user_preferences, :school_groups
+    remove_foreign_key :vacancies, :school_groups
```
david-mears-dfe, line 4: What replaces these removed foreign keys?

cesidio: Nothing for now

lib/tasks/create_organisations.rake
cesidio: Indentation is correct here, it has to be aligned to the task

PR #1881, opened 2020-08-14
Jobs page filters

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1101

## Changes in this PR
- Add filters to the jobs page


app/frontend/src/lib/ui/components/filterGroup.js
```diff
@@ -16,6 +16,11 @@ export const init = (groupContainerSelector, removeButtonSelector, clearButtonSe
   addFilterChangeEvent(document.getElementsByClassName(groupContainerSelector));
 
   if (document.getElementById(closeButtonSelector)) {
+    if (document.getElementById(closeButtonSelector).innerText === 'Close all' && document.getElementsByClassName('govuk-accordion__section--expanded').length === 0) {
+      document.getElementById(closeButtonSelector).innerText = 'Open all';
```
alexbowen, line 20: seeing as you are using these open all and close all string in a few places would be better to make them constants like `const OPEN_ALL_BUTTON_TEXT = 'Open All'` or similar. then you could export these constants and import them into the unit test and use them where those strings are used and makes the test a bit tighter to the code its testing

app/frontend/src/lib/ui/components/filterGroup.test.js
```diff
@@ -218,15 +219,31 @@ describe('filterGroup', () => {
 
   describe('closeAllSectionsHandler', () => {
     test('removes the class selector from all filter groups that makes them visible', () => {
-      document.body.innerHTML = '<div class="govuk-accordion__section"></div><div class="govuk-accordion__section"></div><div class="govuk-accordion__section"></div>';
+      document.body.innerHTML = `<div class="govuk-accordion__section govuk-accordion__section--expanded"></div>
+      <div class="govuk-accordion__section govuk-accordion__section--expanded"></div>
+      <div class="govuk-accordion__section"></div>`;
 
-      const event = { preventDefault: jest.fn() };
+      const event = { preventDefault: jest.fn(), target: { innerText: 'Close all' } };
```
alexbowen, line 226: see above comment about importing constant

app/frontend/src/lib/ui/components/filterGroup.js
```diff
@@ -33,11 +38,29 @@ export const init = (groupContainerSelector, removeButtonSelector, clearButtonSe
...
+};
+
+export const showMobileFilters = (element) => {
+  element.classList.toggle('filters--show-mobile');
 };
 
 export const closeAllSectionsHandler = (e) => {
   e.preventDefault();
-  Array.from(document.getElementsByClassName('govuk-accordion__section')).map((section) => section.classList.remove('govuk-accordion__section--expanded'));
+  if (e.target.innerText === 'Close all') {
```
alexbowen, line 57: something doesnt feel right here as the logic seems very similar as to what you've added in the init function (and what is in the init function isnt tested) and im sure they could be consolidated in some way and hence tested. I could get a PR together to branch if you want and ill give it a bit more thought

app/frontend/src/lib/ui/components/filterGroup.js
```diff
@@ -27,17 +30,41 @@ export const init = (groupContainerSelector, removeButtonSelector, clearButtonSe
...
+export const displayOpenOrCloseText = (targetElement, expandedElements) => {
+  if (targetElement.innerText === CLOSE_ALL_TEXT && expandedElements === 0) {
+    targetElement.innerText = OPEN_ALL_TEXT;
+  } else if (targetElement.innerText === OPEN_ALL_TEXT && expandedElements > 0) {
+    targetElement.innerText = CLOSE_ALL_TEXT;
+  }
+};
+
+export const openOrCloseAllSectionsHandler = (e) => {
   e.preventDefault();
```
alexbowen, line 56: i dont think default needs to be prevented as all of these are buttons now (without type="submit" attribute and not a tags)

PR #1883, opened 2020-08-18
[TEVA-1002] more than one school job location

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1002

## Changes in this PR:

- Migrated job_location to use integers
- Created job_location enum and job_location_options helper
- Removed redundant test and validation
- Split options helpers in VacanciesHelpers module into a new VacanciesOptionsHelpers module. Included VacanciesOptionsHelpers in the VacanciesHelpers module.

Radio Buttons and error message:

Screenshot 2020-08-19 at 12 27 43

### DIFF
--- a/app/helpers/vacancies_options_helper.rb
+++ b/app/helpers/vacancies_options_helper.rb
@@ -1,0 +1,42 @@
+module VacanciesOptionsHelper
+  def hired_status_options
+    Vacancy.hired_statuses.keys.map { |k| [t("jobs.feedback.hired_status.#{k}"), k] }
+  end
+
+  def job_location_options
+    Vacancy::JOB_LOCATION_OPTIONS.map { |key, _value|
+      [I18n.t("helpers.fieldset.job_location_form.job_location_options.#{key}"), key] if
+        MultiSchoolJobsFeature.enabled? || key != :at_multiple_schools
+    }.reject(&:blank?)
+  end
+
+  def job_role_options
+    Vacancy::JOB_ROLE_OPTIONS.except(:nqt_suitable).map do |key, _value|
+      [I18n.t("jobs.job_role_options.#{key}"), key]
+    end
+  end
+
+  def job_sorting_options
+    Vacancy::JOB_SORTING_OPTIONS
+  end
+
+  def listed_elsewhere_options
+    Vacancy.listed_elsewheres.keys.map { |k| [t("jobs.feedback.listed_elsewhere.#{k}"), k] }
+  end
+
+  def radius_filter_options
+    [1, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 70, 80, 90, 100, 200].inject([]) do |radii, radius|
+      radii << [I18n.t('jobs.filters.number_of_miles', count: radius), radius]
+    end
+  end
+
+  def subject_options
+    SUBJECT_OPTIONS
+  end
+
+  def working_pattern_options
+    Vacancy::WORKING_PATTERN_OPTIONS.keys.map do |key|
+      [Vacancy.human_attribute_name("working_patterns.#{key}"), key.to_s]
+    end
+  end
+end
--- a/db/migrate/20200818111239_change_job_location_to_integer_on_vacancies.rb
+++ b/db/migrate/20200818111239_change_job_location_to_integer_on_vacancies.rb
@@ -1,0 +1,9 @@
+class ChangeJobLocationToIntegerOnVacancies < ActiveRecord::Migration[5.2]
+  def up
+    change_column :vacancies, :job_location, :integer, using: 'job_location::integer'
+  end
+
+  def down
+    change_column :vacancies, :job_location, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_08_13_105100) do
+ActiveRecord::Schema.define(version: 2020_08_18_111239) do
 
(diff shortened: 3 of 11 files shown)


david-mears-dfe:
> You don't need to do the migration because the database doesn't have any data yet.

I can see how this applies to the conversion of legacy data, but we do need to migrate job_location from string to integer if we're to use enums.

david-mears-dfe:
> I'm unsure how to change the order of the radio buttons. I can see they're different in the figma. 'At more than one school in the trust' is second in the figma.

This is determined by the order of JOB_LOCATION_OPTIONS (or whatever method is transforming that) - the radio buttons are made based on an array passed into govuk_collection_radio_buttons

app/helpers/vacancies_options_helper.rb
```diff
@@ -0,0 +1,42 @@
+module VacanciesOptionsHelper
+  def job_location_options
```
david-mears-dfe, line 2: Mild preference for these to be in alphabetical order for easy finding, if all else is equal

josephhull676: Will do that now sir!

alexbowen:
seems rather light on tests, well a bit lighter than before as one has been removed?!

david-mears-dfe:
Regardless of whether the flag is added by this PR or a separate one, you will need to use it to prevent the new option being displayed.

app/helpers/vacancies_options_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+    Vacancy::JOB_LOCATION_OPTIONS.map do |key, _value|
+      [I18n.t("helpers.fieldset.job_location_form.job_location_options.#{key}"), key]
+    end
+  end
+
+  def job_role_options
+    options = Vacancy::JOB_ROLE_OPTIONS.except(:nqt_suitable).map do |key, _value|
+      [I18n.t("jobs.job_role_options.#{key}"), key]
+    end
+    options.reject { |option| option == 'at_multiple_schools' } unless MultiSchoolJobsFeature.enabled?
```
cpjmcquillan, line 12: There should be a spec for this helper method asserting whether or not the option is included depending on if the feature is enabled or not

josephhull676: Will add the test now

app/helpers/vacancies_options_helper.rb
```diff
@@ -0,0 +1,42 @@
+module VacanciesOptionsHelper
+  def hired_status_options
+    Vacancy.hired_statuses.keys.map { |k| [t("jobs.feedback.hired_status.#{k}"), k] }
+  end
+
+  def job_location_options
```
cpjmcquillan, line 6:
```ruby
Vacancy::JOB_LOCATION_OPTIONS.map { |key, _value| [I18n.t("jobs.job_location_options.#{key}"), key] if MultiSchoolJobsFeature.enabled? || key != :at_multiple_schools  }.reject(&:blank?)
```

db/schema.rb
```diff
@@ -10,14 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_08_13_105100) do
+ActiveRecord::Schema.define(version: 2020_08_18_111239) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "citext"
```
cpjmcquillan, line 16: You should only commit changes to the schema introduced by the migration itself

spec/helpers/vacancies_options_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe VacanciesOptionsHelper, type: :helper do
+  describe '#job_location_options' do
+    context 'when FEATURE_MULTI_SCHOOL_JOBS flag is set to true' do
```
cpjmcquillan, line 5:
```ruby
context 'when MultiSchoolJobsFeature is enabled' do
```

spec/helpers/vacancies_options_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      it 'returns an array including the multi-school option' do
+        allow(MultiSchoolJobsFeature).to receive(:enabled?).and_return(true)
```
cpjmcquillan, line 7: Normally we would do a before block in like this...
```ruby
describe '#job_location_options' do
  before do
    allow(MultiSchoolJobsFeature).to receive(enabled?).and_return(multi_school_jobs_enabled)
  end

  context 'when MultiSchoolJobsFeature is enabled' do
    let(:multi_school_jobs_enabled) { true }

    it 'returns an array including at_multiple_schools' do
      ...
    end
  end

  ...
end
```

db/schema.rb
```diff
@@ -10,15 +10,15 @@
...
+ActiveRecord::Schema.define(version: 2020_08_18_111239) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "fuzzystrmatch"
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
   enable_extension "uuid-ossp"
-
+
```
cpjmcquillan, line 21: whitespace

db/migrate/20200818111239_change_job_location_to_integer_on_vacancies.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeJobLocationToIntegerOnVacancies < ActiveRecord::Migration[5.2]
+  def up
```
cpjmcquillan, line 2: What do you think of this instead of just a change method @cesidio ?

cpjmcquillan (approved):

PR #1885, opened 2020-08-18
upgrade @rails/webpacker from 4.2.2 to 5.2.1

vigneshmsft:
upgrade `@rails/webpacker` to 5.2.1
remove earlier package resolution for `serializer-javascript`


alexbowen:
why do we need to make the upgrade to webpacker?

PR #1890, opened 2020-08-19
This is an attempt to diagnose a BigQuery failure

david-mears-dfe:
We now catch all errors in order that one table's failure need not
cause us to throw out all tables' data (baby, bathwater).

Exclude location polygons as records are long and this table appears to
be causing errors. PA has not specifically requested access to this
table.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1172

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -45,6 +45,7 @@ class ExportTablesToBigQuery
     ar_internal_metadata
     audit_data
     friendly_id_slugs
+    location_polygons
     schema_migrations
     sessions
   ].freeze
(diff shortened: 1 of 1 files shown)


lib/export_tables_to_big_query.rb
```diff
@@ -69,13 +70,12 @@ def run!
...
       bigquery_load(table.constantize)
-    rescue NameError => e
-      # This is to address an edge-case failure when a table, `PayScale`, didn't get dropped from the production db
-      # but its model has been removed from the codebase.
-      #
-      # The `PayScale` table should be removed in a later release, but this rescue will prevent any future oversights
-      # of this nature from breaking the run.
-      Rails.logger.error({ bigquery_export: 'error', status: 'handled', message: e })
+    rescue StandardError => e
+      binding.pry
```
cesidio, line 74: this should not be committed

david-mears-dfe: woops

cesidio (approved):

PR #1892, opened 2020-08-19
Fix BigQuery data feed

david-mears-dfe:
There were two errors similar to previous BigQuery data feed errors.

Firstly, legacy_job_roles (formerly 'job_roles') was being rejected by BigQuery when it was a list of length 1. We do not need this legacy data so I excluded the column from the feed.

Secondly, job_roles (a new enum column) evaluates to a string because of Rails helper methods, while the column type is integer. BigQuery quite rightly rejects any attempt to put a string into an integer column. This problem applies to several other columns already, so I just had to add 'job_roles' to the ENUM_ATTIBUTES constant.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1172

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -23,6 +23,7 @@ class ExportTablesToBigQuery
     geolocation
     gias_data
     job_summary
+    legacy_job_roles
     qualifications
     supporting_documents
   ].freeze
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
These changes fixed the two errors on the local development -> BigQuery staging pipeline. It remains to be seen whether the production pipeline also works. I am trying that right now.

david-mears-dfe:
> These changes fixed the two errors on the local development -> BigQuery staging pipeline. It remains to be seen whether the production pipeline also works. I am trying that right now (by copying and pasting lib/export_tables_to_big_query.rb into the production rails console).

Running the data feed on production appeared to populate the tables, which is good, and it resurrected [page 27 of the service report](https://datastudio.google.com/u/1/reporting/1X_lrbWUn7Nw5LZnRWynJKtalYg6-L4Oi/page/oemOB). It did not fix all graphs on page 26. I expect this is because the calculations run nightly, so I will check them again in the morning.

PR #1893, opened 2020-08-20
[TEVA 1135] Education phase mappings

cpjmcquillan:
This PR adds a new field to the organisations model that maps school phase to a readable school phase. 

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1135

## Changes in this PR
- Add mapped school phase
- Update school update job
- Index mapped school phase and update filters builder

### DIFF
--- a/lib/update_school_data.rb
+++ b/lib/update_school_data.rb
@@ -28,6 +28,8 @@ class UpdateSchoolData
     town: 'Town',
   }
 
+  READABLE_PHASE_MAPPINGS = School::READABLE_PHASE_MAPPINGS
+
   def run!
     save_csv_file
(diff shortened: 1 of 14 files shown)


david-mears-dfe:
> * Maybe an array_enum would be useful here?

If you think a school can have more than one phase? I don't think that's possible.

lib/update_school_data.rb
```diff
@@ -28,6 +28,17 @@ class UpdateSchoolData
     town: 'Town',
   }
 
+  READABLE_PHASE_MAPPINGS = {
```
cesidio, line 31: Is this the same that has been defined in the school model?

cpjmcquillan: Yes, I guess I was being lazy 😂 . How do I access that in /lib?

PR #1905, opened 2020-08-24
Bigquery benchmark against comparable vacancies

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1067

## Changes in this PR:

- Adds 'in scope' view of vacancy table in BigQuery that excludes non-teaching/leadership roles and unpublished roles.
- Calculates number of live vacancies and vacancies published on an alternative jobs board as a benchmark.
- Compares the two as a proportion


alexbowen (approved):
some pretty badass SQL you got going on here! but looks ok even if a bit over my head

stevenleggdfe:
> some pretty badass SQL you got going on here! but looks ok even if a bit over my head

Why thank you :-) What's not to love about an analytic function? ;-)

PR #1907, opened 2020-08-25
fix styles on hover of homepage search inputs

alexbowen:
no ticket

- styles fixed on focus of search input elements on homepage
- removed unused file


app/frontend/styles/components/search_panel.scss
```diff
@@ -45,7 +43,6 @@
 
   .location-radius-select {
     display: inline-block;
-    float: right;
```
cpjmcquillan, line 48: We need this to look like this 
image and not like this 
image

alexbowen, line 48: this makes so sense alongside inline-block so removed it

cpjmcquillan: Nevertheless, we need the top image not the bottom one. This branch regresses to the bottom image

alexbowen: ah ok, i forgot this appeared, but floating is not the best way to do it and IE might have a problem with it but i havent got ie vm at the moment so will investigate when ive cleared some space on my hard drive

cpjmcquillan (approved):

PR #1908, opened 2020-08-25
Remove unused tables and fields

cesidio:


david-mears-dfe:
Does this change anything in the README about how to import schools and school groups?

cpjmcquillan (approved):

cesidio:
> Does this require any change in the README about how to import schools and school groups?

Nope, everything works as before. The records will be created/updated in a different table, but that does not change anything in the README

PR #1889, opened 2020-08-19
Add filters to service start page

david-mears-dfe:
This branch originally based off 'jobs-page-filters' and now has been rebased on master.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1132

## Changes in this PR:

- Adds filters in an openable details section to the search panel on the service start page
- Available on the dev environment at time of writing

### Before

Screenshot 2020-08-19 at 14 38 45

### After

Screenshot 2020-08-19 at 14 31 30

Open details element (2 images):

Screenshot 2020-08-19 at 14 33 07
Screenshot 2020-08-19 at 14 33 41

## Next steps:

- [x] Add tests

- [x] Rebase on master after jobs-page-filters is merged

PR #1910, opened 2020-08-25
Add back in the 1px border between filter groups on jobs results page

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1201

## Changes

- Remove light-grey border
- Add 1px dark-grey border back in
- Prevent the appearance of effect-less scrolls by (1) making checkbox scrollability configurable with a Boolean 'scroll' parameter which adds a 'scrollable' class and (2) reducing the size of [the govuk frontend touch-target](https://github.com/alphagov/govuk-frontend/blob/2b7a162b0f915483190ad56b5b15c543b8cd36c9/src/govuk/components/checkboxes/_index.scss#L47) by 1px to make it fit within the available space.
- No visible change to service start page or job specification subjects checkbox drop down

## Screenshots of UI changes:

| Before | After |
| ------ | ------ |
| Jobs results page | Note different borders and no scrolls |
| Screenshot 2020-08-25 at 19 13 36 | Screenshot 2020-08-25 at 19 12 18 |
| MAT dashboard filter | Note different borders |
| Screenshot 2020-08-25 at 19 13 57 | Screenshot 2020-08-25 at 19 12 00 |


cpjmcquillan:
It hasn't been deleted, it gets hidden by JS. I'd prefer an approach that hides the correct element instead, rather than adding extra CSS that actually already exists and is present

david-mears-dfe:
> It hasn't been deleted, it gets hidden by JS. I'd prefer an approach that hides the correct element instead, rather than adding extra CSS that actually already exists and is present

You mention hiding, but I'm not sure what you think should be hidden. The border should always be visible.

alexbowen (approved):
this might have gone astray when i was fixing another defect regarding font weights of checkbox labels

david-mears-dfe:
> It hasn't been deleted, it gets hidden by JS. I'd prefer an approach that hides the correct element instead, rather than adding extra CSS that actually already exists and is present

It was deleted accidentally by me in https://github.com/DFE-Digital/teacher-vacancy-service/pull/1889/files

What JS hides this border?

cpjmcquillan:
> > It hasn't been deleted, it gets hidden by JS. I'd prefer an approach that hides the correct element instead, rather than adding extra CSS that actually already exists and is present
> 
> It was deleted accidentally by me in https://github.com/DFE-Digital/teacher-vacancy-service/pull/1889/files
> 
> What JS hides this border?

The border as per the Figma exists when the filter is expanded, so if the `.tv-checkbox__group` element was hidden instead of the `.filter-group__checkboxes` element, then the border would still be there. If this approach matches the design then ok, but this looks like a double border to me

david-mears-dfe:
Good spot on the double border. Looks like a conflict between the ticket and the Figma. The Figma shows this 1px border in every instance, and does not show the thicker, lighter grey border in any instance. I will delete the thicker lighter border when I return to this ticket this afternoon.

david-mears-dfe:
Have dismissed Alex's earlier approval as this PR now contains about 10 times as many changes as it did before.

app/frontend/src/lib/ui/components/filterGroup.scss
```diff
@@ -89,7 +93,7 @@
       background: none;
       border: 1px solid govuk-colour("white");
       padding: 0;
-      border-bottom: 2px solid #b1b4b6;
+      border-bottom: 2px solid $govuk-border-colour;
```
alexbowen, line 96: i used the colour b1b4b6 (which isnt a standard gov colour) as it was the same as the border of the tabs. personally i think should also change the tabs border to govuk-border-colour too

david-mears-dfe: As far as I can tell, the border of the tabs comes from here: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/tabs/_index.scss

Which is alphagov's. And it does seem to use govuk-border-colour already, which translates to #b1b4b6.

david-mears-dfe: https://design-system.service.gov.uk/styles/colour/ govuk-border-colour is #b1b4b6

alexbowen: oh does it, fair enough, maybe i didnt read gds colors through thouroughly

alexbowen (approved):

cpjmcquillan (approved):

PR #1911, opened 2020-08-25
update webpacker as outdated yarn dependencies

alexbowen:
no ticket

webpacker needed updating as was conflicting with an outdated dependency in yarn.lock


alexbowen:
running tests locally suggests that a webpacker config `resolved_paths` is deprecated and has changed to `additional paths` but this change will have to be another PR as i think it introduces a race condition in the build being in the same PR as this

alexbowen:
no it didnt

cpjmcquillan:
What did you change to fix the failing specs?

alexbowen:
the specs were failing because originally i had changed resolved_paths -> additional_paths in the webpacker config. this was because when i ran them locally they all passed but many deprecation warnings. when this key was changed in the PR the failures were saying that webpack couldn't compile but the deprecation warnings disappeared and test passed locally (but i did have to make sure i recompiled assets first).

cpjmcquillan:
> the specs were failing because originally i had changed resolved_paths -> additional_paths in the webpacker config. this was changed because when i ran them locally they all passed but with many deprecation warnings. when this key was changed in the PR the failures were saying that webpack couldn't compile although the deprecation warnings disappeared and test passed locally (but i did have to make sure i recompiled assets first which i dont think the gh action pipeline is able to do).
> 
> Screenshot 2020-08-25 at 12 46 03

Ok, can you add a ticket in for addressing the deprecation warning?

cpjmcquillan (approved):

alexbowen:
my plan was to imediately see if another PR is able to pass with this key change else yes create ticket 👍

PR #1918, opened 2020-08-25
Fix Bigquery export for organisations

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1219

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -103,8 +103,8 @@ class ExportTablesToBigQuery
     end
 
     if table.column_names.include?('geolocation')
-      @bigquery_data['geolocation_x'] = record.geolocation.x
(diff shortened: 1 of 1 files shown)


lib/export_tables_to_big_query.rb
```diff
@@ -105,6 +105,8 @@ def bigquery_data(record, table)
     if table.column_names.include?('geolocation')
       @bigquery_data['geolocation_x'] = record.geolocation.x
```
cpjmcquillan, line 106: Do we still need these lines?

cesidio: we have a `geolocation` field in the organisation table.

cesidio: it is of type `point` and I guess what we are doing here is to extract the lat and long

cpjmcquillan (approved):

PR #1926, opened 2020-08-26
[TEVA-1125] Ensure hint and label are inline

cpjmcquillan:
This addresses the last (I pray) of the css inconsistencies between design and the subject input introduced in the move to using the checkbox component
 
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1125

## Changes in this PR:
- Ensure label and hint are inline for checkboxes

### DIFF
--- a/app/frontend/styles/components/accordion_group.scss
+++ b/app/frontend/styles/components/accordion_group.scss
@@ -43,6 +43,10 @@
   .govuk-hint {
     margin-left: govuk-spacing(2);
   }
+
+  .govuk-checkboxes__hint {
+    margin-left: 1px;
+  }
 }
 
 .accordion-content__group .govuk-accordion__section--expanded {


alexbowen (approved):
just one question, why 1px instead of 0 - not that it really matters

alexbowen:
The only true solution to little design inconsistencies like this creeping in is to introduce snapshot testing. But in my experience is a complex thing to implement  and maintain on an app like TV which is predominantly a server side app, and hence the effort/value doesn't add up. This is my experience anyway - I have worked mostly with react or similar/SPAs for the past 4 years and snapshot testing is a *little* easier but still requires some pretty strict rules for coding markup/styles. Maybe a spike is warranted to see if snapshot testing is viable with recent tooling. On a side note I have had a word with myself to pay 100 rather than 95% attention to figma etc but stuff like this is going to happen every now and, unfortunately, as I/we aren't robots

cpjmcquillan:
> just one question, why 1px instead of 0 - not that it really matters

The govuk frontend checkboxes component has 1px padding on the label, if the margin is 0 in the hint then they don't align

david-mears-dfe (approved):

PR #1927, opened 2020-08-26
[TEVA-1188] Anchor page to vacancies on form submit

cpjmcquillan:
This PR anchors both the job results page and the dashboard to the job results.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1188

## Changes in this PR:
- Form submit url is anchored to `#vacancies-hits` elements

### DIFF
--- a/app/views/hiring_staff/organisations/_vacancies_awaiting_feedback.html.haml
+++ b/app/views/hiring_staff/organisations/_vacancies_awaiting_feedback.html.haml
@@ -1,4 +1,4 @@
-%section.govuk-tabs__panel
+%section.govuk-tabs__panel#vacancies-hits
   %h2.govuk-heading-m= t("jobs.#{selected_type}_jobs_with_count", count: vacancies.count)
 
   - if vacancies.any?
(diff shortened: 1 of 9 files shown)


cesidio (approved):

app/views/hiring_staff/organisations/_vacancies_awaiting_feedback.html.haml
```diff
@@ -1,4 +1,4 @@
-%section.govuk-tabs__panel
+%section.govuk-tabs__panel#vacancies-hits
```
alexbowen, line 1: wouldn't it be better to put this line outside of all these templates in the presumably one parent template so all the duplication here is removed?

alexbowen:
What is done in this PR means that a user using keyboard navigation cannot tab to the search inputs once they have used a filter

alexbowen:
Or to rephrase this pr isn't incorrect but need in addition some clever use of aria tags and possibly extra link somewhere. Let's discuss tomorrow

app/views/shared/_skip_links.html.haml
```diff
@@ -0,0 +1,3 @@
+%a.govuk-skip-link{ href: "#main-content" } Skip to main content
+- if params[:jobs_search_form]
+  %a.govuk-skip-link{ href: "#vacancies-hits" } Skip to search results
```
cpjmcquillan, line 3: New line

alexbowen (approved):

PR #1928, opened 2020-08-26
[TEVA-1165] filter side panel persistent state

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1165

- refactor open close/panel into separate script so is not coupled to filters specifically
- add tests for new panel component
- add persistence to open/closed state using local storage for each page load
- integrated with changes in master to do with column widths

Test Coverage

Screenshot 2020-08-26 at 18 46 28

### DIFF
--- a/app/frontend/src/lib/ui/components/panel.test.js
+++ b/app/frontend/src/lib/ui/components/panel.test.js
@@ -1,0 +1,55 @@
+import { toggleButtonText, isPanelClosed, panelToggle } from './panel';
+
+describe('panel', () => {
+  document.body.innerHTML = '<div id="panel-container"></div><button id="toggle-button">original text</button>';
+
+  const HIDE_BUTTON_TEXT = 'Hide panel';
+  const SHOW_BUTTON_TEXT = 'Show panel';
+  const CLOSED_CLASS = 'closed';
+  const container = document.getElementById('panel-container');
+  const button = document.getElementById('toggle-button');
+  const options = {
+    container,
+    toggleButton: button,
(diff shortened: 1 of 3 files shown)


app/frontend/src/lib/ui/components/panel.test.js
```diff
@@ -0,0 +1,77 @@
...
+      showText: SHOW_BUTTON_TEXT,
+      toggleClass: CLOSED_CLASS,
+      onToggleHandler: jest.fn(),
+      onClosedHandler: jest.fn(),
+      onOpenedHandler: jest.fn(),
+    };
+  });
+
+  describe('toggleButtonText', () => {
+    test('changes button text to show panel state when panel is hidden', () => {
```
cpjmcquillan, line 29: I'm not sure this message matches up with line 31 properly? Or maybe I'm just being confused today 😂

app/frontend/src/lib/ui/components/panel.test.js
```diff
@@ -0,0 +1,77 @@
...
+      toggleButtonText(options);
+      expect(button.innerHTML).toBe(HIDE_BUTTON_TEXT);
+    });
+
+    test('changes button text to hide panel state when panel is visible', () => {
```
cpjmcquillan, line 34: Same as above

cpjmcquillan (approved):

app/frontend/src/lib/ui/components/panel.test.js
alexbowen: you are right they round the wrong way

PR #1931, opened 2020-08-27
Bigquery organisation data model migration

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1070

## Changes in this PR:

- Migrate all references to the now-deleted feb20_school table to use the new feb20_organisation and feb20_organisationvacancy tables which replace it
- Create new school view of the feb20_organisation table which contains only the organisations that are schools and fields that are relevant
- Refactor various field processing functions (like address field concatenation) into the school view and out of individual scheduled queries/views
- Remove (now redundant) metrics that rely on the concept of a number of vacancies being broken down by region, as now vacancies can be from multiple schools it is no longer conceptually possible to break these counts down by region e.g. % vacancies in a particular LA offering flexible working
- Move all queries which define views into a views folder to distinguish them from queries that are scheduled queries
- Simplify school field naming, make it more consistent and reduce the number of aliases used
- Redefine breadth of usage metrics slightly (i.e. the % of (signed up) schools which have published a vacancy within the last year/quarter/ever, possibly broken down by RSC region) so that a vacancy listed as being at multiple schools counts as all those schools having published a vacancy, but a vacancy listed as at a school group (i.e. a MAT) does *not* count as all the schools in the school group having published a vacancy


alexbowen (approved):
i guess the google cloud platform screenshot you posted is the proof of pudding so approving!

PR #1945, opened 2020-09-01
Ensure form renders selected job roles

cpjmcquillan:
Array_enum returns job_roles as strings, so the options need to have string values otherwise they won't be selected in the form.


david-mears-dfe:
(copied from [Slack thread](https://ukgovernmentdfe.slack.com/archives/CQ3UX00P6/p1598954025004200):) Is there a not-very-expensive test that you can add which makes this less likely to happen again?

cesidio (approved):

PR #1950, opened 2020-09-01
[TEVA-1227] Limit some scheduled jobs to only prod environment

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1227


cpjmcquillan (approved):

app/jobs/export_dsi_users_to_big_query_job.rb
```diff
@@ -5,6 +5,8 @@ class ExportDsiUsersToBigQueryJob < ApplicationJob
   queue_as :export_users
 
   def perform
+    return unless Rails.env.production?
```
saliceti, line 8: is this a standard approach to hardcode the environment in Rails? I usually prefer to have a variable set in the environment. Like `if setting.jobs_enabled`. So you can have more freedom to enable/disable per environment.

PR #1955, opened 2020-09-02
Fix bug when creating a vacancy with one school

cesidio:
We did not read the organisation_id from the session correctly


david-mears-dfe (approved):
Excellent

PR #1956, opened 2020-09-02
[TEVA-1003] Multi school vacancies

cpjmcquillan:
This PR allows jobs to be created that are based at more than one school

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1003

## Changes in this PR
- Ensure only distinct vacancies are shown in the dashboard
- Allow jobs at multiple schools to be created

## Next steps
- Every instance of `vacancy.organisation` needs to be refactored


app/controllers/hiring_staff/vacancies/job_location_controller.rb
```diff
@@ -13,7 +13,7 @@ def show
   def create
     @form.vacancy.readable_job_location = readable_job_location(@form.vacancy.job_location)
     store_vacancy_attributes(@form.vacancy.attributes)
-    session[:organisation_id] = current_organisation.id
+    session[:vacancy_attributes]['organisation_id'] = current_organisation.id
```
david-mears-dfe, line 16: I believe Cesi fixed this in the opposite direction in https://github.com/DFE-Digital/teacher-vacancy-service/pull/1955  : he changed the code that reads this session value, but you changed the code that stores it.
```
-       @form.vacancy.organisation_vacancies.build(organisation_id: session[:vacancy_attributes]['organisation_id'])
+.      @form.vacancy.organisation_vacancies.build(organisation_id: session[:organisation_id])
```

david-mears-dfe: Either way seems fine

david-mears-dfe: Yours slightly neater

cpjmcquillan: Also, if it's not stored in vacancy_attributes it leaks when creating new job listings

david-mears-dfe: Actually, since you have written `.except('organisation_id', 'organisation_ids'` twice (in the code I've so far seen), I think Cesi's way is neater.

cpjmcquillan: See above

david-mears-dfe: But as you prefer

cpjmcquillan: I mean that issue could be fixed a different way, but I think I prefer it there

app/controllers/hiring_staff/vacancies/schools_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    return redirect_to next_step if job_location == 'central_office'
+    @form = SchoolsForm.new(attributes.symbolize_keys)
+  end
+
+  def create
+    @form.vacancy.readable_job_location = readable_job_location(job_location, school_name: school&.name,
+                                                                schools_count: @form.organisation_ids&.count)
+    store_vacancy_attributes(@form.vacancy.attributes)
+    session[:vacancy_attributes]['organisation_id'] = @form.organisation_id
+    session[:vacancy_attributes]['organisation_ids'] = @form.organisation_ids
```
david-mears-dfe, line 24: Is there some reason we can't use the same attribute for both cases? A single-element organisation_ids array?

Would save a lot of pieces of logic here and there.

spec/system/hiring_staff_can_publish_a_vacancy_as_a_school_group_spec.rb
```diff
@@ -135,15 +138,113 @@
...
+
+          fill_in_job_location_form_field(vacancy)
+          click_on I18n.t('buttons.continue')
+
+          expect(page).to have_content(I18n.t('jobs.current_step', step: 1, total: 8))
+          within('h2.govuk-heading-l') do
+            expect(page).to have_content(I18n.t('jobs.job_location'))
+          end
+
+          expect(Vacancy.count).to eql(0)
```
david-mears-dfe, line 181: There is also `not_to change` (equivalent to `to_not change`) - maybe a slightly more reliable way of putting it in general, because we are testing the difference in the count, not the absolute value of the count.

david-mears-dfe: I don't mind which you choose to use. There are many `expect(Vacancy.count).to eql(0)`, some of which are written by you, some of which are old, so whatever you choose let's regularize

david-mears-dfe: It would require you to pass in the interaction as a code block:

```ruby
expect { do_something }.to change { object.attribute }
```

app/components/hiring_staff/vacancies_component.rb
```diff
@@ -45,7 +45,7 @@ def vacancy_type_tab_link(vacancy_type)
   def set_vacancies
     @vacancies =
       if @filters[:managed_school_ids]&.any?
-        Vacancy.in_organisation_ids(@filters[:managed_school_ids])
+        Vacancy.in_organisation_ids(@filters[:managed_school_ids]).distinct
```
cesidio, line 48: I would put `.distinct` into the scope
```
scope :in_organisation_ids, -> (ids) do
    joins(:organisation_vacancies).where(organisation_vacancies: { organisation_id: ids }).distinct
  end
```
So that we do not have to put it everywhere

app/controllers/hiring_staff/vacancies/schools_controller.rb
cpjmcquillan: While it might save logic in some places, it would introduce logic in lots of other places too because the fields on the form are distinctly different. I'd prefer to just have one form input and users select 1 or more schools, but here we are 🤷‍♂️

cesidio (approved):

PR #1960, opened 2020-09-02
Refactor vacancy.organisation to vacancy.parent_organisation

cpjmcquillan:
This PR refactors the frequent usage of `vacancy.organisation`.

If a vacancy has many organisations, then the parent organisation is the school group, otherwise it is the organisation

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -348,6 +348,10 @@ class Vacancy < ApplicationRecord
     self.documents.each { |document| DocumentDelete.new(document).delete }
   end
 
+  def parent_organisation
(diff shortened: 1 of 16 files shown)


app/models/vacancy.rb
```diff
@@ -348,6 +348,10 @@ def delete_documents
     self.documents.each { |document| DocumentDelete.new(document).delete }
   end
 
+  def parent_organisation
+    self.organisations.many? ? self.organisations.first.school_groups.first : self.organisation
```
cesidio, line 352: extra `.first` here `self.organisations.first.school_groups.first`

cesidio (approved):

PR #1966, opened 2020-09-03
Teva 797 fix terraform state

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-797

## Changes in this PR:

- Change Makefile to support separate state file per review app
- Change GitHub Actions workflows to support separate state file per review app
- Use `terraform workspace select || terraform workspace new` to ensure workspace is present


saliceti (approved):
Ok

PR #1983, opened 2020-09-07
prevent the return to results button submitting the form in any scenario

alexbowen:
no ticket, as described in title


david-mears-dfe (approved):
Seems fine, but what is the benefit? Performance / keeping Algolia usage to a minimum?

PR #1984, opened 2020-09-07
Fix failing spec

cpjmcquillan:
Merging of #1982 highlighted an issue with our current CI workflow. 

Tests were run locally and in CI on #1982 but it was not up to date with master, and this introduced a breaking change


david-mears-dfe (approved):
If I understand correctly, we need CI to check the resulting master branch, not the feature branch. Master had failing tests.

cesidio (approved):

PR #1986, opened 2020-09-07
Update vacancy-feedback-metrics-by-month.sql

stevenleggdfe:
- Add cumulative totals and rates within each academic year
- Remove now unused (exclusive) hire upper bound calculations
- Begin to add confidence intervals for the extrapolated calculations here given the limited sample size we're working from. (Just some key rates for now; will need further work to extend these to totals and savings.)
- Rename lots of things to improve clarity
- Split comments into multiple rows
- Add additional calculations for metrics over the last year (as opposed to just since the beginning of the academic year)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1064

### DIFF
--- a/bigquery/vacancy-feedback-metrics-by-month.sql
+++ b/bigquery/vacancy-feedback-metrics-by-month.sql
@@ -1,99 +1,143 @@
 SELECT
   *,
-  #calculate the confidence intervals for these key rates, assuming a 95% confidence level (i.e. z=1.96), that they can be modelled as binomial distributions and that the normal distribution approximation can be applied.
-  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusivity_rate*(1-exclusivity_rate),
-      feedback_available_this_AY_so_far)) AS exclusivity_rate_confidence_interval,
+  #calculate the confidence intervals for these key rates,
+  #assuming a 95% confidence level (i.e. z=1.96),
+  #that they can be modelled as binomial distributions
+  #and that the normal distribution approximation can be applied.
+  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusivity_rate_this_month*(1-exclusivity_rate_this_month),
+      sample_size_this_AY_so_far)) AS exclusivity_rate_confidence_interval,
   1.96*SAFE.SQRT(SAFE_DIVIDE(exclusivity_rate_this_AY_so_far*(1-exclusivity_rate_this_AY_so_far),
-      feedback_available_this_AY_so_far)) AS exclusivity_rate_confidence_interval_this_AY_so_far,
-  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusive_hires_rate*(1-exclusive_hires_rate),
-      feedback_available)) AS exclusive_hires_rate_confidence_interval,
+      sample_size_this_AY_so_far)) AS exclusivity_rate_confidence_interval_this_AY_so_far,
+  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusivity_rate_in_the_last_year*(1-exclusivity_rate_in_the_last_year),
+      sample_size_in_the_last_year)) AS exclusivity_rate_confidence_interval_in_the_last_year,
+  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusive_hires_rate_this_month*(1-exclusive_hires_rate_this_month),
+      sample_size_this_month)) AS exclusive_hires_rate_confidence_interval,
   1.96*SAFE.SQRT(SAFE_DIVIDE(exclusive_hires_rate_this_AY_so_far*(1-exclusive_hires_rate_this_AY_so_far),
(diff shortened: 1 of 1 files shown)


bigquery/vacancy-feedback-metrics-by-month.sql
```diff
@@ -1,83 +1,115 @@
...
-
-  #Extrapolations from hiring staff feedback to estimates of the overall proportion of vacancies which were hires, exclusive etc, unnormalised (i.e. assuming all categories of vacancy behave the same)
-  SAFE_DIVIDE(SUM(hires_through_tv),SUM(feedback_available)) AS raw_hires_rate_through_tv,
-  SAFE_DIVIDE(SUM(hires_through_tv_upperbound),SUM(feedback_available)) AS raw_hires_rate_through_tv_upperbound,
-  SAFE_DIVIDE(SUM(exclusive_hires),SUM(feedback_available)) AS raw_exclusive_hires_rate,
-  SAFE_DIVIDE(SUM(exclusive_hires_upperbound),SUM(feedback_available)) AS raw_exclusive_hires_rate_upperbound,
-  SAFE_DIVIDE(SUM(exclusive_listings),SUM(feedback_available)) AS raw_exclusivity_rate,
-
+  *,
+  #calculate the confidence intervals for these key rates, assuming a 95% confidence level (i.e. z=1.96), that they can be modelled as binomial distributions and that the normal distribution approximation can be applied.
```
david-mears-dfe, line 3: It would be easier to review this if the comment was wrapped onto multiple lines. In GitHub I can't read the comment on one line:
Screenshot 2020-09-09 at 14 42 23

david-mears-dfe: Actually, there are a lot of too-long-to-read lines in this file, so I will just check out the branch and read it in my editor, as I don't know how easy it is to limit the length of SQL lines.

bigquery/vacancy-feedback-metrics-by-month.sql
```diff
@@ -1,83 +1,115 @@
...
+  1.96*SAFE.SQRT(SAFE_DIVIDE(exclusivity_rate*(1-exclusivity_rate),
+      feedback_available_this_AY_so_far)) AS exclusivity_rate_confidence_interval,
```
david-mears-dfe, line 5: Is this doing something like making a confidence interval which scales in proportion to (the rate we are interested in), so the shape of the bell curve is squished or stretched out as the rate goes down or up (respectively)? In which case, does feedback_available_this_AY_so_far mean 'the number of feedbacks given so far this year'? If so, then it might be nice to call SUM(feedback_available) `total_feedback_available` so it's a bit clearer what the variable is.

bigquery/vacancy-feedback-metrics-by-month.sql
```diff
@@ -1,83 +1,115 @@
...
-        AND listed_elsewhere IN ("not_listed",
-          "listed_free",
-          "listed_dont_know")) AS exclusive_hires_upperbound,
-      COUNTIF(listed_elsewhere IN ("not_listed","listed_free")) AS exclusive_listings
-    FROM
-      `teacher-vacancy-service.production_dataset.vacancies_published`
+      month,
+      AY_beginning,
+      SUM(vacancies_published) AS vacancies_published,
+      SUM(feedback_available) AS feedback_available,
```
david-mears-dfe, line 44: In SUM(feedback_available) does feedback_available refer directly to a column or something? I would be tempted to say that it's confusing to reassign feedback_available to something other than the column (or whatever) it originally referred to, because that namespace should be reserved for the column called feedback_available.

david-mears-dfe: (Same for vacancies_published)

bigquery/vacancy-feedback-metrics-by-month.sql
stevenleggdfe: @david-mears-dfe I'll split the comments out, but the lines aren't as easy to split - because BigQuery handles the spacing and splitting itself as part of its autoformat function.

bigquery/vacancy-feedback-metrics-by-month.sql
stevenleggdfe: @david-mears-dfe it's taking a table of 'month-categories' with a number of vacancies/feedbacks for each one and collapsing it into a table of months, again with a number of vacancies/feedbacks for each one. I think I'll add in the _this_month prefix for clarity.

bigquery/vacancy-feedback-metrics-by-month.sql
stevenleggdfe: @david-mears-dfe - Yes it is, so this is the 'n' you see in the z sqrt(p(1-p)/n) standard formula, where n is the sample size. But I also want to retain the '_this_AY_so_far' etc. because I'm doing totals over that period, over the previous 12 months and over the month (because the data arrives in days, not months). So - I think I'll go with something like 'sample_size_this_AY_so_far'. Does that sound clearer?

david-mears-dfe (approved):

PR #1991, opened 2020-09-08
[TEVA-1277] Protect production database

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1277

## Changes in this PR:

Add a step in the databse sync job to upload the production backup to S3. This is a safety net in case something terrible happens in paas.

## Review
You can see a successful workflow here: https://github.com/DFE-Digital/teacher-vacancy-service/runs/1085588648?check_suite_focus=true


cpjmcquillan (approved):

nickromneydfe (approved):
Looks good to me. 

Discussed with Colin:
- even if Terraform tried to remove the S3 bucket with the backup, this would fail because of a non-empty bucket
- edge case of sync step running for a second time (with a file name of `yyyy-mm-dd.backup.sql.gz` would still succeed because `s3 cp` overwrites

PR #1994, opened 2020-09-09
Remove unused code

cesidio:


app/form_models/school_form.rb
```diff
@@ -1,10 +0,0 @@
-class SchoolForm < VacancyForm
```
david-mears-dfe, line 1: We still need this form. https://github.com/DFE-Digital/teacher-vacancy-service/blob/master/app/controllers/hiring_staff/vacancies/schools_controller.rb

app/models/concerns/vacancy_job_location_validations.rb
```diff
@@ -1,7 +0,0 @@
-module VacancyJobLocationValidations
-  extend ActiveSupport::Concern
-
-  included do
-    validates :job_location, presence: true
```
david-mears-dfe, line 5: Why get rid of the presence validation?

lib/tasks/export_tables_as_csv_to_big_query.rake
```diff
@@ -1,10 +0,0 @@
-namespace :tables_as_csv do
-  desc 'Exports postgres tables to Big Query'
-  namespace :to_big_query do
-    task export: :environment do
-      # It doesn't use CSV anymore, but I'm not changing the task name for now to avoid having to change more of the
-      # infrasctructure than strictly necessary
-      ExportTablesToBigQueryJob.perform_later
```
david-mears-dfe, line 7: I thought this was how Steven got his data?

lib/tasks/update_database_records_in_big_query.rake
```diff
@@ -1,8 +0,0 @@
-namespace :database_records do
-  desc 'Updates Big Query user and approver tables with latest records from DSI'
-  namespace :in_big_query do
-    task update: :environment do
-      ExportDsiUsersToBigQueryJob.perform_later
```
david-mears-dfe, line 5: Again, I thought this was how Steven got his data?

david-mears-dfe (changes requested):
Some decisions I disagree with, some I am confused by.

app/form_models/school_form.rb
cpjmcquillan: No we don't?

lib/tasks/export_tables_as_csv_to_big_query.rake
cpjmcquillan: We don't use the rake tasks

app/models/concerns/vacancy_job_location_validations.rb
cesidio: I made a search for module VacancyJobLocationValidations, it is not included anywhere. In order to find the code that is not used, I have a look here:
https://sonarcloud.io/component_measures?id=DFE-Digital_teacher-vacancy-service&metric=coverage&view=list

cpjmcquillan: We already do `validates :job_location, presence: true` on the form model

lib/tasks/export_tables_as_csv_to_big_query.rake
cesidio: When we were on AWS, the way to do cron jobs was to create rake tasks and set up some kind of cronjobs that invoked them. When we moved to GOV.UK PaaS, we now have Sidekiq Cron, a gem that invokes directly the jobs, no need to create rake tasks. You could argue that when you have to run them manually, you can invoke them, but calling the job from within the console is not the same amount of typing.

lib/tasks/update_database_records_in_big_query.rake
cesidio: same as before

cpjmcquillan (approved):

app/form_models/school_form.rb
david-mears-dfe: The link is to a file which uses it. This PR doesn't delete that file. So we are still using it. What am I missing?

cesidio: it's `SchoolsForm` not `SchoolForm`

david-mears-dfe (approved):

PR #2008, opened 2020-09-14
Ticket 1266 use build push action v2 buildkit with cache

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1266

## Changes in this PR:

- Update GitHub action `docker/build-push-action` to v2 which has buildkit enabled
- Uses GitHub cache action for Docker layer caching


nickromneydfe:
Evaluated three different approaches with an iterative model. Closing without merging.

PR #2009, opened 2020-09-14
Multi-school maps

david-mears-dfe:
## Jira

https://dfedigital.atlassian.net/browse/TEVA-1021

## Feedback request

I’m not very happy with where the JS lives.

Using inline JS means we can’t use ‘defer’ to make sure that the script defining initMap is run before the GMaps script which uses it as a callback. If the scripts are run in the wrong order (which would happen if the second loads first), no map is shown.

However, after hours of trying, not to say crying, I couldn’t find any way to ensure that the two scripts run in the correct order (other than using the power of prayer, which works in general, because the second script is just longer than the first). If the script could be made not inline then we could use ‘defer’ to do this, but I was unable to find a way to make it not inline (to load it with javascript_include_tag). Open to ideas!

The same problem also applies to the other js partial, ‘vacancies/_school.js.haml’.

## Check it out

Screenshot 2020-09-14 at 18 39 24
Screenshot 2020-09-14 at 18 39 40


app/models/vacancy.rb
```diff
@@ -265,6 +265,10 @@ def organisation
     organisation_vacancies.first&.organisation
   end
 
+  def schools
```
cpjmcquillan, line 268: There is already a scope for this

app/views/vacancies/_school.js.haml
```diff
@@ -3,7 +3,7 @@
     var myLatLng = {lat: #{lat}, lng: #{lng}};
 
     var map = new google.maps.Map(document.getElementById('map_zoom'), {
-      zoom: 17,
+      zoom: 16,
```
david-mears-dfe, line 6: This level of zoom gives a better view of where the school is in context: not just the few streets around it.

app/views/vacancies/_schools.js.haml
```diff
@@ -0,0 +1,51 @@
+:javascript
+  const getSchools = () => {
+    if (document.getElementById('map_zoom').dataset.schools) {
+      return JSON.parse(document.getElementById('map_zoom').dataset.schools);
+    }
+      return false;
```
cpjmcquillan, line 6: Indentation is wrong here

app/models/vacancy.rb
cpjmcquillan, line 268: You don't need to do this, we should just be interested in `vacancy.organisations`

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -103,10 +110,45 @@
   end
 
   it 'renders all the school links to the school website' do
-    [school_2, school_3].each { |school| expect(rendered_component).to include(school.url) }
+    [school_1, school_2, school_3].each { |school| expect(rendered_component).to include(school.url) }
```
cpjmcquillan, line 113: This should fail

app/components/jobseekers/schools_overview_component.rb
```diff
@@ -9,4 +9,17 @@ def initialize(vacancy:)
   def render?
     @vacancy.at_multiple_schools?
   end
+
+  def schools_map_data
+    schools = []
+    @vacancy.schools.select(&:geolocation).each do |school|
```
cpjmcquillan, line 15: organisations not schools

app/components/jobseekers/schools_overview_component.rb
```diff
@@ -9,4 +9,17 @@ def initialize(vacancy:)
...
+      schools.push({ name: school.name,
+                     name_link: link_to(school.name, school.url),
```
cpjmcquillan, line 17: Website or url

spec/components/jobseekers/schools_overview_component_spec.rb
david-mears-dfe: Why? All the schools have a url. In the spec they get it from the from the schools factory, and in our app they get it from GIAS if there isn't one set by the user.

app/components/jobseekers/school_group_overview_component.html.haml
```diff
@@ -36,7 +36,7 @@
   -# TODO: Unable to show the map as we cannot extract geolocation data from GIAS
```
cpjmcquillan, line 36: Can we turn this TODO into done now?

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -99,14 +106,63 @@
   context 'when GIAS-obtained website has been overwritten' do
     it 'renders a link to the school website' do
       expect(rendered_component).to include(school_1.website)
+      expect(rendered_component).not_to include(school_1.url)
```
cpjmcquillan, line 109: Can we put this in it's own `it` block?

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -99,14 +106,63 @@
...
     expect(rendered_component).to include(vacancy.school_visits)
   end
+
+  it 'renders the location heading for multiple schools' do
+    expect(rendered_component).to include('School locations')
+  end
+
+  context 'when at least one school has a geolocation' do
+    it 'shows the map' do
+      expect(rendered_component).to include('School locations')
```
cpjmcquillan, line 127: Let's separate these

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -99,14 +106,63 @@
...
+      expect(rendered_component).to include('map')
+    end
+  end
+
+  context 'when no school has a geolocation' do
+    let(:geolocation_trait) { :no_geolocation }
+
+    it 'does not show the map' do
+      expect(rendered_component).not_to include('School locations')
```
cpjmcquillan, line 136: And these

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -99,14 +106,63 @@
...
+      expect(rendered_component).not_to include('map')
+    end
+  end
+
+  describe '#schools_map_data' do
+    let(:school_1_data) do
+      JSON.parse(described_class.new(vacancy: vacancy_presenter).schools_map_data).each do |school|
```
cpjmcquillan, line 143: This seems overly complicated, let's just put it in the format it should be in

spec/components/jobseekers/schools_overview_component_spec.rb
```diff
@@ -99,14 +106,63 @@
...
+
+  context 'when at least one school has a geolocation' do
+    it 'shows the map' do
+      expect(rendered_component).to include('School locations')
+      expect(rendered_component).to include('map')
+    end
+  end
+
+  context 'when no school has a geolocation' do
+    let(:geolocation_trait) { :no_geolocation }
```
cpjmcquillan, line 133: I'd prefer us to stub `any?(&:geolocation)` and remove the need for `geolocation_trait`

app/components/jobseekers/school_group_overview_component.html.haml
david-mears-dfe: School groups still don't have geolocations yet, but I can uncomment the map for if/when they do.

cpjmcquillan: We should make a ticket to give them geolocations then, because there isn't one at the minute

app/views/vacancies/_schools.js.haml
```diff
@@ -0,0 +1,57 @@
+:javascript
+  const getSchools = () => {
+    if (document.getElementById('map').dataset.schools) {
+      return JSON.parse(document.getElementById('map').dataset.schools);
+    }
+  };
+
+  // Stops working if I change to const initMap = () => {}
```
alexbowen, line 8: reason why is this file will not be transpiled in the usual way as it is inline js

app/views/vacancies/_schools.js.haml
```diff
@@ -0,0 +1,57 @@
...
+      zoom: 16,
+      center: firstLatLng
+    })
+
+    // Use one infoWindow for all markers.
+    // This means that only one infoWindow can be open at a time.
+    const infoWindow = new google.maps.InfoWindow;
+    
+    schools.forEach((school) => {
+      const contentString = `<p>${school.name_link}</p>` +
```
alexbowen, line 30: you dont need to string concat here - a `` js template can be mutliline

app/views/vacancies/_schools.js.haml
```diff
@@ -0,0 +1,57 @@
...
+                            `<p>${school.address}</p>` +
+                            `<p>${school.school_type}</p>`
+
+      const latLng = { lat: school.lat, lng: school.lng };
```
alexbowen, line 34: these shouldnt really be const's as they are in iterative array, bit not a huge issue at all

app/components/jobseekers/school_group_overview_component.html.haml
david-mears-dfe: I'll do it.

david-mears-dfe: https://dfedigital.atlassian.net/browse/TEVA-1297

app/views/vacancies/_schools.js.haml
david-mears-dfe: How would you prefer it?

app/views/vacancies/_schools.js.haml
david-mears-dfe: So `let`?

app/views/vacancies/_schools.js.haml
alexbowen: `<p>${school.name_link}</p>
<p>${school.address}</p>
<p>${school.school_type}</p>`

app/views/vacancies/_schools.js.haml
alexbowen: actually forget it consts are fine

spec/components/jobseekers/schools_overview_component_spec.rb
david-mears-dfe: Not sure how to simplify - see latest changes and let me know if you have a preferred way to do it.

alexbowen (approved):
need a rebase to squah/fixup the commits into some more suuccint else looks good

cpjmcquillan (approved):

PR #2014, opened 2020-09-15
Bigquery fix vacancy tagging complexity

stevenleggdfe:
## Changes in this PR:

- Ensure that the scraped_vacancy and scraped_vacancies_in_scope views only contain DISTINCT vacancies.
- Remove a subquery from a query that was failing due to surpassing BigQuery's complexity limit.
- Didn't fix the root cause - the crawler appears to be recording the same vacancy multiple times. Will work on this next.

### DIFF
--- a/bigquery/calculate-daily-vacancy-tag-time-metrics.sql
+++ b/bigquery/calculate-daily-vacancy-tag-time-metrics.sql
@@ -38,19 +38,6 @@ WITH
     vacancy.id,
     job_roles,
     category),
-  scraped_vacancies_in_scope AS (
-  SELECT
-    scraped_vacancy.scraped_url AS url,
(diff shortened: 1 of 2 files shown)


bigquery/calculate-daily-vacancy-tag-time-metrics.sql
```diff
@@ -84,7 +71,7 @@ WITH
   FROM
     dates
   CROSS JOIN
-    scraped_vacancies_in_scope
+     `teacher-vacancy-service.production_dataset.scraped_vacancies_in_scope` AS scraped_vacancies_in_scope
```
david-mears-dfe, line 74: Is this the name of a table/column? Will this table/column get out of date quickly and will we forget that it is out of date? Worth putting in a comment?

david-mears-dfe (approved):

PR #2016, opened 2020-09-15
Set cookies preferences

cpjmcquillan:
This PR allows users to set and update their cookie preferences

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1232

## Changes in this PR
- User can set cookies preference
- Cookies banner is only rendered if cookies preference has not been set
- Only run GTM script if user has consented to cookies

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -4,7 +4,7 @@
 %html.govuk-template.app-html-class{ lang: "en" }
   /<![endif]
   %head
-    = render partial: 'shared/google_tag_manager_head'
+    = render partial: 'shared/google_tag_manager_head' if !CookiesBannerFeature.enabled? || cookies['consented-to-cookies'] == 'yes'
     = render partial: 'shared/open_graph'
     = render partial: 'shared/meta'
     %meta{ charset: "utf-8" }/
@@ -29,7 +29,7 @@
     %meta{ content: asset_pack_path('media/images/govuk-opengraph-image.png'), property: "og:image" }/
     = csrf_meta_tags
   %body{ class: body_class, data: { vacancy_state: @vacancy.present? ? @vacancy.state : 'create' } }
-    = render partial: 'shared/google_tag_manager_body'
+    = render partial: 'shared/google_tag_manager_body' if !CookiesBannerFeature.enabled? || cookies['consented-to-cookies'] == 'yes'
     :javascript
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     = render 'shared/skip_links'
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -33,7 +33,7 @@
     :javascript
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     = render 'shared/skip_links'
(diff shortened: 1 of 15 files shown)


config/locales/en.yml
```diff
@@ -17,15 +17,94 @@ en:
...
+          like your IP address or your postcode before we share it with Google.
+        details:
+          intro: 'Google Analytics stores information about:'
+          list:
+            - the pages you visit on Teaching Vacancies
+            - how long you spend on each page
+            - how you got to the site
+            - what you click on while you're visiting the site
+        explanation: >-
+          We use Google Analytics software to collection information about how you use Teaching Vacancies.
```
stevenleggdfe, line 52: @cpjmcquillan typo here - 'collection' should be collect. To check, is this the final version of the content from @valentine-carter ?

app/views/layouts/application.html.haml
```diff
@@ -29,11 +29,11 @@
     %meta{ content: asset_pack_path('media/images/govuk-opengraph-image.png'), property: "og:image" }/
     = csrf_meta_tags
   %body{ class: body_class, data: { vacancy_state: @vacancy.present? ? @vacancy.state : 'create' } }
-    = render partial: 'shared/google_tag_manager_body'
+    = render partial: 'shared/google_tag_manager_body' if !CookiesBannerFeature.enabled? || cookies['consented-to-cookies'] == 'yes'
     :javascript
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     = render 'shared/skip_links'
-    = render 'layouts/cookies_banner' if CookiesBannerFeature.enabled? && !current_page?(page_path('cookies'))
+    = render 'layouts/cookies_banner' unless !CookiesBannerFeature.enabled? || controller_name == 'cookies_preferences' || cookies_preference_set?
```
david-mears-dfe, line 36: I don't know how much time it takes for an app to check a little variable for true/false (probably almost none), but we might shave off a millisecond or two by putting the most-frequently-true thing to evaluate first (`cookies_preference_set?`).
🤷

config/locales/activerecord.yml
```diff
@@ -65,6 +65,9 @@ en:
   hiring_staff_user_preference_errors:
     managed_organisations:
       blank: Select which schools' job listings you would like to view
+  cookies_preferences_errors: &cookies_preferences_errors
```
david-mears-dfe, line 68: The ampersand denotes an alias, yes? I don't see a difference here...?

david-mears-dfe (approved):

app/views/layouts/application.html.haml
cpjmcquillan: The order is important. 

If the feature is disabled, **don't render**
If the feature is enabled, but we're at the cookies_preferences create/new action, **don't render**
If the feature is enabled, and we're not at the cookies_preferences create/new action, but cookies preferences are set, **don't render**
If the feature is enabled, and we're not at the cookies_preferences create/new action, and cookies preferences are not set, **render**

david-mears-dfe (changes requested):
Just thought, I'm not sure if you actually initialize CookiesBannerFeature?

cpjmcquillan:
> Just thought, I'm not sure if you actually initialize CookiesBannerFeature?

Been and done and merged to master

david-mears-dfe (approved):

PR #2020, opened 2020-09-16
[TEVA-1100] Refactor new sections notification

cpjmcquillan:
This PR refactors the new sections notification.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1100

## Changes in this PR
- Use notification component to render action required and new sections on review/edit pages
- Add new section message for contact number


app/frontend/styles/components/flashes.scss
```diff
@@ -40,6 +40,10 @@
 
 .govuk-notification--notice {
   border-color: govuk-colour('blue');
+
+  .govuk-notification__list a {
+    color: govuk-colour('blue');
```
david-mears-dfe, line 45: Isn't blue the default for <a>? I'd be tempted to just specify when you need it to not be the default color

app/views/hiring_staff/vacancies/_new_tag.html.haml
```diff
@@ -0,0 +1,4 @@
+- attributes.each do |attribute|
```
david-mears-dfe, line 1: The filename isn't obvious enough for my liking. Can we do e.g. _new_attributes_tag?

david-mears-dfe (approved):
LG to M

app/views/hiring_staff/vacancies/_new_tag.html.haml
cpjmcquillan: Certainly

app/frontend/styles/components/flashes.scss
cpjmcquillan: Indeed it is, copy pasta got out of hand 🍝

PR #2023, opened 2020-09-17
[TEVA-1266] Speed up build times with Docker caching

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1266

## Changes in this PR:

- Add cache metadata to generated Docker images, stored in Docker Hub
- Use `cache-from` in generating new images
- Use v2 of `docker/build-push-action` to default to BUILDKIT

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,6 @@ jobs:
       run: |
         TAG=${GITHUB_SHA}
         echo ::set-env name=TAG::${TAG}
-        echo ::set-env name=DOCKER_BUILDKIT::1
 
     - name: Set up Ruby 2.7
       uses: actions/setup-ruby@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -39,15 +39,37 @@ jobs:
...
         password: ${{ secrets.DOCKER_PASSWORD }}
-        repository: ${{ env.DOCKERHUB_REPOSITORY }}
-        tags: ${{ env.TAG }}
+
+    - name: Build and push docker dev-build image
+      uses: docker/build-push-action@v2
+      with:
+        build-args: BUILDKIT_INLINE_CACHE=1
+        cache-from: |
+          ${{ env.DOCKERHUB_REPOSITORY }}:build-cache-${{ env.BRANCH }}
```
saliceti, line 57: `env.BRANCH` is not set here. Since it will always be master, we can hard code `master` as it will be clearer.

.github/workflows/deploy.yml
```diff
@@ -39,15 +39,37 @@ jobs:
...
+        push: true
+        tags: ${{ env.DOCKERHUB_REPOSITORY }}:build-cache-${{ env.BRANCH }}
```
saliceti, line 59: Can we call it `dev-build` as well? It may be clearer.

.github/workflows/deploy.yml
nickromneydfe: Updated to `builder-master`

.github/workflows/deploy.yml
nickromneydfe: We are following the Docker pattern of name the build stage `builder`

.github/workflows/deploy.yml
```diff
@@ -39,15 +39,36 @@ jobs:
...
+        push: true
+        tags: ${{ env.DOCKERHUB_REPOSITORY }}:builder-master
+        target: builder
+
+    - name: Build and push docker production image
+      uses: docker/build-push-action@v2
+      with:
+        build-args: BUILDKIT_INLINE_CACHE=1
+        cache-from: |
+          ${{ env.DOCKERHUB_REPOSITORY }}:builder-master
```
saliceti, line 67: Shouldn't we add the production target cache as well?
```
      cache-from: |
          ${{ env.DOCKERHUB_REPOSITORY }}:builder-master
          ${{ env.DOCKERHUB_REPOSITORY }}:${{ env.TAG }}
          ${{ env.DOCKERHUB_REPOSITORY }}:master
```

nickromneydfe: Yes, absolutely. Changed the code to reflect that

Makefile
```diff
@@ -51,8 +59,20 @@ production:
 
 .PHONY: build-local-image
 build-local-image:
+		$(eval export DOCKER_BUILDKIT=1)
+		$(eval branch=$(shell git rev-parse --abbrev-ref HEAD))
 		$(eval tag=dev-$(shell git rev-parse HEAD)-$(shell date '+%Y%m%d%H%M%S'))
-		docker build -t $(repository):$(tag) .
+		docker build \
+			--build-arg BUILDKIT_INLINE_CACHE=1 \
+			--cache-from $(repository):builder-master \
```
saliceti, line 67: Can we add:
--cache-from $(repository):builder-$(branch) \
and
--cache-from $(repository):master \
?

nickromneydfe: agreed

saliceti (approved):

PR #2027, opened 2020-09-18
Fix documents controller bug when nil

cpjmcquillan:
This PR fixes an error that would occur when JS disabled hiring staff would choose to submit `Upload files` form without choosing any files to upload...

https://rollbar.com/dfe/teacher-vacancies/items/2177/


david-mears-dfe (approved):
Nice

PR #2037, opened 2020-09-22
Add no-store to Cache-Control header

cpjmcquillan:
This PR will hopefully kill https://rollbar.com/dfe/teacher-vacancies/items/2078/ once and for all

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -103,6 +103,6 @@ Rails.application.configure do
 
   config.action_dispatch.trusted_proxies = AWSIpRanges.cloudfront_ips.map { |proxy| IPAddr.new(proxy) }
 
-  # Browsers must check to see if resource has changed before using browser cache
-  config.action_dispatch.default_headers.merge!('Cache-Control' => 'no-cache')
+  # Ensure browsers don't cache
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -103,6 +103,6 @@
 
   config.action_dispatch.trusted_proxies = AWSIpRanges.cloudfront_ips.map { |proxy| IPAddr.new(proxy) }
 
-  # Browsers must check to see if resource has changed before using browser cache
-  config.action_dispatch.default_headers.merge!('Cache-Control' => 'no-cache')
+  # Ensure browsers don't cache
+  config.action_dispatch.default_headers.merge!('Cache-Control' => 'no-cache, no-store')
```
david-mears-dfe, line 107: Do you have a sense how much the page loading time will be affected by this?

Should we check the mean and median page loading times before and after this change?

cpjmcquillan: It doesn't affect the time it takes to load a page, it just ensures browsers make a request to the server every time. I can't imagine how you could begin to reasonably measure any impact to users

david-mears-dfe (approved):

PR #2045, opened 2020-09-23
[TEVA-1235] Update service start page banner

cpjmcquillan:
This PR implements some of the updates to the service start page.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1235

## Changes in this PR:
- Service start page updates

### Before
image

### After
image


app/frontend/styles/components/search_panel.scss
```diff
@@ -1,10 +1,33 @@
...
+  left: 50%;
+  margin-bottom: govuk-spacing(6);
+  margin-left: -50vw;
+  padding-bottom: govuk-spacing(3);
+  padding-top: govuk-spacing(1);
+  position: relative;
+  width: 100vw;
+}
+
+body {
```
alexbowen, line 12: why the `body` css selector? is `.pages_home` selector on it on not enough?

config/locales/en.yml
```diff
@@ -205,9 +205,9 @@ en:
         description_html: <p>We’re completing planned maintenance on job listings.</p><p>You can search and view vacancies, but you can’t manage them.</p><p>You’ll be able to use the full service again in less than an hour.</p><p><a href='mailto:teaching.vacancies@education.gov.uk'>Contact our support team</a> if you need help.</p>
       signin:
         title: List a teaching job
-        description: You must sign in to your Teaching Vacancies account to list a job at your school.
-        signin_link: how to get one
-        signin_guidance_html: Don't have an account? Find out %{link} if your school is eligible.
+        description: Sign in to your Teaching Vacancies account to list a job at your school or Trust.
```
alexbowen, line 208: Trust should be non caps

config/locales/en.yml
```diff
@@ -347,7 +347,10 @@ en:
       labels:
         action_required: action required
         new: new
-    heading: 'Find a job in teaching'
+    heading: Welcome to Teaching Vacancies
+    intro: >-
+      Teaching Vacancies is a free job-listing service from the Department for Education.
+      You can search for a job at a school or Trust in England and set up job alerts.
```
alexbowen, line 353: "

config/locales/en.yml
cpjmcquillan: @valentine-carter says NO!!

app/frontend/styles/components/search_panel.scss
cpjmcquillan: I just copypastad 🍝  the css block for a different controller class to here, but in fact in a subsequent PR I apply zero padding to `.govuk-main-wrapper` everywhere, so I will just do that in this PR instead

alexbowen (approved):

PR #2048, opened 2020-09-23
[TEVA-1240] Add breadcrumb to job listing page

cpjmcquillan:
This PR adds a breadcrumb to the job listing page

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1240

## Changes in this PR:
- Add breadcrumb to the job listing page

### DIFF
--- a/app/views/shared/vacancy/_jobseeker_view.html.haml
+++ b/app/views/shared/vacancy/_jobseeker_view.html.haml
@@ -7,7 +7,7 @@
   .govuk-width-container
     = render(Shared::BreadcrumbComponent.new(collapse_on_mobile: false,
                                              crumbs: [{ link_path: root_path, link_text: t('breadcrumbs.home') },
-                                                      { link_path: request.host == URI(request.referrer || '').host ? request.referrer : jobs_path, link_text: t('breadcrumbs.jobs') },
+                                                      { link_path: referred_from_host? ? request.referrer : jobs_path, link_text: t('breadcrumbs.jobs') },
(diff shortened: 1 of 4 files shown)


app/views/shared/vacancy/_jobseeker_view.html.haml
```diff
@@ -5,7 +5,10 @@
 
 .jobs-banner
   .govuk-width-container
-    = render 'vacancies/back_link'
+    = render(Shared::BreadcrumbComponent.new(collapse_on_mobile: false,
+                                             crumbs: [{ link_path: root_path, link_text: t('breadcrumbs.home') },
+                                                      { link_path: request.host == URI(request.referrer || '').host ? request.referrer : jobs_path, link_text: t('breadcrumbs.jobs') },
```
david-mears-dfe, line 10: I'm not sure what `request.host == URI(request.referrer || '').host` is checking. (Is it 'Was the request from this page'?) Can it be wrapped in a boolean method with an explanatory name that lives in the view component?

```
def explanatory_name?(request)
  true
end
```

app/views/shared/vacancy/_jobseeker_view.html.haml
```diff
@@ -5,7 +5,10 @@
 
 .jobs-banner
   .govuk-width-container
-    = render 'vacancies/back_link'
```
david-mears-dfe, line 8: This is the only place where the back_link partial is used. Please delete the partial and the partial's tests and translations 🗑️ 🔥

alexbowen (approved):

PR #2056, opened 2020-09-24
Fix wrapper padding regression and try to put class in a better location

cpjmcquillan:
I did a bad thing and this fixes that bad thing. Also _attempts_ to put the controller specific padding in a better place


alexbowen:
i take it this is for the page heading padding on all pages?

alexbowen (approved):

PR #2060, opened 2020-09-25
[TEVA-1346] Redirect to domain if request host is different

cpjmcquillan:
This PR redirects requests with different host names to DOMAIN.

See tv_engineering slack channel for context

**Note:** the rspec/capybara/selenium host/domain config saga was more than a pain the backside, hence not running the redirect before_action in tests

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  before_action :redirect_to_domain
+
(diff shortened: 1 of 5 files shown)


saliceti:
We suspect it fails as the paas internal healthcheck may not pass the right HOST header.
Additionnally, we need to make sure the header is set by cloudfront (in dev , staging, prod).

saliceti (approved):
OK

cesidio (approved):

PR #2064, opened 2020-09-28
for help

alexbowen:
i cant get this to work, although the method in the helper seems to do the desired thing when i run it from console `[["daily", "something.daily"], ["weekly", "something.weekly"]]`

### DIFF
--- a/app/helpers/subscription_helper.rb
+++ b/app/helpers/subscription_helper.rb
@@ -2,4 +2,10 @@ module SubscriptionHelper
   def hex?(reference)
     reference.match(/^\h+$/)
   end
+
+  def frequency_options
+    Subscription::FREQUENCY_OPTIONS.map do |key|
+      [key, I18n.t("subscriptions.frequency.#{key}")]
+    end
+  end
 end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,4 +1,6 @@
 class Subscription < ApplicationRecord
+  FREQUENCY_OPTIONS = %W[daily weekly]
+
   include Auditor::Model
 
   enum frequency: { daily: 0 }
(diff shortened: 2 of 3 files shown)


app/views/subscriptions/new.html.haml
```diff
@@ -28,7 +28,7 @@
         required: true
 
       = f.govuk_collection_radio_buttons :frequency,
-        [['daily', t('subscriptions.frequency.daily')], ['weekly', t('subscriptions.frequency.weekly')]],
+        frequency_options
```
cpjmcquillan, line 31: Subscription::FREQUENCY_OPTIONS

app/helpers/subscription_helper.rb
```diff
@@ -2,4 +2,10 @@ module SubscriptionHelper
   def hex?(reference)
     reference.match(/^\h+$/)
   end
+
+  def frequency_options
```
cpjmcquillan, line 6: You can lose this

app/models/subscription.rb
```diff
@@ -1,4 +1,6 @@
 class Subscription < ApplicationRecord
+  FREQUENCY_OPTIONS = %W[daily weekly]
```
cpjmcquillan, line 2:
```ruby
FREQUENCY_OPTIONS = { 
  daily: 0,
  weekly: 1
}.freeze
```

app/models/subscription.rb
```diff
@@ -1,4 +1,6 @@
...
+
   include Auditor::Model
 
   enum frequency: { daily: 0 }
```
cpjmcquillan, line 6:
```ruby
enum frequency: FREQUENCY_OPTIONS
```

app/helpers/subscription_helper.rb
```diff
@@ -2,4 +2,10 @@ module SubscriptionHelper
...
+    Subscription::FREQUENCY_OPTIONS.map do |key|
```
cpjmcquillan, line 7:
```ruby
Subscription::FREQUENCY_OPTIONS.map do |key, _|
  [key, I18n.t("subscriptions.frequency.#{key}")]
end
```

PR #2067, opened 2020-09-29
Remove unused gem

david-mears-dfe:
We haven't been using this gem.


cesidio:
This is a useful gem to prevent N+1 queries

david-mears-dfe:
> This is a useful gem to prevent N+1 queries

But we don't use it. Let's either use it or get rid of it.

PR #2072, opened 2020-09-29
Remove utilities which duplicate govuk front end classes

david-mears-dfe:
Replacing 'utility' styles for which there are perfectly decent govuk frontend classes available.

Pro: it's overhead and hard to read what these mean e.g. `mt0`

Con: the govuk spacing classes include ! meaning we have to use the lengthy { class: 'govuk-!-margin-bottom-6' } formation in haml


alexbowen (approved):
nice tidy up

PR #2073, opened 2020-09-30
Bigquery mat takeup metrics

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1300

## Changes in this PR:

- Add number_of_mat_access_workaround_users and using_mat_access_workaround fields to table of MATs to allow us to target comms at these MATs
- Add new table of metrics over time to measure MAT level access takeup. This includes:
* #/% MATs with at least one user signed up with MAT level access, broken down by 'MAT size'
* #/% MATs with at least one user still signed up for access to multiple schools within the MAT (the 'workaround'), broken down by 'MAT size'
* #/% MATs who have published at least vacancy within the last year/quarter/ever/currently live via MAT level access, broken down by 'MAT size'
* #/% MATs who have published at least one multi-school vacancy within the last year/quarter/ever/currently live, broken down by 'MAT size'
* #/% MATs who have published at least one trust-level vacancy within the last year/quarter/ever/currently live, broken down by 'MAT size'


bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
```diff
@@ -0,0 +1,346 @@
+  #scheduled query that appends new rows to CALCULATED_daily_MAT_time_metrics table
+WITH
+  dates AS ( #the range of dates we're calculating for
+  SELECT
```
david-mears-dfe, line 4: How strict are .sql files about indentation? My eyes are looking for where this bracket closes. Is it line 270?

david-mears-dfe: Maybe it's 14. If possible please indent the contents of the brackets by 1, for readability.

bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
```diff
@@ -0,0 +1,346 @@
...
+  FROM
+    UNNEST(GENERATE_DATE_ARRAY('2020-09-01', DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY), INTERVAL 1 DAY)) AS date
+  WHERE
+    #only calculate these metrics for dates we don't already have within this range
+    date NOT IN (
+    SELECT
+      DISTINCT date
+    FROM
+      `teacher-vacancy-service.production_dataset.CALCULATED_daily_MAT_time_metrics`) ),
+  trust_metrics AS (
```
david-mears-dfe, line 15: Same here

bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
```diff
@@ -0,0 +1,362 @@
...
+      COUNTIF(had_live_trust_level_vacancies)) AS trusts_which_had_trust_level_vacancies_live
+  FROM (
+    SELECT
+      trusts.uid AS uid,
+      dates.date AS date,
+      trusts.size_bracket AS size_bracket,
+      trusts.using_mat_access_workaround AS using_mat_access_workaround,
+    IF
+      ((trusts.status != "Closed"
+          OR trusts.date_closed > dates.date) #if the trust is not currently closed or if it closed after the date we're calculating for
```
david-mears-dfe, line 87: these comments are useful

bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
```diff
@@ -0,0 +1,362 @@
...
+        AND trusts.date_created <= dates.date,
+        #and if the trust was first listed on GIAS before or on the date we're calculating for
+        TRUE,
+        FALSE) AS in_scope,
+    IF
+      ((trusts.status = "Closed"
+          AND trusts.date_closed <= dates.date)
+        OR trusts.date_opened > dates.date,
+        FALSE,
+        #mark trusts which had closed or had not yet opened as not signed up
```
david-mears-dfe, line 100: If you mark trusts which are closed as not signed up, does that give a misleading impression of the proportion of trusts which we still want to sign up?

david-mears-dfe: (same q applies for trust which are closed marked as not having published)

stevenleggdfe: @david-mears-dfe in the case, we're calculating the trusts which *are* signed up as a proportion of the trusts in scope, both of which exclude closed trusts, so the figures we're producing aren't actually misleading I think. But good question!

stevenleggdfe: Let me see if i can change the comment to make this clearer

david-mears-dfe: is there a way to only need to exclude the closed trusts once rather than repeatedly?

stevenleggdfe: Not simply, unless you've got an idea? The complexity is that it's not as simple as just excluding closed trusts, because which trusts are closed changes for each date we're calculating for - I think we'd need to add a subquery in a WITH statement before this one that produced a table of dates and trusts that were open on those dates. But while that would be nicer to read, it would end up being executed multiple times and so actually be less efficient SQL I think. BigQuery imposes limits on the number of subquery executions you can do in any one query to encourage you to be efficient, hence why I'm trying to keep this kind of thing down.

bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
stevenleggdfe: Essentially, BigQuery has a SQL autoformatter that (if you use it) controls the indentation. Looking at it it seems to prefer intending based on the logic within the keywords, rather than on brackets, which I suspect is a bit of a SQL thing as it's not a scripting language. Although different people's SQL differs, and personally I'd favour as much indentation as possible to make things clearer.

So essentially... either we have more indentation but lose the auto format function, or we do without the extra indentation but keep the auto formatter (which is a big time saver).

bigquery/scheduled_queries/calculate-daily-mat-time-metrics.sql
stevenleggdfe: @david-mears-dfe I found a way :-) Please could you re-review?

david-mears-dfe (approved):

PR #2075, opened 2020-09-30
[TEVA-1295] update alert email copy

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1295

'You can manage your alert here.' text is not added as part of this task but will be added when dashboard is implemented

### DIFF
--- a/app/views/alert_mailer/daily_alert.text.haml
+++ b/app/views/alert_mailer/daily_alert.text.haml
@@ -1,5 +1,4 @@
-\# #{t('app.title')}
-\## #{t('job_alerts.alert.email.daily.summary', count: @vacancies.count, reference: @subscription.reference)}
+\# #{t("job_alerts.alert.email.#{@subscription.frequency}.summary", count: @vacancies.count)}
 \
 \---
 \
@@ -15,7 +14,7 @@
   \---
   \
 \
-Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - the free service for schools in England to list teaching roles and for jobseekers to find them.
+Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - Teaching Vacancies is a free job-listing service from the Department for Education. It helps schools to save money on recruitment and you to take the next step in your career.
 
 \---
 
--- a/app/views/alert_mailer/daily_alert.text.haml
+++ b/app/views/alert_mailer/daily_alert.text.haml
@@ -14,7 +14,13 @@
   \---
   \
 \
-Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - Teaching Vacancies is a free job-listing service from the Department for Education. It helps schools to save money on recruitment and you to take the next step in your career.
+\# #{t('subscriptions.title')}
+= t('subscriptions.intro')
+- @subscription.filtered_search_criteria.each_pair do |filter, value|
+  = "#{filter.humanize}: #{value}"
+= t('subscriptions.frequency.alert_text', frequency: @subscription.frequency)
+= t('subscriptions.link.home_page', home_page_link: notify_link('https://teaching-vacancies.service.gov.uk', t('app.title')))
+= - t('subscriptions.footer')
 
 \---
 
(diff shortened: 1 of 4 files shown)


app/models/subscription.rb
```diff
@@ -6,6 +6,8 @@ class Subscription < ApplicationRecord
 
   include Auditor::Model
 
+  attr_reader :subject
```
david-mears-dfe, line 9: I don't understand how this works - Subscription objects don't have a `subject` column, and I can't see a `subject` method here either.

app/views/alert_mailer/daily_alert.text.haml
```diff
@@ -15,7 +14,13 @@
   \---
   \
 \
-Visit #{notify_link('https://teaching-vacancies.service.gov.uk', t('app.title'))} - the free service for schools in England to list teaching roles and for jobseekers to find them.
+\# #This job alert
```
david-mears-dfe, line 17: I guess this text 'This job alert' (and the below 'Visit') should be in the en.yml.

app/models/subscription.rb
alexbowen: it is needed so that the test can access subject

david-mears-dfe: What is the test accessing? A nil value?

spec/mailers/alert_mailer_spec.rb
```diff
@@ -66,14 +65,16 @@
...
-          ),
+          I18n.t('job_alerts.alert.email.subject'),
         )
         expect(mail.to).to eq([subscription.email])
 
         expect(body).to match(/\[#{first_vacancy_presenter.job_title}\]/)
+        expect(body).to match(
+          /2 new jobs matching your search criteria were listed in the last day./,
+        )
+        expect(body).to match(/Subject: #{subscription.subject}/)
```
david-mears-dfe, line 76: I think `subscription.subject` returns `nil` - correct?

app/models/subscription.rb
david-mears-dfe: Maybe you meant `reference` instead of `subject`?

spec/mailers/alert_mailer_spec.rb
alexbowen: ah yes i think you are right, so the test is a false positive

david-mears-dfe (approved):

spec/mailers/alert_mailer_spec.rb
```diff
@@ -8,7 +8,7 @@
   let(:body) { mail.body.raw_source }
   let(:subscription) do
     subscription = create(:daily_subscription, email: 'an@email.com',
-                                               reference: 'a-reference',
+                                               frequency: 'daily',
                                                search_criteria: {
                                                  subject: 'English',
```
david-mears-dfe, line 13: In Cesi's PR, he rightly corrects 'subject' to 'keyword' here as we don't filter by subject (we used to, a year ago).

spec/mailers/alert_mailer_spec.rb
```diff
@@ -37,18 +37,17 @@
...
         expect(mail.to).to eq([subscription.email])
 
-        expect(body).to match(/# #{I18n.t('app.title')}/)
+        expect(body).to match(/# #{I18n.t('job_alerts.alert.email.daily.summary', count: 1)}/)
         expect(body).to match(
-          /A new job matching your search criteria &#39;#{subscription.reference}&#39; was posted yesterday/,
+          /A new job matching your search criteria was listed in the last day./,
         )
         expect(body).to match(/---/)
+        expect(body).to match(/Subject: English/)
```
david-mears-dfe, line 49: As above - Keyword not Subject

spec/mailers/alert_mailer_spec.rb
```diff
@@ -66,14 +65,16 @@
...
-          ),
+          I18n.t('job_alerts.alert.email.subject'),
         )
         expect(mail.to).to eq([subscription.email])
 
         expect(body).to match(/\[#{first_vacancy_presenter.job_title}\]/)
+        expect(body).to match(
+          /2 new jobs matching your search criteria were listed in the last day./,
+        )
+        expect(body).to match(/Subject: English/)
```
david-mears-dfe, line 76: As above - Keyword not Subject

david-mears-dfe (changes requested):
Sorry, just spotted something - see comments

alexbowen:
@david-mears-dfe yes i saw this change in cesi's. i need to close this PR and create another as the environment deployment is messed up for this one

PR #2080, opened 2020-10-02
[TEVA-1295] alert email copy

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1295

'You can manage your alert here.' text is not added as part of this task but will be added when dashboard is implemented

### DIFF
--- a/spec/mailers/alert_mailer_spec.rb
+++ b/spec/mailers/alert_mailer_spec.rb
@@ -35,15 +35,15 @@ RSpec.describe AlertMailer, type: :mailer do
       it 'shows a vacancy' do
         expect(mail.subject).to eq(
           I18n.t(
-            'job_alerts.alert.email.daily.subject.one',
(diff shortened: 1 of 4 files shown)


spec/mailers/alert_mailer_spec.rb
```diff
@@ -97,16 +98,16 @@
...
-            reference: subscription.reference,
-          ),
+          I18n.t('job_alerts.alert.email.subject'),
         )
         expect(mail.to).to eq([subscription.email])
 
-        expect(body).to match(/# #{I18n.t('app.title')}/)
-        expect(body).to match(/A new job matching your search criteria &#39;#{subscription.reference}&#39; was posted in the last week/)
+        expect(body).to match(
+          /A new job matching your search criteria was listed in the last week./,
```
cesidio, line 106: you can use `I18n` keys here too

PR #2086, opened 2020-10-02
ticket 1329 move smoke test to deploy workflow

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1329

## Changes in this PR:

- Move the smoke test of staging from its own workflow to steps in the deploy pipeline, between deploying to staging and production - verifies that releasing this version of the code doesn't break staging, before progressing to production


.github/workflows/deploy.yml
```diff
@@ -62,6 +62,32 @@ jobs:
         terraform workspace select staging terraform/app || terraform workspace new staging terraform/app
         terraform apply -var-file terraform/workspace-variables/staging.tfvars -auto-approve -input=false terraform/app
 
+    - name: Set up Ruby
```
saliceti, line 65: This repeats what's in the smoke test workflow. Could we try to reuse it instead?
Maybe by using a combination like:
- [Workflow dispatch](https://github.com/marketplace/actions/workflow-dispatch) to trigger the smoke test workflow from the deploy workflow
- [Wait on check](https://github.com/marketplace/actions/wait-on-check) to wait in deploy workflow for smoke test workflow to end

nickromneydfe:
Will re-open when using dependent workflows

PR #2093, opened 2020-10-06
Remove unused API key from docker-compose.env.sample

david-mears-dfe:
We don't use this.


cesidio:
There is more cleanup that we can do on that file, maybe we can remove it and document environment variables somewhere else?

david-mears-dfe:
> There is more cleanup that we can do on that file, maybe we can remove it and document environment variables somewhere else?

Yeah. Opened a convo on Slack but no responses yet https://ukgovernmentdfe.slack.com/archives/G0127R0KXQF/p1602009305001600

PR #2094, opened 2020-10-06
For school groups, save postcode to an attribute on organisation

david-mears-dfe:
As opposed to nesting it inside the gias_data json attribute

This makes the method `organisation.postcode` work for both Schools and SchoolGroups, which is needed for encapsulation.


cesidio:
I am happy to approve it, but I think that while you work at it, it would be good to set all the required fields that you can see in `SchoolGroup` class specified in the TODO

david-mears-dfe:
https://github.com/DFE-Digital/teacher-vacancy-service/pull/2095

PR #2095, opened 2020-10-06
For school groups, save gias_data to an attribute on organisation

david-mears-dfe:
As opposed to sourcing it from inside the gias_data json attribute

For example, this makes the method organisation.postcode work for both Schools and SchoolGroups, which is needed for encapsulation.

### DIFF
--- a/lib/import_school_group_data.rb
+++ b/lib/import_school_group_data.rb
@@ -58,15 +58,18 @@ private
   def convert_to_school_group(row)
     school_group = SchoolGroup.find_or_initialize_by(uid: row['Group UID'])
     set_gias_data_as_json(school_group, row)
-    set_geolocation(school_group, row)
+    school_group.postcode = row['Group Postcode']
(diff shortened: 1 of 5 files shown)


cesidio (approved):

lib/import_school_group_data.rb
```diff
@@ -58,15 +58,18 @@ def save_csv_file(url, location)
...
+    school_group.name = row['Group Name']&.titlecase
+    school_group.address = row['Group Locality']
+    school_group.town = row['Group Town']
+    school_group.county = row['Group County']
+    set_geolocation(school_group)
     school_group
   end
 
-  def set_geolocation(school_group, row)
-    if row['Group Postcode']&.present? && (school_group.geolocation.blank? || school_group.postcode != row['Group Postcode'])
```
cpjmcquillan, line 66: This was intended to prevent unnecessary API usage and us running out of operations

PR #2096, opened 2020-10-06
[TEVA-1331] IAM permissions for deploy user

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1331

## Changes in this PR:

- Amazon Certificate Manager IAM permissions for the `deploy` user to be able to list certificates from a datasource

## Next steps:

- [ ] Terraform deployment of common via Makefile

### DIFF
--- a/terraform/common/iam.tf
+++ b/terraform/common/iam.tf
@@ -89,6 +89,29 @@ resource aws_iam_user_policy_attachment cloudwatch {
   policy_arn = aws_iam_policy.cloudwatch.arn
 }
 
+# ACM
+
+data aws_iam_policy_document acm {
+  statement {
+    actions = [
+      "acm:DescribeCertificate",
+      "acm:ListCertificates",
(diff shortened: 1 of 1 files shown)


terraform/common/iam.tf
```diff
@@ -89,14 +89,40 @@ resource aws_iam_user_policy_attachment cloudwatch {
   policy_arn = aws_iam_policy.cloudwatch.arn
 }
 
+# ACM
+
+data aws_iam_policy_document acm {
+  statement {
+    actions   = ["acm:DescribeCertificate"]
```
saliceti, line 96: Can you put the 3 acm actions in the same list? This would be more consistent with the rest of the file.

nickromneydfe: Thanks. I've grouped as per S3 statements.

saliceti (approved):

PR #2098, opened 2020-10-07
[TEVA-1388] Add lifecycle policy to S3 db backups bucket

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1388

## Changes in this PR:

- Add lifecycle policy to S3 bucket
- Set at 35 days, to allow for a full month, even accounting for weekends including a bank holiday

## Next steps:

### DIFF
--- a/terraform/common/s3.tf
+++ b/terraform/common/s3.tf
@@ -2,6 +2,15 @@ data aws_caller_identity current {}
 
 resource aws_s3_bucket db_backups {
   bucket = "${data.aws_caller_identity.current.account_id}-tv-db-backups"
+
+  lifecycle_rule {
+    id      = "backups"
+    enabled = true
+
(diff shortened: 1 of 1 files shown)


terraform/common/s3.tf
```diff
@@ -2,6 +2,14 @@ data aws_caller_identity current {}
 
 resource aws_s3_bucket db_backups {
   bucket = "${data.aws_caller_identity.current.account_id}-tv-db-backups"
+
+  lifecycle_rule {
+    id      = "backups"
+    enabled = true
+
+    expiration {
+      days = 35
```
saliceti, line 11: This is only for disaster recovery. 7 days would be enough and save money.

saliceti (approved):

PR #2100, opened 2020-10-07
[TEVA-1387] - Exclude folders from triggering certain GitHub workflows

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1387

## Changes in this PR:

- Exclude 5 folders not required to build the app from starting `deploy*` or `test` workflows

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - master
+    paths:
+    - '!bigquery/**'
+    - '!documentation/**'    
+    - '!terraform/common/**'
+    - '!terraform/monitoring/**'
 
 env:
  DOCKERHUB_REPOSITORY: dfedigital/teaching-vacancies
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,12 @@ on:
   push:
     branches:
       - master
+    paths:
+    - '!.github/**'
```
saliceti, line 8: If we make a change to the workflow, we want to run it I guess, so this shouldn't be excluded?

nickromneydfe: My thinking was that as the workflow YML file itself is not part of the eventual app, then the next push which has a change to the app would be deployed using the new workflow. Possibly this decouples the feedback loop, so I'm happy to remove it.

saliceti: Yes I saw that issue in my previous job and it was really annoying.

nickromneydfe: Removed!

saliceti (approved):

PR #2101, opened 2020-10-08
Remove unused method

david-mears-dfe:
From git history, it appears we used to use to_row to convert the feedback objects to a row before adding them to a spreadsheet. We no longer use any spreadsheets.

https://github.com/DFE-Digital/teacher-vacancy-service/pull/766/files
https://github.com/DFE-Digital/teacher-vacancy-service/pull/855/files

Instead of spreadsheets, we have BigQuery storing the records, and also an audit table called activities.


cesidio (approved):
Nice cleanup

PR #2104, opened 2020-10-08
Job Alert Feedback: migration, model, specs

david-mears-dfe:
Intermediary PR for https://dfedigital.atlassian.net/browse/TEVA-1307

## Changes in this PR:

The back end changes that create the JobAlertFeedback model.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -9,6 +9,7 @@ class Subscription < ApplicationRecord
   enum frequency: FREQUENCY_OPTIONS
 
   has_many :alert_runs
+  has_many :job_alert_feedbacks
 
   validates :email, email_address: { presence: true }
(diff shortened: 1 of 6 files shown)


app/models/subscription.rb
```diff
@@ -9,6 +9,7 @@ class Subscription < ApplicationRecord
   enum frequency: FREQUENCY_OPTIONS
 
   has_many :alert_runs
+  has_many :job_alert_feedbacks
```
cesidio, line 12: Can you leave more than one feedback for a subscription?

david-mears-dfe: Yes. You can leave positive feedback. Then maybe 4 months later you decide to leave negative feedback. Maybe you edited the alert in between.

cesidio (approved):

PR #2110, opened 2020-10-09
Move Notify template IDs from env to repository

cesidio:


david-mears-dfe (approved):

nickromneydfe (approved):
I've run a comparison between the existing values stored in Parameter store, and confirm they are the same

PR #2119, opened 2020-10-12
[WIP] add UI to manage filters of a subscription

alexbowen:
need preview environment hence created WIP PR


app/form_models/subscription_form.rb
```diff
@@ -0,0 +1,61 @@
+class SubscriptionForm
```
alexbowen, line 1: realise there is lots thats not needed in here

PR #2133, opened 2020-10-13
Turn on MultiSchoolJobsFeature on production

cpjmcquillan:
What it says on the tin


spec/rails_helper.rb
```diff
@@ -54,7 +54,7 @@
 
   config.before do
     allow(SchoolGroupJobsFeature).to receive(:enabled?).and_return(true)
-    allow(MultiSchoolJobsFeature).to receive(:enabled?).and_return(false)
+    allow(MultiSchoolJobsFeature).to receive(:enabled?).and_return(true)
```
cesidio, line 57: Do we still need these 2 lines?

cpjmcquillan: Until the feature flags are removed I think so

PR #2137, opened 2020-10-14
[DRAFT] front end structure initial draft - feedback encouraged

alexbowen:
this is a first draft of clarification of what the front end structure should be and will drive the implementation/refactoring in upcoming JIRA tickets. this is intended for discussion/ideas/thoughts

my main drivers for this are (and this is what i encourage discussion/comments on)
- consistency
- not over engineering
- finding/using functionality is intuitive
- reusability

### DIFF
--- a/app/components/README.md
+++ b/app/components/README.md
@@ -1,0 +1,19 @@
+## Rails view components
+JS and SCSS files are optional
+```
+/app
+  /components
+    /my_component
+      my_component.js
+      my_component.test.js
+      my_component.scss
+      my_component.html.haml
+    my_component.rb
+```
+### HAML usage
+`= render(MyComponent.new()`
+
+### SCSS usage
+import in styles/application.scss
+
+`@import 'my_component/my_component';`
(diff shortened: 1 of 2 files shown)


app/components/README.md
```diff
@@ -0,0 +1,19 @@
...
+JS and SCSS files are optional
+```
+/app
+  /components
+    /my_component
+      my_component.js
+      my_component.test.js
+      my_component.scss
+      my_component.html.haml
+    my_component.rb
```
david-mears-dfe, line 11: Do we want the .rb in the directory, too? (Saving vertical visual space)

alexbowen: my thinking was thats in an entry point to using the component

alexbowen: in node.js you can have an index file that mean you only need to reference the folder. if can do this consistently with rails too i would favour what you suggest here @david-mears-dfe

david-mears-dfe: "thats in an entry point to using the component"

I'm not sure what you mean?

I'm not too picky about this.

cpjmcquillan (approved):
Looks sensible to me 👍

ddippolito (approved):
Happy with this

PR #2138, opened 2020-10-14
[TEVA-1320] Additional copy for 401 unauth page

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1320

Screenshot 2020-10-14 at 08 37 51

### DIFF
--- a/app/views/errors/unauthorised.html.haml
+++ b/app/views/errors/unauthorised.html.haml
@@ -1,4 +1,9 @@
-- content_for :page_title_prefix, t('error_pages.unauthorised')
+- content_for :page_title_prefix, t('error_pages.unauthorised.title')
 
-%h1.govuk-heading-xl=t('error_pages.unauthorised')
+%h1.govuk-heading-xl=t('error_pages.unauthorised.title')
+%p= t('error_pages.unauthorised.text_html')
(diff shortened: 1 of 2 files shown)


app/views/errors/unauthorised.html.haml
```diff
@@ -1,4 +1,9 @@
-- content_for :page_title_prefix, t('error_pages.unauthorised')
+- content_for :page_title_prefix, t('error_pages.unauthorised.title')
 
-%h1.govuk-heading-xl=t('error_pages.unauthorised')
+%h1.govuk-heading-xl=t('error_pages.unauthorised.title')
+%p= t('error_pages.unauthorised.text_html')
+%ul.govuk-list.govuk-list--bullet
+  -t('error_pages.unauthorised.options').each do |li|
+    %li=li.html_safe
```
cpjmcquillan, line 7: I think this should work without the `.html_safe`

cpjmcquillan (approved):

app/views/errors/unauthorised.html.haml
alexbowen: no because one option has html in it, and i thought this was the least hacky way to do it

PR #2146, opened 2020-10-15
[TEVA-1346] Let cloudfront forward Host header to allow redirects

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1346

## Changes in this PR:

Configure cloudfront to forward Host header so it's passed down to the app in paas. Then the redirect code will work. 
Enabled only for dev now.


nickromneydfe (approved):
Looks good to me. Have cloned the branch, and run a successful `terraform plan` against dev, staging and production

PR #2173, opened 2020-10-20
Put Steven's GSC account verification in the correct place

david-mears-dfe:
Google needs to be able to access this html file at the address www.domain.com/googlef126f8525cc1ddc1.html in order to verify Steven's account. This is a one-off thing which can be removed as soon as it has been used.

Here it is on the review app: https://teaching-vacancies-review-pr-2173.london.cloudapps.digital/googlef126f8525cc1ddc1.html

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,13 @@ class ApplicationController < ActionController::Base
   include AuthenticationConcerns
   include Ip
 
+  # TODO: remove this after Steven has verified his account
+  def google_search_console_verification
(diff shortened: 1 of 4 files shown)


cpjmcquillan (approved):
It's kind of nasty that there isn't a better way to do this, but lets just do it and undo it so Steven is unblocked 👍

csutter:
We've got one of these in `/public` already - this should probably replace it? I think the Google ID has changed and that's why Steven is having issues...

david-mears-dfe:
Gonna do it the Christian way

PR #2179, opened 2020-10-21
Local authority tweaks to the create a job journey

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1407

## Changes in this PR
- Central office job location option removed for LAs
- Content tweaks to labels and hints in create a job journey

### DIFF
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -107,12 +107,13 @@ module VacanciesHelper
   end
 
   def vacancy_about_school_hint_text(vacancy)
-    return I18n.t('helpers.hint.job_summary_form.about_schools') if vacancy.organisations.many?
+    if vacancy.organisations.many?
+      return I18n.t('helpers.hint.job_summary_form.about_schools',
+                    organisation_type: organisation_type_basic(vacancy.parent_organisation))
+    end
 
-    I18n.t(
-      'helpers.hint.job_summary_form.about_organisation',
-      organisation_type: organisation_type_basic(vacancy.parent_organisation).capitalize,
-    )
+    I18n.t('helpers.hint.job_summary_form.about_organisation',
+           organisation_type: organisation_type_basic(vacancy.parent_organisation).capitalize)
   end
 
   def vacancy_about_school_value(vacancy)
@@ -127,5 +128,18 @@ module VacanciesHelper
       vacancy&.job_location == 'at_multiple_schools'
 
     address_join([organisation.name, organisation.town, organisation.county])
+  end
+
+  def vacancy_job_location_heading(vacancy)
+    return I18n.t("school_groups.job_location_heading.#{vacancy.job_location}") unless
+      vacancy.job_location == 'at_multiple_schools'
+
+    I18n.t('school_groups.job_location_heading.at_multiple_schools',
+           organisation_type: organisation_type_basic(vacancy.parent_organisation))
+  end
+
+  def vacancy_school_visits_hint(vacancy)
+    organisation = organisation_type_basic(vacancy.parent_organisation).gsub(' ', '_')
+    I18n.t("helpers.hint.application_details_form.#{organisation}_visits")
   end
 end
--- a/app/helpers/vacancies_options_helper.rb
+++ b/app/helpers/vacancies_options_helper.rb
@@ -3,11 +3,10 @@ module VacanciesOptionsHelper
     Vacancy.hired_statuses.keys.map { |k| [t("jobs.feedback.hired_status.#{k}"), k] }
   end
 
-  def job_location_options
-    Vacancy::JOB_LOCATION_OPTIONS.map { |key, _value|
-      [I18n.t("helpers.fieldset.job_location_form.job_location_options.#{key}"), key] if
-        MultiSchoolJobsFeature.enabled? || key != :at_multiple_schools
-    }.reject(&:blank?)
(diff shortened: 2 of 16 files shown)


app/helpers/vacancies_helper.rb
```diff
@@ -107,12 +107,13 @@ def vacancy_about_school_label_organisation(vacancy)
   end
 
   def vacancy_about_school_hint_text(vacancy)
-    return I18n.t('helpers.hint.job_summary_form.about_schools') if vacancy.organisations.many?
+    if vacancy.organisations.many?
+      return I18n.t('helpers.hint.job_summary_form.about_schools',
+                    organisation_type: vacancy.parent_organisation.group_type.gsub!('_', ' '))
```
david-mears-dfe, line 112: Alternatively, include the method `organisation_type_basic` from organisation_helper.

david-mears-dfe: If you like, you could update that method to use the `gsub` because it's nicer.

app/helpers/vacancies_helper.rb
```diff
@@ -107,12 +107,13 @@ def vacancy_about_school_label_organisation(vacancy)
...
+    end
 
-    I18n.t(
-      'helpers.hint.job_summary_form.about_organisation',
-      organisation_type: organisation_type_basic(vacancy.parent_organisation).capitalize,
-    )
+    I18n.t('helpers.hint.job_summary_form.about_organisation',
+           organisation_type: organisation_type_basic(vacancy.parent_organisation).capitalize)
```
david-mears-dfe, line 116: I think we should either use organisation_type_basic or gsub in both cases (consistently).

app/helpers/vacancies_helper.rb
```diff
@@ -128,4 +129,24 @@ def vacancy_job_location(vacancy)
...
+
+  def vacancy_job_location_heading(vacancy)
+    return I18n.t("school_groups.job_location_heading.#{vacancy.job_location}") unless
+      vacancy.job_location == 'at_multiple_schools'
+
+    if vacancy.parent_organisation.group_type == 'local_authority'
+      I18n.t('school_groups.job_location_heading.at_multiple_schools', organisation_type: 'local authority')
+    else
+      I18n.t('school_groups.job_location_heading.at_multiple_schools', organisation_type: 'trust')
+    end
```
david-mears-dfe, line 141: This could replace the if-else-end:
```
I18n.t('school_groups.job_location_heading.at_multiple_schools', organisation_type: organisation_type_basic(vacancy.parent_organisation))
```

app/helpers/vacancies_helper.rb
```diff
@@ -128,4 +129,24 @@ def vacancy_job_location(vacancy)
...
+  end
+
+  def vacancy_school_visits_hint(vacancy)
+    visits = if vacancy.parent_organisation.group_type == 'local_authority'
+      'local_authority'
+             else
+      organisation_type_basic(vacancy.parent_organisation)
+             end
```
david-mears-dfe, line 149: Alternatively,
```ruby
visits = organisation_type_basic(vacancy.parent_organisation).gsub(' ', '_')
```

app/helpers/vacancies_options_helper.rb
```diff
@@ -24,6 +23,16 @@ def listed_elsewhere_options
     Vacancy.listed_elsewheres.keys.map { |k| [t("jobs.feedback.listed_elsewhere.#{k}"), k] }
   end
 
+  def mapped_job_location_options(organisation)
+    Vacancy::JOB_LOCATION_OPTIONS.map do |k, _v|
+      next unless MultiSchoolJobsFeature.enabled? || k != :at_multiple_schools
```
david-mears-dfe, line 28: We should probably get rid of this feature flag, right?

david-mears-dfe: If I'm reading it correctly, this should be `&&`, not `||` - what would be the point of rejecting every key-value pair where the key is not `:at_multiple_schools`?

config/locales/en.yml
```diff
@@ -825,13 +826,9 @@ en:
...
     job_location_heading:
-      central_office: At head office
+      central_office: Trust head office
       at_one_school: Single school
-      at_multiple_schools: At more than one school in the trust
-      review:
-        central_office: Trust head office
-        at_one_school: Single school
-        at_multiple_schools: At more than one school in the trust
+      at_multiple_schools: At more than one school in the %{organisation_type}
```
david-mears-dfe, line 831: Thanks for clearing this bit up

app/helpers/vacancies_options_helper.rb
cpjmcquillan: Yes I would like to. It would be nice to have some sort of process in place for how long features are live before we remove them, I'll probably raise a PR soon for it, we haven't seen any issues with it anyway and I assume we won't want to turn it off again

cpjmcquillan: The next happens only when MultiSchoolJobsFeature is disabled and k == :at_multiple_schools (false || false). Everything else evaluates to true and the code following is executed

david-mears-dfe: I'm confused by the number of nested negatives there! ('Unless not...' whu?) But you're right.

david-mears-dfe (approved):

PR #2185, opened 2020-10-22
[TEVA-1418] LA authentication fallback

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1418

## Changes in this PR:

- We're letting LAs sign in via the fallback authentication, which is there to be switched on if DSI has an outage.


app/views/hiring_staff/sign_in/email/sessions/choose_organisation.html.haml
```diff
@@ -9,10 +9,16 @@
 
       %p= t('hiring_staff.temp_login.choose_organisation.please_select')
 
-      - @schools.each do |school|
-        = link_to(location(school), auth_email_create_session_path(urn: school.urn), class: 'govuk-link')
+      - @local_authorities.each do |la|
```
cesidio, line 12: it is ok to have links, but a nicer radio buttons interface like DSI would be better here

cesidio (approved):

app/views/hiring_staff/sign_in/email/sessions/choose_organisation.html.haml
david-mears-dfe: It would be but that's a piece of work that might never be used.

PR #2190, opened 2020-10-22
Change command to run with review apps

cesidio:


nickromneydfe (approved):
Reducing scope of command from to exclude `db:create`

PR #2191, opened 2020-10-22
Update append-yesterdays-daily-users-from-cloudfront-logs.sql

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1078
https://dfedigital.atlassian.net/browse/TEVA-1429

## Changes in this PR:

- Add 'medium' field to users table (to emulate the GA medium, but better)
- Add landing page fields to users table
- Pull user UTM parameters out of the landing page query string only (not all query strings accessed by that user that day)
- Limit amount of log data handled by query to only the data the query needs.
- Replicate all these changes in the monthly users table as well as the daily users table.
- Use utm_campaign to detect traffic from job alerts instead of utm_source

### DIFF
--- a/bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
+++ b/bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
@@ -21,7 +21,7 @@ WITH
   SELECT
     *,
     CASE
-      WHEN utm_source="subscription" THEN "Job alert"
+      WHEN utm_campaign LIKE "%alert%" THEN "Job alert"
       WHEN LOWER(utm_medium) LIKE "%email%" THEN "Email"
       WHEN referrer LIKE "%facebook%" OR referrer LIKE "%twitter%" OR referrer LIKE "%t.co%" OR referrer LIKE "%linkedin%" OR referrer LIKE "%youtube%" THEN "Social"
       WHEN (referrer NOT LIKE "%teaching-jobs.service.gov.uk%"
@@ -65,7 +65,7 @@ WITH
       job_alert_destinations,
       vacancies_viewed_slugs IS NOT NULL AS viewed_a_vacancy,
       vacancies_with_gmi_clicks_ids IS NOT NULL AS clicked_get_more_information,
-      first_page.query LIKE "%utm_source=subscription%" AS from_job_alert,
+      first_page.query LIKE "%utm_campaign=%alert%" AS from_job_alert,
       ARRAY_LENGTH(search_parameters) AS unique_searches,
       ARRAY_LENGTH(vacancies_viewed_slugs) AS vacancies_viewed,
       ARRAY_LENGTH(vacancies_with_gmi_clicks_ids) AS vacancies_with_gmi_clicks,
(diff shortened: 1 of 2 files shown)


bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
```diff
@@ -1,96 +1,155 @@
...
-    ARRAY_LENGTH(vacancies_with_gmi_clicks_ids) AS vacancies_with_gmi_clicks
+    DATE_TRUNC(date, MONTH) AS month,
+    DATETIME(date,
+      time) AS datetime,
+    c_ip,
+    cs_uri_stem,
+    cs_referer,
+    c_user_agent,
+    cs_uri_query,
+    sc_content_type
```
david-mears-dfe, line 15: `c`, `sc`, and `cs` are opaque to a newcomer

bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
```diff
@@ -1,96 +1,155 @@
...
+      WHEN LOWER(utm_medium) LIKE "%email%" THEN "Email"
+      WHEN referrer LIKE "%facebook%" OR referrer LIKE "%twitter%" OR referrer LIKE "%t.co%" OR referrer LIKE "%linkedin%" OR referrer LIKE "%youtube%" THEN "Social"
+      WHEN (referrer NOT LIKE "%teaching-jobs.service.gov.uk%"
+      AND referrer NOT LIKE "%teaching-vacancies.service.gov.uk%"
+      AND referrer NOT LIKE "%google%"
+      AND referrer NOT LIKE "%bing%"
+      AND referrer NOT LIKE "%yahoo%"
+      AND referrer NOT LIKE "%aol%"
+      AND referrer NOT LIKE "%ask.co%")
+    OR utm_medium="referral" THEN "Referral"
```
david-mears-dfe, line 34: Is this `OR` correct? Do we want to assume that if the referrer is not like any of the above, then the medium is referral?

bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
david-mears-dfe: Though perhaps they are not used enough to be worth renaming as I see you are just copying them from the table column names.

bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
stevenleggdfe: I think technically it *is* correct, due to a quirk of how the SQL 'not like' handles null values (NULL NOT LIKE '%aol%' returns FALSE, weirdly). But I don't think that's clear to the reader, so I'll add in an explicit NOT NULL check here as well.

Re. the logic itself, the data here is pretty messy, so we're trying to work out two possible ways traffic could be marked as referral traffic:
- the utm_medium parameter, which is where a referrer explicitly chooses to mark a link as a referral link
- the HTTP referrer

The HTTP referrer is often hidden, and most referrers won't add a utm_medium, hence why I'm using an OR here.

bigquery/scheduled_queries/append-monthly-users-from-cloudfront-logs.sql
stevenleggdfe: Yeah, it's how the server log is structured. But agreed, there is meaning there that isn't clear, so I'll put a comment in at least.

david-mears-dfe (approved):

PR #2194, opened 2020-10-23
[TEVA-1409] LA access allow list

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1409

## Changes in this PR
- Enable LA access allow list for DfE sign in
- Enable LA access allow list for authentication fallback

## Next steps
- Update `local_authorities.yml` with organisation `local_authority_code`s for trial


config/locales/en.yml
```diff
@@ -208,6 +208,7 @@ en:
             organisation until DfE Sign In is working as normal. Please %{try_again} later.
         please_select: >-
           You are associated with more than one organisation. Please select the one you wish to sign in with.
+      not_authorised: You are not authorised to log in
```
david-mears-dfe, line 211: I think we need a more informative message here - the user might be trying to log in as an LA whereas they do have access as a school, and we could suggest they try logging in as a school. Something to run by @valentine-carter ?

david-mears-dfe: Oh, but this message is only for authentication fallback, in which case it's rather unlikely this will ever be seen by anyone.

cpjmcquillan: In order to be seen by someone they would need to 
- Have access to TVS as an LA in DSI
- DSI to be down
- Not be in the allowed LA users list (which is intended to be shortlived/for trial purposes only)
which is very unlikely if not impossible!

david-mears-dfe (approved):

cesidio (approved):

PR #2196, opened 2020-10-23
Add host prefix to indexes in Algolia

cesidio:


csutter:
I like what we've come up with!

One thing: We should add `INDEX_PREFIX` to the `.env.example` and the docs to explain how it keeps different local indices apart.

cpjmcquillan (approved):

PR #2199, opened 2020-10-26
[TEVA-1434] Only add link to referrer if referred from jobs path

cpjmcquillan:
Given the ticket is a bug, I think this meets the definition of a fixed bug. Ensuring search parameters are persisted after a user navigates to `Get similar job alert` and back is outside the scope of this ticket, is functionality in it's own right, and not something I'm sure we should necessarily do anyway.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1434

## Changes in this PR:
- Only add referrer link for search results if the referrer path is the jobs page

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::Base
   before_action :detect_device_format
   before_action :set_root_headers
 
-  helper_method :cookies_preference_set?, :referred_from_host?
+  helper_method :cookies_preference_set?, :referred_from_jobs_path?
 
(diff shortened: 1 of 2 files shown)


alexbowen (approved):
looks fine for what the PR is about but for the record im not a fan of little tenary expressions in templates like was already here in the view (related to other PRs ive put similar comments on)

cpjmcquillan:
> looks fine for what the PR is about but for the record im not a fan of little tenary expressions in templates like was already here in the view (related to other PRs ive put similar comments on)

But they're cute! If you can lobby someone for a +1 I'll move it elsewhere

PR #2214, opened 2020-10-27
[TEVA-1197] Change text in heading component if a polygon search was used

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1197

## Changes in this PR:

- Added check for a polygon search
- Added translations for when a polygon search has taken place
- Updated specs

### Location Search

Screenshot 2020-10-27 at 12 53 36

### Location Polygon Search

Screenshot 2020-10-27 at 12 53 13

### DIFF
--- a/spec/components/jobseekers/search_results/heading_component_spec.rb
+++ b/spec/components/jobseekers/search_results/heading_component_spec.rb
@@ -6,54 +6,55 @@ RSpec.describe Jobseekers::SearchResults::HeadingComponent, type: :component do
   let(:vacancies_search) { instance_double(Search::VacancySearchBuilder) }
   let(:keyword) { 'maths' }
   let(:location) { 'London' }
+  let(:location_polygon) { 'London' }
   let(:count) { 10 }
 
   before do
     allow(vacancies_search).to receive(:keyword).and_return(keyword)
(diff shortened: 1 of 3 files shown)


spec/components/jobseekers/search_results/heading_component_spec.rb
```diff
@@ -6,16 +6,40 @@
...
+
+      it 'renders correct heading' do
+        expect(rendered_component).to include(
+          I18n.t('jobs.search_result_heading.keyword_location_polygon_html', keyword: keyword, location: location, count: count),
+        )
+      end
+    end
+  end
+
+  context 'when keyword and location polygon are present' do
```
cpjmcquillan, line 40: copy paste typo

cpjmcquillan (commented):
I don't think there is much benefit in the tests that are 1 job/more than one job, since that is all handled by I18n, so I think we can remove them

spec/components/jobseekers/search_results/heading_component_spec.rb
josephhull676: Rumbled.

cpjmcquillan (approved):

PR #2222, opened 2020-10-28
Specify 2.7 instead of 2.7.1

cpjmcquillan:


cpjmcquillan:
We need to update to 2.7.2

PR #2224, opened 2020-10-28
Randomly add assignees to Dependabot PRs

david-mears-dfe:
This is to avoid the situation where nobody reviews the PR because it's too boring, and then the most altruistic person has to review all Dependabot PRs, and the PRs sit there saying 'high security issue' for weeks.

The auto-assignment bot will trigger unless the PR title contains `wip` or `teva`.

Unfortunately, this bot will trigger on PRs which are made by us humans, but don't contain those keywords, like this one doesn't. We could agree either:

(1) to add a keyword to all human PR titles that don't include `teva`, e.g. `*` (and add this to the list to skip)

or (2) ignore the assignee in these cases.

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,23 @@
+# Set to true to add reviewers to pull requests
+addReviewers: false
+
+# Set to true to add assignees to pull requests
+addAssignees: true
+
+# A list of assignees to be added to pull requests (GitHub user name)
+assignees:
+  - david-mears-dfe
+  - csutter
+  - alexbowen
+  - josephhull676
+  - cpjmcquillan
+  - cesidio
+
+# A number of assignees to add to the pull request
+# Set to 0 to add all of the assignees.
+numberOfAssignees: 1
+
+# A list of keywords to skip the PR if the title includes it
+skipKeywords:
+  - wip
+  - teva


.github/auto_assign.yml
```diff
@@ -0,0 +1,23 @@
+# Set to true to add reviewers to pull requests
+addReviewers: false
+
+# Set to true to add assignees to pull requests
+addAssignees: true
+
+# A list of assignees to be added to pull requests (GitHub user name)
+assignees:
```
csutter, line 8: I'm guessing there's no way of making it pick someone at random from a Github team? Just thinking it might end up being a bit annoying to have _one more_ thing to do when someone joins/leaves the team...

david-mears-dfe: Indeed not.

david-mears-dfe: This would be an alternative, but we would lose the skipKeywords feature: would apply to all PRs. We would need to ask someone with more admin rights than me to do this. https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/managing-code-review-assignment-for-your-team

csutter: Hmm, yeah. Another concern I'd have with random assignment (either through this or Github native functionality) is that the person it's assigned to might be on leave etc.

A more low-tech option (maybe aided by a Slack automation) could just be having a weekly rota of "first point of call" within the team, with a defined list of responsibilities including this, but also other things that are mildly annoying but should be shared within the team?

david-mears-dfe: Sounds good! We used to have a 'BAU' person for the sprint but the role wasn't very clearly defined so it died.

david-mears-dfe: Will keep PR open for another day or so in case anyone else wants to chime in.

cpjmcquillan:
This is cool with me. I don't mind anyone just taking responsibility if it's there to be done too

cpjmcquillan (approved):

PR #2226, opened 2020-10-28
[TEVA-1286] Dropping subjects and leadership tables and related fields in Vacancies, deleting models and references

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1286

## Changes in this PR:

- Dropped the Subjects and Leaderships tables. Removed fields in Vacancies table referencing these tables.
- Deleted Subjects And Leaderships models
- Removed references
- Updated specs 

Disappearance


spec/services/copy_vacancy_spec.rb
```diff
@@ -62,42 +62,11 @@
     end
 
     describe '#subjects' do
```
cpjmcquillan, line 64: We don't need this bit at all now

josephhull676: Will remove it now!

cesidio (approved):

PR #2227, opened 2020-10-28
Widen the scope of Dependabot

david-mears-dfe:
Previously, we limited it to security patches only, because we were getting too many PRs. But we can (and here, we do) set a limit on the number of open PRs.

https://dfedigital.atlassian.net/browse/TEVA-1131


nickromneydfe:
We should ensure we're not duplicating effort. 
We also have `.github/dependabot.yml` which I think is the new replacement - certainly it's at version 2, as per https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

cpjmcquillan (approved):

nickromneydfe (approved):
Happy that we are:
- taking a change which has been trialled on another project and [documented in Conflunce](https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/2072903681/Maintaining+Actions)
- limiting to two concurrent Pull Requests
- starting with weekly intervals

PR #2234, opened 2020-10-28
Remove asdf from readme as it doesn't support Ruby 2.7.2

david-mears-dfe:


cesidio:
It does support ruby 2.7.2

```bash
asdf plugin-update ruby
asdf list-all ruby
asdf install ruby 2.7.2
asdf local ruby 2.7.2
```

david-mears-dfe:
Thanks @cesidio !

PR #2254, opened 2020-10-30
Remove custom routes that conflict with pagination links

cesidio:
we will lose nice URLs like

https://teaching-vacancies.service.gov.uk/teaching-jobs-in-Leeds

that will become:

https://teaching-vacancies.service.gov.uk/jobs?location=Leeds


alexbowen (approved):
so there will be some work to do to the google sitemap? but that can wait till next week if so, important this is fixed for weekend

cpjmcquillan (approved):

PR #2264, opened 2020-11-02
[TEVA-1463] Do not omit nil vacancy attributes when merging to session

david-mears-dfe:
If we omit 'nil' attributes, then no overwriting of previous values happens for `nil`s.

We want to overwrite previous session attributes when the user goes back a step
and reselects a different option.

For example, when selecting 'at one school' on Step 1, the readable_job_location
should be set to nil, otherwise the previous value will be used instead in
the page heading ('Create a job listing at More than one school ()', in this
case.)

## Question

Does this justify writing a regression test?

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1463

### Before

Screenshot 2020-10-28 at 15 39 08

### After

Screenshot 2020-11-02 at 12 38 50


david-mears-dfe:
Need to find better solution

PR #2265, opened 2020-11-02
Create a content security policy

csutter:
- Adds a CSP (Content Security Policy) to the application to stop
  browsers from loading untrusted external resources or inline scripts
- Generate and use nonces for trusted inline scripts
- Adds a basic endpoint for browsers to report CSP violations

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-790


cesidio (approved):

csutter:
Looks like Google Maps embed requires enabling unsafe inline styles 😩  – https://teaching-vacancies-review-pr-2265.london.cloudapps.digital/jobs/geography-teacher

It's not great but I don't think we have a choice. Will add them into the CSP.

PR #2266, opened 2020-11-02
[TEVA 1404] scss refactor pt one

alexbowen:
- all existing style moved into same folder as other assets - as per README
- webpack configuration updated so that styles can be imported via components
- migrate existing component scss files into individual component folders
- perform some renaming for consistency and linting

This task is only going to get bigger hence issuing a part one PR


david-mears-dfe:
I do enjoy how the dismissable notification fades out. Make sure to clear it with UX if it isn't already.

app/frontend/src/components/loader/loader.scss
```diff
@@ -1,10 +1,9 @@
-.js-location-finder {
-
+.location-finder {
```
cesidio, line 1: wasn't it a good practice to prefix with js classes that were affected by javascript changes?

alexbowen: ok funny i was just talking to connor about this

alexbowen: is this good practice in the alphagov guidelines? i happy either way as long as consistent

alexbowen: actually i will revert this to keep the js- and keep it out of this task because there is js- in many other places

david-mears-dfe (approved):

cesidio (approved):

PR #2272, opened 2020-11-03
Remove remaining custom RuboCop rules

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1486


.rubocop.yml
```diff
@@ -5,52 +5,3 @@ inherit_gem:
...
     - Exclude
-
-AllCops:
-  TargetRubyVersion: 2.7.2
-  Exclude:
-    - 'db/**/*'
-    - 'bin/**/*'
-    - 'node_modules/**/*'
-    - 'tmp/**/*'
-    - 'vendor/**/*'
```
david-mears-dfe, line 16: I wonder why we were excluding these before?

cesidio: because we do not want to lint code that is not ours. The same rule is in the defaults I believe

david-mears-dfe (approved):

PR #2282, opened 2020-11-04
Update .editorconfig with GDS recommendations

cesidio:


csutter (approved):
Looks good ❤️ 

Maybe we should add Haml as well?

cesidio:
> Looks good ❤️
> 
> Maybe we should add Haml as well?

Or maybe Slim :)

PR #2289, opened 2020-11-05
[TEVA-1463] hiring staff journey page heading bugfix

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1463

## Changes in this PR:

- This was a lot simpler than I initially thought it was, because the first step (for school group hiring staff) actually only needs to have 'Create a job' as the title, which simplifies the task a lot.
- While trying to do the complicated version of this task I refactored the heading partial into its own view component in order to get the logic all in one place.

### Before (this is what the bug looked like)

Screenshot 2020-11-05 at 13 38 21

Screenshot 2020-11-05 at 13 37 53

### After

Screenshot 2020-11-05 at 14 40 18

### DIFF
--- a/app/components/hiring_staff/vacancy_form_page_heading_component.rb
+++ b/app/components/hiring_staff/vacancy_form_page_heading_component.rb
@@ -1,0 +1,50 @@
+class HiringStaff::VacancyFormPageHeadingComponent < ViewComponent::Base
+  delegate :current_organisation, to: :helpers
+  delegate :current_step, to: :helpers
+  delegate :total_steps, to: :helpers
+
+  def initialize(vacancy, session_vacancy_attributes)
+    unless session_vacancy_attributes.nil?
+      @session_job_location = session_vacancy_attributes["job_location"]
+      @session_readable_job_location = session_vacancy_attributes["readable_job_location"]
+    end
+    @vacancy = vacancy
+  end
+
+  def heading
+    vacancy.present? ? page_title : page_title_no_vacancy
+  end
+
+  def show_current_step?
+    %w[copy edit edit_published].exclude?(vacancy&.state)
+  end
+
+private
+
+  attr_reader :session_job_location, :session_readable_job_location, :vacancy
+
+  def page_title_no_vacancy
+    organisation_for_title =
+      if session_job_location == "at_one_school"
+        session_readable_job_location
+      elsif session_job_location == "at_multiple_schools"
+        "multiple schools"
+      else
+        current_organisation.name
+      end
+
+    organisation_for_title ? I18n.t("jobs.create_a_job_title", organisation: organisation_for_title) : I18n.t("jobs.create_a_job_title_no_org")
+  end
+
+  def page_title
+    return I18n.t("jobs.copy_job_title", job_title: vacancy.job_title) if vacancy.state == "copy"
+    return I18n.t("jobs.create_a_job_title", organisation: organisation_from_job_location) if
+      %w[create review].include?(vacancy.state)
+
+    I18n.t("jobs.edit_job_title", job_title: vacancy.job_title)
+  end
+
+  def organisation_from_job_location
+    vacancy.job_location == "at_multiple_schools" ? "multiple schools" : vacancy.parent_organisation_name
+  end
+end
--- a/app/components/hiring_staff/vacancy_form_page_heading_component.rb
+++ b/app/components/hiring_staff/vacancy_form_page_heading_component.rb
@@ -24,6 +24,8 @@ private
   attr_reader :session_job_location, :session_readable_job_location, :vacancy
 
   def page_title_no_vacancy
+    return I18n.t("jobs.create_a_job_title_no_org") unless current_organisation.is_a?(School) || current_step > 1
+
     organisation_for_title =
       if session_job_location == "at_one_school"
         session_readable_job_location
--- a/app/views/hiring_staff/vacancies/application_details/show.html.haml
+++ b/app/views/hiring_staff/vacancies/application_details/show.html.haml
@@ -1,6 +1,5 @@
 - content_for :page_title_prefix, page_title_prefix(@vacancy, @application_details_form, t('jobs.application_details'))
-= render 'hiring_staff/vacancies/vacancy_form_partials/heading'
-
+= render HiringStaff::VacancyFormPageHeadingComponent.new(@vacancy, session[:vacancy_attributes])
 - if @vacancy.present? && @vacancy&.state != 'create'
   = render 'hiring_staff/vacancies/vacancy_form_partials/cancel_and_return_link'
 - else
--- a/spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
+++ b/spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
@@ -1,0 +1,112 @@
+require "rails_helper"
+
+RSpec.describe HiringStaff::VacancyFormPageHeadingComponent, type: :component do
+  let(:school) { create(:school, name: "Teaching Vacancies Academy") }
+  let(:state) { "create" }
+  let(:vacancy) { create(:vacancy, state: state, job_title: "Test job title") }
+  let(:session_job_location) { nil }
+  let(:session_readable_job_location) { nil }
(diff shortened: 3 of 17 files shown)


app/views/hiring_staff/vacancies/application_details/show.html.haml
```diff
@@ -1,6 +1,5 @@
 - content_for :page_title_prefix, page_title_prefix(@vacancy, @application_details_form, t('jobs.application_details'))
-= render 'hiring_staff/vacancies/vacancy_form_partials/heading'
-
+= render HiringStaff::VacancyFormPageHeadingComponent.new(@vacancy, session[:vacancy_attributes]&.[]('job_location'), session[:vacancy_attributes]&.[]('readable_job_location'))
```
cesidio, line 2: I would use a simpler interface:
```ruby
render HiringStaff::VacancyFormPageHeadingComponent.new(@vacancy, session[:vacancy_attributes])
```
and then have the logic to extract job_location and readable_job_location in the component

david-mears-dfe: Fair enough, it's a bit ugly to look at. I was trying to be specific about what arguments are needed in the component.

app/components/hiring_staff/vacancy_form_page_heading_component.rb
```diff
@@ -0,0 +1,50 @@
+class HiringStaff::VacancyFormPageHeadingComponent < ViewComponent::Base
+  delegate :current_organisation, to: :helpers
+  delegate :current_step, to: :helpers
+  delegate :total_steps, to: :helpers
+
+  def initialize(vacancy, session_vacancy_attributes)
```
cesidio, line 6: I would set a default value of an empty hash so that the code that follows is a bit nicer
```ruby
def initialize(vacancy, session_vacancy_attributes = {})
```

david-mears-dfe: Good idea

david-mears-dfe:
This is weird, the default `= {}` doesn't seem to work:

```
     6: def initialize(vacancy, session_vacancy_attributes = {})
     7:   binding.pry
 =>  8:   @session_job_location = session_vacancy_attributes["job_location"]
     9:   @session_readable_job_location = session_vacancy_attributes["readable_job_location"]
    10:   @vacancy = vacancy
    11: end

[1] pry(#<HiringStaff::VacancyFormPageHeadingComponent>)> session_vacancy_attributes
=> nil
```

david-mears-dfe:
> This is weird, the default `= {}` doesn't seem to work:
> 
> ```
>      6: def initialize(vacancy, session_vacancy_attributes = {})
>      7:   binding.pry
>  =>  8:   @session_job_location = session_vacancy_attributes["job_location"]
>      9:   @session_readable_job_location = session_vacancy_attributes["readable_job_location"]
>     10:   @vacancy = vacancy
>     11: end
> 
> [1] pry(#<HiringStaff::VacancyFormPageHeadingComponent>)> session_vacancy_attributes
> => nil
> ```

https://stackoverflow.com/questions/10506091/why-do-rubys-default-parameter-values-not-get-assigned-to-nil-arguments#:~:text=The%20default%20parameter%20is%20used,yes%2C%20this%20is%20expected%20behavior.

cesidio (approved):

app/components/hiring_staff/vacancy_form_page_heading_component.rb
```diff
@@ -0,0 +1,52 @@
...
+    vacancy.present? ? page_title : page_title_no_vacancy
+  end
+
+  def show_current_step?
+    %w[copy edit edit_published].exclude?(vacancy&.state)
+  end
+
+private
+
+  attr_reader :session_job_location, :session_readable_job_location, :vacancy
```
cpjmcquillan, line 24: stuff like the placement of this line would be nice to be consistent across all components

app/components/hiring_staff/vacancy_form_page_heading_component.rb
```diff
@@ -0,0 +1,52 @@
+class HiringStaff::VacancyFormPageHeadingComponent < ViewComponent::Base
+  delegate :current_organisation, to: :helpers
+  delegate :current_step, to: :helpers
+  delegate :total_steps, to: :helpers
+
+  def initialize(vacancy, session_vacancy_attributes)
```
cpjmcquillan, line 6: also how we pass arguments to the component - probably out of scope of the ticket, but something for us to think about

spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      it "shows the current step" do
+        expect(rendered_component).to include(I18n.t("jobs.current_step", step: current_step, total: 100))
+      end
+    end
+
+    context "when vacancy state is not create or review" do
+      let(:state) { "copy" }
+
+      it "does not show the current step" do
+        expect(rendered_component).not_to include(I18n.t("jobs.current_step", step: 100, total: 101))
```
cpjmcquillan, line 35: I wonder if being less specific here about what not to include might be a more revealing test?

cpjmcquillan: Either way wouldn't we want it to not include `step: current_step`?

spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      context "when the user is not signed in as a single school and the current step is 1" do
+        let(:organisation) { create(:trust) }
+        let(:current_step) { 1 }
+        let(:session_job_location) { "at_one_school" }
+        let(:session_readable_job_location) { "That Specific School" }
+
+        it "returns the create a job title without an organisation, ignoring the session parameters" do
+          expect(
+            page.text.split("Step").first.strip,
```
cpjmcquillan, line 52: Why do we do this bit like this as opposed to `rendered_component`?

david-mears-dfe: I want to avoid a false positive where the incorrect heading `Create a job for X School` passes a test for `Create a job`.
The way I did it was split the rendered text where it goes onto the next content ('Step') to make sure the first line _only_ includes `Create a job`.

spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
david-mears-dfe: I'll change it to `expect(rendered_component.downcase).not_to include('step')` - ok?

app/components/hiring_staff/vacancy_form_page_heading_component.rb
david-mears-dfe: What's the alternative placement?

I want to restrict the interface of the component as far as possible by not keeping attr_readers private where possible.

app/components/hiring_staff/vacancy_form_page_heading_component.rb
david-mears-dfe: Open to other ideas about how to pass arguments in. Originally I had two separate arguments for the two parts of the session[:vacancy_attributes] that are actually used. That resulted in rather ugly views like

```ruby
= render HiringStaff:: VacancyFormPageHeadingComponent.new(@vacancy, session[:vacancy_attributes]&.[]('job_location'), session[:vacancy_attributes]&.[]('readable_job_location'))
```

app/components/hiring_staff/vacancy_form_page_heading_component.rb
cpjmcquillan: That's cool, just would be nice (whatever we do) to do the same across the different components if possible

spec/components/hiring_staff/vacancy_form_page_heading_component_spec.rb
cpjmcquillan: I prefer the first

cpjmcquillan (approved):

alexbowen (approved):

PR #2292, opened 2020-11-06
[TEVA-1484] Remove vacancy scopes not relevant anymore

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1484

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -233,30 +233,16 @@ class Vacancy < ApplicationRecord
                       :expires_on, error_clash_behaviour: :omit_gov_uk_date_field_error
 
   scope :active, (-> { where(status: %i[published draft]) })
-  scope :applicable, (-> { applicable_by_date.or(applicable_by_time) })
-  scope :applicable_by_time, (-> { where("expiry_time IS NOT NULL AND expiry_time >= ?", Time.zone.now) })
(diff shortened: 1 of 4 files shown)


app/models/vacancy.rb
```diff
@@ -233,30 +233,16 @@ def _geoloc
                       :expires_on, error_clash_behaviour: :omit_gov_uk_date_field_error
 
   scope :active, (-> { where(status: %i[published draft]) })
-  scope :applicable, (-> { applicable_by_date.or(applicable_by_time) })
-  scope :applicable_by_time, (-> { where("expiry_time IS NOT NULL AND expiry_time >= ?", Time.zone.now) })
-  scope :applicable_by_date, (-> { where("expiry_time IS NULL AND expires_on >= ?", Time.zone.today) })
+  scope :applicable, (-> { where("expiry_time >= ?", Time.zone.now) })
```
cpjmcquillan, line 236: What happens for vacancies where the expiry_time column is NULL?

cesidio: I just run this on production:

```ruby
Vacancy.where("expiry_time IS NULL AND expires_on IS NOT NULL").find_each { |v| v.update_column :expiry_time, v.expires_on }
```

we do not have vacancies with expiry_time nil anymore

cpjmcquillan: Cool I was wondering if you'd run something like that. Nice!

cpjmcquillan (approved):

PR #2295, opened 2020-11-06
[TEVA-1535] Add GitHub secret use to documentation

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1535

## Changes in this PR:

- In preparing to remove 4 unused secrets from GitHub secrets, I created a document showing which workflows consume which current secrets

### DIFF
--- a/documentation/github-actions.md
+++ b/documentation/github-actions.md
@@ -1,0 +1,97 @@
+# GitHub Actions
+
+## Secrets
+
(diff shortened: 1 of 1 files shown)


nickromneydfe:
The document as it stands will be almost impossible to keep up to date, and the use of secrets may well be refactored to use AWS SSM Parameter Store. Will move the more generic information under TEVA-890.

PR #2300, opened 2020-11-09
[TEVA-1539] Refactor how subscriptions are unsubscribed

cpjmcquillan:
Subscriptions that had associated job alert feedbacks could not be unsubscribed from (deleted). This approach unsubscribes the user, while retaining useful information and maintaining the relationship between the tables.

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1539

## Changes in this PR
- Add active boolean column to subscriptions table
- Add unsubscribed_on datetime column to subscriptions table
- Unsubscribe subscription rather than deleting the record (set email to nil, active to false, unsubscribed_on to now)
- Scope subscriptions to active for job alert emails

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -11,9 +11,7 @@ class Subscription < ApplicationRecord
   has_many :alert_runs
   has_many :job_alert_feedbacks
 
-  validates :email, email_address: { presence: true }
-  validates :frequency, presence: true
-  validates :search_criteria, uniqueness: { scope: %i[email frequency] }
+  scope :active, (-> { where(active: true) })
 
   def self.encryptor
     key_generator_secret = SUBSCRIPTION_KEY_GENERATOR_SECRET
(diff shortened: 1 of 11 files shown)


app/models/subscription.rb
```diff
@@ -44,6 +42,10 @@ def token
     self.class.encryptor.encrypt_and_sign(token_values)
   end
 
+  def unsubscribe
+    update(email: nil, active: false, unsubscribed_on: Time.zone.now)
```
cesidio, line 46: If it is a time, it should be called `unsubscribed_at`

cesidio (approved):

app/models/subscription.rb
cpjmcquillan: Damn it! 😂

david-mears-dfe (approved):

app/models/subscription.rb
stevenleggdfe: @cpjmcquillan at this stage, should you also set any location search criteria to null? Locations can be postcodes, which can be construed as PII, and so probably shouldn't be retained after a user thinks they have 'deleted' their subscription.

PR #2302, opened 2020-11-09
[TEVA-1486] Move search logic into Concern

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1485

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -1,0 +1,137 @@
+module Searchable
+  extend ActiveSupport::Concern
+
+  INDEX_NAME = [ENV["ALGOLIA_INDEX_PREFIX"], DOMAIN, Vacancy].compact.join("-").freeze
+
+  included do
+    include AlgoliaSearch
+
+    scope :unindexed, (-> { live.where(initially_indexed: false) })
+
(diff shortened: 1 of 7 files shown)


app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,147 @@
+module Searchable
+  extend ActiveSupport::Concern
+
+  included do
+    include AlgoliaSearch
+
+    INDEX_NAME = [ENV["ALGOLIA_INDEX_PREFIX"], DOMAIN, name].compact.join("-").freeze
```
csutter, line 7: I think having a constant defined here might be an issue.

It'll be set on the `Searchable` module even though it's in the `included` block, unless we use `base.const_set` (which might then still be an issue with referencing the constant in specs etc). How do you feel about moving the constant to be at the top of this file, so explicitly on the `Searchable` module, and then changing the references to `Vacancy::INDEX_NAME` throughout the app to `Searchable::INDEX_NAME`?

csutter (approved):
🚢

david-mears-dfe:
I'm confused by the `Searchable` naming. Is `-able` some Rubyish convention? I would expect an object's name to be a noun, normally.

cesidio:
> able

Yes it is a Ruby convention. Something similar to what searchable module provides is https://ruby-doc.org/core-2.7.2/Enumerable.html

PR #2307, opened 2020-11-09
Bigquery job alert active unsubscribed at

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1539

## Changes in this PR:
Changes to BigQuery scheduled queries and views to handle / exploit the new active and unsubscribed_at fields in the subscription table:
- Add active, unsubscribed_date and days_active fields to job_alert view of subscription table.
- Add job alert creation/unsubscription counts to daily time metrics table.
- Add job alert creation CTR to daily time metrics table. (now possible because we have this goal set up in the table of daily users from Cloudfront logs)
- Add unsubscription count to daily job alert metrics table.
- Take unsubscriptions into account when calculating the number of live job alerts for each day for daily job alert metrics table.

### DIFF
--- a/bigquery/views/job-alert.sql
+++ b/bigquery/views/job-alert.sql
@@ -2,6 +2,14 @@ SELECT
   id,
   CAST(created_at AS date) AS created_date,
   CAST(updated_at AS date) AS updated_date,
+  active,
+  CAST(unsubscribed_at AS date) AS unsubscribed_date,
+IF
+  (NOT active,
+    DATETIME_DIFF(unsubscribed_at,
+      created_at,
+      DAY),
+    NULL) AS days_active,
   recaptcha_score,
   FIRST_VALUE(id) OVER (PARTITION BY email ORDER BY created_at) AS email_address_id,
   #the ID of the first job alert with this email address - allows us to string job alerts with the same email address together without including the PII in this view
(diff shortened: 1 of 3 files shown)


bigquery/views/job-alert.sql
```diff
@@ -2,6 +2,14 @@ SELECT
   id,
   CAST(created_at AS date) AS created_date,
   CAST(updated_at AS date) AS updated_date,
+  active,
+  CAST(unsubscribed_at AS date) AS unsubscribed_date,
```
david-mears-dfe, line 6: This is a datetime - don't know if that matters.

bigquery/views/job-alert.sql
```diff
@@ -2,6 +2,14 @@ SELECT
...
+IF
+  (NOT active,
+    DATETIME_DIFF(unsubscribed_at,
+      created_at,
+      DAY),
+    NULL) AS days_active,
```
david-mears-dfe, line 12: Re datetime, perhaps this will come out as a strange number like 14.238483 - also don't know if that matters, if true.

bigquery/views/job-alert.sql
stevenleggdfe: Yes, the conversion to a date is deliberate - because downstream this is used to calculate loads of counts by day, and it saves a lot of duplicate type conversion to put it here.

bigquery/views/job-alert.sql
stevenleggdfe: Interesting thought! I just checked the documentation though, and it should come out as a whole number of DAY intervals, not as a decimal: https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_diff

cesidio (approved):

PR #2308, opened 2020-11-10
Remove usused attribute

david-mears-dfe:
Remove usused attribute introduced in commit 8f9766601b28754f21057acf79c705b2c0571456


cpjmcquillan (commented):
See `app/views/shared/search/_location.html.haml` and `app/frontend/src/application/jobSeekers/location.js` (it is used)

david-mears-dfe:
> See `app/views/shared/search/_location.html.haml` and `app/frontend/src/application/jobSeekers/location.js` (it is used)

That uses `Search::VacancySearchBuilder#point_coordinates`, which is set by line 52 in `Search::VacancyLocationBuilder`. But this is `VacanciesPresenter#coordinates`.

cpjmcquillan (approved):
Right you are, burn it 🔥

PR #2309, opened 2020-11-10
Update local_authorities.yml

milimaldedfe:
Added Lincolnshire (925) and North Yorkshire (815) to allowed list for LA dashboard.


lib/allowed_organisations/local_authorities.yml
```diff
@@ -1,3 +1,5 @@
 ---
 - '309'
 - '330'
+- '925'
```
cesidio, line 4: we tend to sort things alphabetically or numerally in this case for easier look-up

cesidio (approved):

lib/allowed_organisations/local_authorities.yml
milimaldedfe: ok thanks! good to know

PR #2315, opened 2020-11-10
[TEVA-1484] Rename expiry_time to expires_at

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1484


app/mailers/alert_mailer.rb
```diff
@@ -5,7 +5,7 @@ class AlertMailer < ApplicationMailer
   def alert(subscription_id, vacancy_ids)
     subscription = Subscription.find(subscription_id)
     @subscription = SubscriptionPresenter.new(subscription)
-    vacancies = Vacancy.where(id: vacancy_ids).order(:expires_on).order(:expiry_time)
+    vacancies = Vacancy.where(id: vacancy_ids).order(:expires_on).order(:expires_at)
```
cpjmcquillan, line 8: Do we need to chain the two attributes, or could we just order by `:expires_at`?

spec/components/jobseekers/vacancy_summary_component_spec.rb
```diff
@@ -42,7 +42,7 @@
 
       context "when expiry time is nil" do
```
cpjmcquillan, line 43: `"when expires_at is nil"`

db/migrate/20180404131534_add_leadership_data.rb
```diff
@@ -1,9 +1,4 @@
 class AddLeadershipData < ActiveRecord::Migration[5.1]
   def change
-    ['Executive Head', 'Headteacher',
```
cpjmcquillan, line 3: Why is this change here?

cpjmcquillan (approved):

PR #2319, opened 2020-11-11
Move away from deprecated `DeliveryJob`

csutter:
This has been deprecated in favour of `MailDeliveryJob` and was causing
Rails 6.1 deprecation warnings.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1488

### DIFF
--- a/app/jobs/alert_mailer_job.rb
+++ b/app/jobs/alert_mailer_job.rb
@@ -1,4 +1,4 @@
-class AlertMailerJob < ActionMailer::DeliveryJob
+class AlertMailerJob < ActionMailer::MailDeliveryJob
   EXPIRES_IN = 4.hours
 
   before_enqueue do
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):

app/jobs/alert_mailer_job.rb
```diff
@@ -32,7 +32,7 @@ def subscription
   end
 
   def subscription_id
-    arguments[3]
+    arguments.last[:args].first
```
csutter, line 35: Moving to `MailDeliveryJob` means we have to mangle job arguments like this now as the method signature of `#perform` has changed:

```ruby
# DeliveryJob
def perform(mailer, mail_method, delivery_method, *args)
def perform(mailer, mail_method, delivery_method, args:, kwargs: nil, params: nil)
# MailDeliveryJob

```

https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/delivery_job.rb#L23
https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/mail_delivery_job.rb#L16

PR #2322, opened 2020-11-11
Add Brakeman gem for security analysis

csutter:
- Set up configuration and ignore false positives caused by
  `view_components` gem
- Fix identified issue with missing `rel="nooopener"` on external links
  opening in new tab
- Add Brakeman to backend linting Github action
- Remove unsafe use of `params.permit!` (and inline into only usage) in
  `HiringStaff::Vacancies::ApplicationController`
- Remove unused `retrieve_job_from_db` method in
  `HiringStaff::Vacancies::ApplicationController`

### DIFF
--- a/app/controllers/hiring_staff/vacancies/application_controller.rb
+++ b/app/controllers/hiring_staff/vacancies/application_controller.rb
@@ -6,10 +6,6 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
 
   before_action :set_vacancy
 
-  def vacancy_id
-    params.permit![:job_id]
-  end
-
   def session_vacancy_id
     session[:vacancy_attributes].present? ? session[:vacancy_attributes]["id"] : false
   end
@@ -100,10 +96,6 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
 
   def redirect_unless_vacancy_session_id
     redirect_to job_specification_organisation_job_path(school_id: current_school.id) unless session_vacancy_id
-  end
-
-  def retrieve_job_from_db
-    current_organisation.all_vacancies.published.find(vacancy_id).attributes
   end
 
   def source_update?
--- a/app/views/hiring_staff/organisations/schools/_school_group.html.haml
+++ b/app/views/hiring_staff/organisations/schools/_school_group.html.haml
@@ -27,9 +27,9 @@
         %strong.strong= t('school_groups.website', organisation_type: organisation_type_basic(current_organisation).humanize)
       %td.govuk-table__cell.change_answer
         - if current_organisation.website.present?
-          = link_to current_organisation.website, current_organisation.website, class: 'govuk-link wordwrap', target: '_blank'
+          = link_to current_organisation.website, current_organisation.website, class: 'govuk-link wordwrap', target: '_blank', rel: 'noopener'
         - elsif current_organisation.url.present?
-          = link_to current_organisation.url, current_organisation.url, class: 'govuk-link wordwrap', target: '_blank'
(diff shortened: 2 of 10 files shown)


app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -102,10 +98,6 @@ def redirect_unless_vacancy_session_id
     redirect_to job_specification_organisation_job_path(school_id: current_school.id) unless session_vacancy_id
   end
 
-  def retrieve_job_from_db
```
csutter, line 105: This is not actually used anywhere.

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -6,10 +6,6 @@ class HiringStaff::Vacancies::ApplicationController < HiringStaff::BaseControlle
 
   before_action :set_vacancy
 
-  def vacancy_id
```
csutter, line 9: This is only used in one place, and might as well just be inlined there.

cesidio (approved):

app/views/hiring_staff/organisations/schools/_school_group.html.haml
```diff
@@ -27,9 +27,9 @@
         %strong.strong= t('school_groups.website', organisation_type: organisation_type_basic(current_organisation).humanize)
       %td.govuk-table__cell.change_answer
         - if current_organisation.website.present?
-          = link_to current_organisation.website, current_organisation.website, class: 'govuk-link wordwrap', target: '_blank'
+          = link_to current_organisation.website, current_organisation.website, class: 'govuk-link wordwrap', target: '_blank', rel: 'noopener'
```
david-mears-dfe, line 30: What would you say to overwriting `link_to` in our app, so that it always has a rel: 'noopener' (so we don't have to rely on remembering to include it)?

csutter: It's only really relevant for a small subset of our links (those that open in a new tab/window using `target="blank"`). We could have a separate helper for it but I think it's probably not worth it given we just have a few of them.

david-mears-dfe: What would be the demerits of doing something like this?

```ruby
def link_to(args, **options)
  if options[:target] == 'blank'
    options.merge({rel: 'noopener'})
  end
  RailsWherever.link_to(args, options)
end
```

csutter: Good question. In general, I tend to override/monkey patch existing framework functionality (as opposed to adding a new explicit helper) only as a last resort, because it's usually tending to be too implicit. In this particular instance, it wouldn't be a huge change in and of itself, but someone who's worked on a fair few Rails projects might not expect `link_to` to behave this way.

It would also force this attribute even in those cases where we may _not_ want that (not that I can think of any, to be fair), possibly lead to headaches if the signature of `link_to` changes in a future Rails upgrade, and possibly cause issues if a library that we use overrides `link_to` as well. You could also argue that if this really was behaviour any link should always adopt, it would already be part of Rails core.

csutter: The `safe_target_blank` gem takes a slightly different approach by overriding `UrlHelper`'s `convert_options_to_data_attributes`:
https://github.com/jvenezia/safe_target_blank/blob/master/lib/safe_target_blank/url_helper.rb

But I feel like adding an extra library for the sake of saving ourselves a handful of characters in the code is probably overkill 😅  Plus we now have Brakeman to keep a watchful eye out for us!

david-mears-dfe: Thanks!

PR #2334, opened 2020-11-12
Enable `NewCops` in Rubocop

csutter:
New rules added to Rubocop between major versions are disabled by
default. `rubocop-govuk` currently makes that setting explicit.

Enabling it means we get new rules as soon as they are added in minor
Rubocop versions, meaning we can deal with them bit by bit (and get
the benefit of nicer code early) instead of having them come through in
a huge wave on a major update.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,3 +5,6 @@ inherit_gem:
 inherit_mode:
   merge:
     - Exclude
+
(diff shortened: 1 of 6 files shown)


cpjmcquillan (approved):
No objections from me

cesidio (approved):

PR #2336, opened 2020-11-12
[TEVA-1532] Update monitoring

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1532

## Changes in this PR:
Terraform monitoring code updated and now deployed continuously by the pipeline to avoid any risk of drift

The pipeline was run successfully here: https://github.com/DFE-Digital/teacher-vacancy-service/runs/1390541883?check_suite_focus=true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,15 +113,19 @@ terraform-common-apply: terraform-common-init ## make terraform-common-apply
 
 ##@ terraform/monitoring. Deploys grafana, prometheus monitoring on Gov.UK PaaS
 
-.PHONY: monitoring-plan
-monitoring-plan: ## Plan changes in terraform/monitoring
-		TF_WORKSPACE=monitoring terraform init -upgrade=true -input=false terraform/monitoring \
-		&& bin/run-in-env -e monitoring -o tf_subshell -- terraform plan -input=false terraform/monitoring
+.PHONY: terraform-monitoring-init
+terraform-monitoring-init:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -113,15 +113,19 @@ terraform-common-apply: terraform-common-init ## make terraform-common-apply
 
 ##@ terraform/monitoring. Deploys grafana, prometheus monitoring on Gov.UK PaaS
 
+.PHONY: monitoring-init
+monitoring-init:
+		$(if $(passcode), , $(error Missing environment variable "passcode"))
+		$(eval export TF_VAR_paas_sso_passcode=$(passcode))
+		terraform init -upgrade=true -input=false terraform/monitoring
```
nickromneydfe, line 120: In [TEVA-1534] I added `-reconfigure` to common.
Would line 120 be improved by being 
```
terraform init -upgrade=true -input=false -reconfigure terraform/monitoring
```

By moving out the init to its own action, you seem to have dropped the workspace (which in `monitoring-plan` and `monitoring-apply` was being set with `TF_WORKSPACE=monitoring`

I suggest a line 121, like `terraform-common-init`, of 
```
terraform workspace select monitoring terraform/monitoring
```

saliceti: As agreed with @nickromneydfe , we're not going to manage workspaces here. We will remove them entirely in a separate story to avoid confusion.

nickromneydfe (approved):
Looks good to me.

PR #2340, opened 2020-11-13
TEVA-1536 GH Action to refresh environment variables in apps

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1536

## Changes in this PR:

- Added Docker image tag as Terraform output
- Created `deploy-manual` workflow to receive tag and environment as inputs, and deploy to an environment
- Created `refresh` workflow to determine current docker image tag on an environment, and redeploy, which should have the effect of refreshing just the environment variables.
- Added `eu-west-2` region to AWS environment variables for `run-in-env` script which retrieves secrets from Parameter Store

### DIFF
--- a/terraform/app/modules/paas/outputs.tf
+++ b/terraform/app/modules/paas/outputs.tf
@@ -1,0 +1,4 @@
+output docker_image {
+  value       = cloudfoundry_app.web_app.docker_image
+  description = "Docker image - repository:tag"
+}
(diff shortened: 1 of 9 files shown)


terraform/app/modules/paas/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output docker_image_tag {
+  value = cloudfoundry_app.web_app.docker_image
```
vigneshmsft, line 2: won't this return the full image name and not just the sha?
the deploy-pipeline is just expecting a sha?

nickromneydfe: Thanks. I've made it clearer that `docker_image` consists of `repository:tag`

saliceti (approved):

PR #2349, opened 2020-11-16
Rename teacher-vacancy-service to teaching-vacancies

cesidio:


alexbowen:
why is this being changed, for consistency?

cesidio:
> why is this being changed, for consistency?

Yes, the product is called Teaching Vacancies, and we should have one name consistently everywhere

cpjmcquillan (approved):

PR #2350, opened 2020-11-16
Add 'git secrets'

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1593

## Changes in this PR:

- The idea is to add another line of defence for these secrets, as we have twice lost time this year to accidental publication of secrets.
- Added [git secrets](https://github.com/awslabs/git-secrets). The patterns were based on the contents of `.env` and nothing else.
- Try setting this up locally and let me know if you have any issues.
- If a pattern is matched when committing, git secrets offers the option to ignore the warning (temporarily or permanently).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ make -s local print-env > .env
 
 To run the command above you need [AWS credentials](#aws-credentials).
 
+[Git secrets](/documentation/secrets-detection.md) offers an easy way to defend against accidentally publishing these secrets.
+
 ### Install dependencies
 
 Install Ruby dependency libraries:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -91,6 +91,27 @@ make -s local print-env > .env
 
 To run the command above you need [AWS credentials](#aws-credentials).
 
+#### Git secrets
```
cesidio, line 94: I think that the main README file should contain the minimal information to let you start develop. This information can be in a separate Markdown document in the `/documents` folder

david-mears-dfe: Fair point. I would want to ensure that all new joiners do set this up though. So shall I add it to the documentation folder and then link it from the README?

cesidio: Yes I think it is better in that folder with just a link from the README

documentation/git-secrets.md
```diff
@@ -0,0 +1,20 @@
+# Git secrets
```
cesidio, line 1: I would call this document by what it does more than what tool it uses.

for instance instead of DfE Sign-in we have Authentication
instad of Algolia we have Search

This could be called secrets detection, or something like that

cesidio (approved):

PR #2352, opened 2020-11-16
MOST WICKED PR EVER

cpjmcquillan:
## Motivation for using wicked gem
- The name
- Massive simplification of the create a job journey for hiring staff (from a code perspective). This is especially important I think, as it should help us define and refine a pattern for a multi-step process in our service (before we implement applications for jobseekers)
- 1 controller instead of 9 different ones (fact-check this)
- We don't use the session for things unnecessarily*
- We've removed all validations from the model*

*Could have been achieved without this gem (but they haven't been until now)

## Changes in this PR
- Use wicked
- Refactor the entire create a job listing journey (routes, controllers, views et al)
- Remove validations from the model
- Add job to remove stale vacancies (those created that don't have a job title)


app/controllers/concerns/create_a_job_concerns.rb
```diff
@@ -0,0 +1,116 @@
+module CreateAJobConcerns
```
cesidio, line 1: Usually these concerns are called with something `able`, like `stepable` maybe?

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -62,6 +62,11 @@ def summary
 
 private
 
+  def all_steps_valid?
+    step_valid?(JobDetailsForm) && step_valid?(PayPackageForm) && step_valid?(ImportantDatesForm) &&
+      step_valid?(SupportingDocumentsForm) && step_valid?(ApplyingForTheJobForm) && step_valid?(JobSummaryForm)
+  end
+
   def step_valid?(step_form)
     validation = step_form.new(@vacancy.attributes)
```
cesidio, line 71: this should be called `form`

app/controllers/concerns/create_a_job_concerns.rb
cpjmcquillan: Createabjobable 😂

config/routes.rb
```diff
@@ -76,77 +76,21 @@
     scope constraints: { type: /(published|draft|pending|expired|awaiting_feedback)/ } do
       get "jobs(/:type)", to: "hiring_staff/organisations#show", defaults: { type: :published }, as: :jobs_with_type
     end
-    resources :jobs, only: %i[new edit destroy delete show], controller: "hiring_staff/vacancies" do
-      get "review",
-          defaults: { create_step: 8, step_title: I18n.t("jobs.review_heading") }
+
+    resources :jobs, only: %i[create edit destroy delete show], controller: "hiring_staff/vacancies" do
+      resources :build, only: %i[show update], controller: "hiring_staff/vacancies/build"
+      resource :documents, only: %i[create destroy show], controller: "hiring_staff/vacancies/documents"
```
cesidio, line 82: why resource is singular here?

app/components/hiring_staff/sidebar_component.html.haml
```diff
@@ -4,15 +4,15 @@
 
 #step-by-step-navigation.app-step-nav.app-step-nav--active
   %ol.app-step-nav__steps
-    - steps_to_display.each do |step|
-      %li.app-step-nav__step.js-step{ class: "#{set_active_step_class(step[:number])} #{set_visited_step_class(step[:number], @vacancy)}", 'aria-current': 'step', data: { show: '' } }
+    - steps_to_display.uniq { |_, v| v }.to_h.each do |step_name, step_values|
```
csutter, line 7: Wouldn't this do the trick here as it never uses `step_name`?
```ruby
steps_to_display.values.each do |step_values|
```

app/controllers/hiring_staff/vacancies/build_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+    publish_in_past = @vacancy.published? && @vacancy.reload.publish_on.past?
+    delete_publish_on_params if @vacancy.published? && @vacancy.reload.publish_on.past?
+    dates_to_convert = publish_in_past ? %i[starts_on expires_on] : %i[starts_on publish_on expires_on]
+    @date_errors = convert_multiparameter_attributes_to_dates(:important_dates_form, dates_to_convert)
+  end
+
+  def delete_publish_on_params
+    params.require(:important_dates_form).delete("publish_on(3i)")
+    params.require(:important_dates_form).delete("publish_on(2i)")
+    params.require(:important_dates_form).delete("publish_on(1i)")
```
csutter, line 86: Might be a bit tidier with [`ActionController::Parameters#extract!`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-extract-21)?

app/controllers/hiring_staff/vacancies_controller.rb
```diff
@@ -74,47 +67,46 @@ def redirect_if_published
...
-  end
-
   def set_completed_step
-    @vacancy.update(completed_step: current_step)
+    @vacancy.update(completed_step: current_step_number)
   end
 
   def update_vacancy_state
-    state = if params[:edit_draft] == "true" || @vacancy&.state == "edit"
+    state = if params[:edit_draft] == "true" || @vacancy.state == "edit"
```
csutter, line 94: I know it's not part of this PR, but why is this not an enum? 😕

config/routes.rb
cpjmcquillan: Truthfully I don't know (they were always like that = copy 🍝 ) But after reading [this](https://rubyinrails.com/2019/04/16/rails-routes-difference-between-resource-and-resources/) I still think the singular resource is suitable here

app/components/hiring_staff/sidebar_component.html.haml
cpjmcquillan: Classic case of completely changing something and never revisiting the other thing once I changed it...

app/controllers/hiring_staff/vacancies_controller.rb
cpjmcquillan: Mainly because enums weren't used as much across the codebase when this was hacked together (for GA purposes), so there's no good reason it shouldn't be

app/models/concerns/vacancy_important_date_validations.rb
```diff
@@ -3,7 +3,7 @@ module VacancyImportantDateValidations
 
   included do
     validates :publish_on, presence: true, unless: proc { |v| v.status == "published" }
-    validate :publish_on_must_not_be_before_today
+    validate :publish_on_must_not_be_before_today, unless: proc { |v| v.status == "published" }
```
csutter, line 6:
```ruby
validate :publish_on_must_not_be_before_today, unless:  :published?
```

db/migrate/20201111115249_remove_vacancies_constraints.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveVacanciesConstraints < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :vacancies, :job_title, true
```
csutter, line 3: When would we want to have a `NULL` job title?

cpjmcquillan: When we have to create a vacancy without one 😞 . Designers!!! 😡 😆

app/controllers/hiring_staff/vacancies/application_controller.rb
```diff
@@ -116,38 +67,29 @@ def remove_google_index(job)
...
-    @form = form_class.new(form_params)
+  def review_path_with_errors(vacancy)
+    organisation_job_review_path(job_id: vacancy.id, anchor: "errors", source: "publish")
   end
 
-  def set_up_url
-    @form_submission_url_method = @vacancy&.persisted? ? "patch" : "post"
-    @form_submission_url = form_submission_path(@vacancy&.id)
+  def set_vacancy
+    if params[:job_id]
```
cesidio, line 82: This could just be:
```ruby
@vacancy = current_organisation.all_vacancies.find(params[:job_id].presence || params[:id].presence)
```

app/controllers/hiring_staff/vacancies/build_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+    return unless step == :important_dates
+
+    publish_in_past = @vacancy.published? && @vacancy.reload.publish_on.past?
+    delete_publish_on_params if @vacancy.published? && @vacancy.reload.publish_on.past?
+    dates_to_convert = publish_in_past ? %i[starts_on expires_on] : %i[starts_on publish_on expires_on]
+    @date_errors = convert_multiparameter_attributes_to_dates(:important_dates_form, dates_to_convert)
+  end
+
+  def delete_publish_on_params
+    params.require(:important_dates_form).extract!("publish_on(3i)")
```
cesidio, line 84: Can't it be done in one line?

```ruby
params.require(:important_dates_form).extract!("publish_on(3i)", "publish_on(2i)", "publish_on(1i)")
```

app/views/hiring_staff/vacancies/build/job_summary.html.haml
```diff
@@ -0,0 +1,30 @@
+- content_for :page_title_prefix, page_title_prefix(@vacancy, @form, t("jobs.job_summary"))
+= render HiringStaff::VacancyFormPageHeadingComponent.new(@vacancy)
+
+- unless @vacancy.state == 'create'
```
cesidio, line 4: not a big fan of unless else

cesidio (approved):
Best refactoring since I joined the team

PR #2361, opened 2020-11-18
[TEVA-1459] Part 1 - Save establishment status when importing schools data

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1459

## Changes in this PR
- Save establishment status name as a string on the organisations table when importing schools from GIAS


david-mears-dfe (commented):
It seems like exactly the kind of thing an integer enum would be good at. That would give us something like the not_closed scope for free (`organisation.closed?`). But I guess we can't rely on GIAS' spelling remaining constant over time?

cesidio (approved):

PR #2363, opened 2020-11-18
[TEVA-1459] Part 2 - Filter closed schools

cpjmcquillan:
**TODO** 
- [x] Merge Part 1
- [x] Run Import schools data job

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1459

## Changes in this PR
- Add not_closed scope for organisations
- Only show not_closed schools in managed organisations form
- Only show not_closed schools in dashboard filters form
- Only show not_closed schools in school options in the schools step of the create a job journey

### DIFF
--- a/spec/components/hiring_staff/vacancies_component_spec.rb
+++ b/spec/components/hiring_staff/vacancies_component_spec.rb
@@ -23,43 +23,61 @@ RSpec.describe HiringStaff::VacanciesComponent, type: :component do
   end
 
   context "when organisation has active vacancies" do
-    let(:organisation) { create(:school, name: "A school with jobs") }
-    let(:vacancy) { create(:vacancy, :published) }
+    context "when organisation is a school" do
+      let(:organisation) { create(:school, name: "A school with jobs") }
+      let(:vacancy) { create(:vacancy, :published) }
 
-    before { vacancy.organisation_vacancies.create(organisation: organisation) }
+      before { vacancy.organisation_vacancies.create(organisation: organisation) }
 
-    let!(:inline_component) { render_inline(subject) }
+      let!(:inline_component) { render_inline(subject) }
 
-    it "renders the vacancies component" do
-      expect(inline_component.css(".govuk-tabs").to_html).not_to be_blank
-    end
+      it "renders the vacancies component" do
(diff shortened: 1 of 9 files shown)


cesidio (approved):

spec/components/hiring_staff/vacancies_component_spec.rb
```diff
@@ -76,13 +94,43 @@
...
+      it "does not render the closed school as a filter option" do
+        expect(inline_component.css(".new_managed_organisations_form").to_html).not_to include("Closed school")
+      end
     end
 
     context "when the organisation is a local authority" do
       let(:organisation) { create(:local_authority) }
+      let(:open_school) { create(:school, name: "Open school") }
+      let(:closed_school) { create(:school, :closed, name: "Closed school") }
       let!(:vacancy) { create(:vacancy, :published, :at_one_school) }
```
david-mears-dfe, line 111: Possibly change the vacancy to be `at_multiple_schools`, since the test is currently testing an impossible state (a listing associated directly to an LA)

```ruby
# line 117
vacancy.organisation_vacancies.create(organisation: organisation)
```

david-mears-dfe (commented):
I worry about, in the future, naively typing `organisation.schools` and not remembering that we have to exclude the closed ones. We could either:

1) Not store closed schools in the database in the first place
2) Stop creating associations between school groups and schools when the schools are closed
3) Catch the `schools` method on its way and tell it to also check for the schools being not closed
4) Accept the risk.

cpjmcquillan:
> I worry about, in the future, naively typing `organisation.schools` and not remembering that we have to exclude the closed ones. We could either:
> 
> 1. Not store closed schools in the database in the first place
> 2. Stop creating associations between school groups and schools when the schools are closed
> 3. Catch the `schools` method on its way and tell it to also check for the schools being not closed
> 4. Accept the risk.

- 3 isn't possible since `.schools` refers to an organisation having many schools
- 1 and 2 are a combined possibility and - provided we definitely don't need the information - the best option. 

Depending on how important the information is/why it's useful, this will probably fall into 'decoupling data for analytics and data for the application', which is I think out of scope of the ticket, so for now we can do 4

PR #2365, opened 2020-11-18
[TEVA-1470] Add timestamps to JobAlertFeedbacks table

david-mears-dfe:
Because records already exist in this table, we need to give them a default created_at and updated_at.

In this migration, after applying that default to all existing records, we then reinstate the default 'null: false' restriction which raises an error if you try to create a feedback without one of these timestamps.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1470

### DIFF
--- a/db/migrate/20201118111330_add_timestamps_to_job_alert_feedbacks.rb
+++ b/db/migrate/20201118111330_add_timestamps_to_job_alert_feedbacks.rb
@@ -1,0 +1,7 @@
+class AddTimestampsToJobAlertFeedbacks < ActiveRecord::Migration[6.0]
+  def change
+    add_timestamps :job_alert_feedbacks, default: Time.new(2020, 10, 16)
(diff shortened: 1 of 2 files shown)


db/migrate/20201118111330_add_timestamps_to_job_alert_feedbacks.rb
```diff
@@ -0,0 +1,7 @@
+class AddTimestampsToJobAlertFeedbacks < ActiveRecord::Migration[6.0]
+  def change
+    add_timestamps :job_alert_feedbacks, default: DateTime.new(2020, 1, 1)
```
stevenleggdfe, line 3: @david-mears-dfe what would you think of setting this to the launch date of the feedback feature in October instead? Might make the URs' feedback dashboard easier to navigate.

david-mears-dfe: If it makes the dashboard easier to use, then yes let's.

cpjmcquillan (approved):

PR #2366, opened 2020-11-18
TEVA-1604 GH Actions use deploy app workflow

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1604

## Changes in this PR:

- Use `deploy_app` workflow from other workflows (`deploy`, `deploy_branch`, `review`)
- Access variables with `${{ github.X }}` not `$GITHUB_X`
- Access variables with `${{ env.VAR }}` not `${VAR}`
- Get branch with `${{ github.head_ref }}` instead of `$(echo ${GITHUB_REF} | cut -d/ -f3)`
- Remove additional outputs where these were already available as environment variables

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Set environment variables
       id: set_env_vars
       run: |
-        TAG=${GITHUB_SHA}
+        TAG=${{ github.sha }}
         echo "TAG=${TAG}" >> $GITHUB_ENV
 
     - name: Set up Ruby
@@ -95,7 +95,7 @@ jobs:
       with:
         workflow: Deploy App to Environment # Workflow name
         token: ${{ secrets.PERSONAL_TOKEN }}
-        inputs: '{"environment": "staging", "tag": "${TAG}"}'
+        inputs: '{"environment": "staging", "tag": "${{ env.TAG }}"}'
 
     - name: Wait for Deploy App Workflow for staging
       id: wait_for_deploy_app_staging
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     - name: Set environment variables
       id: set_env_vars
       run: |
-        TAG=${GITHUB_SHA}
+        TAG=${{ github.sha }}
         echo "TAG=${TAG}" >> $GITHUB_ENV
```
vigneshmsft, line 44:
```suggestion
        echo "TAG=${{ github.sha }}" >> $GITHUB_ENV
```

.github/workflows/deploy.yml
```diff
@@ -120,50 +125,55 @@ jobs:
...
+      if: ${{ steps.wait_for_deploy_app_production.outputs.conclusion == 'failure' }} && github.ref == 'refs/heads/master'
       uses: rtCamp/action-slack-notify@v2.1.1
       env:
         SLACK_CHANNEL: twd_tv_dev
         SLACK_USERNAME: CI Deployment
         SLACK_ICON_EMOJI: ':cry:'
         SLACK_TITLE: Deployment failure
-        SLACK_MESSAGE: 'Deployment to production failed'
+        SLACK_MESSAGE: 'Deployment of Docker tag ${{ env.TAG }} to production - failure'
         SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
```
vigneshmsft, line 179: this can be combined
```
if: ${{ always() && github.ref == 'refs/heads/master' }}
...
# and use ${{ job.status }} Possible values are success, failure, or cancelled.
SLACK_TITLE: Deployment ${{ job.status }}
```
and same in other workflows as well

.github/workflows/deploy.yml
nickromneydfe: Thanks. I've folded that in.

.github/workflows/deploy.yml
nickromneydfe: Thanks. I've combined the notification step in 4 workflows.

saliceti (approved):

vigneshmsft (approved):

PR #2367, opened 2020-11-18
Remove protocol from URL generation

cesidio:
We used to run development environment in HTTPS, so we had to specify the protocol to test URL generation. Now it just uses the correct protocol, HTTP for development, HTTPS for production environments


david-mears-dfe:
Why?

cesidio:
> Why?

added a description

david-mears-dfe (approved):

PR #2368, opened 2020-11-18
Bigquery string functions

stevenleggdfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1424

## Changes in this PR:

- Refactor queries that process data from Cloudfront by moving string processing functions into 'user defined functions'. This significantly reduces duplication, making these functions easier to maintain. For example - we'll be able to refine how we determine device categories and mediums in one place instead of four.
- Reduce number of subqueries in some of these cases to simplify things.
- Fix minor bug in the 'from job alert' field - it used to be NULL if utm_medium was NULL. Now it's FALSE if utm_medium is NULL.

### DIFF
--- a/bigquery/user_defined_functions/string-functions.sql
+++ b/bigquery/user_defined_functions/string-functions.sql
@@ -1,0 +1,109 @@
+# Contains useful string processing functions that can be reused across multiple scheduled queries and views in BigQuery.
+# This script must be run in BigQuery for these functions to become available, and re-run to overwrite the available functions with any new version.
+  
+# Take a client user agent from a Cloudfront server log and use it to categorise that traffic into a device category
+CREATE OR REPLACE FUNCTION
+  `teacher-vacancy-service.production_dataset.convert_client_user_agent_to_device_category`(client_user_agent STRING)
+  RETURNS STRING AS (
+    CASE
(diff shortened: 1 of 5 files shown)


bigquery/user_defined_functions/string-functions.sql
```diff
@@ -0,0 +1,109 @@
+# Contains useful string processing functions that can be reused across multiple scheduled queries and views in BigQuery.
+# This script must be run in BigQuery for these functions to become available, and re-run to overwrite the available functions with any new version.
+  
+# Take a client user agent from a Cloudfront server log and use it to categorise that traffic into a device category
+CREATE OR REPLACE FUNCTION
+  `teacher-vacancy-service.production_dataset.convert_client_user_agent_to_device_category`(client_user_agent STRING)
+  RETURNS STRING AS (
+    CASE
+      WHEN LOWER(client_user_agent) LIKE "%bot%" OR LOWER(client_user_agent) LIKE "%http%" OR LOWER(client_user_agent) LIKE "%python%" OR LOWER(client_user_agent) LIKE "%scan%" OR LOWER(client_user_agent) LIKE "%check%" OR LOWER(client_user_agent) LIKE "%spider%" OR LOWER(client_user_agent) LIKE "%curl%" OR LOWER(client_user_agent) LIKE "%trend%" OR LOWER(client_user_agent) LIKE "%fetch%" THEN "bot"
```
david-mears-dfe, line 9: Could you split the bot line out onto separate lines like the mobile line?

david-mears-dfe: Helps my reading brain make an analogy between the two if they look similar

stevenleggdfe: Done!

david-mears-dfe (approved):
I love a good tidy up.

PR #2369, opened 2020-11-18
Foundations for Devise-based jobseeker accounts

csutter:
- Add `devise` gem to application
- Create Devise-protected `Jobseeker` account model and configure with
  some initial settings
- Add feature flag and enable in dev and review environments

_Output of our mob session this afternoon. Note that this is just the very basic setup - a lot of plumbing remains to be done._

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1506


cesidio (approved):

config/initializers/devise.rb
```diff
@@ -0,0 +1,311 @@
...
+  # If true, requires any email changes to be confirmed (exactly the same way as
+  # initial account confirmation) to be applied. Requires additional unconfirmed_email
+  # db field (see migrations). Until confirmed, new email is stored in
+  # unconfirmed_email column, and copied to email column on successful confirmation.
+  config.reconfirmable = true
+
+  # Defines which key will be used when confirming an account
+  # config.confirmation_keys = [:email]
+
+  # ==> Configuration for :rememberable
```
david-mears-dfe, line 165: Can you remove this section, because we aren't going to use it?

config/locales/devise.en.yml
```diff
@@ -0,0 +1,65 @@
...
+      confirmation_instructions:
+        subject: "Confirmation instructions"
+      reset_password_instructions:
+        subject: "Reset password instructions"
+      unlock_instructions:
+        subject: "Unlock instructions"
+      email_changed:
+        subject: "Email Changed"
+      password_change:
+        subject: "Password Changed"
```
david-mears-dfe, line 29: I Don't Like Boomer Case.

cpjmcquillan: I presume content will go to town on this anyway. However, I would very much love if we removed all the unnecessary double quotes

spec/factories/jobseekers.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
+  factory :jobseeker do
```
cpjmcquillan, line 2: Do we want to flesh this out?

spec/models/jobseeker_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe Jobseeker, type: :model do
```
cpjmcquillan, line 3: And do we want to add anything here yet?

cpjmcquillan (approved):

spec/models/jobseeker_spec.rb
csutter: I don't think we need to add anything just yet, as the model is just empty (other than including Devise functionality, which I think we don't need to test at this low a level, instead once we have a rudimentary "dashboard" we'll build system tests for it).

config/initializers/devise.rb
csutter: I think if we leave the other commented out bits in there, it might make sense to keep this one too? Plus we don't know yet if we might change our minds on `rememberable` later...

PR #2370, opened 2020-11-18
Refactor DSI API consumption when signing in

david-mears-dfe:
Splitted out from https://github.com/DFE-Digital/teaching-vacancies/pull/2358 and refactored further

It's more robust against edge-cases to check for e.g. the LA category '002' in the DSI API response, than to assume that an organisation is LA if it's neither trust or school, or to assume that if the JSON contains a URN then it's a school.

I asked DSI for the mapping from id to category and submitted a PR https://github.com/DFE-Digital/login.dfe.public-api/pull/60 adding that to the API README.

### DIFF
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -68,18 +68,37 @@ private
   end
 
   def school_urn
+    return "" unless user_category == :school
+
     auth_hash.dig("extra", "raw_info", "organisation", "urn") || ""
   end
 
   def trust_uid
+    return "" unless user_category == :multi_academy_trust
+
     auth_hash.dig("extra", "raw_info", "organisation", "uid") || ""
   end
 
   def local_authority_code
+    return unless LocalAuthorityAccessFeature.enabled?
+    return "" unless user_category == :local_authority
+
     # All organisations have an establishmentNumber, but we only want this for identifying LAs by.
-    # Assume that if and only if the organisation has no URN or UID, it is a Local Authority.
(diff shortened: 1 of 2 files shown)


david-mears-dfe:
For reference, [here](https://github.com/david-mears-dfe/login.dfe.public-api/blob/b1b97d3d394e6ecc8a69bf7e5accdc3b82b94b34/README.md#how-do-ids-map-to-categories-and-types) are the mappings from id to organisation category.

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -68,18 +68,37 @@ def identifier
...
 
   def trust_uid
+    return "" unless user_category == :multi_academy_trust
+
     auth_hash.dig("extra", "raw_info", "organisation", "uid") || ""
   end
 
   def local_authority_code
+    return "" unless user_category == :local_authority
+    return unless LocalAuthorityAccessFeature.enabled?
```
cpjmcquillan, line 84: Maybe this statement should go first

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -68,18 +68,37 @@ def identifier
...
   def school_urn
+    return "" unless user_category == :school
+
     auth_hash.dig("extra", "raw_info", "organisation", "urn") || ""
   end
 
   def trust_uid
+    return "" unless user_category == :multi_academy_trust
+
     auth_hash.dig("extra", "raw_info", "organisation", "uid") || ""
```
cpjmcquillan, line 79: Do we need `|| ""` if we have the previous line?

cpjmcquillan (approved):

PR #2376, opened 2020-11-19
Alert team to failed table exports to BigQuery

david-mears-dfe:
With a little tweaking to the Rollbar-Slack integration this should alert us.

### DIFF
--- a/lib/export_tables_to_big_query.rb
+++ b/lib/export_tables_to_big_query.rb
@@ -76,7 +76,7 @@ class ExportTablesToBigQuery
     rescue StandardError => e
       # If any table causes an uncaught error, no data from any later table is sent.
       # Catch errors and skip the failing tables
-      # TODO: alert the team (devs + PA) to any failing tables.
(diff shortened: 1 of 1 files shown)


lib/export_tables_to_big_query.rb
```diff
@@ -76,7 +76,7 @@ def run!
     rescue StandardError => e
       # If any table causes an uncaught error, no data from any later table is sent.
       # Catch errors and skip the failing tables
-      # TODO: alert the team (devs + PA) to any failing tables.
+      Rollbar.error("The table '#{table}' failed during the BigQuery export job.", details: e)
```
cpjmcquillan, line 79: "The '#{table}' table" would probably be easier to read for my small brain

cpjmcquillan (approved):

lib/export_tables_to_big_query.rb
david-mears-dfe: Table table. (Table)

PR #2379, opened 2020-11-19
Update local_authorities.yml

milimaldedfe:
Added Phase 2 LAs to the allowed list so they can access the LA dashboard.


cpjmcquillan (approved):
They're even in order 🤩

PR #2380, opened 2020-11-19
[TEVA-1458] Sign in users with trust or LA instead of school

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1458

### DIFF
--- a/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
@@ -51,6 +51,7 @@ private
       multiple_organisations: authorisation_permissions.many_organisations?,
       id_token: id_token,
     )
+    use_school_group_if_available
     Rails.logger.info(updated_session_details)
     audit_successful_authorisation
   end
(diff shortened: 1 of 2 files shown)


david-mears-dfe:
Are you doing LAs in a separate PR?

cesidio:
> Are you doing LAs in a separate PR?

Done in this PR

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
```diff
@@ -138,4 +139,20 @@ def allowed_user?
     school_urn.present? || trust_uid.present? || (local_authority_code.present? &&
       (Rails.env.staging? || Rails.env.development? || ALLOWED_LOCAL_AUTHORITIES.include?(local_authority_code)))
   end
+
+  def use_school_group_if_available
+    user = User.find_by(email: identifier)
+    user_trusts = user&.dsi_data&.fetch("trust_uids", [])
+    user_local_authorities = user&.dsi_data&.fetch("la_codes", [])
+    return unless user_trusts&.any? || user_local_authorities&.any?
```
cpjmcquillan, line 147: My python brain is screaming that `return if ...none?` is clearer, but old habits die hard

spec/system/hiring_staff_can_sign_in_with_dfe_spec.rb
```diff
@@ -126,6 +126,41 @@
...
+
+      visit root_path
+      sign_in_user
+    end
+
+    context "with trust" do
+      let(:school_group) { create(:trust, uid: "14323") }
+
+      scenario "it associates the user with the trust instead of the school" do
+        expect(current_path).to eql(organisation_managed_organisations_path)
```
cpjmcquillan, line 151: I quite like `expect(page).to have_current_path(...)`

cpjmcquillan: Also, is it worth checking the heading is for the Trust/LA respectively? (I think possibly).

And I also thought about whether we should test LA feature enabled/disabled? (I thought probably not because the flag will get removed soon)

cpjmcquillan (approved):

spec/system/hiring_staff_can_sign_in_with_dfe_spec.rb
cesidio: first one is a bit more concise though:
`expect(current_path).to eql(organisation_managed_organisations_path)`
vs
`expect(page).to have_current_path(organisation_managed_organisations_path)`

cesidio: > Also, is it worth checking the heading is for the Trust/LA respectively? (I think possibly).
added

app/controllers/hiring_staff/sign_in/dfe/sessions_controller.rb
david-mears-dfe: I think `if there are no X and no Y` is clearer in English than `unless there are any X or any Y`

PR #2407, opened 2020-11-24
Strip whitespaces when looking up LocationPolygons

david-mears-dfe:
Currently, 'london' searches a polygon and 'london ' searches a point.

Compare "90 jobs found near 'London '" and "134 jobs found in 'London'". The former fails to use a polygon because of the trailing whitespace.

This has also been messing up searches in Birmingham (127 counts of searching with a whitespace), Manchester, Liverpool, Sheffield, etc.

In our sample, there were 168 searches for "london " and 6376 for "london"

### DIFF
--- a/app/models/location_polygon.rb
+++ b/app/models/location_polygon.rb
@@ -5,6 +5,7 @@ class LocationPolygon < ApplicationRecord
   scope :regions, -> { where(location_type: "regions") }
 
   def self.with_name(location)
+    location.strip!
     find_by(name: (MAPPED_LOCATIONS[location.downcase].presence || location).downcase)
   end
 end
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):

app/models/location_polygon.rb
```diff
@@ -5,6 +5,7 @@ class LocationPolygon < ApplicationRecord
   scope :regions, -> { where(location_type: "regions") }
 
   def self.with_name(location)
+    location.strip!
```
alexbowen, line 8: wouldnt the form model be a better place to do this?

david-mears-dfe: Yes it would

david-mears-dfe:
Reimplemented in https://github.com/DFE-Digital/teaching-vacancies/pull/2433

PR #2409, opened 2020-11-24
Rename concept of "user" to "publisher"

csutter:
We now have multiple distinct types of users, so the original naming of
hiring staff members as "user" is confusing.

This renames the `User` model to `Publisher` and refactors most
occurrences of "user" in the codebase to "publisher", with the exception
of where they specifically refer to DSI users.

### DIFF
--- a/app/controllers/hiring_staff/base_controller.rb
+++ b/app/controllers/hiring_staff/base_controller.rb
@@ -20,32 +20,32 @@ class HiringStaff::BaseController < ApplicationController
   end
 
   def check_terms_and_conditions
-    redirect_to terms_and_conditions_path unless current_user&.accepted_terms_and_conditions?
+    redirect_to terms_and_conditions_path unless current_publisher&.accepted_terms_and_conditions?
   end
 
-  def current_user
+  def current_publisher
     return if current_session_id.blank?
 
-    @current_user ||= User.find_or_create_by(oid: current_session_id)
+    @current_publisher ||= Publisher.find_or_create_by(oid: current_session_id)
   end
 
-  def current_user_preferences
+  def current_publisher_preferences
     return unless current_organisation.is_a?(SchoolGroup)
 
-    UserPreference.find_by(user_id: current_user.id, school_group_id: current_organisation.id)
+    PublisherPreference.find_by(publisher_id: current_publisher.id, school_group_id: current_organisation.id)
   end
 
   def check_user_last_activity_at
-    return redirect_to logout_endpoint if current_user&.last_activity_at.blank?
+    return redirect_to logout_endpoint if current_publisher&.last_activity_at.blank?
 
-    if Time.current > (current_user.last_activity_at + TIMEOUT_PERIOD)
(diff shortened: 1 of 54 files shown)


alexbowen:
this sounds sensible but does it not mean that really should rename hiringstaff which is used in file/folder naming a lot to publisher too? 'We now have multiple distinct types of users, so the original naming of
hiring staff members as "user" is confusing.' - is the naming of hiring staff memebers as publishers not also confusing? if you get my drift

app/controllers/hiring_staff/base_controller.rb
```diff
@@ -20,32 +20,32 @@ def check_session
...
-    redirect_to terms_and_conditions_path unless current_user&.accepted_terms_and_conditions?
+    redirect_to terms_and_conditions_path unless current_publisher&.accepted_terms_and_conditions?
   end
 
-  def current_user
+  def current_publisher
     return if current_session_id.blank?
 
-    @current_user ||= User.find_or_create_by(oid: current_session_id)
+    @current_publisher ||= Publisher.find_or_create_by(oid: current_session_id)
```
david-mears-dfe, line 29: If I am logged in as a jobseeker using the session id, then this line will create a Publisher. How can we ensure that ids of Publishers and Jobseekers do not collide?

csutter: > this sounds ok but does it not mean that really should rename hiringstaff which is used in file/folder naming a lot to publisher too? 'We now have multiple distinct types of users, so the original naming of
> hiring staff members as "user" is confusing.' - is the naming of hiring staff members as publishers not also confusing? if you get my drift. maybe im overthinking :|

Excellent point @alexbowen – I was thinking of renaming that too, but possibly as a separate follow-on PR so that this one doesn't get too overwhelming! 👍

csutter:
> this sounds ok but does it not mean that really should rename hiringstaff which is used in file/folder naming a lot to publisher too? 'We now have multiple distinct types of users, so the original naming of
> hiring staff members as "user" is confusing.' - is the naming of hiring staff members as publishers not also confusing? if you get my drift. maybe im overthinking :|

Excellent point @alexbowen – I was thinking of renaming that too, but possibly as a separate follow-on PR so that this one doesn't get too overwhelming! 👍

alexbowen:
> > this sounds ok but does it not mean that really should rename hiringstaff which is used in file/folder naming a lot to publisher too? 'We now have multiple distinct types of users, so the original naming of
> > hiring staff members as "user" is confusing.' - is the naming of hiring staff members as publishers not also confusing? if you get my drift. maybe im overthinking :|
> 
> Excellent point @alexbowen – I was thinking of renaming that too, but possibly as a separate follow-on PR so that this one doesn't get too overwhelming! 👍

absolutley. but yes i like the work publisher rather than user

alexbowen (approved):

cesidio (approved):
Very nice! Much clearer the distinction of publisher/jobseeker now

PR #2412, opened 2020-11-24
[TEVA-1604] deploy_branch workflow - fix branch name

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1604

## Changes in this PR:

Derive branch names from GITHUB_REF, using additional pattern `#refs/heads/` which will still work even on longer branch names (such as `feature/foo` rather than simply `staging` or `dev`)


csutter (approved):
🎉

PR #2417, opened 2020-11-24
Add location polygon usage counter

david-mears-dfe:
This is to help us to detect when there is something wrong with
our polygons. If a polygon is never or very rarely used - which
here means 'searched in' - then one possible explanation is that
it is misnamed. For example, we map 'herefordshire' to our polygon
called 'herefordshire, county of' because no human would ever
choose to input the latter when they mean 'herefordshire'.

To find issues it might be clever to compare the analytics of
which location search terms are commonly used (currently available
on our [dashboard](https://datastudio.google.com/u/1/reporting/1X_lrbWUn7Nw5LZnRWynJKtalYg6-L4Oi/page/vo6kB) page 30) and the actual recorded usage of the
corresponding location polygon.

### DIFF
--- a/app/services/search/location_builder.rb
+++ b/app/services/search/location_builder.rb
@@ -38,7 +38,10 @@ private
 
   def initialize_location_polygon
     @location_polygon = LocationPolygon.with_name(@location_category)
-    @location_polygon_boundary = [@location_polygon.boundary] if @location_polygon.present?
+    if @location_polygon.present?
+      @location_polygon_boundary = [@location_polygon.boundary]
+      @location_polygon.increment!(:usage_count)
+    end
     if location_polygon.nil? && (DOWNCASE_REGIONS + DOWNCASE_COUNTIES).include?(@location_category.downcase)
       # If a location category that we expect to have a polygon actually does not,
       # append the location category to the text search as a fallback.
--- a/db/migrate/20201124133720_add_usage_count_to_location_polygons.rb
+++ b/db/migrate/20201124133720_add_usage_count_to_location_polygons.rb
@@ -1,0 +1,5 @@
+class AddUsageCountToLocationPolygons < ActiveRecord::Migration[6.0]
+  def change
+    add_column :location_polygons, :usage_count, :integer, default: 0
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_24_120747) do
+ActiveRecord::Schema.define(version: 2020_11_24_133720) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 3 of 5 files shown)


cpjmcquillan (commented):
If we export the table to big query, can this analytics information be aggregated by the logs instead? It feels a bit strange to add a column to a table that doesn't serve any other purpose. What does everybody else think? @cesidio @csutter @stevenleggdfe

david-mears-dfe:
> If we export the table to big query, can this analytics information be aggregated by the logs instead? It feels a bit strange to add a column to a table that doesn't serve any other purpose. What do other people think? @cesidio @csutter @stevenleggdfe

Do you mean something like tell Rails to log each time a polygon is used, and then tell Performance Analysis to look for the log output?

cpjmcquillan:
> > If we export the table to big query, can this analytics information be aggregated by the logs instead? It feels a bit strange to add a column to a table that doesn't serve any other purpose. What do other people think? @cesidio @csutter @stevenleggdfe
> 
> Do you mean something like tell Rails to log each time a polygon is used, and then tell Performance Analysis to look for the log output? @cpjmcquillan

That, or infer it by comparing existing search logs with the polygon table and any mappings that take place

david-mears-dfe:
> That, or infer it by comparing existing search logs with the polygon table and any mappings that take place

If I understand this option, it means the monitoring would have to reimplement all the code it is attempting to test. With the coming tranch of location search bug tickets, this mapping might have several components. So to trust the numbers, I would want to either merge this PR, or:

> tell Rails to log each time a polygon is used, and then tell Performance Analysis to look for the log output

csutter:
Just looking through this – do you reckon the triggering an event for location polygon usage using the soon-to-be-merged-I-swear events functionality in #2492 would be a good approach for this?

stevenleggdfe:
> Just looking through this – do you reckon the triggering an event for location polygon usage using the soon-to-be-merged-I-swear events functionality in #2492 would be a good approach for this?

Yes, but I think we might thank ourselves if we come up with a more abstract approach that picks up other things that happen at the same time in the event. Is the event in question viewing a page of site search results? If so we could come up with a site search results event data model that stores polygon information as well as a load of other stuff we'll want to know at the same time.

david-mears-dfe:
> Yes, but I think we might thank ourselves if we come up with a more abstract approach that picks up other things that happen at the same time in the event. Is the event in question viewing a page of site search results? If so we could come up with a site search results event data model that stores polygon information as well as a load of other stuff we'll want to know at the same time.

I think for the original purpose of this PR, what you're talking about (a request) should be considered a separate event (a RequestEvent in #2492's terminology) from what I want to track (a basic back-end Event). These events are related but importantly distinct. We want/need to know how often a location polygon object has in fact been used (which is distinct from how often a polygon ought to have been used), in order to pinpoint issues in our mapping from search parameters to polygon objects.

Let's create a ticket for the request event. I'll update this PR to consume #2492 code once that is merged.

david-mears-dfe:
Ticket: https://dfedigital.atlassian.net/browse/TEVA-1709

PR #2431, opened 2020-11-25
[TEVA-1391] Zero search results creating buffer area

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1391

## Changes in this PR:

- Buffers are now added to location polygons when location polygons are imported

- Users will now see suggestions for expanding their polygon search when they receive zero search results

- Tests have been added for the buffer suggestions builder service

- Other tests have been updated to reflected changes

Screenshot 2020-12-03 at 16 05 21


app/services/import_polygons.rb
```diff
@@ -41,4 +43,57 @@ def location_categories_include?(region_name)
...
+    buffer_distances_in_miles = [5, 10, 15, 20, 25]
+    buffered_boundaries = {}
+    buffer_distances_in_miles.each do |distance|
+      # convert 1D array, to 2D array for arcgis
+      polygon_coords = points.each_slice(2).to_a
+      response = HTTParty.get(buffer_api_endpoint(polygon_coords, convert_miles_to_metres(distance)))
+      buffer_coords = response.dig("geometries", 0, "rings").flatten
+      # algolia requires a 2D containing 1 element with all the coordinates contained within it
+      # so buffer_coords array has been wrapped in another array
+      buffered_boundaries[distance.to_s] = format_buffer_coords_for_algolia(buffer_coords)
```
cpjmcquillan, line 58: You don't need to do this. We store the 1d array and in `app/services/search/location_polygon.rb:41` we put it in the format algolia requires

app/services/import_polygons.rb
```diff
@@ -41,4 +43,57 @@ def location_categories_include?(region_name)
...
+               'f' => 'json' }.to_param
+
+    buffer_api_url = 'https://ons-inspire.esriuk.com/arcgis/rest/services/Utilities/Geometry/GeometryServer/buffer?'
+
+    api_endpoint = buffer_api_url + params
+
+    if api_endpoint.length > 2000
+      # Reduce number of coordinates in order to have fewer than 2000 characters in request url
+      coords = coords.each_with_index
+      .map { |item, index| item if (index % 5).zero? }
```
cpjmcquillan, line 86: I don't feel particularly comfortable downsampling the initial polygon this crudely unless we can clearly show it is acceptable for all cases. Not sure how everyone else feels

app/services/search/search_builder.rb
```diff
@@ -49,7 +48,8 @@ def build_search_replica
   end
 
   def build_suggestions
-    @wider_radius_suggestions = Search::SuggestionsBuilder.new(search_params, @params_hash[:radius]).radius_suggestions
+    return @wider_radius_suggestions = Search::RadiusSuggestionsBuilder.new(search_params, @params_hash[:radius]).radius_suggestions if @point_coordinates.present?
+    @wider_radius_suggestions = Search::BuffersSuggestionsBuilder.new(@params_hash[:location], search_params).buffer_suggestions
```
cpjmcquillan, line 52: We only want to do this if a polygon search has been carried out

app/services/import_polygons.rb
josephhull676: @david-mears-dfe

david-mears-dfe: We tested this by checking two buffered polygons which have high numbers of points in their original boundaries (1000s).

```ruby
# These numbers should be halved in order to get the number of points; these are the numbers of lats and lngs.
pry(main)> LocationPolygon.with_name('portsmouth').boundary.length
=> 1886
[2] pry(main)> LocationPolygon.with_name('wakefield').boundary.length
=> 4410
```

Screenshot 2020-11-25 at 11 28 35

Screenshot 2020-11-25 at 11 32 20

I think that going down from 1000s to 100s of points (or even less) would not result in an appreciable difference in the shape of the buffer.

@josephhull676 Would you be able to tell us what is the length of the buffers displayed above (I forget if they were 1 miles or 5 miles), and what is the shortest length of any buffer? Something like:

```ruby
LocationPolygon.pluck(:buffers) # will give you only the buffers column, as an array.
  .map { |polygon_buffers_list|
    polygon_buffers_list.map(&:length) # or is it :&? will give you a list of the length of each buffer on this polygon
  }.flatten # flatten all the buffer lengths into a 1-D array
  .sort # get them in length order
```

app/services/search/search_builder.rb
```diff
@@ -49,7 +48,11 @@ def build_search_replica
   end
 
   def build_suggestions
-    @wider_radius_suggestions = Search::SuggestionsBuilder.new(search_params, @params_hash[:radius]).radius_suggestions
+    @wider_radius_suggestions = if @point_coordinates.present?
```
david-mears-dfe, line 51: I wonder if having `radius` in the variable name will be confusing, because we are referring to radiuses and buffers differently in the code. How about `wider_search_suggestions`?

app/views/vacancies/index.html.haml
```diff
@@ -41,7 +41,10 @@
               .govuk-heading-m= t('jobs.no_jobs.radius_suggestions.heading.no_keyword')
             %ul.govuk-list.govuk-list--bullet
               - @vacancies_search.wider_radius_suggestions.each do |wider_radius|
-                %li= link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
+                - if @vacancies_search.location_search.location_polygon.present?
+                  %li= link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(buffer_radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
```
alexbowen, line 45: would be slightly better to pass `wider_radius.first` into a template in yml like `radius_link: {radius} miles`

app/views/vacancies/index.html.haml
```diff
@@ -41,7 +41,10 @@
...
+                - else
+                  %li= link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
```
alexbowen, line 47: and here

app/services/import_polygons.rb
josephhull676: I **think** the Portsmouth buffer was 5 miles and the Wakefield buffer was 10 miles.

The 5 mile buffer polygon for Portsmouth is made of 238 points.

The 10 mile buffer polygon for Wakefield is made of 264 points.

The smallest amount of points for any buffer is 216.

app/services/search/search_builder.rb
josephhull676: I think that makes sense, yeah!

app/services/search/search_builder.rb
josephhull676: I believe this has been changed now take another look and see if you're happy

app/controllers/vacancies_controller.rb
```diff
@@ -40,7 +40,7 @@ def algolia_search_params
     %w[job_roles phases working_patterns].each do |facet|
       params[facet] = params[facet].split(" ") if params[facet].is_a?(String)
     end
-    params.permit(:keyword, :location, :location_category, :radius, :job_roles, :phases, :working_patterns,
+    params.permit(:keyword, :location, :location_category, :radius, :job_roles, :phases, :working_patterns, :buffer_radius,
```
david-mears-dfe, line 43: Is there an argument for storing both the size of the buffer and the size of the radius in one param? Because for the user, they are one and the same concept (they don't know we use polygons). Or does that make the code more complicated?

david-mears-dfe: We would interpret the param as a buffer radius if the search was in a polygon, and we would interpret it as a normal radius if the search was on a point.

david-mears-dfe: That would allow a job alert created before the introduction of polygons (or before the introduction of a specific polygon) to seamlessly continue having a radius.

E.g. in March 2018 I create a job alert for London with radius 10. In mid-2020, polygons are introduced, so my search criteria are interpreted as a search within London. The radius is discarded because I'm doing a polygon search now. So the search has become less relevant to me: it only searches London, not London + 10 miles.

app/services/import_polygons.rb
```diff
@@ -41,4 +43,51 @@ def location_categories_include?(region_name)
...
+  def get_buffers(points)
+    # ArcGIS API documentation: https://ons-inspire.esriuk.com/arcgis/sdk/rest/index.html#//02ss0000003z000000
+    buffer_distances_in_miles = [5, 10, 15, 20, 25]
+    buffered_boundaries = {}
+    buffer_distances_in_miles.each do |distance|
+      # convert 1D array, to 2D array for arcgis
+      polygon_coords = points.each_slice(2).to_a
+      response = HTTParty.get(buffer_api_endpoint(polygon_coords, convert_miles_to_metres(distance)))
+      buffer_coords = response.dig("geometries", 0, "rings").flatten
+      # algolia requires a 2D containing 1 element with all the coordinates contained within it
```
david-mears-dfe, line 56: "a 2D containing"

app/services/import_polygons.rb
```diff
@@ -41,4 +43,51 @@ def location_categories_include?(region_name)
...
+      # so buffer_coords array has been wrapped in another array
```
david-mears-dfe, line 57: Does this wrapping happen here or elsewhere? It's confusing to talk about some wrapping that happens elsewhere.

app/services/import_polygons.rb
```diff
@@ -41,4 +43,51 @@ def location_categories_include?(region_name)
...
+               "inSR" => "4326",
+               "outSR" => "4326",
+               "bufferSR" => "3857",
+               "distances" => distance.to_s,
+               "unit" => "",
+               "unionResults" => "true",
+               "geodesic" => "false",
+               "f" => "json" }.to_param
+
+    buffer_api_url = "https://ons-inspire.esriuk.com/arcgis/rest/services/Utilities/Geometry/GeometryServer/buffer?"
```
david-mears-dfe, line 79: We could afford to extract this url as a constant belonging to `ImportPolygons`.

app/services/search/buffer_suggestions_builder.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :location, :buffer_suggestions, :search_params
+
+  def initialize(location, search_params)
+    @location = location
+    @search_params = search_params
+    get_buffers_suggestions
+  end
+
+  def get_buffers_suggestions
+    buffer_counts = LocationPolygon.with_name(location).buffers.each_with_object([]) do |buffer, buffer_counts|
```
david-mears-dfe, line 13: `buffer_counts` is a little opaque. What is being counted?

app/controllers/vacancies_controller.rb
cpjmcquillan: I think this is definitely outside the scope of the ticket. Also, a job alert created with `London + 10 miles` was created based on the centre of London + 10 miles, so the job alert is actually more relevant now rather than less relevant

app/services/search/location_builder.rb
```diff
@@ -38,7 +35,13 @@ def location_category_search?
 
   def initialize_location_polygon
     @location_polygon = LocationPolygon.with_name(@location_category)
-    @location_polygon_boundary = [@location_polygon.boundary] if @location_polygon.present?
+    if @location_polygon.present?
+      @location_polygon_boundary = if @buffer_radius.present?
+                                     [@location_polygon.buffers[@buffer_radius]]
+                                   else
+                                     [@location_polygon.boundary]
```
david-mears-dfe, line 42: Is this array wrapping for Algolia? (Is it what the comment in ImportPolygon is referring to?)

We might want to find a way to do the array wrapping only in one place, e.g. by writing it in the getter method (replacing attr_reader : location_polygon_boundary). But we haven't yet hit the 'Rule of Three': three strikes and refactor.

spec/services/import_polygons_spec.rb
```diff
@@ -1,85 +1,127 @@
 require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: This doesn't need to be here

spec/services/search/alert_builder_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: as above

app/services/search/location_builder.rb
```diff
@@ -38,7 +35,13 @@ def location_category_search?
 
   def initialize_location_polygon
     @location_polygon = LocationPolygon.with_name(@location_category)
-    @location_polygon_boundary = [@location_polygon.boundary] if @location_polygon.present?
+    if @location_polygon.present?
+      @location_polygon_boundary = if @buffer_radius.present?
```
david-mears-dfe, line 39: I'm uncomfortable about naming the buffered polygon the 'location polygon boundary' - that sounds like it refers to the original boundary of the LocationPolygon. Would @search_polygon_boundary or @search_polygon suit your purposes, or @polygon_boundary, or @boundary, or @search_boundary...?

app/services/search/search_builder.rb
```diff
@@ -49,7 +48,9 @@ def build_search_replica
   end
 
   def build_suggestions
-    @wider_radius_suggestions = Search::SuggestionsBuilder.new(search_params, @params_hash[:radius]).radius_suggestions
+    return @wider_search_suggestions = Search::RadiusSuggestionsBuilder.new(search_params, @params_hash[:radius]).radius_suggestions if @point_coordinates.present?
+
+    @wider_search_suggestions = Search::BufferSuggestionsBuilder.new(@params_hash[:location], search_params).buffer_suggestions if @location_search.location_category_search?
```
david-mears-dfe, line 53: It would be safer to check whether the @location_search has a polygon.

`location_category_search?` just checks whether it's the kind of search that *ought* to have a polygon. But knowing us, we might have lost the polygon in transit.

david-mears-dfe: `if @location_search.location_polygon.present?`

david-mears-dfe: I find it a little hard to read with the `if`s far off to the right. Might be worth going

```ruby
if x
  do y
elsif z
  do a
end
```

app/services/search/buffer_suggestions_builder.rb
cpjmcquillan, line 13: This would look cleaner imo if you just mapped `LocationPolygon.with_name(location).buffers`

spec/services/import_polygons_spec.rb
```diff
@@ -1,85 +1,127 @@
...
+          "15" => buffer_coordinates,
+          "20" => buffer_coordinates,
+          "25" => buffer_coordinates,
+        }
       end
 
-      it "imports Bath boundaries" do
-        expect(LocationPolygon.cities.first.boundary).to eq(points)
+      it "imports buffers" do
+        expect(LocationPolygon.regions.first.buffers).to eq(imported_buffers)
```
david-mears-dfe, line 124: Can you change `LocationPolygon.regions.first` to `LocationPolygon.with_name('bath')`? The `first` threw me off when I was trying to read it - I thought this was only testing the first buffer.

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        "20" => buffer_coordinates,
+        "25" => buffer_coordinates,
+      }
+    end
+
+    before do
+      allow(LocationPolygon).to receive_message_chain(:with_name, :buffers).and_return(buffer_hash)
+    end
+
+    context "when there is one vacancy within all of the buffer polygons" do
```
david-mears-dfe, line 29: Ambiguous
"when there is only one vacancy in any of the buffer polygons"
it "suggests only the relevant buffer"

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      before do
+        allow_any_instance_of(Search::AlgoliaSearchRequest).to receive_message_chain(:stats, :last).and_return(1)
+      end
+
+      it "suggests one buffer radius" do
+        expect(subject.buffer_suggestions).to eql([["5", 1]])
+      end
+    end
+
+    context "when there are more vacancies within the wider buffer polygons" do
```
david-mears-dfe, line 39: "when there is more than one" - for clarity?

app/services/import_polygons.rb
josephhull676: Ha nice spot

app/controllers/vacancies_controller.rb
josephhull676: For the reasons Connor stated above, I'll leave this as it is for now.

app/services/import_polygons.rb
josephhull676: The wrapping happens in the location builder. I've now remove the comment about the wrapping and have kept a comment mentioning that the buffer coordinates are stored as a 1D array.

app/services/search/location_builder.rb
josephhull676: I think it does only happens in the location builder.

app/services/search/location_builder.rb
josephhull676: I think search_polygon_boundary is probably the best option.

app/services/search/location_builder.rb
david-mears-dfe: I meant lines 40 and 42. But this is an acceptable level of duplication, for me.

spec/services/import_polygons_spec.rb
josephhull676: It's used on line 99, no?

spec/services/search/alert_builder_spec.rb
josephhull676: Isn't this used on line 13?

app/services/import_polygons.rb
cpjmcquillan: Given we're downsampling the polygon for the purpose of creating buffers (and the accuracy of these is far less important) and not downsampling the base polygon itself, this is ok

app/services/import_polygons.rb
```diff
@@ -41,4 +45,48 @@ def location_categories_include?(region_name)
...
+               "geodesic" => "false",
+               "f" => "json" }.to_param
+
+    api_endpoint = BUFFER_API_URL + params
+
+    if api_endpoint.length > 2000
+      # Reduce number of coordinates in order to have fewer than 2000 characters in request url
+      coords = coords.each_with_index
+      .map { |item, index| item if (index % 5).zero? }
+      .compact
```
cpjmcquillan, line 86: Can you indent `.map` and `.compact` with `.each_with_index` though, please?

cpjmcquillan: Actually, why are you using both `.each_with_index` and `.map` like that?

app/services/import_polygons.rb
```diff
@@ -29,7 +33,7 @@ def call
       end
 
       LocationPolygon.find_or_create_by(name: region_name, location_type: location_type.to_s)
-                     .update(boundary: points)
+                    .update(boundary: points, buffers: get_buffers(points))
```
cpjmcquillan, line 36: Indentation

app/services/import_polygons.rb
```diff
@@ -1,12 +1,16 @@
 class ImportPolygons
+  include DistanceHelper
+
+  BUFFER_API_URL = "https://ons-inspire.esriuk.com/arcgis/rest/services/Utilities/Geometry/GeometryServer/buffer?".freeze
```
cpjmcquillan, line 4: I think we tend to define constants first in classes/modules

app/services/search/radius_suggestions_builder.rb
```diff
@@ -19,7 +21,7 @@ def get_radius_suggestions
         [
           wider_radius,
           Search::AlgoliaSearchRequest.new(
```
cpjmcquillan, line 23: This can probably fit on the same line a bit more nicely now the helper is included

spec/services/import_polygons_spec.rb
cpjmcquillan: It's already in `spec/rails_helper.rb`

spec/services/import_polygons_spec.rb
```diff
@@ -1,85 +1,127 @@
...
-      let(:file_name) { "ons_cities.json" }
-
-      it "imports Bath city" do
-        expect(LocationPolygon.cities.first.name).to eq("bath")
+    context "when not stubbing #get_buffers" do
+      before do
+        distances = [5, 10, 15, 20, 25]
+        distances.each do |distance|
+          allow(HTTParty).to receive(:get).with(
+            "https://ons-inspire.esriuk.com/arcgis/rest/services/Utilities/Geometry/GeometryServer/buffer?"\
```
cpjmcquillan, line 98: This endpoint is defined elsewhere isn't it?

spec/services/search/alert_builder_spec.rb
cpjmcquillan: it's already included in `spec/rails_helper.rb`

spec/services/search/alert_builder_spec.rb
```diff
@@ -9,7 +10,7 @@
   let(:location) { "SW1A 1AA" }
   let(:date_today) { Date.current.to_time }
   let(:location_point_coordinates) { Geocoder::DEFAULT_STUB_COORDINATES }
-  let(:location_radius) { Search::LocationBuilder.convert_radius_in_miles_to_metres(10) }
+  let(:location_radius) { DistanceHelper.convert_miles_to_metres(10) }
```
cpjmcquillan, line 13: this then changes to `helper.convert_miles_to_metres` or just `convert_miles_to_metres`, I can't remember which

spec/services/search/algolia_search_request_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: same as above

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -0,0 +1,59 @@
+require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: +1

spec/services/search/location_builder_spec.rb
```diff
@@ -5,24 +5,36 @@
     expect(subject.location_category).to eql(options&.dig(:location)&.presence || polygonable_location)
     expect(subject.location_polygon).to eql(location_polygon)
     expect(subject.location_filter).to eql({})
+    expect(subject.buffer_radius).to eql(buffer_radius)
   end
 end
 
 RSpec.describe Search::LocationBuilder do
-  subject { described_class.new(location, radius, location_category) }
+  include DistanceHelper
```
cpjmcquillan, line 13: +1

spec/services/search/location_builder_spec.rb
```diff
@@ -72,27 +91,34 @@
     context "when a non-polygonable location is specified" do
       context "and no radius specified" do
         let(:location) { point_location }
+        let(:radius) { 10 }
+        let(:expected_radius) { 16_093 }
+
+        before { allow_any_instance_of(DistanceHelper).to receive(:convert_miles_to_metres).with(10).and_return(expected_radius) }
```
cpjmcquillan, line 97: I don't really see the point of doing this?

spec/services/search/radius_suggestions_builder_spec.rb
```diff
@@ -1,14 +1,15 @@
 require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: +1

spec/services/search/search_builder_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+include DistanceHelper
```
cpjmcquillan, line 2: +1

spec/system/jobseekers_can_subscribe_to_a_job_alert_spec.rb
```diff
@@ -135,8 +135,20 @@
     end
 
     context "when a location category search is carried out" do
+      let(:boundary_points) { [1.004562496029994090, 51.50833566307333, 1.005710530794815389, 51.5051084208278] }
```
cpjmcquillan, line 138: What's the purpose of this?

app/services/import_polygons.rb
david-mears-dfe: This has made me realise we should test this part of the code - it won't be run by the specs because we don't give the specs incredibly long lists of points. But we still should test that we aren't effing up larger polygons. If this part of the code could be extracted to a method of its own, that would make it easier to unit test it.

david-mears-dfe: I think the easiest bit to extract to a method would be 
```ruby
coords = coords.each_with_index.map { |item, index| item if (index % 5).zero? }.compact
```
because it doesn't get involved with recursion.

david-mears-dfe: Makes for a bit more readable code at the same time - the method name will help explain what is going on

david-mears-dfe: > Actually, why are you using both .each_with_index and .map like that?

We need the index. Apparently, while there is no `map_with_index`, there is a `map.with_index`.

josephhull676: I don't think map is necessary. @david-mears-dfe ?

spec/services/import_polygons_spec.rb
josephhull676: It is, in part, in `import_polygons.rb`, but for the purposes of this test it needs to be different for each distance. How else would I achieve this?

spec/system/jobseekers_can_subscribe_to_a_job_alert_spec.rb
josephhull676: You know what, I don't know. I think it may have been some attempt to not rely on anything else outside of what's being tested, but I assume that's unnecessary for a helper?

spec/services/search/location_builder_spec.rb
josephhull676: You know what, I don't know. I think it may have been some attempt to not rely on anything else outside of what's being tested, but I assume that's unnecessary for a helper?

spec/system/jobseekers_can_subscribe_to_a_job_alert_spec.rb
josephhull676: Before adding buffer_coordinates and buffer_hash, when this test got to the stage where buffer suggestions were being built, it would fail as `LocationPolygon.with_name(location).buffers` was nil:

```
Failure/Error:
       buffer_vacancy_count = LocationPolygon.with_name(location).buffers.map do |buffer|
         [buffer.first, Search::AlgoliaSearchRequest.new(search_params.merge(polygon: [buffer.last])).stats.last]
       end

     NoMethodError:
       undefined method `map' for nil:NilClass
```

I also came across another failure which was caused by boundary being nil too, but I can't seem to recreate this, so I've removed `let(:boundary_points) { [1.004562496029994090, 51.50833566307333, 1.005710530794815389, 51.5051084208278] }`

cpjmcquillan: I would just stub that call (you do something similar in a different spec), since we don't care/ are not testing that it happens

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      it "suggests the buffer with the smallest radius" do
+        expect(subject.buffer_suggestions).to eql([["5", 1]])
+      end
+    end
+
+    context "when there are vacancies in the wider buffer polygons not found in the smaller polygons" do
+      before do
+        allow(Search::AlgoliaSearchRequest).to receive_message_chain(:new, :stats, :last).and_return(1, 2, 3, 4, 5)
```
cpjmcquillan, line 41: Really cool, didn't know you could do this. But I feel like the approach in `radius_suggestions_builder_spec.rb` might be more valuable.

```ruby
mock_algolia_search(
  vacancies_1, 1, "maths",
  arguments_for_algolia.merge(aroundRadius: Search::LocationBuilder.convert_radius_in_miles_to_metres(5))
)
```

for example is testing the correct parameters are called as well, which I think we lose out on here otherwise

app/services/import_polygons.rb
david-mears-dfe: Can you prove it with a unit test, please?

app/views/vacancies/index.html.haml
```diff
@@ -33,15 +33,18 @@
...
             - else
-              .govuk-heading-m= t('jobs.no_jobs.radius_suggestions.heading.no_keyword')
+              .govuk-heading-m= t('jobs.no_jobs.wider_search_suggestions.heading.no_keyword')
             %ul.govuk-list.govuk-list--bullet
-              - @vacancies_search.wider_radius_suggestions.each do |wider_radius|
-                %li= link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
+              - @vacancies_search.wider_search_suggestions.each do |wider_radius|
+                - if @vacancies_search.location_search.location_polygon.present?
+                  %li= link_to(t('jobs.no_jobs.wider_search_suggestions.radius_distance', radius: wider_radius.first), jobs_path(@vacancies_search.params_hash.merge(buffer_radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.wider_search_suggestions.results', count: wider_radius.last)
```
cpjmcquillan, line 45: This should probably be a single html translation with the link_to and count passed as arguments

app/views/vacancies/index.html.haml
```diff
@@ -33,15 +33,18 @@
...
+                - else
+                  %li= link_to(t('jobs.no_jobs.wider_search_suggestions.radius_distance', radius: wider_radius.first), jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "govuk-link wider-radius-gtm") + ' ' + t('jobs.no_jobs.wider_search_suggestions.results', count: wider_radius.last)
```
cpjmcquillan, line 47: As above

app/views/vacancies/index.html.haml
```diff
@@ -33,15 +33,18 @@
...
             - else
-              .govuk-heading-m= t('jobs.no_jobs.radius_suggestions.heading.no_keyword')
+              .govuk-heading-m= t('jobs.no_jobs.wider_search_suggestions.heading.no_keyword')
             %ul.govuk-list.govuk-list--bullet
-              - @vacancies_search.wider_radius_suggestions.each do |wider_radius|
-                %li= govuk_link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
+              - @vacancies_search.wider_search_suggestions.each do |wider_radius|
+                - if @vacancies_search.location_search.location_polygon.present?
+                  = t('jobs.no_jobs.wider_search_suggestions.suggestion_html', search_link: link_to(t('jobs.no_jobs.wider_search_suggestions.radius_distance', radius: wider_radius.first), jobs_path(@vacancies_search.params_hash.merge(buffer_radius: wider_radius.first))), count: t('jobs.no_jobs.wider_search_suggestions.results', count: wider_radius.last))
```
cpjmcquillan, line 45: These are quite unreadable, maybe try relative translations or something to make them shorter

config/locales/jobs.yml
```diff
@@ -46,10 +46,12 @@ en:
     no_jobs:
       heading: Try another search
       intro: 'You may get more results if you:'
-      radius_suggestions:
+      wider_search_suggestions:
         heading:
           keyword_html: Try browsing for <span class='text-wrap-apostrophe'>%{keyword}</span> in a wider location search
           no_keyword: Try browsing in a wider location search
+        suggestion_html: <li class="govuk-link wider-radius-gtm">%{search_link} %{count}</li>
```
cpjmcquillan, line 53: We should be passing in `govuk_link_to` helper, not writing html in the translation

spec/services/search/search_builder_spec.rb
```diff
@@ -81,6 +88,8 @@
         }
       end
 
+      before { allow(Search::BufferSuggestionsBuilder).to receive_message_chain(:new, :buffer_suggestions) }
```
david-mears-dfe, line 91: I'm not sure what this line (and 113) is achieving?

spec/services/import_polygons_spec.rb
```diff
@@ -2,84 +2,151 @@
...
+              "bufferSR=3857&distances=#{convert_miles_to_metres(distance)}&f"\
+              "=json&geodesic=false&geometries=%7B%22geometryType%22%3D%3E%22esriGeometryPolygon%22%2C+%22"\
+              "geometries%22%3D%3E%5B%7B%22rings%22%3D%3E%5B%5B%5B52%2C+0%5D%2C+%5B52%2C+0%5D%5D%5D%7D%5D%7D"\
+              "&inSR=4326&outSR=4326&unionResults=true&unit=",
+            ).and_return(buffer_response)
+          end
+          subject.call
+        end
+
+        it "reduces the number of coordinates before constructing the API endpoint" do
```
david-mears-dfe, line 147: Re-reading this description, it feels too unrelated to what's happening on line 148. Someone could read it and think 'how does checking that the buffers are there prove that the coordinates got reduced'?

I suggest:

```ruby
it "is still able to import the buffers, by reducing the number of coordinates in the params to the API endpoint" do
```

spec/services/import_polygons_spec.rb
```diff
@@ -2,84 +2,151 @@
...
+          "10" => buffer_coordinates,
+          "15" => buffer_coordinates,
+          "20" => buffer_coordinates,
+          "25" => buffer_coordinates,
+        }
       end
 
-      it "does not import non-existent city" do
-        expect(LocationPolygon.cities.count).to eq(1)
+      context "when api_endpoint is shorter than URL_MAXIMUM_LENGTH" do
```
david-mears-dfe, line 105: Let's make the two context descriptions more informative to explain why the api_endpoint length could vary. We should also remove references to implementation details such as the variable name and the constant name.

E.g. "when the length of the params caused the API endpoint length to exceed the maximum"

spec/services/import_polygons_spec.rb
```diff
@@ -2,84 +2,151 @@
...
+          expect(LocationPolygon.regions.first.boundary).to eq(boundary_points)
+        end
       end
 
-      it "does not import non-existent county" do
-        expect(LocationPolygon.counties.count).to eq(1)
-      end
+      context "when location type is counties" do
+        let(:location_type) { :counties }
+        let(:boundary_points) { [52.0652732493605, 0.7031775640868, 52.0486871974891, 0.7162633765872] }
```
david-mears-dfe, line 39: If you changed this, you could probably combine the ons_counties.json fixture and the new fixture into one fixture.

spec/services/import_polygons_spec.rb
```diff
@@ -2,84 +2,151 @@
...
-      let(:location_type) { :counties }
-      let(:points) { [52.0652732493605, 0.7031775640868, 52.0486871974891, 0.7162633765872] }
-      let(:file_name) { "ons_counties.json" }
+        it "does not import non-existent region" do
+          expect(LocationPolygon.regions.count).to eq(1)
+        end
 
-      it "imports Essex county" do
-        expect(LocationPolygon.counties.first.name).to eq("essex")
+        it "imports Nort East boundaries" do
```
david-mears-dfe, line 32: North

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2431.london.cloudapps.digital>

app/views/vacancies/index.html.haml
```diff
@@ -33,15 +33,22 @@
...
+              .govuk-heading-m= t('.wider_search_suggestions.heading.keyword_html', keyword: @vacancies_search.keyword)
             - else
-              .govuk-heading-m= t('jobs.no_jobs.radius_suggestions.heading.no_keyword')
+              .govuk-heading-m= t('.wider_search_suggestions.heading.no_keyword')
             %ul.govuk-list.govuk-list--bullet
-              - @vacancies_search.wider_radius_suggestions.each do |wider_radius|
-                %li= govuk_link_to("#{wider_radius.first} miles", jobs_path(@vacancies_search.params_hash.merge(radius: wider_radius.first)), class: "wider-radius-gtm") + ' ' + t('jobs.no_jobs.radius_suggestions.results', count: wider_radius.last)
+              - @vacancies_search.wider_search_suggestions.each do |wider_radius|
+                - if @vacancies_search.location_search.location_polygon.present?
+                  %li.govuk-link.wider-radius-gtm= t('.wider_search_suggestions.suggestion_html',
```
cpjmcquillan, line 45: We shouldn't be adding classes to the list element here. We don't use `link_to` anymore, instead use `govuk_link_to` which applies "govuk-link" class to links by default, and any other classes can be passed as keyword arguments to the helper

cpjmcquillan (changes requested):

cesidio (approved):

spec/services/search/search_builder_spec.rb
cpjmcquillan: It's about isolating what we're actually testing. We don't want either radius suggestions or buffer suggestions to actually be called, because we don't care about that in these tests - and if they are called we need to be mocking other unrelated things

cpjmcquillan: At least that's why I did it in the first place

spec/system/jobseekers_can_subscribe_to_a_job_alert_spec.rb
```diff
@@ -135,8 +135,19 @@
...
+          "10" => buffer_coordinates,
+          "15" => buffer_coordinates,
+          "20" => buffer_coordinates,
+          "25" => buffer_coordinates,
+        }
+      end
+
       before do
-        LocationPolygon.create(name: "london")
+        LocationPolygon.create(name: "london", buffers: buffer_hash)
```
cpjmcquillan, line 150: I think this would probably be better if we made a location polygon factory object

app/views/vacancies/index.html.haml
josephhull676: Didn't know this. Cheers.

PR #2433, opened 2020-11-26
Strip whitespaces when searching

david-mears-dfe:
Currently, 'london' searches a polygon and 'london ' searches a point.

Compare "90 jobs found near 'London '" and "134 jobs found in 'London'". The former fails to use a polygon because of the trailing whitespace.

This has also been messing up searches in Birmingham (127 counts of searching with a whitespace), Manchester, Liverpool, Sheffield, etc.

In our sample, there were 168 searches for "london " and 6376 for "london"

### DIFF
--- a/app/form_models/vacancy_algolia_search_form.rb
+++ b/app/form_models/vacancy_algolia_search_form.rb
@@ -9,10 +9,10 @@ class VacancyAlgoliaSearchForm
               :total_filters
 
   def initialize(params = {})
-    @keyword = params[:keyword]
+    @keyword = params[:keyword].strip!
 
-    @location = params[:location] || params[:location_category]
-    @location_category = params[:location_category]
+    @location = params[:location].strip! || params[:location_category].strip!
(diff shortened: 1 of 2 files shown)


app/form_models/vacancy_algolia_search_form.rb
```diff
@@ -9,10 +9,10 @@ class VacancyAlgoliaSearchForm
               :total_filters
 
   def initialize(params = {})
-    @keyword = params[:keyword]
+    @keyword = params[:keyword].strip!
```
cesidio, line 12: How about just doing?

`params.values.each(&:strip!)`

cpjmcquillan: I can't think of any scenario where we wouldn't want to strip whitespace, so maybe we should do this at a controller level so params are stripped by default?

david-mears-dfe:
Reimlemented in https://github.com/DFE-Digital/teaching-vacancies/pull/2437

PR #2436, opened 2020-11-26
Rename session keys to be unambiguous

csutter:
- Rename `session_id` to `publisher_oid`
- Prefix `urn`, `uid`, `la_code`, `multiple_organisations`, `id_token`,
  and `signing_out_for_inactivity` with `publisher_` for consistency
- Tweak i18n keys for signing out notifications to match

The existing naming is a bit confusing, and the `session_id` key
conflicts with Devise (which we are now using for jobseeker accounts).

Caution: Deploying this will have the side effect of logging out all
currently logged in users (as the code checks for presence of the new
`publisher_oid` value in the session as part of the activity check in
`Publishers::BaseController#check_user_last_activity_at`)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1634

### DIFF
--- a/app/controllers/interests_controller.rb
+++ b/app/controllers/interests_controller.rb
@@ -8,7 +8,7 @@ private
 
   def audit_click
     VacancyGetMoreInfoClick.new(vacancy).track
-    Auditor::Audit.new(vacancy, "vacancy.get_more_information", current_session_id).log
+    Auditor::Audit.new(vacancy, "vacancy.get_more_information", current_publisher_oid).log
     AuditExpressInterestEventJob.perform_later(
       datestamp: Time.current.iso8601.to_s,
       vacancy_id: vacancy.id,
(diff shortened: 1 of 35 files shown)


david-mears-dfe:
`publisher_oid` and `publisher_uid` are worryingly close!
What about `organisation_uid` and `organisation_urn` and `organisation_la_code`?

cesidio (approved):

app/controllers/interests_controller.rb
```diff
@@ -8,7 +8,7 @@ def new
 
   def audit_click
     VacancyGetMoreInfoClick.new(vacancy).track
-    Auditor::Audit.new(vacancy, "vacancy.get_more_information", current_session_id).log
+    Auditor::Audit.new(vacancy, "vacancy.get_more_information", current_publisher_oid).log
```
cpjmcquillan, line 11: I'd vote for getting rid of all of these while you're here rather than updating them

csutter: Agree they need to go sharpish, but I'd like to keep them around just a tiny bit longer – I want to replace them with analytics events as part of the events-to-Bigquery spike that's on the horizon.

PR #2437, opened 2020-11-26
Strip whitespaces from all form params, and from all existing job alert search criteria

david-mears-dfe:
The reason for doing this was that, currently, 'london' is handled differently by search than 'london     ' (the latter searches around a point, the former in a polygon).

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1653

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,8 +9,8 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
 
-  before_action :set_headers
-  before_action :set_root_headers
+  before_action :set_headers, :set_root_headers
+  before_action { strip_nested_param_whitespaces(request.params) }
 
   helper_method :cookies_preference_set?, :referred_from_jobs_path?, :utm_parameters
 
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_24_120747) do
+ActiveRecord::Schema.define(version: 2020_11_26_201055) do
```
cesidio, line 13: THis should not be here

app/controllers/application_controller.rb
```diff
@@ -96,4 +95,17 @@ def set_headers
   def invalid_recaptcha_score?
     recaptcha_reply["score"] < SUSPICIOUS_RECAPTCHA_THRESHOLD
   end
+
+  def strip_param_whitespaces
```
cpjmcquillan, line 99: Does this strip nested params? Also is there a reason why you iterate over the keys like this?

db/schema.rb
david-mears-dfe: Why?

app/controllers/application_controller.rb
david-mears-dfe: - Well caught, I only went down one nesting level, which only gets us down to the top-level form params.
- I iterate over the keys to find any hashes, because any form params will be nested in a hash. But this can be made neater.

cpjmcquillan: If you want you can do something recursive, which tends to make heads hurt a little bit but means we account for any level of nesting

david-mears-dfe: I had some fun making it recursive. Thanks for the idea.

app/controllers/application_controller.rb
```diff
@@ -93,6 +93,19 @@ def invalid_recaptcha_score?
     recaptcha_reply["score"] < SUSPICIOUS_RECAPTCHA_THRESHOLD
   end
 
+  def strip_nested_param_whitespaces(object)
```
david-mears-dfe, line 96: Not sure about this name. I want to be explicit but not confusing. `string_string_whitespaces`?

cesidio (approved):
Nice 👍 I like the recursive function

PR #2443, opened 2020-11-27
Import some types of local authorities special schools

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1598


lib/organisation_import/import_school_data.rb
```diff
@@ -57,6 +56,11 @@ def datestring
     Time.current.strftime("%Y%m%d")
   end
 
+  def school_in_local_authority_scope?(row)
```
csutter, line 59: I like the helper method being split out, but is it worth making these separate methods so we know what each of these magic GIAS codes does? e.g.:

```ruby
def school_in_local_authority_scope?(row)
  school_is_local_authority_maintained?(row) || def school_is_community_or_foundation_special_school?(row)
end

def school_is_local_authority_maintained?(row)
  row["EstablishmentTypeGroup (code)"] == "4"
end

def school_is_community_or_foundation_special_school?(row)
  row["EstablishmentTypeGroup (code)"] == "5" && %w[7 12].include?(row["TypeOfEstablishment (code)"])
end
```

cesidio: addressed

cpjmcquillan (approved):

csutter (approved):

PR #2453, opened 2020-11-30
Force signout of opposite user type on sign in

csutter:
When signing in as a publisher or jobseeker, we need to sign out the
respective opposite user type to avoid getting the application into a
weird state or confuse our users. This is very much an edge case (we
don't expect either kind of user to have an account of the opposite
type), but needs to be covered nonetheless.

- Rename `Publishers::BaseController#check_user_last_activity_at` to
  `authenticate_publisher!` to be more consistent with Devise-style
  controller authentication enforcement
- Rename `Publishers::BaseController#update_user_last_activity_at` to
  to `update_publisher_last_activity_at`
- Add `Publishers::AuthenticationConcerns#sign_out_publisher!` to
  clear publisher-related session data (and call on jobseeker sign in)
- Add `Publishers::SignIn::BaseSessionsController::sign_out_jobseeker!`
  (and call on publisher sign in)
- Rename `sign_in_user` test helper to `sign_in_publisher`

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1634

### DIFF
--- a/app/controllers/jobseekers/sessions_controller.rb
+++ b/app/controllers/jobseekers/sessions_controller.rb
@@ -1,3 +1,4 @@
 class Jobseekers::SessionsController < Devise::SessionsController
   after_action :replace_devise_notice_flash_with_success!, only: %i[create destroy]
+  before_action :sign_out_publisher!, only: %i[create]
 end
--- a/app/controllers/publishers/base_controller.rb
+++ b/app/controllers/publishers/base_controller.rb
@@ -1,8 +1,8 @@
 class Publishers::BaseController < ApplicationController
   TIMEOUT_PERIOD = 60.minutes
 
-  before_action :check_user_last_activity_at,
-                :update_user_last_activity_at,
+  before_action :authenticate_publisher!,
+                :update_publisher_last_activity_at,
                 :redirect_to_root_if_read_only,
                 :check_session,
                 :check_terms_and_conditions
@@ -34,8 +34,9 @@ class Publishers::BaseController < ApplicationController
     PublisherPreference.find_by(publisher_id: current_publisher.id, school_group_id: current_organisation.id)
   end
 
-  def check_user_last_activity_at
-    return redirect_to logout_endpoint if current_publisher&.last_activity_at.blank?
+  def authenticate_publisher!
+    return redirect_to(new_identifications_path) unless current_publisher
+    return redirect_to(logout_endpoint) if current_publisher.last_activity_at.blank?
 
     if Time.current > (current_publisher.last_activity_at + TIMEOUT_PERIOD)
       session[:publisher_signing_out_for_inactivity] = true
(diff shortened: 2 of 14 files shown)


app/controllers/jobseekers/sessions_controller.rb
```diff
@@ -1,3 +1,4 @@
 class Jobseekers::SessionsController < Devise::SessionsController
   after_action :replace_devise_notice_flash_with_success!, only: %i[create destroy]
+  before_action :sign_out_publisher!, only: %i[create]
```
david-mears-dfe, line 3: So what do these bangs denote?

csutter: Destructive, I suppose. It changes stuff in-place. Bit of a grey area though. Happy to have it without if that's nicer?

app/controllers/publishers/base_controller.rb
```diff
@@ -34,16 +34,17 @@ def current_publisher_preferences
     PublisherPreference.find_by(publisher_id: current_publisher.id, school_group_id: current_organisation.id)
   end
 
-  def check_user_last_activity_at
-    return redirect_to logout_endpoint if current_publisher&.last_activity_at.blank?
+  def authenticate_publisher!
```
cesidio, line 37: Not sure about the name of this method. are we authenticating the publisher or just checking the last activity?

csutter: I wasn't sure about the name at first, but actually this method was implicitly authenticating publishers already – in the first line, if `current_publisher` is `nil`, it also redirects to the logout endpoint because of the `&.`. The `#check_session` method only checks organisation keys in the session.

I also changed this to go to the login page if you don't already have a session, because it seems weird to redirect the user to DSI in that case. But maybe it should?

cesidio (approved):

app/controllers/jobseekers/sessions_controller.rb
david-mears-dfe: I don't mind, just wanted to make sure we have a common understanding of what they mean

spec/system/users_can_only_be_signed_in_to_one_type_of_account_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  context "when a jobseeker is signed in" do
+    before do
+      login_as(jobseeker, scope: :jobseeker)
+    end
+
+    context "when email fallback is disabled" do
+      let(:dsi_email_address) { Faker::Internet.email }
+
+      before(:each) do
```
cpjmcquillan, line 24: is `before(:each) do` not the same as `before do`?

csutter: Yes it is – I seem to have managed to copypasta my way into using both `before` and `before(:each)` here 🍝 

I have a mild preference for explicitly stating `:each` normally, but I can see we mix and match throughout TV. Do we have any preference?

csutter: I've changed them all to `:each` 🦘

PR #2454, opened 2020-11-30
Add back pretty URLs for location categories

cesidio:


cpjmcquillan (approved):
It looks good to me, but it makes me sad that we can't have much confidence in whether or not it will break pagination 😔

cesidio:
> It looks good to me, but it makes me sad that we can't have much confidence in whether or not it will break pagination 😔

Yes, we need to do something on that front, not sure if in this PR though

alexbowen (approved):

PR #2455, opened 2020-11-30
Fix bug preventing importing of community special schools

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1598


lib/organisation_import/import_school_data.rb
```diff
@@ -65,7 +65,7 @@ def school_is_local_authority_maintained?(row)
   end
 
   def school_is_community_or_foundation_special_school?(row)
-    row["EstablishmentTypeGroup (code)"] == "5" && %w[7 12].include?(row["TypeOfEstablishment (code)"])
+    row["EstablishmentTypeGroup (code)"] == "5" && %w[07 12].include?(row["TypeOfEstablishment (code)"])
```
cpjmcquillan, line 68: This is why I did the `.to_i` before if I recall correctly. The "5" is probably an "05" but ruby just goes with the flow

cpjmcquillan (approved):

lib/organisation_import/import_school_data.rb
cesidio: I think it is consistent, it is always `5` or `05`. But yeah doing `to_i` is not too much work

alexbowen (approved):

PR #2456, opened 2020-12-01
A miniscule refactor (set before_action argument as default)

david-mears-dfe:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,17 +92,13 @@ private
     recaptcha_reply["score"] < SUSPICIOUS_RECAPTCHA_THRESHOLD
   end
 
-  def strip_nested_param_whitespaces(object = request.params)
+  def strip_nested_param_whitespaces(param = request.params)
     # Recursively find strings and strip them of trailing whitespaces
-    if object.is_a?(String)
-      return object.strip
-    elsif object.is_a?(Hash)
-      object.each do |key, value|
-        object[key] = strip_nested_param_whitespaces(value)
-      end
+    if param.respond_to?(:strip!)
+      param.strip!
+    elsif param.respond_to?(:each_value)
+      param.each_value(&method(:strip_nested_param_whitespaces))
     end
-
-    object
   end
 
   def replace_devise_notice_flash_with_success!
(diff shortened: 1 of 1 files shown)


cesidio (approved):

cpjmcquillan (approved):

app/controllers/application_controller.rb
```diff
@@ -93,7 +92,7 @@ def invalid_recaptcha_score?
     recaptcha_reply["score"] < SUSPICIOUS_RECAPTCHA_THRESHOLD
   end
 
-  def strip_nested_param_whitespaces(object)
+  def strip_nested_param_whitespaces(object = request.params)
     # Recursively find strings and strip them of trailing whitespaces
     if object.is_a?(String)
       return object.strip
```
csutter, line 98: Not part of this commit, but `String#strip` is not an in-place operation – so depending on the argument this method would either modify its `object` argument (if a `Hash`) or return a new object while leaving `object` untouched (if a `String`). Which feels a bit weird.

Also, `request.params` would be an instance of `ActionController::Parameters`, so not a `Hash` – in which case this method would just do nothing 🤔

csutter (changes requested):

app/controllers/application_controller.rb
david-mears-dfe: I'm not sure there is a neat way around your first problem. Let me know if you think of one.

csutter: In Ruby, I would suggest duck typing instead of checking for a specific type most of the time. In this case, do you care that your argument is a specific thing, or do you just want it to be able to strip itself of whitespace or enumerate itself? 🦆 

It'll probably save your bacon if at some point there are new Rails wrapper objects and suddenly the keys at the bottom of your params hierarchy aren't `String`s anymore, or `ActionController::Parameters` becomes `ActionController::SuperShinyParametersWithBaconAndCheese` etc etc etc.

 How would you feel about something like this instead?

```ruby
def strip_nested_param_whitespaces(param = request.params)
  if param.respond_to?(:strip!)
    param.strip!
  elsif param.respond_to?(:each_value)
    param.each_value(&method(:strip_nested_param_whitespaces))
  end
end
```

david-mears-dfe: Very nice explanation. Also, I didn't realise `strip!` existed. That works nicely.

csutter (approved):

PR #2457, opened 2020-12-01
Fix some random rollbar errors from bots that annoy me

cpjmcquillan:
Bad bots annoying

### DIFF
--- a/app/services/search/filters_builder.rb
+++ b/app/services/search/filters_builder.rb
@@ -60,6 +60,6 @@ private
   end
 
   def normalize_array_params(params)
-    params&.reject(&:blank?)&.map { |value| value.to_s.gsub('["', "").gsub('"]', "") }
+    params&.reject(&:blank?)&.map { |value| value.to_s.delete(",[]\'\"") }&.reject(&:blank?)
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/search/filters_builder.rb
```diff
@@ -60,6 +60,6 @@ def expired_now_filter
   end
 
   def normalize_array_params(params)
-    params&.reject(&:blank?)&.map { |value| value.to_s.gsub('["', "").gsub('"]', "") }
+    params&.reject(&:blank?)&.map { |value| value.to_s.delete("[").delete("]").delete('"').delete("'").delete("") }&.reject(&:blank?)
```
csutter, line 63: `String#delete` takes multiple arguments, and deleting an empty string won't do anything, so this could be simplified to:
```ruby
params&.reject(&:blank?)&.map { |value| value.to_s.delete("[", "]", '"', "'") }&.reject(&:blank?)
```

Or perhaps even a regex that specifies exactly what a valid filter looks like? Rather than disallowing invalid params?

cpjmcquillan: I've removed the errorneous empty string delete, but delete with multiple arguments doesn't work as expected. I'll have a try with some regex though that would be nicer

cpjmcquillan: To delete multiple things you just chuck them all together 👍

csutter (approved):

PR #2458, opened 2020-12-01
Don't titlecase acronyms in SchoolGroup names

david-mears-dfe:
### ⚠️ **You'll want to review by looking at the [first commit](https://github.com/DFE-Digital/teaching-vacancies/pull/2458/commits/757a86d2d7c2fbcbe34dac6ab6f1b527f84c9915). The second is just putting the methods in a different order.**

Hypothesis: Trusts will be more willing to use TV if we get their names right.

(I started doing this as something mindless yet productive to do in a ceremony-filled day that wouldn't require too much of my brain to be distracted.)

We receive names for Trusts from Get Information About Schools in a state where all the letters are uppercase.

Exclude acronyms from titlecase transformation, so that for example 'GLF SCHOOLS' is stored as 'GLF Schools', not 'Glf Schools'.

Checking that there are no vowels will not catch all acronyms. I hard-coded acronyms that the vowel-filter didn't catch.

### Before

Screenshot 2020-12-01 at 16 21 17

### After

```ruby
pry(main)> Organisation.find_by(uid: 3190).name
=> "GLF Schools"
```

### DIFF
--- a/lib/organisation_import/import_trust_data.rb
+++ b/lib/organisation_import/import_trust_data.rb
@@ -5,6 +5,12 @@ class ImportTrustData < ImportOrganisationData
 
   MEMBERSHIP_TEMP_LOCATION = "./tmp/school-group-membership-data.csv".freeze
 
+  WORDS_TO_UPPERCASE = %w[
+    aces bmat ce csia ebn ekc epa hcat hcuk hisp hti inmat kwest link mat nas own pa pace pact polymat qed qegsmat rmet rnib rsa sdbe seax sen sendat sfaet step tbap tcat timu tvi uk utc vi xxiii
+  ].freeze
+  WORDS_TO_TITLECASE = %w[st ltd].freeze
+  WORDS_TO_DOWNCASE = %w[of and].freeze
+
   def run!
     import_data(trust_csv_url, TRUST_TEMP_LOCATION, :create_organisation)
     import_data(membership_csv_url, MEMBERSHIP_TEMP_LOCATION, :create_school_group_membership)
@@ -24,7 +30,7 @@ private
 
   def column_value_transformations
     {
-      name: :titlecase,
+      name: ->(name) { detailed_titlecase(name) },
     }
   end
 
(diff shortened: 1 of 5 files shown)


cpjmcquillan:
I would be nicer if GIAS turned off caps lock for school groups 😂

david-mears-dfe:
> It would be nicer if GIAS turned off caps lock for school groups 😂

Yes it would. I'll suggest it to them and link this PR.

david-mears-dfe:
> > It would be nicer if GIAS turned off caps lock for school groups 😂
> 
> Yes it would. I'll suggest it to them and link this PR.

The answer is pretty much that they source the data from Companies House, which is in all caps.

lib/organisation_import/import_organisation_data.rb
```diff
@@ -30,6 +30,8 @@ def set_complex_properties(organisation, row)
       organisation[attribute_name] = if attribute_name == :url
                                        # Addressable::URI ensures we store a valid URL.
                                        Addressable::URI.heuristic_parse(row[row_key]).to_s
+                                     elsif attribute_name == :name && organisation.is_a?(SchoolGroup)
```
david-mears-dfe, line 33: Check if the whole name is uppercase.

lib/organisation_import/import_trust_data.rb
```diff
@@ -5,6 +5,12 @@ class ImportTrustData < ImportOrganisationData
 
   MEMBERSHIP_TEMP_LOCATION = "./tmp/school-group-membership-data.csv".freeze
 
+  WORDS_TO_UPPERCASE = %w[
+    aces bmat ce csia ebn ekc epa hcat hcuk hisp hti inmat kwest link mat nas own pa pace pact polymat qed qegsmat rmet rnib rsa sdbe seax sen sendat sfaet step tbap tcat timu tvi uk utc vi xxiii
```
cesidio, line 9: should `own` be in this list? are these all acronyms?

cesidio (approved):
As it i's been already said, I would prefer GIAS to take care of this. We should not be doing it, but if business deemed this to be required, then let's merge it.

lib/organisation_import/import_trust_data.rb
david-mears-dfe: https://www.own-trust.education/

Yes, all acronyms - I checked.

david-mears-dfe:
> As it i's been already said, I would prefer GIAS to take care of this. We should not be doing it, but if business deemed this to be required, then let's merge it.

In fact, I decided to do this, there was no Jira ticket. So you should feel free to say that we shouldn't do it.

csutter:
So I'm in two minds about this and struggling to settle on one side – I love the UX of not haphazardly capitalising names, but like @cesidio says, a) this should ultimately be a GIAS concern (even though they just get the data from CH and thus have the same problem as us) and b) it's a lot of faff and no matter what we do we'll never be able to get it 100% right.

Devil's advocate semi-facetious suggestion: What about giving up on our current attempt to titlecase, and just presenting names in their original SHOUTING CASE? It's not pretty, but DSI already does this and it saves us from having to do guesswork 😕

david-mears-dfe:
> Devil's advocate semi-facetious suggestion: What about giving up on our current attempt to titlecase, and just presenting names in their original SHOUTING CASE? It's not pretty, but DSI already does this and it saves us from having to do guesswork 😕

That's not a bad idea. I think it will be less offensive to trusts such as GLF SCHOOLS.

david-mears-dfe:
Closing in order to reduce number of open PRs and start conversation with GIAS

PR #2466, opened 2020-12-02
[TEVA-1517] add jobseeker account job alert table markup

alexbowen:
Screenshot 2020-12-02 at 15 53 57

### DIFF
--- a/app/views/jobseekers/subscriptions/index.html.haml
+++ b/app/views/jobseekers/subscriptions/index.html.haml
@@ -4,7 +4,28 @@
 
 .govuk-width-container
   .govuk-grid-row
-    .govuk-grid-column-full-width
+    .govuk-grid-column-full
       %h1.govuk-heading-l= t(".page_title")
+
+      %table.govuk-table
+        %thead.govuk-table__head
+          %th.govuk-table__header{ class: 'govuk-!-width-two-thirds' }= t(".heading.criteria")
+          %th.govuk-table__header= t(".heading.frequency")
+          %th.govuk-table__header.govuk-table__header--numeric= t(".heading.actions")
+        %tr.govuk-table__row
+          %td.govuk-table__cell
+            %ul.govuk-list
+              %li
+                %span{ class: 'govuk-!-font-weight-bold' }= 'key'
+                = 'value'
+              %li
+                %span{ class: 'govuk-!-font-weight-bold' }= 'key'
+                = 'value'
+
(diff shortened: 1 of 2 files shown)


app/views/jobseekers/subscriptions/index.html.haml
```diff
@@ -3,8 +3,31 @@
...
+                = 'value'
+              %li
+                %span{ class: 'govuk-!-font-weight-bold' }= 'key'
+                = 'value'
+
+          %td.govuk-table__cell= 'Daily, at approx 3pm'
+          %td.govuk-table__cell.govuk-table__header--numeric
+            %ul.govuk-list
+              %li
+                %a.govuk-link{ href: '/' }= 'Manage alert'
```
cpjmcquillan, line 29: We tend to use `link_to` helpers (and soon `govuk_link_to` helpers)

alexbowen: ok i see that as being added a subsequent PR though

config/locales/jobseekers.yml
```diff
@@ -43,6 +43,11 @@ en:
         page_title: Saved jobs
     subscriptions:
       index:
+        heading_actions: Actions
```
cpjmcquillan, line 46: If I was being super anal I might say we could nest these like
```
heading:
  actions: Actions
``` 
etc, but in reality I don't mind too much

cpjmcquillan (approved):

app/views/jobseekers/subscriptions/index.html.haml
```diff
@@ -3,8 +3,29 @@
 = render partial: 'account_header'
 
 .govuk-width-container
-  .govuk-grid-row
+  .govuk-grid-row{ class: 'govuk-!-margin-left-0 govuk-!-margin-right-0' }
     .govuk-grid-column-full-width
```
cpjmcquillan, line 7: I think this class is actually called `govuk-grid-column-full`

app/views/jobseekers/subscriptions/index.html.haml
```diff
@@ -3,8 +3,29 @@
 = render partial: 'account_header'
 
 .govuk-width-container
-  .govuk-grid-row
+  .govuk-grid-row{ class: 'govuk-!-margin-left-0 govuk-!-margin-right-0' }
```
cpjmcquillan, line 6: Which would then mean you don't need these margin classes (I think)

PR #2471, opened 2020-12-03
♻️ Refactor GIAS import data transformations

csutter:
- Refactored "simple" and "complex" property setting to use a single
  mapping of GIAS keys to our attributes, and a separate set of
  transformations to be applied (using any object that can act as a
  `Proc` and be called with a value to be transformed)
- Call `#presence` on all values by default (instead of specifically
  just for a few fields)
- Removed specs testing internal implementation details of import
  classes

Co-authored-by: David Mears <david.mears@digital.education.gov.uk>

### DIFF
--- a/lib/organisation_import/import_school_data.rb
+++ b/lib/organisation_import/import_school_data.rb
@@ -22,23 +22,43 @@ class ImportSchoolData < ImportOrganisationData
 
 private
 
-  # These are the attributes that require additional transformation before being added to the model. The first value of
-  # the array is the row key name, the second is the method used for the transformation. URL is the exception, as it
-  # requires an external function call - this is handled in the set_complex_properties method.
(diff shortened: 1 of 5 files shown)


lib/organisation_import/import_school_data.rb
```diff
@@ -22,23 +22,43 @@ def run!
...
+      school_type: "EstablishmentTypeGroup (name)",
+      town: "Town",
+      url: "SchoolWebsite",
     }.freeze
   end
 
+  def column_value_transformations
+    {
+      phase: :to_i,
+      url: ->(url) { Addressable::URI.heuristic_parse(url).to_s },
```
cesidio, line 51: very nice!

cesidio (approved):
👍  love it

PR #2476, opened 2020-12-04
Reduce Devise patching

david-mears-dfe:
The original Devise method looks like this:

      def after_sign_in_path_for(resource_or_scope)
        stored_location_for(resource_or_scope) || signed_in_root_path(resource_or_scope)
      end

Devise lets us specify the jobseeker_root_path in routes, for it to find with its `after_sign_in_path_for` method.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,6 +7,8 @@ Rails.application.routes.draw do
       sessions: "jobseekers/sessions",
     }
 
+    get "/jobseekers/saved_jobs", to: "jobseekers/saved_jobs#index", as: :jobseeker_root
+
     namespace :jobseekers do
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -6,6 +6,8 @@
       sessions: "jobseekers/sessions",
     }
 
+    get "/jobseekers/saved_jobs", to: "jobseekers/saved_jobs#index", as: :jobseeker_root
```
david-mears-dfe, line 9: Can't go in `namespace :jobseekers` because that would prepend 'jobseekers' to the path helper method name.

alexbowen (approved):
makes sense i think, approving pending rebase

PR #2479, opened 2020-12-04
[TEVA-1510] Add reset password functionality

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1510

## Review app
https://teaching-vacancies-review-pr-2479.london.cloudapps.digital

## Screenshots of UI changes:
Screenshot 2020-12-04 at 16 56 32
Screenshot 2020-12-04 at 16 57 04
Screenshot 2020-12-04 at 16 57 24
Screenshot 2020-12-04 at 16 58 47
Screenshot 2020-12-04 at 16 59 07
Screenshot 2020-12-04 at 17 01 04
Screenshot 2020-12-04 at 17 03 46
Screenshot 2020-12-04 at 16 56 04


app/controllers/jobseekers/passwords_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  before_action :check_reset_password_token, only: %i[edit]
+  after_action :remove_devise_notice_flash!, only: %i[create]
+
+protected
+
+  def reset_password_token
+    Devise.token_generator.digest(resource_class, :reset_password_token, params[:reset_password_token] || params[resource_name][:reset_password_token])
+  end
+
+  def set_jobseeker_from_token
```
cpjmcquillan, line 12: I'm surprised we don't get more of this for free

config/initializers/devise.rb
```diff
@@ -133,7 +133,7 @@
   config.send_email_changed_notification = true
 
   # Send a notification email when the user's password is changed.
-  config.send_password_change_notification = true
+  config.send_password_change_notification = false
```
cpjmcquillan, line 136: I thought we wanted this to be true?

app/controllers/jobseekers/passwords_controller.rb
cesidio: the problem is that at that stage it does not retrieve the model by default. Only when you submit the token and the new password.

config/initializers/devise.rb
cesidio: this is not when we request to change the password, this is when a new password has been set. For the moment we do not have a template for that, but we should probably add it back.

spec/support/mailer_helpers.rb
```diff
@@ -0,0 +1,13 @@
+module MailerHelpers
+  def last_email
+    ActionMailer::Base.deliveries.last
+  end
+
+  def delivered_emails
```
cpjmcquillan, line 6: Do you want to update `jobseekers_can_sign_up_to_an_account_spec.rb` please? 🙏

cpjmcquillan (approved):

PR #2483, opened 2020-12-07
Update years given in example dates in date form

david-mears-dfe:
These should be updated around December 14th each year to be the following year, said Valentine, our outgoing content designer.

https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1603900573348700


csutter:
I wonder how easy (or not) it would be to have these dynamic 🤔

david-mears-dfe:
We could do it, but then I feel like it might produce more clutter (code for people to read and decipher) than it's really worth.

cpjmcquillan (approved):

PR #2485, opened 2020-12-07
Remove table component and use GovukComponent::SummaryList

cpjmcquillan:
Pretty much what it says on the tin


alexbowen (approved):
lgtm and this should play nicely with mobile, good work

PR #2486, opened 2020-12-07
[TEVA-1684] fix notification bg color/spacing/layout for all pages Part 1

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1684

There is more work to do to decouple dashboard styles from 'mast' area styles but this will be done in subsequent PR

### DIFF
--- a/app/frontend/src/styles/application/dashboard.scss
+++ b/app/frontend/src/styles/application/dashboard.scss
@@ -7,8 +7,7 @@
   width: 100vw;
 
   > .govuk-width-container {
-    background-color: govuk-colour('white');
-    padding-top: govuk-spacing(7);
+    background-color: inherit;
   }
 
   .moj-primary-navigation {
--- a/app/views/publishers/organisations/schools/edit.html.haml
+++ b/app/views/publishers/organisations/schools/edit.html.haml
@@ -1,4 +1,4 @@
-= content_for :page_title_prefix, @organisation.name.titlecase
+- content_for :page_title_prefix, @organisation.name.titlecase
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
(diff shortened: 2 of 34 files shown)


app/frontend/src/styles/application/dashboard.scss
```diff
@@ -7,8 +7,7 @@
   width: 100vw;
 
   > .govuk-width-container {
-    background-color: govuk-colour('white');
-    padding-top: govuk-spacing(7);
+    background-color: inherit;
```
cpjmcquillan, line 10: nice 👍

app/views/publishers/organisations/schools/edit.html.haml
```diff
@@ -1,4 +1,6 @@
-= content_for :page_title_prefix, @organisation.name.titlecase
+.govuk-grid-row{ class: 'govuk-!-margin-top-7' }
```
cpjmcquillan, line 1: I'd imagine this should actually just be
```
- content_for :page_title_prefix, @organisation.name.titlecase
```

app/views/publishers/organisations/schools/index.html.haml
```diff
@@ -1,4 +1,6 @@
-= content_for :page_title_prefix, current_organisation.name
+.govuk-grid-row{ class: 'govuk-!-margin-top-7' }
```
cpjmcquillan, line 1: Same as above

david-mears-dfe:
Why do some pages have margin-top-7 and others have margin-top-4?

alexbowen:
> Why do some pages have margin-top-7 and others have margin-top-4? I'm sure there's a reason but perhaps it could be made clearer either in the commit message or with some clever class naming.

sorted this out, all done in css rather than repetitious classes being added

csutter (approved):

PR #2492, opened 2020-12-08
Send page view events to BigQuery

csutter:
This is an initial version of sending events to BigQuery from within
our application (using the Streaming Insert API). It sends one event to
BigQuery for every (non-error) page view. In the future, we will want to
send other types of events as well, such as specific user actions or
errors, as well as including additional data, such as selected variants
for currently running AB tests.

- Create `Event` class to encapsulate an arbitrary event within the app
  (to be used in non-request lifecycle situations, e.g. within a
  background job)
- Create `RequestEvent` class to represent an `Event` that happens as
  part of the Rails request lifecycle (i.e. is triggered from a
  controller) and includes information about the context of the request
- Create `SendEventToDataWarehouseJob` to send event data to BigQuery
  asynchronously
- Trigger a basic `page_view` event for every successful request (using
  `RequestEvent` with no additional metadata)

To try it out, do something on the review app then visit the BigQuery console and run:
```sql
SELECT *
FROM `teacher-vacancy-service.staging_dataset.events`
ORDER BY occurred_at DESC
LIMIT 1000;
```

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1443

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,6 +12,8 @@ class ApplicationController < ActionController::Base
   before_action :set_headers, :set_root_headers
   before_action { strip_nested_param_whitespaces(request.params) }
 
+  after_action :trigger_page_visited_event, unless: :request_is_paas_healthcheck?
+
   helper_method :cookies_preference_set?, :referred_from_jobs_path?, :utm_parameters
 
   include Publishers::AuthenticationConcerns
@@ -70,11 +72,11 @@ private
   end
 
   def request_host_is_invalid?
-    !Rails.env.test? && request_is_not_paas_healthcheck? && request.host_with_port != DOMAIN
+    !Rails.env.test? && !request_is_paas_healthcheck? && request.host_with_port != DOMAIN
   end
 
-  def request_is_not_paas_healthcheck?
-    request.headers.env["HTTP_USER_AGENT"] != "diego-healthcheck"
+  def request_is_paas_healthcheck?
+    request.headers["User-Agent"] == "diego-healthcheck"
   end
 
   def set_root_headers
@@ -108,5 +110,13 @@ private
 
   def remove_devise_notice_flash!
     flash.discard(:notice) if flash[:notice].present?
+  end
+
+  def request_event
+    @request_event ||= RequestEvent.new(request, response, session, current_jobseeker, current_publisher_oid)
+  end
+
+  def trigger_page_visited_event
+    request_event.trigger(:page_visited)
   end
 end
--- a/app/services/event.rb
+++ b/app/services/event.rb
@@ -1,0 +1,35 @@
+##
+# Represents events occurring within the application that we are interested in tracking.
+#
+# At a minimum, an event has a type and a timestamp at which it occurred. It may also have optional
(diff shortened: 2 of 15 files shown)


app/services/event.rb
```diff
@@ -0,0 +1,36 @@
...
+# warehouse using a background job. An instance of `Event` can trigger an arbitrary number of
+# events (to allow for potentially expensive computation on initialization of subclasses).
+class Event
+  TABLE_NAME = "events".freeze
+
+  ##
+  # Asynchronously sends an event and its metadata to the data warehouse
+  #
+  # @param [Symbol, String] event_type The type of event (e.g. `:page_view`) to trigger
+  # @param [Hash{Symbol => Object}] data An optional hash of data to include with the event
```
csutter, line 15: Interested to hear how people feel about this level of documentation (using YARD) for "library" style services like this one – too verbose?

app/services/event.rb
```diff
@@ -0,0 +1,36 @@
+##
+# Represents events occurring within the application that we are interested in tracking.
```
cpjmcquillan, line 2: I don't mind so much documentation like this that adds some context to a library style service

app/services/event.rb
```diff
@@ -0,0 +1,36 @@
...
+  #   (Important: if present, values will be coerced into Strings)
+  def trigger(event_type, data = {})
+    data = base_data.merge(
+      type: event_type,
+      occurred_at: Time.now.utc.iso8601(6),
+      data: data.map { |key, value| { key: key.to_s, value: value&.to_s } },
+    )
+    SendEventToDataWarehouseJob.perform_later(TABLE_NAME, data)
+  rescue StandardError => e
+    # We do not want failing to fire an event to impact fulfilling a request, but we do want to be aware
```
cpjmcquillan, line 25: But things like this I find less useful/like less

app/controllers/application_controller.rb
```diff
@@ -109,4 +111,14 @@ def replace_devise_notice_flash_with_success!
...
   end
+
+  def request_event
+    @request_event ||= RequestEvent.new(request, response, session, current_jobseeker, current_publisher_oid)
+  end
+
+  def trigger_page_view_event
+    return if request_is_paas_healthcheck?
+
+    request_event.trigger(:page_view)
```
cpjmcquillan, line 122: I quite like `request_event.trigger(:page_view) unless request_is_paas_healthcheck?`

app/services/request_event.rb
```diff
@@ -0,0 +1,54 @@
...
+  ##
+  # Hashes potentially identifiable information. Uses "bubblebabble" algorithm to make the
+  # resulting SHA hash more human-readable.
+  #
+  # @param [#to_s] identifier An item of identifiable information to anonymise
+  # @return [String] a human-readable anonymised identifier
+  def anonymise(identifier)
+    return nil if identifier.blank?
+
+    Digest::SHA256.bubblebabble(identifier.to_s)
```
cpjmcquillan, line 52: and `Digest::SHA256.bubblebabble(identifier.to_s) unless identifier.blank?`, but I understand the guard clause is a bit more explicit

cpjmcquillan (approved):

PR #2499, opened 2020-12-09
Refactor table sorting

cpjmcquillan:
This PR abstracts away table sorting from the publisher jobs dashboard and adds sorting to the jobseeker saved jobs table. It should also make everybody else's lives easier when they need to add sorting to any other tables.

**FYI**
Brakeman doesn't like the ordering of saved jobs. I'm not sure of a cleaner way to do this when we want to order based on an association's attributes, however given that `RecordSort` explicitly allows only certain sort columns and orders, I don't think there is an SQL injection risk

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1516 (kind of)

## Changes in this PR
- Refactor `table_header_sort_by` helper
- Add sorting to jobseeker saved jobs table


app/controllers/jobseekers/saved_jobs_controller.rb
```diff
@@ -16,7 +16,8 @@ def destroy
   end
 
   def index
-    @saved_jobs = current_jobseeker.saved_jobs
+    @sort = SavedJobSort.new.update(column: sort_column, order: sort_order)
+    @saved_jobs = current_jobseeker.saved_jobs.includes(:vacancy).order("#{@sort.column} #{@sort.order}")
```
csutter, line 20: Stabbing in the dark but would using the Hash form of `#order` arguments help with your Brakeman woes?

```ruby
.order(@sort.column => @sort.order)
```

app/services/record_sort.rb
```diff
@@ -0,0 +1,22 @@
+class RecordSort
+  attr_reader :column, :order
+
+  def initialize
+    @column = default_column
+    @order = default_order
+  end
+
+  def update(column:, order:)
+    @column = valid_sort_columns.include?(column) ? column : @column
```
csutter, line 10: I think this would be more readable:

```ruby
@column = column if valid_sort_columns.include?(column)
@order = order if valid_sort_orders.include?(order)
```

app/helpers/sort_helper.rb
```diff
@@ -0,0 +1,15 @@
+module SortHelper
+  def table_header_sort_by(heading, path, column:, sort:)
+    order = column == sort.column ? sort.reverse_order : sort.order
+    extra_classes = "sortby--#{order}"
+    extra_classes += " active" if column == sort.column
+
+    govuk_link_to(heading, "#{path}?#{params_with_sort(column, order).to_query}",
```
csutter, line 7: Is the path for sorting always going to be the same as the path the user is currently on? If so it might be easier to just leave the path off entirely and do:

```ruby
govuk_link_to(heading, params_with_sort(column, order), ...)
```

... I think. It's been a long day ☕

app/controllers/jobseekers/saved_jobs_controller.rb
cpjmcquillan: That's what I would normally do but it doesn't work for association attributes

csutter: 😞  In which case this is fine, given we're explicitly allowing a bunch of attributes.

csutter: Hang on, I just spotted the Brakeman error. I think what it's _actually_ complaining about is the call to `SavedJobSort#update`, it thinks that `SavedJobSort` is an ActiveRecord model.

In which case you can just tell Brakeman to ignore it: https://brakemanscanner.org/docs/ignoring_false_positives/ (tl;dr: run `brakeman -I`, pick option 2 (Hide previously ignored warnings), and choose `n` (note) as the action, write something poetic, save).

alexbowen (approved):
pending fixing the test(s), this LGTM

app/controllers/jobseekers/saved_jobs_controller.rb
```diff
@@ -35,4 +36,12 @@ def set_up_saved_job
     end
     @saved_job = SavedJob.find_or_initialize_by(jobseeker_id: current_jobseeker.id, vacancy_id: @vacancy.id)
   end
+
+  def sort_column
```
cesidio, line 40: How about moving these methods in a concern like `Sortable`?

app/helpers/sort_helper.rb
```diff
@@ -0,0 +1,15 @@
+module SortHelper
```
cesidio, line 1: helpers names should be plural like controller names. Not sure in this case

cesidio (approved):

PR #2503, opened 2020-12-09
[TEVA-1503] Jobseeker sign in: Part 1, Version 3

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1503

## Changes in this PR:

This reimplements https://github.com/DFE-Digital/teaching-vacancies/pull/2489.

- Add sign in page with form validation errors for email format, email and password presence

    Content has deemed it to be important for us to do some extension of Devise in order to make this very frequently used form more user-friendly. I don't believe extending Devise in order to include a form model introduces too much code complexity relative to the need for this page to be as user-friendly as we can afford. We'll be allowed more leeway to keep code simple on other less important forms. Significantly different code, so a different PR.

Also:

- Put back the errors on the sign up page that got broken
- Fix the font size for page headings on various jobseeker pages
- Put the form error summaries between the title and the description on various jobseeker pages by request of UX
- Add a missing description to the change password page

## Questions

I'm not 100% happy with my solution to the problem of how to use govuk error summary components in place of the Devise flash alerts. Currently, I am solving it by, after being returned to the new sessions page after an attempted form submission, checking for any Devise flash alerts and then adding their text as an error on the form model.

An alternative idea would be to discard any flash messages, and then use a custom validation on the form to directly check whether `jobseeker.valid_password?(password)`, and then add the error on the form model.

What would be best?

## Screenshots of UI changes:

Screenshot 2020-12-09 at 20 06 41

Screenshot 2020-12-09 at 20 07 26

## Next steps:

- [ ] Locking users out after 10 failed attempts and unlocking the account

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -105,7 +105,9 @@ private
   end
 
   def replace_devise_notice_flash_with_success!
-    flash[:success] = flash.discard(:notice) if flash[:notice].present?
+    return unless flash[:notice].present?
+
+    flash[:success] = flash.discard(:notice)
   end
 
   def remove_devise_notice_flash!
--- a/app/controllers/jobseekers/sessions_controller.rb
+++ b/app/controllers/jobseekers/sessions_controller.rb
@@ -3,11 +3,10 @@ class Jobseekers::SessionsController < Devise::SessionsController
   before_action :sign_out_publisher!, only: %i[create]
 
   def new
-    if [I18n.t("devise.failure.invalid"), I18n.t("devise.failure.not_found_in_database")].include? flash[:alert]
-      self.resource = resource_class.new(sign_in_params)
-      resource.errors.add(:email, flash[:alert])
+    @sign_in_form = JobseekerSignInForm.new(flash[:alert], sign_in_params)
+    if params[:action] == "create" && @sign_in_form.invalid?
       flash.clear
-      render :new and return
+      render :new
     else
       super
     end
--- a/app/controllers/jobseekers/sessions_controller.rb
+++ b/app/controllers/jobseekers/sessions_controller.rb
@@ -1,14 +1,26 @@
 class Jobseekers::SessionsController < Devise::SessionsController
   after_action :replace_devise_notice_flash_with_success!, only: %i[create destroy]
   before_action :sign_out_publisher!, only: %i[create]
+  before_action :render_form_with_errors, only: %i[new]
 
-  def new
-    @sign_in_form = JobseekerSignInForm.new(flash[:alert], sign_in_params)
-    if params[:action] == "create" && @sign_in_form.invalid?
-      flash.clear
-      render :new
-    else
-      super
+  AUTHENTICATION_FAILURE_MESSAGES = %w[
+    invalid not_found_in_database locked last_attempt
+  ].map { |error| I18n.t("devise.failure.#{error}") }.freeze
+
+private
+
+  def render_form_with_errors
+    self.resource = JobseekerSignInForm.new(sign_in_params)
+    if params[:action] == "create" && resource.invalid?
+      clear_flash_and_render(:new)
+    elsif AUTHENTICATION_FAILURE_MESSAGES.include?(flash[:alert])
+      resource.errors.add(:email, flash[:alert])
+      clear_flash_and_render(:new)
     end
+  end
+
+  def clear_flash_and_render(view)
+    flash.clear
+    render view
   end
 end
--- a/app/controllers/jobseekers/sessions_controller.rb
+++ b/app/controllers/jobseekers/sessions_controller.rb
@@ -1,4 +1,15 @@
 class Jobseekers::SessionsController < Devise::SessionsController
   after_action :replace_devise_notice_flash_with_success!, only: %i[create destroy]
   before_action :sign_out_publisher!, only: %i[create]
+
+  def new
+    if [I18n.t("devise.failure.invalid"), I18n.t("devise.failure.not_found_in_database")].include? flash[:alert]
+      self.resource = resource_class.new(sign_in_params)
+      resource.errors.add(:email, flash[:alert])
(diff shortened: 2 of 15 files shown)


app/controllers/application_controller.rb
```diff
@@ -103,7 +103,9 @@ def strip_nested_param_whitespaces(object)
   end
 
   def replace_devise_notice_flash_with_success!
-    flash[:success] = flash.discard(:notice) if flash[:notice].present?
+    return unless flash[:notice].present?
+
+    flash[:success] = flash.discard(:notice)
```
david-mears-dfe, line 108: This just looks nicer to me with a guard clause.

app/views/jobseekers/confirmations/new.html.haml
```diff
@@ -2,7 +2,7 @@
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
-    %h1.govuk-heading-l= t(".title")
+    %h1.govuk-heading-xl= t(".title")
```
cesidio, line 5: I think we should start with the habit of small PRs. This should be another small PR that does not distract the reviewer from what is the main purpose of the PR

david-mears-dfe: I have moved it to a separate PR.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2503.london.cloudapps.digital>

app/form_models/jobseeker_sign_in_form.rb
```diff
@@ -0,0 +1,30 @@
+class JobseekerSignInForm
+  include ActiveModel::Model
+
+  AUTHENTICATION_FAILURE_MESSAGES = [
```
cesidio, line 4: This could just be:

`AUTHENTICATION_FAILURE_MESSAGES = %w[invalid not_found_in_database locked last_attempt].map { |error| I18n.t("devise.failure.#{error}" }`

app/views/jobseekers/passwords/edit.html.haml
```diff
@@ -9,6 +9,6 @@
 
       = f.govuk_error_summary
 
-      = f.govuk_password_field :password, label: { text: t(".new_password"), size: "s" }, width: "two-thirds"
+      = f.govuk_password_field :password, label: { text: t(".new_password"), size: "s" }, hint: { text: t("helpers.hint.jobseeker_sign_up.password") }, width: "two-thirds"
```
cesidio, line 12: You shouldn't define the hint here. It should come directly from the the translations

app/controllers/jobseekers/sessions_controller.rb
```diff
@@ -1,4 +1,14 @@
 class Jobseekers::SessionsController < Devise::SessionsController
   after_action :replace_devise_notice_flash_with_success!, only: %i[create destroy]
   before_action :sign_out_publisher!, only: %i[create]
+
+  def new
```
cesidio, line 5: how about doing this instead:

```
  def replace_devise_error_flash_with_form_errors
    return unless AUTHENTICATION_FAILURE_MESSAGES.include?(flash[:alert])

    message = flash[:alert]
    flash.clear
    render_resource_with_error(:new, :email, message)
  end
```

You would need to move the method `render_resource_with_error` from `RegistrationsController` to `ApplicationController` and take as parameter the view as well

david-mears-dfe: There's two problems with this, but it's a helpful idea that will help me come up with a better solution.

The main problem is that the before_action skips the form validation errors - it immediately checks whether Devise could authenticate and renders if it can't.

Secondly, I can't use resource_class here. I have to use a form object, in order to keep separate the different validations on the sign in / sign up pages (e.g. non-unique email address). But that problem can easily be solved by writing a `render_form_with_error` method.

How about this?

```ruby
def new
    @sign_in_form = JobseekerSignInForm.new(sign_in_params)

    if params[:action] == "create" && @sign_in_form.invalid?
      clear_flash_and_render(:new)
    elsif AUTHENTICATION_FAILURE_MESSAGES.include?(flash[:alert])
      message = flash[:alert]
      @sign_in_form.errors.add(:email, message)
      clear_flash_and_render(:new)
    else
      super
    end
  end

private

  def clear_flash_and_render(view)
    flash.clear
    render view
  end
```

spec/system/jobseekers_can_save_a_job_spec.rb
```diff
@@ -68,7 +68,7 @@
     it "saves the job after signing up" do
       save_job
       expect(page).to have_content(I18n.t("messages.jobseekers.saved_jobs.unauthenticated"))
-      click_on "Sign up"
+      click_on "create an account"
```
cesidio, line 71: I would use a I18n key for this

cesidio (approved):

app/controllers/jobseekers/sessions_controller.rb
```diff
@@ -1,4 +1,26 @@
...
+  ].map { |error| I18n.t("devise.failure.#{error}") }.freeze
+
+private
+
+  def render_form_with_errors
+    self.resource = JobseekerSignInForm.new(sign_in_params)
+    if params[:action] == "create" && resource.invalid?
+      clear_flash_and_render(:new)
+    elsif AUTHENTICATION_FAILURE_MESSAGES.include?(flash[:alert])
+      resource.errors.add(:email, flash[:alert])
```
cesidio, line 17: You could still reuse the `render_resource_with_error` in registrations_controller if you move it to the application controller and take the view as parameter as well

app/controllers/application_controller.rb
cesidio: to me it was more readable the way it was before, but it is personal taste. I can live with this

david-mears-dfe: I hope you don't mind.

PR #2507, opened 2020-12-10
Fix flaky reset password spec

cesidio:
The way we parsed the token in the reset password email sent was not considering the fact that there could be non characters signs


cpjmcquillan (approved):
Well done 🎉

PR #2517, opened 2020-12-11
Fix wrong IPs in logs, take two

csutter:
So it turns out my previous PR (#2501) was a bit overenthusiastic – while the setting of the trusted proxies _was_ indeed the root cause, we still need the AWS IP ranges to be excluded when determining a user's IP. 😞 

However, these IP ranges need to be _appended_ to the default set instead of replacing them, otherwise we do not clean out private PaaS IPs from `X-Forwarded-For`.

I initially thought it worked because I was looking at the review app, which wasn't behind Cloudfront (duh 🤦).


csutter:
SonarQube is complaining about duplicated lines, maybe we can just kill off the `staging` Rails env as part of the next "firebreak"? 😅

spec/lib/modules/aws_ip_ranges_spec.rb
```diff
@@ -0,0 +1,115 @@
+require "rails_helper"
+
+RSpec.describe AWSIpRanges do
+  describe ".cloudfront_ips" do
+    before(:each) do
```
cpjmcquillan, line 5: I might be being anal, but I prefer us to just use `before do` since they do the same thing now

csutter: Fair – although this is just reverting a commit to restore existing code!

spec/lib/modules/aws_ip_ranges_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      aws_ip_ranges = File.read(Rails.root.join("spec/fixtures/aws_ip_ranges.json"))
```
cesidio, line 6: You can use `file_fixture("aws_ip_ranges.json")`

cesidio (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2517.london.cloudapps.digital>

PR #2520, opened 2020-12-12
[TEVA-1625] fixes lingering page layout issues

alexbowen:
no Jira ticket

unfortunatley still some layout problems as a result of recent page reworking to accomodate dashboard page layouts. you will see in jobseeker mode when you resize the screen its not quite right at certain widths

- max-width of the main `govuk-width-container` was changed as it was conditional to user signed in status. this was the right thing to do but the max width needs to be less than the main app container or the margin disappears on smaller desktop views. so now the `govuk-width-container` max-width is calculated as main app container width minus 10% up to `$desktop-container-width`
- the main header navigation was not switching at correct breakpoint and this meant the navigation was not appearing correctly in menu dropdown for tablet screen size
- there were conflicting overrides of some inline GDS styles that was leading to inconsistency of vertical spacing on certain pages - these have been removed

### DIFF
--- a/app/components/shared/navbar_component/navbar_component.scss
+++ b/app/components/shared/navbar_component/navbar_component.scss
@@ -1,6 +1,6 @@
 .govuk-header {
   .govuk-header__navigation {
-    @include govuk-media-query($from: tablet) {
+    @include govuk-media-query($from: desktop) {
       display: flex;
 
       .govuk-header__navigation-spacer {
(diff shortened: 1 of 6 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2520.london.cloudapps.digital>

david-mears-dfe (approved):

cesidio (approved):

PR #2532, opened 2020-12-15
Controller refactoring

david-mears-dfe:
- Rewrote a method to read more analogously to its neighbours
- Removed method that needlessly reimplements Devise functionality

### DIFF
--- a/app/controllers/jobseekers/passwords_controller.rb
+++ b/app/controllers/jobseekers/passwords_controller.rb
@@ -14,9 +14,4 @@ protected
   def after_sending_reset_password_instructions_path_for(_resource)
     jobseekers_check_your_email_password_path
   end
-
-  def after_resetting_password_path_for(resource)
-    sign_in(resource)
-    jobseekers_saved_jobs_path
-  end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/jobseekers/passwords_controller.rb
```diff
@@ -14,9 +14,4 @@ def ensure_reset_password_period_valid
   def after_sending_reset_password_instructions_path_for(_resource)
     jobseekers_check_your_email_password_path
   end
-
-  def after_resetting_password_path_for(resource)
```
cesidio, line 18: This is needed to sign in the and redirect the user to `jobseekers_saved_jobs_path`

david-mears-dfe: See the commit message! It's not needed.

david-mears-dfe: Maybe when you were writing this controller, the routes.rb did not yet contain 'jobseeker_root', which is what Devise uses to determine the path for after sign in and for after resetting the password.

cesidio (approved):

app/controllers/jobseekers/passwords_controller.rb
cpjmcquillan: We need to sign in though still, don't we?

PR #2533, opened 2020-12-15
Move homepage to new HomeController

csutter:
The app homepage currently lives in the `PagesController`, which made
sense originally when it was just a totally static page. Now that it
has a fair bit of logic behind it and no longer depends on HighVoltage,
it deserves a place of its own (not to mention that `pages/home` exists
but is broken).

- Create `HomeController` and move homepage logic there
- Adjust translations to live under correct keys for partials
- Reuse `ApplicationController#not_found` in `PagesController` when page
  isn't found (instead of redirecting to /404)
- Remove header setting for root path and long gone pages from
  `PagesController`
- Simplify robots header setting


cpjmcquillan (approved):

app/views/home/_signin_publishers.html.haml
```diff
@@ -2,8 +2,8 @@
 
 - if publisher_signed_in?
   %div.manage-vacancies
-    %p= "#{t('pages.home.signed_in.description')} #{current_organisation.name}"
-    = link_to t('pages.home.signed_in.manage_link'), organisation_path, class: 'govuk-button govuk-button--secondary manage-vacancies'
+    %p= "#{t('home.index.signed_in.description')} #{current_organisation.name}"
```
cesidio, line 5: Can we use the short lookup `t('.description')` and put the translation under
```
home:
  signin_publishers:
    description: ....
```

csutter: 👍  I was sure these were used somewhere else and that's why they had absolute paths but they're actually not!

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2533.london.cloudapps.digital>

PR #2538, opened 2020-12-16
[TEVA-1511] Update jobseeker email

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1511

## Changes in this PR
- Refactor edit/update jobseeker registration to account for updating email **or** password
- Add email changed notification email
- Reuse confirmation instructions email for reconfirming email address (with different translations)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2538.london.cloudapps.digital>

app/controllers/jobseekers/registrations_controller.rb
```diff
@@ -1,38 +1,52 @@
...
+  end
+
   def render_resource_with_error(field, error)
     self.resource = resource_class.new
     resource.errors.add(field, error)
     render :edit
   end
 
+  def set_update_password
+    @update_password = params[:update_password] == "true" || params[:commit] == t("buttons.update_password")
```
david-mears-dfe, line 42: Does rubocop object to:
```ruby
    @update_password = (params[:update_password] == "true") || (params[:commit] == t("buttons.update_password"))
```

Bit easier on the old eyes imo.

david-mears-dfe:
You don't mention whether or not Design has given the thumbs up to your straying off the path laid out for us on [Figma (Screen 6)](https://www.figma.com/file/IB5MUgCC5Kf084XxafUjG1/Sprint-66?node-id=2%3A1347). Or perhaps (given this is WIP PR) this is not a final product?

Screenshot 2020-12-16 at 18 16 50

config/locales/forms.yml
```diff
@@ -58,7 +58,7 @@ en:
         expires_at: For example, 9:30 am or 11:59 pm (midnight)
         publish_on: For example, 01 01 2021
         starts_on: For example, 01 09 2021
-      jobseeker_sign_up:
+      jobseeker:
```
david-mears-dfe, line 61: 👍

app/views/jobseekers/accounts/show.html.haml
```diff
@@ -19,7 +19,7 @@
         %dt.govuk-summary-list__key= t(".summary_list.password")
         %dd.govuk-summary-list__value= t(".summary_list.password_placeholder")
         %dd.govuk-summary-list__actions
-          = govuk_link_to edit_jobseeker_registration_path do
+          = govuk_link_to edit_jobseeker_registration_path(update_password: true) do
```
david-mears-dfe, line 22: What do you think about setting `to_update: 'password'` or `to_update: 'email'`, instead of having email as the default and password as the exception?

It would be more verbose in the controller, but more explicit in the view (as there isn't a reason for email to be the default field to update except for historical contingency - the email part was written first).

```ruby
class Jobseekers::RegistrationsController < Devise::RegistrationsController
  def updating_password?
    params[:to_update] == "password" || params[:commit] == t("buttons.update_password")
  end
  
  def updating_email?
    params[:to_update] == "email" || params[:commit] == t("whatever")
  end  
end
```

david-mears-dfe: Or, with a little edit to the .yml, in the view you could have e.g.
```ruby
t(".#{ @to_update }.title")
```
Instead of various conditionals.

app/mailers/jobseeker_mailer.rb
```diff
@@ -9,11 +9,21 @@ def confirmation_instructions(record, token, _opts = {})
...
     )
   end
 
+  def email_changed(record, _opts = {})
+    @jobseeker = record
+
+    view_mail(
+      NOTIFY_JOBSEEKER_EMAIL_CHANGED_TEMPLATE,
+      to: @jobseeker.email,
+      subject: I18n.t("jobseeker_mailer.email_changed.subject"),
```
cesidio, line 23: You can use the short lookup here too, just use t(".subject") and put the translation in the right place

app/controllers/jobseekers/registrations_controller.rb
cesidio: I prefer minimal code, and operators precedence is something that should be well known

app/views/jobseekers/accounts/show.html.haml
cesidio: It is clear enough to me what this code is doing and being more explicit and more verbose do not add much here. But that is just my opinion

app/views/jobseekers/registrations/edit.html.haml
```diff
@@ -4,16 +4,20 @@
...
       = f.govuk_error_summary
-    
-      %p.govuk-body= t(".description", email: resource.email)
 
-      = f.govuk_password_field :current_password, label: { text: t(".current_password"), size: "s" }, width: "two-thirds"
-      = f.govuk_password_field :password, label: { text: t(".new_password"), size: "s" }, hint: { text: t("helpers.hint.jobseeker_sign_up.password") }, width: "two-thirds"
-
-      = f.govuk_submit t("buttons.update_password"), classes: "govuk-!-margin-bottom-0"
+      - if @update_password
+        %p.govuk-body= t(".update_password.description", email: resource.email)
```
cesidio, line 13: Here you could avoid the `.update_password.xxx` translations by having 2 different partials

app/controllers/jobseekers/registrations_controller.rb
david-mears-dfe: It's more that I have to re-parse the sentence a few times as I go along. https://en.wikipedia.org/wiki/Garden-path_sentence#Examples

I'll leave it to Connor's taste.

app/views/jobseekers/registrations/edit.html.haml
cpjmcquillan: Nice idea, thanks

david-mears-dfe (approved):

spec/system/jobseekers_can_change_email_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(delivered_emails.second.subject).to eq(I18n.t("jobseeker_mailer.confirmation_instructions.reconfirmation.subject"))
+      expect(delivered_emails.second.to.first).to eq("new@email.com")
+      expect(current_path).to eql(jobseekers_check_your_email_path)
+    end
+  end
+
+  describe "confirming email" do
+    it "confirms the email, updates the email, and redirects to saved_jobs page" do
+      update_jobseeker_email("new@email.com", jobseeker.password)
+      confirm_email_address
```
david-mears-dfe, line 30: Should we use `visit first_link_from_last_mail` here, so that we can be sure the link and token are included in the email?

cpjmcquillan: Going to raise a separate PR for implementing the use of that helper across jobseeker system specs

cesidio (approved):

PR #2539, opened 2020-12-16
[TEVA-1503] Part 2: Jobseekers can unlock their accounts

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1503

## Changes in this PR:

- Uses the Devise module Lockable.

- Add 'too many attempts' page. We redirect to this page with a before_action hook if the jobseeker's account is locked, replacing the inbuilt Devise functionality (which is to reload the sign in page with a flash notification.)

- Adapt the styling of Devise's `views/unlocks/new` view to match our site. This is the page which Devise redirects to if an unlock token is invalid. I have done the minimal adaptations to make this page presentable and consistent with our site: e.g. not using Times New Roman. This page is awaiting some content design which can be inserted into the translations file without breaking things.

## Screenshots of UI changes:

Screenshot 2020-12-16 at 17 19 28

Screenshot 2020-12-16 at 17 19 56

Screenshot 2020-12-16 at 17 20 32

Screenshot 2020-12-17 at 17 13 38

Screenshot 2020-12-16 at 17 21 50

## Next steps:

- [x] Update content placeholders. This can be a separate PR since the page is hidden behind a feature flag. However, we must not forget to do this.

- [x] Reviews from Content and UX Design


david-mears-dfe:
Review app: https://teaching-vacancies-review-pr-2539.london.cloudapps.digital/

spec/support/mailer_helpers.rb
```diff
@@ -7,7 +7,7 @@ def delivered_emails
     ActionMailer::Base.deliveries
   end
 
-  def devise_token_from_last_mail(token_name)
-    last_email.body.to_s[/#{token_name}_token=([^)]*)/, 1]
+  def first_link_from_last_mail
```
cesidio, line 10: For now this works because the first link is always what we need in the test, which might still be the case in the future as these emails are pretty standard. But the previous method was specific in finding the specific token you need.

cesidio (approved):

app/mailers/jobseeker_mailer.rb
```diff
@@ -24,4 +24,15 @@ def reset_password_instructions(record, token, _opts = {})
...
   end
+
+  def unlock_instructions(record, token, _opts = {})
+    @jobseeker = record
+    @token = token
+
+    view_mail(
+      NOTIFY_JOBSEEKER_LOCKED_ACCOUNT_TEMPLATE,
+      to: @jobseeker.email,
+      subject: I18n.t("jobseeker_mailer.unlock_instructions.subject"),
```
cesidio, line 35: You can use the short lookup here too, just use `t(".subject")` and put the translation in the right place

spec/support/mailer_helpers.rb
david-mears-dfe: That's true. I think there is a slight benefit in replacing a test that the email contains the correct token, with a test that the email contains the correct link: the token could be inserted into the wrong link.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2539.london.cloudapps.digital>

app/mailers/jobseeker_mailer.rb
```diff
@@ -24,4 +24,15 @@ def reset_password_instructions(record, token, _opts = {})
       subject: I18n.t("jobseeker_mailer.reset_password_instructions.subject"),
     )
   end
+
+  def unlock_instructions(record, token, _opts = {})
+    @jobseeker = record
+    @token = token
+
+    view_mail(
+      NOTIFY_JOBSEEKER_LOCKED_ACCOUNT_TEMPLATE,
```
cpjmcquillan, line 33: This all fits on one line now nicely if you're into that sort of thing

spec/system/jobseekers_can_unlock_their_account_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  before do
+    allow(JobseekerAccountsFeature).to receive(:enabled?).and_return(true)
+    visit root_path
+    within("nav") do
+      click_on I18n.t("buttons.sign_in")
+    end
+  end
+
+  context "when the jobseeker has one sign-in attempt remaining" do
+    before { jobseeker.update!(failed_attempts: Devise.maximum_attempts) }
```
cpjmcquillan, line 15: This things make me a bit torn. 

One the one hand, one line is nice and clean and tidy
But on the other hand, I like consistency and the block notation is super clear

Just commenting so we think about it really

spec/system/jobseekers_can_reset_password_spec.rb
```diff
@@ -41,9 +38,7 @@
       fill_in "jobseeker[email]", with: jobseeker.email
       click_on I18n.t("buttons.continue")
       travel_to(3.hours.from_now) do
-        raw_reset_password_token = devise_token_from_last_mail(:reset_password)
-        reset_password_url = edit_jobseeker_password_url(reset_password_token: raw_reset_password_token)
-        visit reset_password_url
+        visit first_link_from_last_mail
```
cpjmcquillan, line 41: This is nice and can be used in a couple places I think (an alternative approach I used was to visit the path with the token in the param, rather than actually clicking the link in the email)

cpjmcquillan (approved):

spec/system/jobseekers_can_unlock_their_account_spec.rb
david-mears-dfe: I like using the one-liners where possible. Decreasing the vertical space between different specs makes it that bit easier to quickly compare them. However, there are also readability costs with using inconsistent code style.

spec/system/jobseekers_can_reset_password_spec.rb
david-mears-dfe: When you merge, I can add this method to your specs. Mine isn't being merged until I chat with Chris.

david-mears-dfe: Or you could pre-base with a copy paste of this method.

PR #2546, opened 2020-12-17
update_2020_12_17

RoseMackworth:
This PR adds the latest update (release note) to the service.


david-mears-dfe:
I believe the review app might be failing due to no write access, so I asked Rose to be given write access.

PR #2553, opened 2020-12-21
[TEVA-1597] Add as soon as possible to important dates form

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1597

## Changes in this PR:

- Added boolean starts_asap column to vacancies table
- Added checkbox to important dates form
- Added check for start_asap in _key_dates_sidebar.html.haml and _edit_important_dates.html.haml . If starts_asap is present, it displays 'As soon as possible' under the start date.
- Added javascript to disable and remove content from govuk-date-input fields when checkbox checked

Screenshot 2020-12-22 at 16 41 54

Screenshot 2020-12-22 at 16 42 24

Screenshot 2020-12-22 at 16 43 09

## Questions:

- No instructions where given on where the checkbox should be on smaller viewports. The figma - https://www.figma.com/file/wyEeHQ1U9aCXOJdzh8lF8a/Sprint-72?node-id=0%3A1 . What does everyone think about how it looks? I assume I should ask Chris...

- I couldn't get govuk-media-queries to behave how I wanted it to, so I went with a standard media query. Does this need to change? If so and anyone knows their way around this stuff help would be appreciated!

- At it stands, this is how it looks on smaller viewports

### Tablet

Screenshot 2020-12-22 at 16 50 00

### Mobile 

Screenshot 2020-12-22 at 16 50 26

### DIFF
--- a/app/frontend/src/components/form/form.js
+++ b/app/frontend/src/components/form/form.js
@@ -1,0 +1,37 @@
+document.addEventListener('DOMContentLoaded', () => {
+  Array.from(document.getElementsByClassName('clear-form')).forEach((el) => {
+    el.querySelector('input[type="checkbox"]').addEventListener('click', (event) => {
+      checkboxClickHandler(el, event.target.checked);
+    });
+  });
+});
+
+export const disableInputs = (inputs) => {
+  inputs.forEach((input) => {
+    input.disabled = true;
+    input.value = '';
+  });
+};
+
+export const enableInputs = (inputs) => {
+  inputs.forEach((input) => {
+    input.disabled = false;
+  });
+};
+
+export const checkboxClickHandler = (element, checked) => {
+  const fields = Array.from(element.querySelectorAll('input[type="text"]'));
+  if (checked) {
+    form.disableInputs(fields);
+  } else {
+    form.enableInputs(fields);
+  }
+};
+
+const form = {
+  disableInputs,
+  enableInputs,
+  checkboxClickHandler,
+};
+
+export default form;
--- a/app/frontend/src/components/form/form.test.js
+++ b/app/frontend/src/components/form/form.test.js
@@ -1,0 +1,61 @@
+import form from './form';
+
+describe('form', () => {
+  document.body.innerHTML = '<div class="clear-form"> <input id="test-text-input" type="text"/> <input id="test-checkbox" type="checkbox"/> </div>';
+  const textField = document.querySelector('#test-text-input');
+  textField.value = '01';
+  const fields = Array.from(document.querySelectorAll('input[type="text"]'));
+
+  describe('toggling inputs', () => {
+    beforeEach(() => {
+      form.disableInputs(fields);
+    });
+
+    describe('disableInputs', () => {
+      test('removes the value from the input element', () => {
+        expect(document.querySelector('#test-text-input').value).toBe('');
+      });
+
+      test('disables the input element', () => {
+        expect(document.querySelector('#test-text-input').disabled).toBe(true);
+      });
+    });
+
+    describe('enableInputs', () => {
+      beforeEach(() => {
+        form.enableInputs(fields);
+        textField.value = '01';
+      });
+
+      test('allows for a value to be added to the input element', () => {
+        expect(document.querySelector('#test-text-input').value).toBe('01');
+      });
+
+      test('enables the input element', () => {
+        expect(document.querySelector('#test-text-input').disabled).toBe(false);
+      });
+    });
+  });
+
+  describe('checkboxClickHandler', () => {
+    const formElement = document.querySelector('.clear-form');
+
+    describe('when checkbox is checked', () => {
+      test('calls disableInputs', () => {
+        form.disableInputs = jest.fn();
+        const disableInputsMock = jest.spyOn(form, 'disableInputs');
+        form.checkboxClickHandler(formElement, true);
+        expect(disableInputsMock).toHaveBeenCalled();
+      });
+    });
+
+    describe('when checkbox is not checked', () => {
+      test('calls enablesInputs', () => {
+        form.enableInputs = jest.fn();
+        const enableInputsMock = jest.spyOn(form, 'enableInputs');
+        form.checkboxClickHandler(formElement, false);
+        expect(enableInputsMock).toHaveBeenCalled();
+      });
+    });
+  });
+});
--- a/app/frontend/src/styles/application/form.scss
+++ b/app/frontend/src/styles/application/form.scss
@@ -19,3 +19,30 @@
   color: $govuk-link-hover-colour;
   text-decoration: underline;
 }
+
+.clear-form {
+  display: flex;
+  @media only screen and (max-width: 882px) {
+    flex-wrap: wrap;
+
+    .govuk-form-group {
+      flex-basis: 100%;
+    }
+  }
+
+  .govuk-checkboxes__item {
+    margin-bottom: govuk-spacing(6);
+    margin-left: govuk-spacing(4);
+  }
+
+  .clear-form__checkbox {
+    align-items: baseline;
+    align-self: flex-end;
+    display: flex;
+
+    &::before {
+      @include govuk-font($size: 19);
+      content: 'Or';
+    }
+  }
+}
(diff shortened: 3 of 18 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2553.london.cloudapps.digital>

app/frontend/src/components/form/form.test.js
```diff
@@ -0,0 +1,37 @@
+import { disableInputs, enableInputs } from './form';
```
alexbowen, line 1: nice, well done 👍  few minor comments below

alexbowen: you could add a test for the click event and that it calls the correct method enable/disable

app/frontend/src/styles/application/form.scss
```diff
@@ -19,3 +19,30 @@
   color: $govuk-link-hover-colour;
   text-decoration: underline;
 }
+
```
alexbowen, line 22: i would be inclined to make the naming here a bit more BEM like and abstract rather than specifically `starts asap`

alexbowen: maybe `clear-form` ?

app/frontend/src/styles/application/form.scss
```diff
@@ -19,3 +19,30 @@
...
+      flex-basis: 100%;
+    }
+  }
+
+  .govuk-checkboxes__item {
+    margin-bottom: govuk-spacing(6);
+    margin-left: govuk-spacing(4);
+  }
+
+  .checkbox-starts-asap {
```
alexbowen, line 38: this would become `clear-form__checkbox`

app/views/publishers/vacancies/build/important_dates.html.haml
```diff
@@ -33,7 +33,16 @@
 
       = render 'publishers/vacancies/expires_at_field', f: f
 
-      = f.govuk_date_field :starts_on
+      .start-date
```
alexbowen, line 36: see class naming suggestions above

app/frontend/src/components/form/form.test.js
alexbowen: remember we used spys to spy on another function so you can see if/how many times/what arguments the function you are spying on is called with

app/frontend/src/components/form/form.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+describe('form', () => {
+  document.body.innerHTML = '<input id="test-text-input" type="text"/>';
+  const textField = document.querySelector('#test-text-input');
```
alexbowen, line 5: getElementById is always preference to select DOM elements - if its possible to use it in a sensible way

alexbowen: `document. getElementById('test-text-input');`

app/frontend/src/components/form/form.js
```diff
@@ -0,0 +1,25 @@
+document.addEventListener('DOMContentLoaded', () => {
+  Array.from(document.getElementsByClassName('start-date')).forEach((el) => {
+    el.querySelector('input[type="checkbox"]').addEventListener('click', (event) => {
+      const dateFields = Array.from(el.querySelectorAll('input[type="text"]'));
```
alexbowen, line 4: try not to use the word 'date' to make this a more abstract form utility

app/frontend/src/styles/application/form.scss
alexbowen: BEM = `block__element--modifier`

alexbowen: whilst we dont really harness the full benefit of BEM in TV app, some kind of naming convention is better than none. I do have plans to improve the naming of everything at some point soon and trying to use BEM now will aid that task

app/frontend/src/components/form/form.test.js
```diff
@@ -0,0 +1,37 @@
...
+  textField.value = '01';
+  const dateFields = Array.from(document.querySelectorAll('input[type="text"]'));
+
+  beforeEach(() => {
+    disableInputs(dateFields);
+  });
+
+  describe('disableInputs', () => {
+    test('removes the value from the input element', () => {
+      expect(document.querySelector('#test-text-input').value).toBe('');
```
alexbowen, line 15: you should be able to re-use what getElementById is above as it selects a 'live' element unlike querySelector

david-mears-dfe (commented):
Should there be some ruby tests? (I think so!)

app/views/vacancies/_key_dates_sidebar.html.haml
```diff
@@ -7,9 +7,15 @@
       - if @vacancy.starts_on.present?
         %li.key-dates-timeline__item
           %h3.key-dates-timeline__title.govuk-heading-s
-            = t('jobs.starts_on')
+            = t('jobs.expected_start')
           %p.key-dates-timeline__by
             = format_date(@vacancy.starts_on)
+      - elsif @vacancy.starts_asap.present?
```
david-mears-dfe, line 13: Should this be `elsif @vacancy.starts_asap?` ?

david-mears-dfe: This is a candidate for a ruby test

david-mears-dfe: I believe `@vacancy.starts_asap.present?` is always true, right?

alexbowen (approved):
from my point of view this is lgtm

config/locales/jobs.yml
```diff
@@ -158,8 +158,9 @@ en:
       today: Deadline is today
       tomorrow: Deadline is tomorrow
     publication_date: Date job will be listed
-    start_on: Date job starts
-    starts_on: Expected start date
+    starts_on: Date job starts
+    starts_asap: As soon as possible
+    expected_start: Expected start date
```
cpjmcquillan, line 163: Alphabetical order please

config/locales/forms.yml
```diff
@@ -190,6 +190,7 @@ en:
         expires_at_html: Time application is due (<span class='text-red'>Required</span>)
         publish_on_html: Date job will be listed (<span class='text-red'>Required</span>)
         starts_on: Date job starts
+        starts_asap: As soon as possible
```
cpjmcquillan, line 193: Alphabetical order please

cpjmcquillan (changes requested):
We shouldn't allow `starts_asap` and `starts_on` to both be set

app/models/concerns/vacancy_important_date_validations.rb
```diff
@@ -12,6 +12,8 @@ module VacancyImportantDateValidations
     validate :starts_on_must_not_be_before_today, if: :starts_on?
     validate :starts_on_must_not_be_before_publish_on, if: :starts_on?
     validate :starts_on_must_not_be_before_expires_on, if: :starts_on?
+
+    validate :starts_on_and_starts_asap_must_not_be_both_given, if: :starts_on? && :starts_asap?
```
cpjmcquillan, line 16: `validate :starts_on_and_starts_asap_must_not_both_be_present` would read better I think. Also, you don't need to add the if part here when that's identical to what the validation is doing

app/views/vacancies/_key_dates_sidebar.html.slim
```diff
@@ -7,9 +7,15 @@
       - if @vacancy.starts_on.present?
         li.key-dates-timeline__item
           h3.key-dates-timeline__title.govuk-heading-s
-            = t("jobs.starts_on")
+            = t("jobs.expected_start")
           p.key-dates-timeline__by
             = format_date(@vacancy.starts_on)
+      - elsif @vacancy.starts_asap.present?
```
cpjmcquillan, line 13: Why are you doing `@vacancy.starts_asap.present?` instead of `@vacancy.starts_asap?`?

config/locales/jobs.yml
cpjmcquillan: Alphabetical order please...

app/views/vacancies/_key_dates_sidebar.html.slim
josephhull676: Mistake. I got rid of this at some point and then seem to have got rid of those changes. Will change now.

config/locales/forms.yml
cpjmcquillan: Asap comes before on

cpjmcquillan (approved):

PR #2576, opened 2021-01-04
Basic AB testing prototype

csutter:
- Add `AbTests` service to select and persist variants
- Add `AbTestable` concern with exposed helpers for accessing `AbTests`
- Update `RequestEvent` to include AB variants
- Add fake testing AB tests for now to validate functionality
- Add overview page displaying current AB testing variants

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1443

## Before merging
- [ ] Requires `request_ab_tests` record field to be created on production BQ dataset

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,6 +15,7 @@ class ApplicationController < ActionController::Base
   helper_method :cookies_preference_set?, :referred_from_jobs_path?, :utm_parameters
 
   include Publishers::AuthenticationConcerns
+  include AbTestable
 
   def check
     render json: { status: "OK" }, status: :ok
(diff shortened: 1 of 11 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2576.london.cloudapps.digital>

cpjmcquillan (approved):
Probably out of the scope of this PR, but it would be nice to see a dummy test in action and an established convention for how we should use the variants in the wild.

Simple and I like it 🚀

csutter:
> Probably out of the scope of this PR, but it would be nice to see a dummy test in action and an established convention for how we should use the variants in the wild.

Up soon! 🔼

PR #2577, opened 2021-01-04
[TEVA-1600] Update job alert confirmation link

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1600

## Changes in this PR
- Render different link on job alert confirmation pages when jobseeker accounts are enabled
- Update job alert from search system spec


spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      activities = PublicActivity::Activity.all
+      keys = activities.pluck(:key)
+      expect(keys).to include("subscription.alert.new")
+      expect(keys).to include("subscription.daily_alert.create")
+    end
+  end
+
+  def and_click_job_alert_link
+    if page.has_css?("#job-alert-link")
+      click_on("Receive a job alert")
```
cesidio, line 128: Should we be consistent and use I18n keys here too?

cesidio (approved):

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
cpjmcquillan: I did think about it, but also didn't want to change too much things I was just copying across. In specs I find sometimes there is a balance to be had between something that is easy to read and understand (the string itself that I click on) and consistency and not repeating ourselves (using the i18n translation). Let me smash my head about it for a bit 😂

cpjmcquillan: I did it (the translation needed updating anyway to something a bit cleaner)

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2577.london.cloudapps.digital>

PR #2579, opened 2021-01-05
Kill stats page

cpjmcquillan:
🔥 baby 🔥


csutter (approved):
🥇

cesidio:
Maybe we can get rid of all the code that logs those events too??

PR #2583, opened 2021-01-05
Refactor ApplicationController

csutter:
Our `ApplicationController` and its tests have gotten a bit bloated and
deserved some general TLC.

- Make methods that don't need to be public private
- Move Devise-specific methods to a new `Jobseekers::DeviseController`
  and configure this to be the parent class for Devise controllers
- Rename `#redirect_to_domain` to `#redirect_to_canonical_domain` to be
  more explicit about its purpose
- Stop anonymising remote IPs in logs (we have a legitimate interest in
  having these in the logs for security purposes)
- Refactor tests to not rely on existence of check endpoint (or any
  public action directly on `ApplicationController`) and use RSpec
  anonymous controller instead
- Move tests of check endpoint and redirecting to canonical domain out
  into separate request specs
- Simplify tests for `#strip_empty_checkboxes`


spec/controllers/application_controller_spec.rb
```diff
@@ -1,114 +1,57 @@
...
+      }
+    end
 
-      it "removes empty checkbox from the array without error" do
-        subject.strip_empty_checkboxes(%i[array_field string_field], :test_form)
-        expect(controller.params[:test_form][:array_field]).to eql(%w[first_box])
-        expect(controller.params[:test_form][:string_field]).to eql("")
-      end
+    it "removes empty checkboxes as expected" do
+      expect(controller.params[:test_form][:nothing_to_remove_field]).to eql(%w[first_box second_box])
```
cpjmcquillan, line 52: I think I'm guilty of more than a few of these, but do we want to move towards `eq` across the board unless we explicitly want to check a bit more strictly?

cpjmcquillan (approved):

spec/controllers/application_controller_spec.rb
csutter: Ah I didn't even spot that – it was on my list of little Firebreak things to do 🚒 

I'll leave it in this PR and do that next!

cesidio (approved):

PR #2586, opened 2021-01-05
[TEVA-1643] Conduct generic search when user searches nationwide

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1643

## Changes in this PR:

- Added a check for terms users use/may use for nationwide searches. If the user does use one of these searches, a generic search is conducted, returning all vacancies.

## Questions:

- Not sure about the name of the method
- The coverage is at 90.2%. I checked the report but not too sure why the lines highlighted as not covered have been included. 
- I can see there's also a code smell: 'Merge this "unless" statement with the nested one.' I included the unless so that @location.downcase will not throw an error when @location is nil. I'm not too sure what to do here...

https://user-images.githubusercontent.com/30624173/103671982-8e2bca80-4f73-11eb-9322-ac1368081f47.mov

### DIFF
--- a/app/services/search/location_builder.rb
+++ b/app/services/search/location_builder.rb
@@ -2,6 +2,7 @@ require "geocoding"
 
 class Search::LocationBuilder
   DEFAULT_RADIUS = 10
+  NATIONWIDE_LOCATIONS = ["england", "uk", "united kingdom", "britain", "great britain"].freeze
 
   include DistanceHelper
 
@@ -19,7 +20,9 @@ class Search::LocationBuilder
                            location_category
                          end
 
-    if location_category_search?
+    if NATIONWIDE_LOCATIONS.include?(@location&.downcase)
+      initialize_nationwide_search
+    elsif location_category_search?
       initialize_location_polygon
     elsif @location.present?
       @location_filter = build_location_filter(@location, @radius)
@@ -62,5 +65,9 @@ private
       point_coordinates: Geocoding.new(location).coordinates,
       radius: convert_miles_to_metres(radius),
     }
+  end
+
+  def initialize_nationwide_search
+    @location = nil
   end
 end
--- a/spec/services/search/location_builder_spec.rb
+++ b/spec/services/search/location_builder_spec.rb
@@ -100,5 +100,16 @@ RSpec.describe Search::LocationBuilder do
         end
       end
     end
+
+    context "when a nationwide location is specified" do
+      let(:location) { Search::LocationBuilder::NATIONWIDE_LOCATIONS.sample }
+
+      it "does not set location filters" do
+        expect(subject.location).to be nil
+        expect(subject.location_category).to be nil
+        expect(subject.location_polygon).to be nil
+        expect(subject.location_filter).to eq({})
+      end
+    end
   end
 end


cpjmcquillan (changes requested):
I think this should definitely be part of the location builder service personally

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2586.london.cloudapps.digital>

app/services/search/location_builder.rb
```diff
@@ -2,6 +2,7 @@
 
 class Search::LocationBuilder
   DEFAULT_RADIUS = 10
+  NATIONWIDE_LOCATIONS = ["england", "uk", "united kingdom", "britain", "great britain"].freeze
```
cpjmcquillan, line 5: `%w[blah blah blah]`

app/services/search/location_builder.rb
```diff
@@ -63,4 +64,12 @@ def build_location_filter(location, radius)
       radius: convert_miles_to_metres(radius),
     }
   end
+
+  def set_location_to_nil_for_nationwide_search
+    unless @location.nil?
+      if NATIONWIDE_LOCATIONS.include?(@location.downcase)
```
cpjmcquillan, line 70: Do you need to use an if nested inside an unless?

spec/services/search/location_builder_spec.rb
```diff
@@ -100,5 +100,13 @@
         end
       end
     end
+
+    context "when a non-polygonable location is specified" do
+      let(:location) { Search::LocationBuilder::NATIONWIDE_LOCATIONS.sample }
+
+      it "sets location filter around the location with the default radius" do
+        expect(subject.location).to be nil
```
cpjmcquillan, line 108: We could check the other attributes too like the other specs do

app/services/search/location_builder.rb
josephhull676: As we have "united kingdom" and "great britain", it can't be done it seems. It'd create separate elements for each word.

cpjmcquillan: Oh yeah good point lol

spec/services/search/location_builder_spec.rb
```diff
@@ -100,5 +100,16 @@
         end
       end
     end
+
+    context "when a non-polygonable location is specified" do
```
cpjmcquillan, line 104: when a nationwide location is specified

spec/services/search/location_builder_spec.rb
```diff
@@ -100,5 +100,16 @@
...
+      let(:location) { Search::LocationBuilder::NATIONWIDE_LOCATIONS.sample }
+
+      it "sets location filter around the location with the default radius" do
```
cpjmcquillan, line 107: does not set any location filters

app/services/search/location_builder.rb
```diff
@@ -63,4 +64,10 @@ def build_location_filter(location, radius)
       radius: convert_miles_to_metres(radius),
     }
   end
+
+  def set_location_to_nil_for_nationwide_search
```
cpjmcquillan, line 68:
```ruby
def initialize_nationwide_search
  @location = nil
end
```

app/services/search/location_builder.rb
```diff
@@ -18,7 +19,7 @@ def initialize(location, radius, location_category, buffer_radius)
                          else
                            location_category
                          end
-
+    set_location_to_nil_for_nationwide_search unless @location.nil?
```
cpjmcquillan, line 22:
```ruby
if NATIONWIDE_LOCATIONS.include?(@location&.downcase)
  initialize_nationwide_search
elsif location_category_search?
  initialize_location_polygon
elsif @location.present?
  @location_filter = build_location_filter(@location, @radius)
end
```

spec/services/search/location_builder_spec.rb
josephhull676: Sorry about that. Silly mistake.

app/services/search/location_builder.rb
josephhull676: What if @location is nil?

cpjmcquillan (approved):

PR #2588, opened 2021-01-05
TEVA-1763 Make GH Actions deploys less brittle

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1763

## Changes in this PR:

- Used Docker tag (which is GitHub SHA) to create unique GH Action job names to ensure the correct environment and SHA/tag are used for:
-- wait for check workflow action
-- smoke test for staging
- Added waiting with turnstyle action on the `deploy_app.yml` workflow to handle multiple merges in quick succession
- Fixed accidentally hard-coded PR number in Slack Message on review app creation

### DIFF
--- a/.github/workflows/deploy_app.yml
+++ b/.github/workflows/deploy_app.yml
@@ -14,6 +14,17 @@ env:
  DOCKERHUB_REPOSITORY: dfedigital/teaching-vacancies
 
 jobs:
+  turnstyle:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 20
+    steps:
+    - uses: softprops/turnstyle@v1
+      name: Check workflow concurrency
+      with:
+        poll-interval-seconds: 20
+        same-branch-only: true
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   deploy-app:
     name: Deploy ${{ github.event.inputs.tag }} to ${{ github.event.inputs.environment }}
     runs-on: ubuntu-20.04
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy_app.yml
```diff
@@ -14,8 +14,19 @@ env:
...
+    - uses: softprops/turnstyle@v1
+      name: Check workflow concurrency
+      with:
+        poll-interval-seconds: 20
+        same-branch-only: true
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   deploy-app:
-    name: Deploy app to environment
+    name: Deploy ${{tag}} to ${{environment}}
```
cpjmcquillan, line 29: Are `tag` and `environment` defined somewhere? And should we put spaces on the inside of the brackets? (we tend to elsewhere)

cpjmcquillan (approved):

.github/workflows/deploy_app.yml
nickromneydfe: `tag` and `environment` are defined as inputs to this workflow in lines 3-11:
```
on:
  workflow_dispatch:
    inputs:
      environment:
        description: Environment
        required: true
      tag:
        description: Docker tag
        required: true
```

In looking at the spacing inside the brackets, I've made the values more explicit:
```
name: Deploy ${{ github.event.inputs.tag }} to ${{ github.event.inputs.environment }}
```

sfawcett123 (approved):
That LGTM, Though I generally create the whole reference as variable and pass that, so that the check and the reference use the same variable and no spaces or other accidents can creep in. Also means if you want to change the name you do it in one place.

vigneshmsft (approved):

PR #2593, opened 2021-01-06
(for reference) The Big Location Polygon Refactor all-in-one PR

david-mears-dfe:
## Jira tickets

https://dfedigital.atlassian.net/browse/TEVA-1642
https://dfedigital.atlassian.net/browse/TEVA-1649
https://dfedigital.atlassian.net/browse/TEVA-1652
https://dfedigital.atlassian.net/browse/TEVA-1739
https://dfedigital.atlassian.net/browse/TEVA-1858

# Changes in this PR:

You can check out this branch to visualize the polygons on a map: teva-1642-composite-locations-with-map-3

## Use multiple boundaries per location where applicable

Instead of a location's boundary being an array, it's now an array of arrays.

This is partly to include island schools (see [jira](https://dfedigital.atlassian.net/browse/TEVA-1858) for more details) and partly to make the below simpler ('composite locations').

## Composite locations

Previously we did not have polygons to search in these locations: Bedfordshire, Berkshire, Cheshire, Greater Manchester, Merseyside, Tyne and Wear, West Yorkshire, Yorkshire.

These locations, mostly counties, are often searched, but they are not included in ONS' API for county and unitary authority boundaries, which we use to import most county boundaries. That API concerns local government divisions, rather than what people commonly think of counties, and so excludes what are known as 'ceremonial' or 'geographical' counties: https://en.wikipedia.org/wiki/Ceremonial_counties_of_England

I could not find a usable source of ceremonial county boundaries, so I decided to make use of Algolia's feature which lets you search in multiple polygons at once. 

Luckily, it is possible to construct these missing counties by using the unitary authorities which together compose each county. (We can get these from the ONS counties and unitary authorities API mentioned above.) For example, Cheshire is treated as the union of Cheshire East, Cheshire West and Chester, Halton, and Warrington.

I am also including 'Yorkshire', which is not a ceremonial county, because it's a common search location.

## Also in this commit:

- Import all unitary authorities and use them instead of cities where applicable. There are some visible examples here: https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1610037810065800
- Do not distinguish between London boroughs and counties/unitary authorities. These all come from the same API endpoint.
- Omit cities that have the same name as a unitary authority so that we search in the larger area.
- Do not fall back to a text-based search if the polygon is missing; use the point instead. (This is necessitated by us no longer distinguishing between unitary authorities and counties.)
- Reduce number of API calls to buffer service: When there is no change in the boundary, we don't need to ask ONS to recalculate the buffers around it.
- Update import task to assign LocationPolygon location_type differently: see Connor's helpful comment [here](https://ukgovernmentdfe.slack.com/archives/G0127R0KXQF/p1611054934024900). Our app needs two different senses of 'county' and 'city', one for ONS, and one for humans. The human-friendly one is needed for the vacancy facets on the front page. The ONS one is needed to define which locations in the API response are in scope. ONS' API endpoints don't separate places in the same way the English language does, but using their own definitions. I'm adding to the import task a mapping from the ONS categorisation to regular human categorisation; the latter is what we store against 'location_type' on the LocationPolygon in the database (these types are not used very much at the moment.)

# Screenshots

## Example composite search area

Screenshot 2021-01-05 at 22 01 40

## Buffer suggestions still work (I use the union of the buffers of the polygons):

Screenshot 2021-01-06 at 10 05 38

## For more visual checking, use the branch teva-1642-composite-locations-with-map-3

# Remaining steps

- [ ] Devise a plan for merging this code and running the task, which does not break search.
- [ ] Test the plan on a pre-prod environment.
- [ ] Afterwards, on production (etc), delete the cities and london boroughs which duplicate unitary authorities, e.g. `LocationPolygon.where(polygons: nil).delete_all`


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2593.london.cloudapps.digital>

app/services/search/buffer_suggestions_builder.rb
```diff
@@ -8,10 +8,20 @@ def initialize(location, search_params)
   end
 
   def get_buffers_suggestions
-    buffer_vacancy_count = LocationPolygon.with_name(location).buffers.map do |buffer|
-      [buffer.first, Search::AlgoliaSearchRequest.new(search_params.merge(polygon: [buffer.last])).stats.last]
-    end
+    if DOWNCASE_COMPOSITE_LOCATIONS.include?(location.downcase)
```
cpjmcquillan, line 11: A la incoming rubocop changes, could do `return unless DOWNCASE_COMPOSITE_LOCATIONS.include?(location.downcase)`

app/services/search/location_builder.rb
```diff
@@ -20,29 +20,39 @@ def initialize(location, radius, location_category, buffer_radius)
...
-    if location_polygon.nil? && (DOWNCASE_REGIONS + DOWNCASE_COUNTIES).include?(@location_category.downcase)
+  def initialize_search_polygon_boundaries
+    location_polygon = LocationPolygon.with_name(location_category)
+    if location_polygon.present?
+      @search_polygon_boundaries = if buffer_radius.present?
+                                     [location_polygon.buffers[buffer_radius]]
+                                   else
+                                     [location_polygon.boundary]
+                                   end
+    elsif DOWNCASE_COMPOSITE_LOCATIONS.include?(location_category.downcase)
```
cpjmcquillan, line 44: I'll admit being a bit bamboozled by all these different constants that mean slightly different things, and having to try and figure out what means what and why we're using which bit where. Not sure how to overcome that though

app/services/search/location_builder.rb
```diff
@@ -20,29 +20,39 @@ def initialize(location, radius, location_category, buffer_radius)
...
+      initialize_search_polygon_boundaries
     elsif @location.present?
       @location_filter = build_location_filter(@location, @radius)
     end
   end
 
   def location_category_search?
-    (@location_category && LocationCategory.include?(@location_category)) ||
-      (@location && LocationCategory.include?(@location))
+    (location_category && LocationCategory.include?(location_category)) ||
```
cpjmcquillan, line 30: Nice 👍

app/services/search/buffer_suggestions_builder.rb
```diff
@@ -8,10 +8,20 @@ def initialize(location, search_params)
...
 
+      buffer_vacancy_count = ImportPolygons::BUFFER_DISTANCES_IN_MILES.map do |distance|
+        buffers = polygons.map { |polygon| polygon.buffers[distance.to_s] }
+        [distance.to_s, Search::AlgoliaSearchRequest.new(search_params.merge(polygons: buffers)).stats.last]
+      end
+    else
+      buffer_vacancy_count = LocationPolygon.with_name(location).buffers.map do |buffer|
+        [buffer.first, Search::AlgoliaSearchRequest.new(search_params.merge(polygons: [buffer.last])).stats.last]
+      end
+    end
```
cpjmcquillan, line 24: My heart says we can trim this method down a bit, but my head hasn't caught up with me yet 😂

cpjmcquillan (commented):
I like it a lot and I think it gives us a lot of options and makes use of useful Algolia functionality. I'm wary the similar but not quite constant names adds a little bit of cognitive complexity for developers but nothing too dramatic. 

I'm happy for us to do this and merge it and learn from it, but as with most of the location polygon functionality I really feel we need more ownership from the wider team for the feature to be useful and continue to be useful in the long term. Until we have that commitment and it's something that is iterated on (along with search in general) we're kind of fighting a losing battle as developers who are trying to keep search relevant and useful.

N.B. I could be wrong though

app/services/search/buffer_suggestions_builder.rb
alexbowen: hadnt quite scrolled to see connors comment but my first thought was also that the method is fairly long and could/should be broken down in to testable parts

david-mears-dfe:
> I'm happy for us to do this and merge it and learn from it, but as with most of the location polygon functionality I really feel we need more ownership from the wider team for the feature to be useful and continue to be useful in the long term. Until we have that commitment and it's something that is iterated on (along with search in general) we're kind of fighting a losing battle as developers who are trying to keep search relevant and useful.

I think there is a problem whereby the behaviour is basically invisible slash it takes developer know-how to visualize it. For BAs to own the feature, they need to be able to inspect it.

A step towards that could be to create a page on the site that is hidden with credentials (like Nick's /sidekiq page https://github.com/DFE-Digital/teaching-vacancies/pull/2627) which does something like (1) mimic the search page as well as (2) display the polygons (and radius) used in the search (reusing code from https://github.com/DFE-Digital/teaching-vacancies/pull/2571 and https://github.com/DFE-Digital/teaching-vacancies/pull/2330).

david-mears-dfe:
> * [ ]  Devise a plan for merging this code and running the task, which does not break search.

Hypothesis:

[0) Get rid of current migration file.]
1) Merge a new migration: add polygons column to LocationPolygon.
2) Adapt import task to make sure it adds polygon-column data and does not destroy boundary-column data.
3) Merge in the import task and run it on the target environment.
4) Merge code which reads the polygon data instead of the boundary data.
5) Delete the cities and london boroughs which duplicate unitary authorities, e.g. LocationPolygon.where(polygons: nil).delete_all
6) Merge a new migration: drop boundary column on LocationPolygon.

cpjmcquillan:
> > * [ ]   Devise a plan for merging this code and running the task, which does not break search.
> 
> Hypothesis:
> 
> [0) Get rid of current migration file.]
> 
> 1. Merge a new migration: add polygons column to LocationPolygon.
> 2. Adapt import task to make sure it adds polygon-column data and does not destroy boundary-column data.
> 3. Merge in the import task and run it on the target environment.
> 4. Merge code which reads the polygon data instead of the boundary data.
> 5. Delete the cities and london boroughs which duplicate unitary authorities, e.g. LocationPolygon.where(polygons: nil).delete_all
> 6. Merge a new migration: drop boundary column on LocationPolygon.

I didn't read this and just came here to suggest pretty much exactly what you propose here (I think we can switch 4 and 5 though) - so it must at least be sensible enough that we both think it's a sensible approach.

app/models/location_polygon.rb
```diff
@@ -1,8 +1,8 @@
 class LocationPolygon < ApplicationRecord
   scope :cities, -> { where(location_type: "cities") }
   scope :counties, -> { where(location_type: "counties") }
-  scope :london_boroughs, -> { where(location_type: "london_boroughs") }
   scope :regions, -> { where(location_type: "regions") }
+  scope :other, -> { where(location_type: "other") }
```
cpjmcquillan, line 5: I think it might pay to keep the location type of `london_boroughs`, as well as an `other` type

app/services/import_polygons.rb
```diff
@@ -1,67 +1,80 @@
...
-      points = []
-      geometry_rings[ring_index].each do |point|
-        # API returns coords in an unconventional lng,lat order
-        # Coordinates rounded as they are stored as double precision floats which have a precision of
-        # 15 decimal digits. All UK coordinates only have a maximum of 2 digits before decimal point.
-        points.push(*point.reverse.map { |coord| coord.round(13) })
+      #
+      # Store them in a hash because Postgres doesn't support multidimensional arrays of varying extents.
+
+      polygons = []
```
cpjmcquillan, line 32: Can we do something this?

```ruby
polygons = geometry_rings.map { |ring| ring.map { |point| *point.reverse.map { |coord| coord.round(13) } } }
```

Although now I've written it the map-ception is confusing me and the each blocks are probably clearer 😂

app/services/search/location_builder.rb
```diff
@@ -21,44 +20,37 @@ def initialize(location, radius, location_category, buffer_radius)
...
-      :error,
-      "A location category search was performed as a text search as no LocationPolygon could
-      be found with the name '#{@location_category}'.",
-    )
-    @missing_polygon = true
+    @polygon_boundaries = []
+    locations.each do |location|
+      polygons = buffer_radius.present? ? location.buffers[buffer_radius] : location.polygons["polygons"]
+      polygons.each do |polygon|
+        @polygon_boundaries.push(polygon)
```
cpjmcquillan, line 51: `polygon_boundaries`

app/models/location_polygon.rb
david-mears-dfe: The reason I got rid of it is that some London boroughs are named after towns they contain, and I thought it was preferable to have those places come up under 'Towns and cities' rather than nowhere. As I don't want to deal with giving places multiple types I got rid of the london_borough type, as it's not being used.

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -9,116 +9,77 @@
...
+          expect(subject.buffer_suggestions).to eq([["25", 5]])
+        end
       end
     end
 
-    context "when there are zero vacancies in a buffer polygon" do
+    context "when searching in a composite location" do
+      let(:location) { "Bedfordshire" }
+      let(:districts) { ["Bedford", "Central Bedfordshire", "Luton"] }
+      let(:search_hits) { [2, 0, 1, 3, 0] }
```
cpjmcquillan, line 64: Can we put the search hits in a realistic order e.g `[0, 1, 1, 2, 3]`

spec/services/search/buffer_suggestions_builder_spec.rb
```diff
@@ -9,116 +9,77 @@
...
+            double("vacancies"), search_hits_count, nil,
+            arguments_for_algolia.merge(insidePolygon: buffer_coordinates_for_all_districts)
+          )
+        end
       end
 
-      it "only returns buffer radii that include vacancies" do
-        expect(subject.buffer_suggestions).to eq([["25", 5]])
+      it "returns the appropriate buffer suggestions" do
+        expect(subject.buffer_suggestions).to eq([["5", 2], ["15", 1], ["20", 3]])
```
cpjmcquillan, line 82: And update this to eq `[["10", 1], ["20", 2], ["25", 3]]`

spec/services/import_polygons_spec.rb
```diff
@@ -1,151 +1,181 @@
...
+            distances = [5, 10, 15, 20, 25]
+            distances.each do |distance|
+              allow(HTTParty).to receive(:get).with(
+                "https://tasks.arcgisonline.com/arcgis/rest/services/Geometry/GeometryServer/buffer?bufferSR=3857&"\
+                "distances=#{convert_miles_to_metres(distance)}&f=json&geodesic=false&geometries=%7B%22geometryType"\
+                "%22%3D%3E%22esriGeometryPolygon%22%2C+%22geometries%22%3D%3E%5B%7B%22rings%22%3D%3E%5B%5B%5B"\
+                "55.8110853660943%2C+-2.0343575091738%5D%2C+%5B55.7647624900862%2C+-1.9841097397706%5D%5D%5D%7D%5D%7D"\
+                "&inSR=4326&outSR=4326&unionResults=true&unit=",
+              ).and_return(buffer_response)
+              allow(HTTParty).to receive(:get).with(
```
cpjmcquillan, line 166: Is this bit a copy pasta?

cpjmcquillan (approved):
(pending migration and merge plan) lgtm

app/models/location_polygon.rb
cpjmcquillan: Yeah forget this lol I thought we still had these on the home page 😆

app/services/search/location_builder.rb
david-mears-dfe: Warum? I'm trying to change the attributes value.

spec/services/import_polygons_spec.rb
david-mears-dfe: It's different each time I'm afraid. I'm reluctant to generate the params stuff programmatically because then I'm replicating the logic in the test for the logic.

david-mears-dfe:
Converting to draft so I can split out the relevant bits for the merging plan above.

app/services/search/location_builder.rb
cpjmcquillan: Maybe I should learn to **read** before I open my mouth 😂

spec/services/import_polygons_spec.rb
cpjmcquillan: I guess my mere mortal brain just looks at the massive strings and nopes out so they look exactly the same, maybe worth a comment that _yes, they are actually different human_

david-mears-dfe:
> 1. Merge a new migration: add polygons column to LocationPolygon.
> 2. Adapt import task to make sure it adds polygon-column data and does not destroy boundary-column data.
> 3. Merge in the import task and run it on the target environment.
> 4. Merge code which reads the polygons column instead of the boundary column.
> 5. Delete the cities and london boroughs which duplicate unitary authorities, e.g. LocationPolygon.where(polygons: nil).delete_all
> 6. Merge a new migration: drop boundary column on LocationPolygon.

Closing in favour of 5 PRs which follow this plan (5+6 squashed into one PR) while not wanting there to be too many PRs open at once.

PR #2598, opened 2021-01-06
Use govuk_button_to to avoid js disabled bug when creating jobs

cpjmcquillan:
~~So... linking to the same path as the current path with a post action obviously causes some problems when JS is disabled. This circumnavigates that scenario by using a new action instead of a create action (not very restful I know). This brought to my attention a similar issue with deleting jobs.~~

~~This fixes the first issue at hand now in the quickest way possible, however nobody has complained about this _yet_, so I'm going to take the liberty of spending a bit more time ironing this out for deleting jobs too - and also use a create action but just with a different route.~~

Forget all that nonsense, just use a lovely button in a form

## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1811

## Changes in this PR
- Use `govuk_button_to` for create a job links

### DIFF
--- a/app/components/publishers/no_vacancies_component.html.slim
+++ b/app/components/publishers/no_vacancies_component.html.slim
@@ -3,7 +3,7 @@
     .govuk-grid-column-full
       h2.govuk-heading-m = t("schools.no_jobs.heading")
 
-      = link_to t("buttons.create_job"), organisation_jobs_path, method: :post, class: "govuk-button govuk-!-margin-bottom-0"
+      = govuk_button_to t("buttons.create_job"), organisation_jobs_path, class: "govuk-!-margin-bottom-0"
 
       p = t("schools.no_jobs.intro")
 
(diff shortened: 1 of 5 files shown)


david-mears-dfe (approved):

csutter (changes requested):
Mmmm but this ends up with us having a `GET` action that changes state, which isn't idempotent (i.e. a user can change state on the server with what should be a "safe" request).

Wouldn't a better way of fixing this be using `button_to` instead of `link_to`? That should generate a button that can POST/DELETE/whatever without needing JS.

cpjmcquillan:
> Mmmm but this ends up with us having a `GET` action that changes state, which isn't idempotent (i.e. a user can change state on the server with what should be a "safe" request).
> 
> Wouldn't a better way of fixing this be using `button_to` instead of `link_to`? That should generate a button that can POST/DELETE/whatever without needing JS.

Yes I agree. I didn't know about `button_to`, but I knew I changed it from a `new` action to `create` a couple months ago!

csutter (approved):
🚀

PR #2606, opened 2021-01-07
Fix missing leading whitespace in job alert link text

cesidio:
### Before
Screenshot 2021-01-07 at 15 29 53

### After
Screenshot 2021-01-07 at 15 30 10

### DIFF
--- a/app/components/jobseekers/search_results/job_alerts_link_component.html.slim
+++ b/app/components/jobseekers/search_results/job_alerts_link_component.html.slim
@@ -2,5 +2,4 @@
   .job-alert-cta__fade
   .job-alert-cta__content class="govuk-heading-s govuk-!-font-weight-regular govuk-!-margin-0" id="job-alert-cta"
     .icon.icon--left.icon--alert-white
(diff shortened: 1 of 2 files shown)


app/components/jobseekers/search_results/job_alerts_link_component.html.slim
```diff
@@ -3,4 +3,4 @@
   .job-alert-cta__content class="govuk-heading-s govuk-!-font-weight-regular govuk-!-margin-0" id="job-alert-cta"
     .icon.icon--left.icon--alert-white
       = link_to t("subscriptions.link.text"), new_subscription_path(search_criteria: @vacancies_search.only_active_to_hash), id: "job-alert-link-sticky-gtm"
-      = t("subscriptions.link.help_text")
+      =< t("subscriptions.link.help_text")
```
cpjmcquillan, line 6: What about passing the link into the translation instead? I think it's cleaner that way

cpjmcquillan (approved):

PR #2607, opened 2021-01-07
Use vanilla Rubocop instead of `rubocop-govuk`

csutter:
We were using `rubocop-govuk`, but it moves very slowly and is somewhat
GDS-specific, catering for idiosyncracies specific to their apps. We
should move closer to the Ruby community styleguide, with a handful of
overrides for rules where as a team we agree to deviate.

- Autocorrect all autocorrectable offences
- Manually correct the following offences:
  - Lint/ConstantDefinitionInBlock: Constants should not be defined
    inside blocks (replacing them with appropriate `let`s)
  - Lint/EmptyBlock: Empty block detected
  - Lint/TopLevelReturnWithArgument: Top-level return statements cannot
    have arguments
  - Naming/VariableNumber: Use normalcase for symbol numbers
  - Style/GuardClause: Use a guard clause instead of wrapping code inside
    a conditional expression.
  - Style/HashLikeCase: case-when represents a simple 1:1 mapping and can
    be replaced with a hash lookup
  - Style/OptionalBooleanParameter: Use keyword arguments when defining
    method with boolean argument
- Other refactorings:
  - Rename `Publishers::SignIn::Dfe::SessionsController#user_category` to
    `#publisher_category` for consistency

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1812

### DIFF
--- a/app/controllers/publishers/sign_in/dfe/sessions_controller.rb
+++ b/app/controllers/publishers/sign_in/dfe/sessions_controller.rb
@@ -1,4 +1,10 @@
 class Publishers::SignIn::Dfe::SessionsController < Publishers::SignIn::BaseSessionsController
+  PUBLISHER_CATEGORIES = {
+    "001" => :school,
+    "002" => :local_authority,
+    "010" => :multi_academy_trust,
+  }.freeze
+
   include SignInAuditConcerns
 
   skip_before_action :check_session, only: %i[create new]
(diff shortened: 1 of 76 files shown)


app/controllers/publishers/sign_in/dfe/sessions_controller.rb
```diff
@@ -69,38 +75,32 @@ def identifier
...
-    case auth_hash.dig("extra", "raw_info", "organisation", "category", "id")
-    when "001"
-      :school
-    when "002"
-      :local_authority
-    when "010"
-      :multi_academy_trust
-    end
+    category_id = auth_hash.dig("extra", "raw_info", "organisation", "category", "id")
+    PUBLISHER_CATEGORIES[category_id]
```
cesidio, line 103: much nicer!

app/controllers/publishers/sign_in/email/sessions_controller.rb
```diff
@@ -63,19 +64,19 @@ def update_session_except_org_id(options)
   end
 
   def get_uid
-    params.dig(:uid)
+    params[:uid]
```
cesidio, line 67: maybe we can get rid of the methods?

app/models/concerns/vacancy_copy_validations.rb
```diff
@@ -13,10 +13,10 @@ module VacancyCopyValidations
...
 
   def job_title_has_no_tags?
-    unless job_title == sanitize(job_title, tags: [])
-      errors.add(
-        :job_title, I18n.t("job_details_errors.job_title.invalid_characters")
-      )
-    end
+    return if job_title == sanitize(job_title, tags: [])
+
+    errors.add(
```
cesidio, line 18: I would prefer this to be a one-liner, but that is just a personal preference

cesidio (approved):

app/models/concerns/vacancy_job_details_validations.rb
```diff
@@ -13,10 +13,10 @@ module VacancyJobDetailsValidations
...
 
   def job_title_has_no_tags?
-    unless job_title == sanitize(job_title, tags: [])
-      errors.add(
-        :job_title, I18n.t("job_details_errors.job_title.invalid_characters")
-      )
-    end
+    return if job_title == sanitize(job_title, tags: [])
+
+    errors.add(
```
cpjmcquillan, line 18: Same as above

cpjmcquillan (approved):

PR #2615, opened 2021-01-11
Miscellaneous post-Rails upgrade tasks

csutter:
- Update all gems
- New framework defaults: Enable "painless" new defaults and remove
  those for frameworks we don't use anyway
- Move from classic autoloader to Zeitwerk (and add preloading for STI)


config/application.rb
```diff
@@ -26,7 +26,6 @@ module TeacherVacancyService
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
-    config.autoloader = :classic
```
cesidio, line 29: ❤️

cesidio (approved):

cpjmcquillan (approved):

config/application.rb
csutter: So long, classic autoloader. Thanks for all the memories! 😢

PR #2618, opened 2021-01-11
[TEVA-1654] Use a jsonb serialiser for search_criteria

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1654

### DIFF
--- a/app/form_models/subscription_form.rb
+++ b/app/form_models/subscription_form.rb
@@ -13,11 +13,7 @@ class SubscriptionForm
   validate :criteria_selected
 
   def initialize(params = {})
-    search_criteria = if params[:search_criteria].present?
-                        JSON.parse(params[:search_criteria]).symbolize_keys
-                      else
-                        {}
-                      end
+    search_criteria = params[:search_criteria].presence || {}
 
     @email = params[:email]
     @frequency = params[:frequency]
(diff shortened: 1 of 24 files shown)


app/form_models/subscription_form.rb
```diff
@@ -13,11 +13,7 @@ class SubscriptionForm
   validate :criteria_selected
 
   def initialize(params = {})
-    search_criteria = if params[:search_criteria].present?
-                        JSON.parse(params[:search_criteria]).symbolize_keys
-                      else
-                        {}
-                      end
+    search_criteria = params[:search_criteria].present? ? params[:search_criteria] : {}
```
cpjmcquillan, line 16: can we do `params[:search_criteria].presence || {}`

cpjmcquillan (approved):

david-mears-dfe:
Why do we need a serializer for this jsonb column, but not for the other jsonb columns: the LocationPolygon.buffers column and the Publisher.dsi_data column?

cesidio:
> Why do we need a serializer for this jsonb column, but not for the other jsonb columns: the LocationPolygon.buffers column and the Publisher.dsi_data column?

Because in the beginning we started storing values as JSON strings. so we need to convert them. Also it is nice to have hash with indifferent access when retrieving them

PR #2619, opened 2021-01-11
[TEVA-1650]   TEVA-1650  'Ilford' should search in London area, not village in Devon

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1650

## Changes in this PR:

- Created a new mapped_point_locations.yml file and changed existing mapped_locations.yml file to mapped_polygon_locations.yml. 
- Created new MAPPED_POLYGON_LOCATION and MAPPED_POINT_LOCATIONS constants
- Updated references

`[6] pry(main)> MAPPED_POINT_LOCATIONS['ilford']
=> "ilford, london"
[7] pry(main)> Geocoding.new("ilford, london").coordinates
=> [51.557609801634506, 0.073491794712767]`

Screenshot 2021-01-11 at 16 19 28

## Questions:

- How does everyone feel about this? Splitting point location mapping out into another file seemed to make sense to me, but I wonder what everyone else thinks? 
- Not sure of `map_location_with_incorrect_coordinates` as a method name

### DIFF
--- a/app/models/location_category.rb
+++ b/app/models/location_category.rb
@@ -1,5 +1,5 @@
 class LocationCategory
   def self.include?(location)
-    ALL_LOCATION_CATEGORIES.include?((MAPPED_LOCATIONS[location.downcase].presence || location).downcase)
+    ALL_LOCATION_CATEGORIES.include?((MAPPED_POLYGON_LOCATIONS[location.downcase].presence || location).downcase)
   end
 end
--- a/app/services/search/location_builder.rb
+++ b/app/services/search/location_builder.rb
@@ -10,7 +10,7 @@ class Search::LocationBuilder
               :location_polygon, :search_polygon_boundary, :missing_polygon, :radius, :buffer_radius
 
   def initialize(location, radius, location_category, buffer_radius)
-    @location = location || location_category
+    @location = MAPPED_POINT_LOCATIONS[location] || location || location_category
     @radius = (radius || DEFAULT_RADIUS).to_i
     @buffer_radius = buffer_radius
     @location_filter = {}
--- a/lib/tasks/data/mapped_polygon_locations.yml
+++ b/lib/tasks/data/mapped_polygon_locations.yml
@@ -35,7 +35,6 @@
 - ['haringey, london', 'haringey']
 - ['herefordshire', 'herefordshire, county of']
 - ['hull', 'kingston upon hull']
-- ['ilford', 'ilford, london']
 - ['islington, london', 'islington']
 - ['isle of wight, newport', 'isle of wight']
 - ['kensington', 'kensington and chelsea']
--- a/lib/tasks/data/mapped_polygon_locations.yml
+++ b/lib/tasks/data/mapped_polygon_locations.yml
@@ -35,6 +35,7 @@
 - ['haringey, london', 'haringey']
 - ['herefordshire', 'herefordshire, county of']
 - ['hull', 'kingston upon hull']
+- ['ilford', 'ilford, london']
 - ['islington, london', 'islington']
 - ['isle of wight, newport', 'isle of wight']
 - ['kensington', 'kensington and chelsea']
--- a/app/models/location_polygon.rb
+++ b/app/models/location_polygon.rb
@@ -5,6 +5,6 @@ class LocationPolygon < ApplicationRecord
   scope :regions, -> { where(location_type: "regions") }
 
   def self.with_name(location)
-    find_by(name: (MAPPED_LOCATIONS[location.downcase].presence || location).downcase)
+    find_by(name: (MAPPED_POLYGON_LOCATIONS[location.downcase].presence || location).downcase)
   end
 end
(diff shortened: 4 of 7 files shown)


app/models/location_category.rb
```diff
@@ -1,5 +1,5 @@
 class LocationCategory
   def self.include?(location)
-    ALL_LOCATION_CATEGORIES.include?((MAPPED_LOCATIONS[location.downcase].presence || location).downcase)
+    ALL_LOCATION_CATEGORIES.include?((MAPPED_POLYGON_LOCATIONS[location.downcase].presence || location).downcase)
```
david-mears-dfe, line 3: Why only polygon locations?

app/services/search/location_builder.rb
```diff
@@ -25,6 +25,7 @@ def initialize(location, radius, location_category, buffer_radius)
     elsif location_category_search?
       initialize_location_polygon
     elsif @location.present?
+      map_location_with_incorrect_coordinates if MAPPED_POINT_LOCATIONS.include?(@location)
```
david-mears-dfe, line 28: It would be safer to do the reassignment of `@location` earlier in the process. It's already been referenced 3 times before this line. That happens to be OK at the moment but it might not always be ok. Let's map it as soon as we assign it for the first time (unless there is a reason we shouldn't?)

cpjmcquillan (commented):
Personally I don't think this is something we should necessarily be doing. 

I would advocate for aligning our autocomplete suggestion and geolocation APIs first to iron out issues. 

Potentially then better surfacing where exactly a user is searching to empower users to better grasp _where_ exactly they are searching for jobs.

I would only ever want us to do something like this as a last resort and in combination with the two aforementioned points. It is one thing to map `london, uk` to `london` but a completely different kettle of fish mapping different geographical locations. How do we determine which is the most relevant result for a user and how do we articulate this to them?

This is just my opinion though and happy to be outvoted/overrulled

app/services/search/location_builder.rb
```diff
@@ -71,4 +72,8 @@ def build_location_filter(location, radius)
   def initialize_nationwide_search
     @location = nil
   end
+
+  def map_location_with_incorrect_coordinates
```
david-mears-dfe, line 76: I find the method name potentially confusing. I know what you mean because I know what the ticket is about. A more explicit method name could be `map_location_name`.

However, given my previous comment, I'd recommend combining the behaviour in line 13 (the first assignment) with this method to make a `initialize_location` method that handles both pieces of assignment logic at once.

lib/tasks/data/mapped_polygon_locations.yml
```diff
@@ -35,6 +35,7 @@
 - ['haringey, london', 'haringey']
 - ['herefordshire', 'herefordshire, county of']
 - ['hull', 'kingston upon hull']
+- ['ilford', 'ilford, london']
```
david-mears-dfe, line 38: remove line

josephhull676:
> Personally I don't think this is something we should necessarily be doing.
> 
> I would advocate for aligning our autocomplete suggestion and geolocation APIs first to iron out issues.
> 
> Potentially then better surfacing where exactly a user is searching to empower users to better grasp _where_ exactly they are searching for jobs.
> 
> I would only ever want us to do something like this as a last resort and in combination with the two aforementioned points. It is one thing to map `london, uk` to `london` but a completely different kettle of fish mapping different geographical locations. How do we determine which is the most relevant result for a user and how do we articulate this to them?
> 
> This is just my opinion though and happy to be outvoted/overrulled

What do you mean by more closely aligning the two? I understand what you mean about determining the most relevant result before deciding which one we'd like the location to point to, but in the case of Ilford I think it's safe for us to make that decision, no? Perhaps not for other locations (like Stratford for eg.) as we discussed but then we could leave the mapping for these locations until it's decided? I don't know...

cpjmcquillan:
> > Personally I don't think this is something we should necessarily be doing.
> > I would advocate for aligning our autocomplete suggestion and geolocation APIs first to iron out issues.
> > Potentially then better surfacing where exactly a user is searching to empower users to better grasp _where_ exactly they are searching for jobs.
> > I would only ever want us to do something like this as a last resort and in combination with the two aforementioned points. It is one thing to map `london, uk` to `london` but a completely different kettle of fish mapping different geographical locations. How do we determine which is the most relevant result for a user and how do we articulate this to them?
> > This is just my opinion though and happy to be outvoted/overrulled
> 
> What do you mean by more closely aligning the two? I understand what you mean about determining the most relevant result before deciding which one we'd like the location to point to, but in the case of Ilford I think it's safe for us to make that decision, no? Perhaps not for other locations (like Stratford for eg.) as we discussed but then we could leave the mapping for these locations until it's decided? I don't know...

We use Google Places Autocomplete API for location suggestions, and we use Ordnance Survey Places API (I think) for geocoding locations into coordinates. I think using Google for both would give a better user experience.

app/services/search/location_builder.rb
```diff
@@ -71,4 +71,8 @@ def build_location_filter(location, radius)
   def initialize_nationwide_search
     @location = nil
   end
+
+  def map_location_name
```
david-mears-dfe, line 75: This method can be deleted now, n'est pas?

josephhull676: Of course. Don't know how that slipped through!

PR #2620, opened 2021-01-11
Refactor publisher vacancies sorting options

csutter:
We previously assigned translations to a constant on app boot time,
which does not reliably work (as the I18n translations may not have been
loaded yet). Instead, this is extracted to a `VacancySortingOptions`
object which loads translations at runtime (and acts as an `Enumerable`
so can be directly used by the `collection_select` view helper).

### DIFF
--- a/spec/services/publishers/vacancy_sorting_options_spec.rb
+++ b/spec/services/publishers/vacancy_sorting_options_spec.rb
@@ -1,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe Publishers::VacancySortingOptions do
+  subject { described_class.new(organisation, vacancy_type) }
+
+  context "for a School" do
+    let(:organisation) { build_stubbed(:school) }
+
+    context "by default" do
+      let(:vacancy_type) { :any_old_type }
+
+      it "returns the base set of options" do
+        expect(subject.map(&:column)).to eq(%w[expires_on job_title])
+      end
+    end
+
+    %i[pending draft].each do |status|
+      context "for a `#{status}` status" do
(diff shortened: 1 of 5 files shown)


spec/services/publishers/vacancy_sorting_options_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe Publishers::VacancySortingOptions do
+  subject { described_class.new(organisation, vacancy_type) }
+
+  context "for a School" do
+    let(:organisation) { build_stubbed(:school) }
```
cpjmcquillan, line 7: Nice, I'd never read about that before 👍

spec/services/publishers/vacancy_sorting_options_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe Publishers::VacancySortingOptions do
+  subject { described_class.new(organisation, vacancy_type) }
+
+  context "for a School" do
```
cpjmcquillan, line 6: I like the safety net the consistency of "contexts always use `when`" provides, but if we say it's ok to just use sensible descriptions then I am ok with that

cpjmcquillan (approved):

spec/services/publishers/vacancy_sorting_options_spec.rb
csutter: Good point - will refactor this momentarily when I do the same for the public-facing sort options!

PR #2626, opened 2021-01-12
Refactor sorting

cpjmcquillan:
This was supposed to be straightforward but it grew horns and turned a bit nasty. I've tried to break it up a bit by commits but it is a bit chunky. 

https://www.figma.com/file/wyEeHQ1U9aCXOJdzh8lF8a/Sprint-72?node-id=91%3A1092

## Changes in this PR
- Fix and refactor sorting for publisher jobs dashboard
- Update jobseeker saved jobs dashboard to accommodate refactored sorting
- Add sorting to jobseeker subscriptions dashboard while I'm here


app/services/publishers/vacancy_sort.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def initialize(organisation, vacancy_type)
+    @organisation = organisation
+    @vacancy_type = VALID_VACANCY_TYPES.include?(vacancy_type) ? vacancy_type : "published"
+    @column = send("#{@vacancy_type}_options").first.column
+    @order = send("#{@vacancy_type}_options").first.order
+  end
+
+  def options
+    send("#{@vacancy_type}_options").tap do |options|
```
csutter, line 12: Not sure how I feel about this in terms of readability – I tend to restrict metaprogramming to libraries, or as a last resort. I reckon something like this might be more faffy but clearer in terms of intent:

```ruby
def options
  base_options.tap do |options|
    # ...
  end
end

def base_options
  case @vacancy_type
  when "pending"
    [
      SortOption.new("publish_on", "desc", I18n.t("jobs.sort_by.published_date.descending")),
      SortOption.new("expires_on", "asc", I18n.t("jobs.sort_by.expires_on.ascending")),
      SortOption.new("job_title", "asc", I18n.t("jobs.sort_by.job_title.ascending")),
    ]
  when "draft"
  # ...
  end
end
```

csutter (approved):
Other than my comments, looks good! There still feels like a refactoring or two left in it somewhere, but let's get this in first.

cpjmcquillan:
> Other than my comments, looks good! There still feels like a refactoring or two left in it somewhere, but let's get this in first.

The whole journey from where we were to now has been a hell of a ride, and we probably would have benefitted from having a dedicated ticket from the outset about sorting dashboards. 

We've gone from the original sorting service which was exclusively for vacancies and tables, to a sorting service that could handle more than just vacancies, to a dashboard that no longer used tables, and eventually to this. Hopefully a separate ticket  will help iron out the kinks we've accumulated along the way!

PR #2627, opened 2021-01-12
[TEVA-1762] Add Sidekiq dashboard with basic auth

nickromneydfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1762

## Changes in this PR:

- Expose sidekiq dashboard, but protect behind basic auth, with credentials pulled from AWS Parameter Store

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,12 @@
+require "sidekiq/web"
+
 Rails.application.routes.draw do
+  Sidekiq::Web.use Rack::Auth::Basic do |username, password|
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -1,4 +1,11 @@
+require "sidekiq/web"
```
csutter, line 1: One tiny stylistic thing – this `require` should have a blank line after it

csutter (approved):
👍

config/routes.rb
nickromneydfe: Have added in the stylistic blank line

PR #2631, opened 2021-01-13
Increase notification component padding when there is an icon

cpjmcquillan:
### Before
image

### After
image


cesidio (approved):

david-mears-dfe:
Nice

PR #2639, opened 2021-01-13
[TEVA-1846] Update to use latest ONS API endpoint

david-mears-dfe:
We were directed by ONS to use the https://services1.arcgis.com API instead of https://ons-inspire.esriuk.com when we reported a bug with ons-inspire (the old API).

Updating the API will solve some issues peculiar to the old API, such as Liverpool being in Wales.

In the process of updating this I found some mistakes we had made when we first implemented polygon importing. We previously took the first ring from each set of rings. This is not always the right ring to use: there are a dozen or so exceptions.

When the ONS API returns multiple rings for a county, some of these are small islands, one is the mainland of the county, and some are cities that are excluded because they are locally governed.

There is no pattern as to which of the rings is relevant to us, so we have to individually specify the index of the ring we want
(generally, but not reliably, the largest).

The way we import polygons is not sophisticated enough to include all schools for each county. The schools which we are failing to capture are:
- In Essex, which has 10 rings, the 5th (I am not 0-indexing here) is Mersea Island, containing Mersea Island School.
- In Hampshire, which has 2 rings, the 1st is Hayling Island, containing at least one school.
- In Kent, which has 4 rings, the 3rd is Sheppey Island, containing at least one school.

The Isles of Scilly only has one school at the time of writing, so we use the ring for St Mary's Island:
https://en.wikipedia.org/wiki/List_of_schools_on_the_Isles_of_Scilly

This is (I hope) an interim solution to updating the API, until we decide how best to deal with the omitted island schools.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1846

### Subset of counties that have multiple rings to choose from

Screenshot 2021-01-13 at 17 38 41
Screenshot 2021-01-13 at 17 39 39
Screenshot 2021-01-13 at 17 40 24
Screenshot 2021-01-13 at 17 42 22
Screenshot 2021-01-13 at 17 43 18
Screenshot 2021-01-13 at 17 44 23
Screenshot 2021-01-13 at 17 45 23
Screenshot 2021-01-13 at 17 46 19
Screenshot 2021-01-13 at 17 47 18
Screenshot 2021-01-13 at 17 48 21

### A county that only has one ring

Screenshot 2021-01-13 at 17 49 28

### A non-county ring

Screenshot 2021-01-13 at 17 55 38

## Next steps:

- [ ] Deal with the omitted island schools. This will be easier when we are doing requests with multiple polygons e.g. for composite locations. https://dfedigital.atlassian.net/browse/TEVA-1858


alexbowen (approved):

config/initializers/location_data.rb
```diff
@@ -14,21 +14,24 @@
 
 LOCATION_POLYGON_SETTINGS = {
   regions: {
-    boundary_api: "https://ons-inspire.esriuk.com/arcgis/rest/services/Administrative_Boundaries/Regions_December_2019_Boundaries_EN_BUC/MapServer/0/query?where=1%3D1&outFields=rgn19nm,shape&outSR=4326&f=json",
-    name_key: "rgn19nm",
+    boundary_api: "https://services1.arcgis.com/ESMARspQHYMw9BZ9/arcgis/rest/services/regions/FeatureServer/0/query?where=1%3D1&outFields=GOR10NM&outSR=4326&f=json",
```
cpjmcquillan, line 17: I have reservations about how stable these endpoints are now, and what `ESMARspQHYMw9BZ9` represents/means

config/initializers/location_data.rb
```diff
@@ -14,21 +14,24 @@
...
   },
   cities: {
-    boundary_api: "https://ons-inspire.esriuk.com/arcgis/rest/services/Other_Boundaries/Major_Towns_and_Cities_December_2015_Boundaries_V2/MapServer/0/query?where=1%3D1&outFields=tcity15nm,shape&outSR=4326&f=json",
-    name_key: "tcity15nm",
+    boundary_api: "https://services1.arcgis.com/ESMARspQHYMw9BZ9/arcgis/rest/services/Major_Towns_and_Cities_December_2015_EW_BGG/FeatureServer/0/query?where=1%3D1&outFields=tcity15nm,shape&outSR=4326&f=json",
+    name_key: "TCITY15NM",
   },
 }.freeze
 
+DOWNCASE_COUNTIES_WITH_RING_INDICES = YAML.load_file(base_path.join("counties.yml")).map { |line|
```
cpjmcquillan, line 34: If you wanted, I think a YAML structure of `- ['County name', <index>]` might be more straightforward to parse 
```ruby
YAML.load_file(base_path.join("counties.yml")).to_h.transform_keys(&:downcase)
```

app/services/import_polygons.rb
```diff
@@ -11,7 +11,7 @@ def initialize(location_type:)
   end
 
   def call
-    response = HTTParty.get(LOCATION_POLYGON_SETTINGS[location_type][:boundary_api])
+    response = JSON.parse(HTTParty.get(LOCATION_POLYGON_SETTINGS[location_type][:boundary_api]))
 
     response.fetch("features", []).each do |region_response|
       region_name = region_response.dig("attributes", LOCATION_POLYGON_SETTINGS[location_type][:name_key]).downcase
```
cpjmcquillan, line 17: Maybe this would be better off as `polygon_name` now

cpjmcquillan (approved):
I think this is the best we can do for now. I'm curious if we might have the same/similar issue with other types of boundaries (e.g cities, regions)

david-mears-dfe:
> I think this is the best we can do for now. I'm curious if we might have the same/similar issue with other types of boundaries (e.g cities, regions)

See https://dfedigital.atlassian.net/browse/TEVA-1858

We will have similar issues for the other boundaries. I think we won't be using this solution for very long (if we're allowed to work on 1858).

PR #2640, opened 2021-01-13
Job alerts account hint

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1515
https://dfedigital.atlassian.net/browse/TEVA-1700
https://dfedigital.atlassian.net/browse/TEVA-1697

## Changes in this PR
- Add redirect_to functionality
- Add cancel link to edit job alert page
- Remove confirm_update template
- Create jobseeker account prompt partial
- Show jobseeker account prompt when jobseeker creates/edits a job alert and they are signed in/out/don't have an account
- Rename job alert system spec files for some consistency


spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -44,82 +49,99 @@
...
+    context "when a location category search is carried out" do
+      let(:location_category_search?) { true }
+      let(:location) { "London" }
+      let!(:location_polygon) { create(:location_polygon, name: "london") }
 
-    scenario "can successfuly subscribe to a new alert" do
-      visit jobs_path
-      and_perform_a_search
-      and_click_job_alert_link
+      scenario "can successfuly subscribe to a new alert" do
```
cesidio, line 93: I think that the latest rubocop recommendations are that we change `scenario` with `it`. Not 100% sure on that though

cesidio (approved):

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
cpjmcquillan: That's cool with me anyway

app/views/subscriptions/_jobseeker_account_prompt.html.slim
```diff
@@ -0,0 +1,10 @@
+- if @jobseeker_account_exists
+  #sign-in-prompt
```
alexbowen, line 2: i am not a fan of using id's like this i.e just as hooks for the tests which as far as i can tell is all this used for. in this day and age id's should only be used for anchor identifiers. possibly in some cases for JS hooks but now we have things like `getElementsByClassName` available you could argue a good case that we shouldnt use id's for this purpose either

alexbowen: nothing to do with this PR but using id's in CSS is definatley an anti-pattern to unless you really really have to

cpjmcquillan: Would you prefer a class or data attribute?

alexbowen: good question in that i have very much used data-attributes in the past for test hooks. so if i was going set a precedent for a consistent approach id say data attribute but we dont have a consistent approach. maybe we should. for now i dont mind class or attribute but not id, im saying in a very roundabout way

alexbowen: in fact the class/data attribute is a part of the to page object/not to page object conversation we started

app/controllers/application_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
 
   before_action :redirect_to_canonical_domain, :set_headers
+  before_action :store_jobseeker_redirect_to!, if: :redirect_to_present?
```
david-mears-dfe, line 11: if: -> { redirect_to_param.present? }

app/controllers/subscriptions_controller.rb
```diff
@@ -22,9 +22,12 @@ def create
       Auditor::Audit.new(subscription, "subscription.#{subscription.frequency}_alert.create", current_publisher_oid).log
       AuditSubscriptionCreationJob.perform_later(@subscription.to_row)
       SubscriptionMailer.confirmation(subscription.id).deliver_later
+
       if jobseeker_signed_in?
         redirect_to jobseekers_subscriptions_path, success: t(".success")
       else
+        @confirmation_type = "create"
```
david-mears-dfe, line 29: Or try `params[:action]` or `action_name`

config/locales/en.yml
```diff
@@ -489,25 +489,15 @@ en:
...
-      header: Your email subscription has started
+      header:
+        create: Your email subscription has started
+        update: Your job alert has been updated
       next_step_details: 'We’ll email you details of any new teaching jobs that match the following criteria:'
       next_step: What happens next
-      title: Subscription confirmation
+      title:
+        create: Subscription confirmation
+        update: Subscription update
```
david-mears-dfe, line 499: I mildly prefer `Subscription update confirmation` as it quickly summarizes the content of the page for a screen-reader user

david-mears-dfe: I've Slacked Rose to ask if she considers titles when writing content

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -3,23 +3,30 @@
...
+  let(:jobseeker) { build_stubbed(:jobseeker) }
 
-  context "when recaptcha score is invalid" do
-    before do
-      allow_any_instance_of(ApplicationController).to receive(:verify_recaptcha).and_return(true)
-      allow_any_instance_of(ApplicationController).to receive(:recaptcha_reply).and_return({ "score" => 0.1 })
-    end
+  before { allow(JobseekerAccountsFeature).to receive(:enabled?).and_return(jobseeker_accounts_enabled?) }
+
+  describe "#recaptcha" do
```
david-mears-dfe, line 12: `#asdf` normally refers to a method called `asdf`. I think remove the `#` to avoid confusion.

app/controllers/subscriptions_controller.rb
cpjmcquillan: I forgot about action_name, nice one

app/views/subscriptions/_jobseeker_account_prompt.html.slim
cpjmcquillan: Let me go data attribute for now pending page object spike

PR #2642, opened 2021-01-14
refactor similar jobs to use gds containers

alexbowen:
no ticket

Wasn't happy with the previous fix for the misalignment of similar jobs. turns out a combination of @cpjmcquillan @csutter suggestions and my original is much more sensible and neater.

flex-box isn't designed to solve this type of layout in a horizontal sense - as horizontally it isn't flexing - so use standard GDS columns for this and use flex-box for the vertical layout which does need to flex. nice.


cpjmcquillan (approved):
as you say - nice

csutter (approved):
Hooray! 🎉

PR #2643, opened 2021-01-14
Move to Rails 6.1 defaults

csutter:
We've not had any more issues with Rails 6.1 so we can flip to use 6.1
defaults across the board – this only affects the cookie `SameSite`
settings, and the way CSRF tokens are generated. No way back now!

- Change `load_defaults` to 6.1
- Remove now superfluous transitionary new framework defaults
  initializer


cpjmcquillan (approved):
💥

cesidio (approved):

PR #2645, opened 2021-01-14
Raise in test on missing translations and fix them

csutter:
We have a handful of missing translations in the app. This enables
`config.i18n.raise_on_missing_translations` in the test environment so
that tests will fail if they encounter any missing translations, and
fixes the missing translations I've come across.


david-mears-dfe (approved):
❤️

PR #2653, opened 2021-01-15
Simplify DSI sign in error handling

csutter:
Apparently DSI will sometimes redirect to our callback URL outside of
the OIDC authentication flow, which will cause OmniAuth to think a CSRF
attempt has been made, and leave the user stranded on an error page.

This was previously handled by tweaked copy-and-pasted code from the
`omniauth_openid_connect` library in order to identify when a request
comes in without the requisite flow parameters. This doesn't make much
sense however, as `omniauth` will handle errors by default anyway by
redirecting to `/auth/failure`, which we already route to an appropriate
place.

- Remove custom OmniAuth strategy and use default OIDC strategy
- Redirect `/auth/failure` route to DSI instead of routing directly

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,53 +1,3 @@
-module OmniAuth
-  module Strategies
-    class OpenIDConnect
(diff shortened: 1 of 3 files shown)


cesidio (approved):
❤️

PR #2655, opened 2021-01-15
[TEVA-1806] Change heading to identify user

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1806

## Screenshots of UI changes:
Screenshot 2021-01-15 at 17 38 00

### DIFF
--- a/app/components/publishers/vacancies_component.rb
+++ b/app/components/publishers/vacancies_component.rb
@@ -1,10 +1,11 @@
 class Publishers::VacanciesComponent < ViewComponent::Base
-  def initialize(organisation:, sort:, selected_type:, filters:, filters_form:, sort_form:)
+  def initialize(organisation:, sort:, selected_type:, filters:, filters_form:, sort_form:, email:)
     @organisation = organisation
     @sort = sort
     @filters = filters
     @filters_form = filters_form
     @sort_form = sort_form
+    @email = email
     @vacancy_types = %w[published pending draft expired awaiting_feedback]
     @selected_type = @vacancy_types.include?(selected_type) ? selected_type : "published"
 
@@ -22,14 +23,6 @@ class Publishers::VacanciesComponent < ViewComponent::Base
 
   def vacancy_sort_options
     Publishers::VacancySort.new(@organisation, @selected_type)
-  end
-
-  def heading
-    if @organisation.group_type == "local_authority"
-      t("schools.jobs.local_authority_index_html", organisation: @organisation.name)
-    else
-      t("schools.jobs.index_html", organisation: @organisation.name)
-    end
   end
 
   def vacancy_links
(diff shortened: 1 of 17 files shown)


app/views/publishers/organisations/show.html.slim
```diff
@@ -2,8 +2,8 @@
 
 .govuk-grid-row
   .govuk-grid-column-full
-    = render(Publishers::VacanciesComponent.new(organisation: @organisation, sort: @sort, selected_type: @selected_type, filters: @filters, filters_form: @managed_organisations_form, sort_form: @sort_form))
-    = render(Publishers::NoVacanciesComponent.new(organisation: @organisation))
+    = render(Publishers::VacanciesComponent.new(organisation: @organisation, sort: @sort, selected_type: @selected_type, filters: @filters, filters_form: @managed_organisations_form, sort_form: @sort_form, email: @publisher_email))
```
cpjmcquillan, line 5: Can we just use `current_publisher.email` here if it's a helper method?

spec/components/publishers/vacancies_component_spec.rb
```diff
@@ -186,4 +187,17 @@
...
     end
   end
+
+  describe "heading" do
+    let(:organisation) { create(:school) }
+    let(:vacancy) { create(:vacancy, :published) }
+
+    before { vacancy.organisation_vacancies.create(organisation: organisation) }
+
+    let!(:inline_component) { render_inline(subject) }
```
cpjmcquillan, line 197: Do we need this line?

app/views/publishers/organisations/show.html.slim
cesidio: `current_jobseeker` is a helper method provided by devise. Will add a helper method to be accessible by the views for `current_publisher` too

spec/components/publishers/vacancies_component_spec.rb
cesidio: Yes we do need that line. Could remove the setup for this and move the test for the heading inside the other tests, which are specific to the organisation type though, so would have to replicate it 3 times.

app/components/publishers/vacancies_component.rb
```diff
@@ -25,11 +26,7 @@ def vacancy_sort_options
   end
 
   def heading
-    if @organisation.group_type == "local_authority"
-      t("schools.jobs.local_authority_index_html", organisation: @organisation.name)
-    else
-      t("schools.jobs.index_html", organisation: @organisation.name)
-    end
+    t("organisation.show.heading_html", organisation: @organisation.name, email: @email)
```
david-mears-dfe, line 29: Same question.

app/components/publishers/vacancies_component/vacancies_component.html.slim
```diff
@@ -1,4 +1,4 @@
-= render(Shared::DashboardComponent.new(heading: heading, navigation_items: vacancy_links, link: { url: organisation_jobs_path, text: t("buttons.create_job") } ))
+= render(Shared::DashboardComponent.new(heading: t("organisation.show.heading_html", organisation: @organisation.name, email: @email), navigation_items: vacancy_links, link: { url: organisation_jobs_path, text: t("buttons.create_job") } ))
```
david-mears-dfe, line 1: Dashboard Component can also be refactored to have the heading in the template rather than a long attribute.

cesidio: the dashboard is used by publishers and jobseekers. in the case of jobseekers, we do not display the organisation

david-mears-dfe (approved):

PR #2669, opened 2021-01-19
[TEVA-1799] shared process steps component

alexbowen:
convert publisher sidebar component -> shared process steps component

https://dfedigital.atlassian.net/browse/TEVA-1799


app/controllers/publishers/vacancies/application_controller.rb
```diff
@@ -1,7 +1,32 @@
...
+
+  def steps_adjust
+    current_publisher_is_part_of_school_group? ? 0 : 1
+  end
+
+  def step_current
+    @step_current = defined?(step) ? step : :job_location
+  end
+
+  def set_service
```
david-mears-dfe, line 27: The name of this method is quite opaque; maybe `set_steps_service`

app/controllers/publishers/vacancies/application_controller.rb
```diff
@@ -1,7 +1,32 @@
 require "indexing"
 
 class Publishers::Vacancies::ApplicationController < Publishers::BaseController
-  include Publishers::JobCreationHelper
+  def steps_config
```
david-mears-dfe, line 4: I wonder if this config would prefer to live in `ProcessSteps`; saves various before_actions and passings of arguments. It feels like a constant that belongs to that class.

app/controllers/publishers/vacancies/application_controller.rb
```diff
@@ -1,7 +1,32 @@
...
+      review: { number: 8, title: I18n.t("jobs.review_heading") },
+    }.freeze
+  end
+
+  def steps_adjust
+    current_publisher_is_part_of_school_group? ? 0 : 1
+  end
+
+  def step_current
+    @step_current = defined?(step) ? step : :job_location
```
david-mears-dfe, line 24: I don't see how it's possible for `defined?(step)` to be true, as it's not been defined within this method.

app/controllers/publishers/vacancies/application_controller.rb
david-mears-dfe: Or `process_steps`, like the name of the class.

app/controllers/publishers/vacancies/application_controller.rb
david-mears-dfe: Ah: I can see `def step; :documents; end` in the documents controller. Ignore me.

app/services/process_steps.rb
```diff
@@ -0,0 +1,40 @@
+class ProcessSteps
+  def initialize(steps:, step:, adjust:)
+    @steps = steps
+    @step = step
+    @adjust = adjust
+  end
+
+  def current_step_number
+    current_step = if defined?(@step)
```
david-mears-dfe, line 9: When is @step undefined? I think never; it comes from `step_current` which defaults to `:job_location`.

app/controllers/publishers/vacancies/application_controller.rb
alexbowen: this config is specific to the create vacancy process, a different config will be needed for application

app/services/process_steps.rb
csutter: I don't think it would be – `@step` is initialised in the constructor.

Also, undefined instance variables are always `nil` so you don't need to check whether they're defined.

csutter: Also the `params[:controller]` below won't work anyway as this is not a controller

app/services/process_steps.rb
```diff
@@ -0,0 +1,40 @@
...
+                     @step
+                   else
+                     params[:controller].split("/").last.to_sym == :documents ? :documents : :review
+                   end
+    @steps[current_step][:number] - @adjust
+  end
+
+  def steps_to_display
+    steps = @steps.dup
+    if @adjust.positive?
```
csutter, line 19: Not sure if I follow completely – the steps below seem to be specific to _one_ particular process, but this service looks like it should be generic 😕

app/controllers/publishers/vacancies/application_controller.rb
david-mears-dfe: I see. Carry on.

csutter: I don't think any of these should be a `before_action`, rather they can be memoized helpers:

```ruby
def steps_config
  @steps_config ||= { ... }
end
helper_method :steps_config

```

app/controllers/publishers/vacancies/application_controller.rb
```diff
@@ -1,7 +1,32 @@
...
+      supporting_documents: { number: 5, title: I18n.t("jobs.supporting_documents") },
+      documents: { number: 5, title: I18n.t("jobs.supporting_documents") },
+      applying_for_the_job: { number: 6, title: I18n.t("jobs.applying_for_the_job") },
+      job_summary: { number: 7, title: I18n.t("jobs.job_summary") },
+      review: { number: 8, title: I18n.t("jobs.review_heading") },
+    }.freeze
+  end
+
+  def steps_adjust
+    current_publisher_is_part_of_school_group? ? 0 : 1
```
csutter, line 20: This method doesn't have any side effects, so can just be a helper method - also not sure this is the best way of handling the adjustment, will have a think after my meeting 💡

app/controllers/publishers/vacancies/application_controller.rb
csutter: As above, should probably be a memoized helper – a nicer pattern than `set_xxx`

csutter (commented):
Looks good for the most part but I'm not sure about the way we adjust steps - will have a think to see if I have a better idea, or someone else does 😄

app/services/process_steps.rb
alexbowen: i think this might explain my failing system spec, thanks!

app/components/shared/process_steps_component.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def steps_to_display
+    @service.steps_to_display
+  end
+
+  def total_steps
+    @service.total_steps
+  end
+
+  def set_active_step_class(step_number, current_step_number)
```
csutter, line 23: These two can probably just be `active_step_class` and `visited_step_class`.

I'm on a crusade against the word "set" today 😂

app/controllers/publishers/vacancies/build_controller.rb
```diff
@@ -14,6 +14,11 @@ class Publishers::Vacancies::BuildController < Publishers::Vacancies::Applicatio
   before_action :set_up_show_form, only: %i[show]
   before_action :set_up_update_form, only: %i[update]
 
+  before_action :steps_config
+  before_action :steps_adjust
+  before_action :step_current
+  before_action :process_steps
```
csutter, line 20: With the changes you've made you can now get rid of these!

csutter (approved):
Tiny comment about the method names, other than that looks good! 👍 

My "hmm" feeling about having the `steps_to_display` logic in `ProcessSteps` still nags me, but I'm happy for us to look into that later.

PR #2675, opened 2021-01-21
Replace delete link_to with delete button_to

cpjmcquillan:
A small PR that replaces the last of the `link_to, method: :delete` components with `button_to` (styled appropriately)

### DIFF
--- a/app/views/publishers/organisations/_vacancy_draft.html.slim
+++ b/app/views/publishers/organisations/_vacancy_draft.html.slim
@@ -13,4 +13,4 @@
       ul.actions.govuk-list
         li = govuk_link_to(t("jobs.edit_link"), vacancy.edit_path)
         li = govuk_link_to(t("jobs.copy_link"), vacancy.copy_path, method: :get)
-        li = govuk_link_to(t("jobs.delete_link"), vacancy.delete_path, method: :delete, data: { confirm: t("jobs.are_you_sure", job_title: vacancy.job_title) })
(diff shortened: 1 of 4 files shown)


app/views/publishers/organisations/_vacancy_draft.html.slim
```diff
@@ -13,4 +13,4 @@
       ul.actions.govuk-list
         li = govuk_link_to(t("jobs.edit_link"), vacancy.edit_path)
         li = govuk_link_to(t("jobs.copy_link"), vacancy.copy_path, method: :get)
-        li = govuk_link_to(t("jobs.delete_link"), vacancy.delete_path, method: :delete, data: { confirm: t("jobs.are_you_sure", job_title: vacancy.job_title) })
+        li = button_to(t("jobs.delete_link"), vacancy.delete_path, method: :delete, data: { confirm: t("jobs.are_you_sure", job_title: vacancy.job_title) }, class: "govuk-delete-link")
```
cesidio, line 16: Why don't we use:

https://github.com/DFE-Digital/govuk-components/blob/master/app/helpers/govuk_link_helper.rb#L10

cpjmcquillan: It makes the button look like a `govuk-button`, but we don't want that

cesidio: ah ok. fair enough

cesidio (approved):

PR #2676, opened 2021-01-21
Lint slim files in app/components too

cesidio:


app/views/shared/_checkbox_group.html.slim
```diff
@@ -8,7 +8,7 @@
     options,
     value_method,
     text_method,
-    hint_method = hint_method,
+    hint_method,
```
cpjmcquillan, line 11: Does this definitely work with the form builder?

cesidio: yes it does, tested it. you can have a look at the review app too in case you want to double check it

cpjmcquillan (approved):

PR #2677, opened 2021-01-21
Require confirmation to unsubscribe from job alert

csutter:
The “unsubscribe” link in job alert emails is a simple link via HTTP
`GET` that does not require confirmation from the user. This means it's
vulnerable to CSRF attacks, email scanning software that follows links,
or simply accidental clicks.

- Add new confirmation page that asks users if they are sure they want
  to cancel their subscription
- Create `SubscriptionDetailsComponent` to DRY up subscription details
  box, with an option to have the phrasing for a "current" job alert or
  a "future" job alert
- Refactor unsubscribe feedbacks into a separate controller for more
  RESTful resource-y goodness
- Make `Subscription` factory produce active subscriptions by default
- Make the content more... better


app/controllers/unsubscribe_feedbacks_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+
+  def subscription_id
+    ParameterSanitiser.call(params).require(:subscription_id)
+  end
+
+  def unsubscribe_feedback_params
+    ParameterSanitiser.call(params.require(:unsubscribe_feedback_form).permit(:reason, :other_reason, :additional_info))
```
cpjmcquillan, line 30: Do we want to decide a convention for which way round we do the sanitisation? (i.e `require.permit` inside here vs outside on line 26)

app/components/jobseekers/subscription_details_component/subscription_details_component.html.slim
```diff
@@ -0,0 +1,11 @@
+.grouped-text-block
+  ul.govuk-list
+    - @subscription.filtered_search_criteria.each_pair do |filter, value|
+      li
+        - if filter.present?
+          span> class="govuk-!-font-weight-bold" #{filter.humanize}:
```
josephhull676, line 6: Ha humanize. I like that.

config/locales/en.yml
```diff
@@ -533,17 +535,28 @@ en:
       description: You can change the search criteria you have used by amending the keywords, location or filters before you create your alert.
       title: Create your job alert
     unsubscribe:
+      title: Unsubscribe from this job alert
```
cpjmcquillan, line 538: As the man who once spent days reorganising and segmenting our translations from one giant file, can we stick to the TV mantra of 

> it's almost impossible to agree on sensible ordering so let's just alphabetise it

please? Unless there is a better convention we can adopt that we all agree on and adopt that everywhere!

cpjmcquillan (approved):

app/controllers/unsubscribe_feedbacks_controller.rb
csutter: Hmm, maybe! Also part of me thinks we have this parameter sanitisation in enough places that there must be a nicer way of doing it  😕

csutter: Have changed it to be consistent in this file for now.

PR #2679, opened 2021-01-22
Refactor form models

cpjmcquillan:
Happy Friday 🍾 
-
In my head, before I started, this seemed like a good sensible idea (especially as we build out the job application journey). However now I've finished I'm more like _why the hell did I do that?_

Let me know what you think!


spec/form_models/publishers/job_listing/copy_vacancy_form_spec.rb
```diff
@@ -1,26 +1,25 @@
...
-      let(:copy_form) { CopyVacancyForm.new(job_title: job_title) }
-
       context "when title is blank" do
         let(:job_title) { nil }
 
         it "requests an entry in the field" do
-          expect(copy_form.valid?).to be false
-          expect(copy_form.errors.messages[:job_title][0])
-            .to eq("Enter a job title")
+          expect(subject.valid?).to be false
```
cesidio, line 12: This, and in many other places, we could have a shorter, more readable spec with:

it { is_expected.to validate_presence_of(:job_title).with_message("Enter a job title") }

spec/form_models/publishers/job_listing/important_dates_form_spec.rb
```diff
@@ -0,0 +1,202 @@
+require "rails_helper"
+
+RSpec.describe Publishers::JobListing::ImportantDatesForm, type: :model do
+  subject { described_class.new(params) }
+
+  context "validations" do
+    let(:params) { {} }
+
+    it { should validate_presence_of(:publish_on).with_message(I18n.t("important_dates_errors.publish_on.blank")) }
```
cesidio, line 9: We should be consistent everywhere and either use `should` or `is_expected.to`. I prefer the latter, but if the team agrees on the former, I would be happy too

cesidio (approved):

spec/form_models/publishers/job_listing/copy_vacancy_form_spec.rb
cpjmcquillan: I think that makes sense, let's agree an approach and make a ticket for it 👍

PR #2697, opened 2021-01-25
Use specific service gem for Google Drive API

csutter:
Using `google-api-client` directly is deprecated and emits a warning on
bundling - we should be using the service-specific gem
`google-apis-drive_v3` instead.


cpjmcquillan (approved):

cpjmcquillan:
Although, I think we use `google-api-client` for a couple of other things too (e.g indexing)

csutter:
> Although, I think we use `google-api-client` for a couple of other things too (e.g indexing)

The BigQuery indexing happens through the `google-cloud-bigquery` gem (which ironically uses `google-api-client` under the hood :trollface: ), so shouldn't be impacted I don't think... At least from my brief digging through!

csutter:
Ooooh I see now, the actual indexing. How did I miss that 🤦

cpjmcquillan:
> Ooooh I see now, the actual indexing. How did I miss that 🤦

I was just looking at the scope array in `config/initializers/google_api.rb`. I'm not sure if we're using `analytics` or not

csutter:
It doesn't _look_ like we are – all I could find was some reference in the docs. Will merge this and see if Steven starts screaming 😁

PR #2699, opened 2021-01-25
[TEVA-1592] AB test job alert link

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1592

## Changes in this PR
- Add variant for legacy vs current job alert link (50:50 weighting)
- Render different link depending on variant


app/components/jobseekers/search_results/job_alerts_link_component.rb
```diff
@@ -1,5 +1,6 @@
 class Jobseekers::SearchResults::JobAlertsLinkComponent < ViewComponent::Base
-  def initialize(vacancies_search:, count:, origin:)
+  def initialize(variant:, vacancies_search:, count:, origin:)
+    @variant = variant
```
csutter, line 3: In theory, `ab_variant_for` is available as a helper so we wouldn't need to pass a variant in to a component explicitly – but in this case I kinda like it as it makes sense for the component to have two subtypes 👍

csutter (approved):
❤️

PR #2706, opened 2021-01-25
Polygons step 2: Changes to ImportPolygon task

david-mears-dfe:
Step 2 of 'Big Location Polygon Refactor' PR merging plan. See plan here:
https://github.com/DFE-Digital/teaching-vacancies/pull/2593\#issuecomment-764615951

Manually tested search both before and after running the ImportPolygons task locally on a production dump.

## Description in case you missed the original all-in-one PR

The most significant change is to import ALL rings from the API, not just the 'first'
(which, de facto, is an arbitrary choice of ring).

When the ONS API returns multiple rings for a location, some of these are small islands or similar. Some of
these islands are inhabited and have a school on, and others do not. (E.g. Mersea Island in Essex, Hayling Island
in Hampshire, Sheppey Island in Kent, St Mary's in the Isles of Scilly.)

If we import all rings, and search in them all at once (as Algolia permits us to) for
a location that has multiple rings, we can catch vacancies/schools that are on islands
as well as ones on the mainland, in one query.

Store the rings in a hash because Postgres doesn't support multidimensional arrays of varying extents.
(In other words, a postgres multi-dimensional array's base arrays must all be the same length.)

my_polygon.polygons will be equal to something like { "polygons" => [[1, 2], [3, 4, 5, 6]] }

The changes that enable us to search in the new rings will follow in a separate
commit, as we need to merge and run the import task before adding the code which
depends on the new data and new data structure.

Other changes to the import task:

- Import all unitary authorities
- Do not distinguish between London boroughs and counties/unitary authorities. These all come
from the same API endpoint.
- Omit cities that have the same name as a unitray authority so that we search in the larger
area.
- Reduce number of API calls to buffer service:
When there is no change in the boundary, we don't need to ask ONS
to recalculate the buffers around it.
- Update import task to assign LocationPolygon location_type differently:
Our app needs two different senses of 'county' and 'city', one for ONS, and one for humans. The human-friendly one is needed for the vacancy facets on the front page. The ONS one is needed to define which locations in the API response are in scope. ONS' API endpoints don't separate places in the same way the English language does, but using their own definitions. Therefore, this commit adds to the import task a mapping from the ONS categorisation to regular
human categorisation; the latter is what we store against 'location_type' on the
LocationPolygon in the database (these types are not used very much at the moment.)
- Update location_type separately in ImportPolygons task
to reduce chance of creating duplicate LocationPolygons for the same location
when we change a location's location-type mapping.
- Use same format across all ons_[location_type].yml files


app/services/search/buffer_suggestions_builder.rb
```diff
@@ -9,7 +9,16 @@ def initialize(location, search_params)
 
   def get_buffers_suggestions
     buffer_vacancy_count = LocationPolygon.with_name(location).buffers.map do |buffer|
-      [buffer.first, Search::AlgoliaSearchRequest.new(search_params.merge(polygon: [buffer.last])).stats.last]
+      # In this step (2) of the LocationPolygon refactor, the format of buffers will be different
+      # before and after running the import task. Before it's a 1D array; after, it's 2D. So I am temporarily
+      # rewriting this method to cope with both formats. This will be reverted in step 3.
+      polygon = if buffer.last.first.is_a?(Float)
```
cpjmcquillan, line 15: `buffer.last.first.is_a?(Float)` for a 2D array? What is `buffer.last.first` for a 1D array? array -> float -> ?

cpjmcquillan (approved):

app/services/search/buffer_suggestions_builder.rb
david-mears-dfe:
```
buffer => { "5" => [] }
buffer.last => []
buffer.last.first => Float or [Float Float Float]
```

PR #2710, opened 2021-01-26
Add map to debug/inspect polygons/point searches

david-mears-dfe:
For business analysts and others to visualize the polygons searched in/points searched around.

To see a map of the polygon(s)/point(s) searched in, just add '+map' to the end of the search query in the location field. E.g.:

Keywords
[english]
Location
[Hampshire+map]

This is for internal use only, but we agreed that security was unnecessary.

## Screenshots

Screenshot 2021-01-26 at 13 40 05
Screenshot 2021-01-26 at 13 39 33


app/frontend/src/application/jobseekers/map.js
```diff
@@ -87,7 +100,44 @@ window.initMap = () => {
       infoWindow.close();
     });
 
-    /* eslint-enable */
+  } else if (polygonBoundaries == "no polygons") {
```
cpjmcquillan, line 103: If no polygons are used in the search can we just not show a map? Every little helps 💰

cpjmcquillan (approved):
I don't think it's much use showing a map that doesn't show anything when polygons aren't used - unless I'm missing something?

Apart from that lgtm

app/components/jobseekers/search_results/polygon_visualizer_component.rb
```diff
@@ -0,0 +1,21 @@
+class Jobseekers::SearchResults::PolygonVisualizerComponent < ViewComponent::Base
+  def initialize(vacancies_search:, auth:)
+    @auth = auth
+    @vacancies_search = vacancies_search
+  end
+
+  def render?
+    @auth == true
```
csutter, line 8: Could just be `@auth`

app/controllers/vacancies_controller.rb
```diff
@@ -75,4 +79,13 @@ def smoke_test?
   def audit_row
     @jobs_search_form.to_hash.merge(total_count: @vacancies_search.vacancies.raw_answer["nbHits"])
   end
+
+  def authenticate_for_map
+    return unless params[:location]&.include?(SEARCH_AFFIX_FOR_MAP)
+
+    params[:location].gsub!(SEARCH_AFFIX_FOR_MAP, "")
+    @authenticated_for_map = authenticate_or_request_with_http_basic do |user, password|
+      user == ENV.fetch("MAP_USERNAME") && password == ENV.fetch("MAP_PASSWORD")
```
csutter, line 88: Could we re-use the Sidekiq auth username/password so we don't need to have another separate set of credentials?

app/frontend/src/application/jobseekers/map.js
```diff
@@ -14,30 +14,43 @@ const getSchool = () => {
   return false;
 };
 
+// For search result map
+const getPolygonBoundaries = () => {
```
csutter, line 18: Not sure how I feel about this pattern but I can see it's the same as already exists in this file - maybe one to refactor a bit later.

app/controllers/vacancies_controller.rb
csutter: On second thought, is this really so sensitive we need to authenticate at all? IMHO just the params bit would be good enough...

app/views/vacancies/index.html.slim
```diff
@@ -20,6 +20,7 @@
   .govuk-grid-column-one-third class="govuk-!-margin-bottom-4"
     = render "filters"
   .govuk-grid-column-two-thirds
+    = render(Jobseekers::SearchResults::PolygonVisualizerComponent.new(vacancies_search: @vacancies_search, auth: @authenticated_for_map))
```
csutter, line 23: Maybe `display` would be a better fit for this param name than `auth`?

spec/components/jobseekers/search_results/polygon_visualizer_component_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    render_inline(subject)
+  end
+
+  let(:auth) { true }
+  let(:vacancies_search) { instance_double(Search::SearchBuilder) }
+  let(:polygon_boundaries) { [[1, 2, 1, 2], [3, 4, 3, 4]] }
+
+  describe "#render?" do
+    context "when unauthenticated" do
+      context "when auth is not exactly true but is truthy" do
```
csutter, line 18: I don't think this is worth testing tbh – would it ever be not a boolean, and if so, do we really care? See comment about just doing

```ruby
def render?
  @auth
end
```

csutter (changes requested):
Would like to discuss if we really need auth, plus some small comments 👀

david-mears-dfe:
After checking with @milimaldedfe I'm going to get rid of the auth!

app/frontend/src/application/jobseekers/map.js
david-mears-dfe: Good point, I was over-eager

app/controllers/vacancies_controller.rb
```diff
@@ -1,6 +1,8 @@
 class VacanciesController < ApplicationController
   include ParameterSanitiser
 
+  before_action :set_map_display, only: :index
```
csutter, line 4: Given this is only for one single action, do you reckon it could just live in the action itself?

app/components/jobseekers/search_results/search_visualizer_component.rb
```diff
@@ -0,0 +1,25 @@
+class Jobseekers::SearchResults::SearchVisualizerComponent < ViewComponent::Base
+  def initialize(vacancies_search:, render:)
+    @render = render
```
csutter, line 3: Nitpick: Let's move these into the order they are in the arguments

csutter (approved):
Tiny comments, then it's good to go!

PR #2711, opened 2021-01-26
add rule for correct bg color to all application pages

alexbowen:
Screenshot 2021-01-26 at 11 02 21

### DIFF
--- a/app/frontend/src/styles/application.scss
+++ b/app/frontend/src/styles/application.scss
@@ -92,10 +92,6 @@ body {
     }
   }
 
-  &.jobseekers_job_applications_build_show,
-  &.jobseekers_job_applications_new,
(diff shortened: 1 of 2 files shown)


cpjmcquillan (commented):
Can you update lines 95 - 98 in `application.scss` too? 

Also, maybe the controller class is sufficient and we don't need `jobseeker` too?

alexbowen:
> Can you update lines 95 - 98 in `application.scss` too?
> 
> Also, maybe the controller class is sufficient and we don't need `jobseeker` too?

yes ill sort the application css but id rather leave the `.jobseeker` bit, as although you are correct that the controller class is sufficient, im bit by bit making all sass files namespaced to what they deal with

cpjmcquillan (approved):

PR #2712, opened 2021-01-26
Refactor job application

cpjmcquillan:
Some small things

- Update job application heading component to use the correct link
- Refactor job application controllers to not use before actions to set instance variables

### DIFF
--- a/app/controllers/jobseekers/job_applications/build_controller.rb
+++ b/app/controllers/jobseekers/job_applications/build_controller.rb
@@ -4,24 +4,24 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::ApplicationCont
 
   steps :personal_details, :professional_status, :personal_statement, :ask_for_support, :declarations
 
-  before_action :set_up_job_application, :set_up_current_step_number
-  before_action :set_up_show_form, only: %i[show]
-  before_action :set_up_update_form, only: %i[update]
+  helper_method :back_link_path, :current_step_number, :job_application
 
   def show
+    @form = FORMS[step].new unless step == "wicked_finish"
     render_wizard
   end
 
   def update
-    application_data = @job_application.application_data.presence || {}
+    @form = FORMS[step].new(form_params)
+    application_data = job_application.application_data.presence || {}
 
-    if @form.valid?
-      @job_application.assign_attributes(application_data: application_data.merge(form_params))
-      if params[:commit] == t("buttons.save_as_draft")
-        save_job_application_as_draft
-      else
-        render_wizard @job_application
-      end
+    if params[:commit] == t("buttons.save_as_draft")
+      job_application.assign_attributes(application_data: application_data.merge(form_params))
+      job_application.save
+      redirect_to jobseekers_saved_jobs_path, success: t(".saved_job_application")
+    elsif @form.valid?
+      job_application.assign_attributes(application_data: application_data.merge(form_params))
+      render_wizard job_application
     else
       render_wizard
     end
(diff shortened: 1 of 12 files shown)


app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -4,59 +4,66 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::ApplicationCont
...
-    @job_application.save
-    redirect_to jobseekers_saved_jobs_path, success: t(".saved_job_application")
+  def current_step_number
+    @current_step_number ||= STEPS[step]
   end
 
-  def set_up_current_step_number
-    @current_step_number = STEPS[step]
+  def form
+    case action_name
```
csutter, line 44: I reckon these might be better off staying as instance variables, but set in the individual actions – so it's obvious what they relate to.

i.e.

```ruby
def show
  @form = FORMS[step].new unless step == "wicked_finish"
  # ...
end
```

app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -4,59 +4,66 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::ApplicationCont
...
 
-    @form = FORMS[step].new
+  def finish_wizard_path
+    jobseekers_job_application_review_path(job_application)
   end
 
-  def set_up_update_form
-    @form = FORMS[step].new(form_params)
+  def job_application
+    @job_application ||= current_jobseeker.job_applications.find(params[:job_application_id])
```
csutter, line 61: Ditto this - far as I can see it's only used in `update`?

app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -4,59 +4,66 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::ApplicationCont
...
   end
 
-  def set_up_job_application
-    @job_application = current_jobseeker.job_applications.find(params[:job_application_id])
+  def save_job_application_as_draft
```
csutter, line 64: I like that this method clearly communicates intent in its name, but on the other hand it feels like those three lines could just live up there in `#update` as they're not used elsewhere. What do you think? 😕

app/views/jobseekers/job_applications/review.html.slim
```diff
@@ -1,13 +1,13 @@
 - content_for :page_title_prefix, t(".title")
 
-= render(Jobseekers::JobApplicationHeadingComponent.new(vacancy: @job_application.vacancy))
+= render(Jobseekers::JobApplicationHeadingComponent.new(vacancy: job_application.vacancy, back_path: jobseekers_job_application_build_path(job_application, :declarations)))
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
-    .govuk-caption-m Step 8 of 8
+    .govuk-caption-m Step 10 of 10
```
csutter, line 7: Not for this PR, but I reckon we should rethink our step numbering entirely at some point so we don't need to hardcode this!

csutter:
Added some random thoughts, not necessarily wedded to them. Would love to hear more from the rest of the team about whether they like this pattern, and how aggressively we want to apply it (e.g. only shared logic _between_ actions, or even for the resource itself). 💡

app/controllers/jobseekers/job_applications/build_controller.rb
cpjmcquillan: I like it because it stops the action getting too massive (although we're probably not there yet)

app/views/jobseekers/job_applications/review.html.slim
cpjmcquillan: I envisaged it happening as part of implementing the process steps sidebar. We have something similar for publishers, but I can't remember if that got refactored as well yet or not?

app/controllers/jobseekers/job_applications/build_controller.rb
cpjmcquillan: It's used in the show action too (all of the views for each of the steps references it)

app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -4,59 +4,53 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::ApplicationCont
 
   steps :personal_details, :professional_status, :personal_statement, :ask_for_support, :declarations
 
-  before_action :set_up_job_application, :set_up_current_step_number
-  before_action :set_up_show_form, only: %i[show]
-  before_action :set_up_update_form, only: %i[update]
+  helper_method :back_link_path, :current_step_number, :job_application
```
cesidio, line 7: nice

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -1,23 +1,24 @@
 class Jobseekers::JobApplicationsController < Jobseekers::ApplicationController
-  before_action :set_up_vacancy, only: %i[new create]
-  before_action :set_up_job_application, only: %i[review submit]
+  helper_method :job_application, :vacancy
 
   def create
-    job_application = current_jobseeker.job_applications.create(status: :draft, vacancy: @vacancy)
-    redirect_to jobseekers_job_application_build_path(job_application, :personal_details)
+    new_job_application = current_jobseeker.job_applications.find_or_initialize_by(vacancy: vacancy)
```
cesidio, line 5: these 3 lines of code can just be
```
current_jobseeker.job_applications.find_or_initialize_by(vacancy: vacancy).draft!
```

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -1,23 +1,24 @@
...
+    new_job_application.status = :draft
```
cesidio, line 6: new_job_application.draft!

cesidio (approved):
looks good to me

PR #2732, opened 2021-01-28
Convert search_criteria from JSON string to Hash

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1850

## Changes in this PR:
This is the first step in refactoring the logic around search_criteria. As soon as this code is deployed I will manually run the following command on production:

```bash
rails subscriptions:convert_search_criteria
```

I have tested it on a production DB dump and it works. I prefer to manually run it instead of creating a migration that will stay in our code for some time.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -26,6 +26,8 @@ class Subscription < ApplicationRecord
   end
 
   def search_criteria_to_h
+    return search_criteria if search_criteria.is_a?(Hash)
+
     parsed_criteria = JSON.parse(search_criteria) if search_criteria.present?
     parsed_criteria.is_a?(Hash) ? parsed_criteria : {}
   rescue JSON::ParserError
--- a/app/form_models/jobseekers/nqt_job_alerts_form.rb
+++ b/app/form_models/jobseekers/nqt_job_alerts_form.rb
@@ -21,7 +21,7 @@ class Jobseekers::NqtJobAlertsForm
     {
       email: email,
       frequency: :daily,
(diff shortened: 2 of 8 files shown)


csutter:
> As soon as this code is deployed I will manually run the following command on production

Agree 100% on not having it in a migration 💯 

That said, how about creating a Rake task for it (that we can later delete), just so that _at some point_ it existed in the codebase and we can go through Git history to figure out what we did?

cesidio:
> > As soon as this code is deployed I will manually run the following command on production
> 
> Agree 100% on not having it in a migration 💯
> 
> That said, how about creating a Rake task for it (that we can later delete), just so that _at some point_ it existed in the codebase and we can go through Git history to figure out what we did?

I like it!

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2732.london.cloudapps.digital>

csutter (approved):
Looks good - I'm assuming these are currently Strings of JSON, but stored as a JSON String? 🤯

app/models/subscription.rb
```diff
@@ -26,6 +26,8 @@ def self.find_and_verify_by_token(token)
   end
 
   def search_criteria_to_h
```
cpjmcquillan, line 28: Once all the search criterias have been converted, can we lose this method?

cpjmcquillan (approved):

app/models/subscription.rb
cesidio: Yes, this is just the first step

PR #2734, opened 2021-01-28
Improve handling of VCAP_SERVICES

csutter:
In preparation for having multiple different Redis instances, this
tidies up how we load service configuration from Cloudfoundry's
`VCAP_SERVICES` env variable.

(Shamelessly stolen from a previous project I worked on!)

PR #2736, opened 2021-01-28
Add additional Redis instance for caching

csutter:
We want to have two Redis instances for the app in order to keep
persistent (e.g. Sidekiq) and volatile (e.g. caching) data separate.

- Add new `redis_cache_instance` Cloudfoundry service instance with an
  `allkeys-lru` eviction policy
- Separate service plans for queue and cache, and start the cache out
  on `tiny` for now
- Rename CF name of the existing Redis instance to `redis-queue`
  to be able to refer to it separately (but keep TF name the same
  until a later change to minimise disruption)
- Use `VcapServices#named_service_url` in application config to
  ensure Sidekiq gets the correct one of the two Redis instances
- Change `VcapServices#named_service_url` to look for name anywhere
  in the service name to avoid having to specify full prefix of
  service name

## Before merge
- [x] Rename CF service name (and double check that doesn't cause issues)
- [x] Make sure app picks up correct instances from `VCAP_SERVICES` (#2734 merged, _then_ update application config)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2736.london.cloudapps.digital>

cpjmcquillan (approved):

PR #2738, opened 2021-01-28
[TEVA-1757] Consolidate 5 existing feedback tables into one

david-mears-dfe:
Part 1 of several on this ticket.

This new consolidated table is not yet used by the code, so the
task will have to be run after that code is updated.

I tested the task on production data locally and checked
feedbacks of all types.

- Migration for new feedback table
- Task for one-off transfer of data from old tables
- Do not export feedbacks table because we will send data as
events
- There is a sixth feedback type coming up, for application
submission
- Drop unused data in ratings columns on GeneralFeedback
and VacancyPublishFeedback - we stopped adding to these
in 2019
- Rename not_interested scopes to uninterested to avoid
warning messages about scope conflict whenever you run
the console
- Rename some other columns to be specific
- Consolidate various 1200-char columns into a single column
called 'comment'
- Remove unused scope

## Jira ticket URL

### DIFF
--- a/lib/tasks/copy_feedback_tables_into_one_consolidated_table.rake
+++ b/lib/tasks/copy_feedback_tables_into_one_consolidated_table.rake
@@ -1,0 +1,75 @@
+namespace :consolidate_feedback_tables do
+  desc "Copy data from feedback tables into one table"
+  task consolidate_feedback_tables: :environment do
+    AccountFeedback.all.in_batches(of: 100).each_record do |account_feedback|
+      Feedback.create(
+        feedback_type: "jobseeker_account",
+        created_at: account_feedback.created_at,
+        rating: account_feedback.rating,
+        jobseeker_id: account_feedback.jobseeker_id,
+        comment: account_feedback.suggestions,
+      )
+    end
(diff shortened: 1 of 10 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2738.london.cloudapps.digital>

cpjmcquillan (approved):

lib/tasks/copy_feedback_tables_into_one_consolidated_table.rake
```diff
@@ -0,0 +1,75 @@
+namespace :consolidate_feedback_tables do
+  desc "Copy data from feedback tables into one table"
+  task consolidate_feedback_tables: :environment do
+    AccountFeedback.all.in_batches(of: 100).each_record do |account_feedback|
```
cesidio, line 4: Instead of `.all.in_batches(of: 100).each_record` you can just use `find_each`
https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/Batches.html#method-i-find_each

PR #2751, opened 2021-02-01
[TEVA-1942] job alert radius bug

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1942

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -2,7 +2,7 @@ class SubscriptionsController < ApplicationController
   def new
     @origin = origin_param if origin_param&.start_with?(%r{/\w})
     session[:subscription_origin] = @origin
-
+    @point_coordinates = params[:search_criteria] ? Geocoding.new(params[:search_criteria][:location]).coordinates : nil
     @subscription_form = Jobseekers::SubscriptionForm.new(params[:search_criteria].present? ? search_criteria_params : email)
   end
 
--- a/app/frontend/src/application/jobseekers/location.js
+++ b/app/frontend/src/application/jobseekers/location.js
@@ -1,6 +1,18 @@
 import { enableRadiusSelect, disableRadiusSelect } from './radius';
 
-const INPUT_ELEMENT = document.getElementById('location-field');
+export const INPUT_ELEMENT_CLASSNAME = 'js-location-finder__input';
+
+export const onLoad = (el) => {
+  if (el) {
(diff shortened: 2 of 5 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2751.london.cloudapps.digital>

david-mears-dfe:
As you comment, the radius from the search should be present on page load - so I'll re-review then.

david-mears-dfe:
I added a comment on the ticket:

> Just to muddy the waters, I think the ‘desired’ radius input behaviour of only switching on if there’s a postcode might pre-date our use of polygons, and it’s better if we offer radiuses to all point-based searches.

The ticket defining the postcode-based radius behaviour was done in the same sprint as adding the polygon-based search.

app/controllers/subscriptions_controller.rb
```diff
@@ -4,7 +4,7 @@ class SubscriptionsController < ApplicationController
   def new
     @origin = origin_param if origin_param&.start_with?(%r{/\w})
     session[:subscription_origin] = @origin
-
+    @point_coordinates = params[:search_criteria] ? Geocoding.new(params[:search_criteria][:location]).coordinates : nil
```
cpjmcquillan, line 7: Rather than Geocode something twice, why don't we just pass the coordinates through?

app/frontend/src/application/jobseekers/location.js
```diff
@@ -10,21 +22,20 @@ export const onChange = (value) => {
   }
 };
 
-export const getCoords = () => {
-  if (INPUT_ELEMENT.dataset.coordinates) {
-    return INPUT_ELEMENT.dataset.coordinates.split(' ').join(',');
+export const getCoords = (el) => {
```
cpjmcquillan, line 25: Given we don't do anything with them, should we change the name of this method and make it return true or false?

david-mears-dfe:
When I check out the branch locally, on the home page, the radius is there by default on pageload. That's not how it works on production. Nor on the review app, oddly! I don't know which behaviour is the correct/best one.

david-mears-dfe:
Also, locally, when I commit an empty search on the home page, the radius drop-down is there by default on the following page, which again is unlike production (and the review app - perhaps the review app is out of date?).

Again, not sure which behaviour is preferable.

david-mears-dfe:
Deleted some comments that applied to bugs caused by me having merge conflicts...

PR #2756, opened 2021-02-02
Update Redis service plans

csutter:
This is a cost neutral change (`micro-ha` costs the same as `tiny`)

- Change cache Redis plan from `tiny` to `small-ha` (for more caching
  space and redundancy in case of PaaS failures)
- Change queue Redis plan from `small-ha` to `micro-ha` (as this is now
  only used as a queue and will require significantly less memory) and
  upgrade to Redis 5
- Change eviction policy for queue instance to `noeviction` (as it
  should now not randomly fill up, and if it does, it's because of a
  serious error)

## Before Deploying
- [x] Clear database `1` on queue Redis instance (which is now no longer used and has stale cache data in it blocking a downgrade)


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2756.london.cloudapps.digital>

csutter:
Turns out this won't work 😞 

New, faffier PR inbound shortly.

PR #2760, opened 2021-02-02
Rename existing queue Redis to "old"

csutter:
This is in preparation for migrating our queue Redis to a smaller plan
and a higher Redis version - we can't do this on the existing instance
so will need to temporarily add another one.

By renaming this it will be less confusing when we add another instance.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2760.london.cloudapps.digital>

cesidio (approved):

PR #2767, opened 2021-02-02
Upgrade Redis cache instance to bigger size and HA

csutter:
What it says on the tin.

## Before merge
- [x] Taint or remove existing instance from state another way


nickromneydfe (approved):
Looks good to me

PR #2768, opened 2021-02-02
Use Rails cache for geocoding

csutter:
We currently use the `geocoder` gem's built-in caching, but that has a
number of issues:
- it stores the *entire* HTTP response from the geocoding service in the
  cache, when all we care about is the coordinates, taking up massive
  amounts of cache memory
- it needs custom code to be able to handle key expiry
- it relies on the connection pool of the `redis-objects` gem (which we
  no longer use for anything else)

Now that we can use Rails's built in cache, we can do better.

- Add `hiredis` gem for better Redis performance
- Remove no longer used `redis-objects` gem
- Use Rails cache in `Geocoding` instead of using the `geocoder`
  gem's caching
- Increase cache duration to 30 days


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2768.london.cloudapps.digital>

cesidio (approved):
❤️

cpjmcquillan (approved):

PR #2772, opened 2021-02-03
[TEVA-1757] Part 3: Update general and publisher feedback forms to use new Feedback model

david-mears-dfe:
## Jira ticket URL

## Changes in this PR:

- Part of the ongoing refactor of feedback tables
- Created generic feedback event trigger method and used it for three types of feedback

## Later steps

- [ ] Run the task that transfers data from old feedback tables to new Feedback table
- [ ] Create and run a task that converts the old data in the Feedback table into an Event and sends it to BigQuery (without creating duplicates)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2772.london.cloudapps.digital>

app/controllers/application_controller.rb
```diff
@@ -82,6 +82,17 @@ def request_event
     @request_event ||= RequestEvent.new(request, response, session, current_jobseeker, current_publisher_oid)
   end
 
+  def trigger_feedback_provided_event
+    feedback_data = feedback_params.to_h.map.each { |key, value|
```
cesidio, line 86: `feedback_params` should be already a hash, no need to do `to_h.map` you should be able to just do `feedback_params.each`

app/controllers/application_controller.rb
```diff
@@ -82,6 +82,17 @@ def request_event
...
+      if key.to_s[-2..] == "id"
```
cesidio, line 87:
```
if key.ends_with?("id")
```

spec/controllers/general_feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
+require "rails_helper"
+
+RSpec.describe GeneralFeedbacksController, type: :controller do
```
cesidio, line 3: It would be good to start writing requrest specs and get rid of controller specs that are not supported anymore

spec/form_models/general_feedback_form_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  describe "#visit_purpose_comment" do
+    it { is_expected.not_to validate_presence_of(:visit_purpose_comment) }
+
+    context "when the visit purpose is 'other_purpose'" do
+      let(:visit_purpose) { "other_purpose" }
+
+      context "and the visit purpose comment is provided" do
+        let(:visit_purpose_comment) { "Just browsing" }
+
+        it "is valid" do
```
cesidio, line 75: I would not test for things when are valid. only when they are not valid

app/controllers/application_controller.rb
david-mears-dfe:
```ruby
undefined method `map' for #<ActionController::Parameters:0x00007fb8e9a4d680>
```

spec/controllers/general_feedbacks_controller_spec.rb
david-mears-dfe: I will replace this with a request spec, in a separate PR.

david-mears-dfe: In https://github.com/DFE-Digital/teaching-vacancies/pull/2774

app/controllers/application_controller.rb
```diff
@@ -82,6 +82,17 @@ def request_event
     @request_event ||= RequestEvent.new(request, response, session, current_jobseeker, current_publisher_oid)
   end
 
+  def trigger_feedback_provided_event
+    feedback_data = feedback_params.each { |key, value|
```
cesidio, line 86: I believe you need `map` here not `each`

or if you want something a bit more rubish you can do:

```
feedback_params.each_with_object({}) do |(key, value), params| 
   %w[jobseeker_id publisher_id].include?(key) ? params["anonymised_#{key}"] = StringAnonymiser.new(value) :  params[key] = value }
end
```

config/routes.rb
```diff
@@ -68,7 +68,7 @@
     resources :interests, only: %i[new]
   end
 
-  resource :feedback, controller: "general_feedback", only: %i[new create]
+  resource :feedback, controller: "general_feedbacks", only: %i[new create]
```
cesidio, line 71: we might want to have a standard
```
resources :feedback
```
and have a refactoring everywhere. not for this PR

cesidio (approved):

app/controllers/application_controller.rb
david-mears-dfe: feedback_params doesn't have each_with_object so I have changed this to feedback_params.to_h.each_with_object

`undefined method `each_with_object' for #<ActionController::Parameters:0x000055904ae666c8>`

PR #2780, opened 2021-02-03
searchable collection scroll logic

alexbowen:
no ticket

- add scrollable class if a) it is supplied as option set to true or b) the collection is searchable
- also update style of collection border to current figma

### DIFF
--- a/app/components/shared/searchable_collection_component.rb
+++ b/app/components/shared/searchable_collection_component.rb
@@ -1,7 +1,7 @@
 class Shared::SearchableCollectionComponent < ViewComponent::Base
-  attr_accessor :form, :attribute_name, :collection, :value_method, :text_method, :hint_method, :threshold, :small
+  attr_accessor :form, :attribute_name, :collection, :value_method, :text_method, :hint_method, :threshold, :small, :scrollable
 
-  def initialize(form:, attribute_name:, collection:, value_method:, text_method:, hint_method:, threshold: 10)
+  def initialize(form:, attribute_name:, collection:, value_method:, text_method:, hint_method:, threshold: 10, scrollable: false)
     @form = form
     @threshold = threshold
     @attribute_name = attribute_name
@@ -11,9 +11,18 @@ class Shared::SearchableCollectionComponent < ViewComponent::Base
     @hint_method = hint_method
 
     @small = searchable
+    @scrollable = scrollable || searchable
   end
 
   def searchable
     collection.count >= threshold
+  end
+
+  def scrollable_class
+    return "collection-component--scrollable" if scrollable
+  end
+
+  def border_class
+    return "collection-component--border" if searchable
   end
 end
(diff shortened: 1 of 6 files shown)


app/components/shared/searchable_collection_component/searchable_collection_component.html+checkbox.slim
```diff
@@ -1,4 +1,4 @@
-.collection-component.scrollable class=("collection-component--border" if searchable)
+.collection-component class=("collection-component--border" if searchable) class=("collection-component--scrollable" if scrollable)
```
cpjmcquillan, line 1: does this work? nice

app/components/shared/searchable_collection_component/searchable_collection_component.scss
```diff
@@ -12,7 +12,7 @@
 
   &.collection-component--border {
     background-color: #fff;
-    border: 5px solid #f3f2f1;
+    border: 2px solid #b1b4b6;
```
cpjmcquillan, line 15: what's the `govuk-colour`?

app/components/shared/searchable_collection_component/searchable_collection_component.html+checkbox.slim
alexbowen: work as in being able to add 2 conditional classes? yes definatley. work as in fixes the problem of scrolling at correct times? i think so but i want to test on dev and havent been able to deploy it there yet

app/components/shared/searchable_collection_component/searchable_collection_component.scss
alexbowen: ah i couldnt use the gov color before i relocated the scss file yesterday, but now i can...

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2780.london.cloudapps.digital>

app/components/shared/searchable_collection_component/searchable_collection_component.html+checkbox.slim
alexbowen: yes this works as expected @cpjmcquillan

app/components/shared/searchable_collection_component/searchable_collection_component.scss
```diff
@@ -3,27 +3,27 @@
     margin-bottom: 0;
   }
 
-  &.scrollable {
+  &.collection-component--scrollable {
     .govuk-form-group {
-      max-height: 194px;
+      max-height: 400px;
```
alexbowen, line 8: this makes the height of the subjects select larger but have to live with it for the moment, i dont see it as a huge problem. chris more keen on at least 5 schools being visible

app/components/shared/searchable_collection_component/searchable_collection_component.html+checkbox.slim
cesidio: wouldn't it be clearer if we have a method `classes` in the component?

app/components/shared/searchable_collection_component.rb
```diff
@@ -11,9 +11,18 @@ def initialize(form:, attribute_name:, collection:, value_method:, text_method:,
...
 
     @small = searchable
+    @scrollable = scrollable || searchable
   end
 
   def searchable
     collection.count >= threshold
   end
+
+  def scrollable_class(scrollable)
```
cpjmcquillan, line 21: you don't need to pass an argument in because the method will already have access to them (same below too)

cpjmcquillan (approved):

PR #2781, opened 2021-02-03
Update VacancyFacets to use Rails cache

csutter:
- Change `VacancyFacets` to use Rails cache instead of manually
  interacting with Redis
- Make `VacancyFacets` always return 0 when caching is disabled so we
  don't keep hitting Algolia in scenarios where it's expensive
  (e.g. dev)
- Do not show facet if count is zero
- Simplify how `VacancyFacets` is used in controller/view
- Get rid of `RefreshVacancyFacets` job and remove it from the schedule


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2781.london.cloudapps.digital>

cpjmcquillan (approved):

csutter:
Closing and re-opening as a new PR as it's changed a fair bit.

PR #2783, opened 2021-02-03
Fix for task and subscription presenter

josephhull676:
##The Issue:

Due to task below, when run, if a subscription did not have a location category, the location in the search criteria was set to nil. This is because `subscription.search_criteria.delete("location_category")` returns nil if there is no location_category key. 

```
namespace :subscription do
  desc "remove location_category from search_criteria"
  task remove_location_category_from_search_criteria: :environment do
    Subscription.find_each do |subscription|
        subscription.search_criteria["location"] = subscription.search_criteria.delete("location_category")
        subscription.save
      end
    end
  end
end
```

As some subscriptions now had location set to nil, in the subscription presenter, when line 48 in the `render_location_filter ` was hit, a NoMethodError was raised(correct term?). 

##The Solution:

-  Replaced `.empty?` with `.blank?` in the subscription presenter as `nil.blank?` does not raise a NoMethodError and instead returns `true`
- Will create a new PR later with the shiny new task that doesn't break everything :)

### DIFF
--- a/app/presenters/subscription_presenter.rb
+++ b/app/presenters/subscription_presenter.rb
@@ -45,7 +45,7 @@ class SubscriptionPresenter < BasePresenter
   end
 
   def render_location_filter(location, radius)
-    return if location.empty?
+    return if location.blank?
 
     return { location: I18n.t("subscriptions.location_polygon_text", location: location) } if LocationPolygon.include?(location)
     return { location: I18n.t("subscriptions.location_radius_text", radius: radius, location: location) } if radius


lib/tasks/remove_location_cateogory_from_subscription_search_criteria.rake
```diff
@@ -2,8 +2,10 @@ namespace :subscription do
   desc "remove location_category from search_criteria"
   task remove_location_category_from_search_criteria: :environment do
     Subscription.find_each do |subscription|
-      subscription.search_criteria["location"] = subscription.search_criteria.delete("location_category")
-      subscription.save
+      if subscription.search_criteria["location_category"].present? && subscription.search_criteria["location"].blank?
+        subscription.search_criteria["location"] = subscription.search_criteria.delete("location_category")
```
david-mears-dfe, line 6: Would it be OK to not delete the `location_category` for now (would that break anything)? I prefer idempotent tasks where possible in case they break half way through. https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation#:~:text=at%204%3A12-,12,as%20making%20a%20single%20request.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2783.london.cloudapps.digital>

cesidio (approved):

csutter (approved):
:shipit:

cpjmcquillan (approved):

PR #2789, opened 2021-02-04
[TEVA-1757] Update to Part 1: Update feedback copy task

david-mears-dfe:
## Jira ticket URL

Part of https://dfedigital.atlassian.net/browse/TEVA-1757

## Changes in this PR:

- Updates the work in https://github.com/DFE-Digital/teaching-vacancies/pull/2738
- Send Events to BigQuery at the same time as moving feedback data
- This updates the task which sends feedback from old feedback tables to the
new feedback table, so that we are creating Events (sent to BigQuery) at
the same time. These events will be used alongside newer RequestEvents,
representing controller events that create Feedbacks, to store/manage/
monitor/analyse/collate all feedback of a certain 'feedback_type'.
- The task has not yet been run. It will be run when all code has been
updated to use the new Feedback table rather than the old feedback table.
- I'm using find_or_create_by in case GovUK PaaS explodes and we need to
run the task a second time.
- I have also updated the Event model so that any data containing a
'created_at' attribute is reflected in the 'occurred_at' column of the
Event (rather than Time.now); and so that fewer data value types are
converted to string (which can be an annoyance for someone wanting to
manipulate the data after it's received).

- When the task is run, I'll remove the code for the task, as well as the occurred_at method in Event.

### DIFF
--- a/app/services/event.rb
+++ b/app/services/event.rb
@@ -17,8 +17,8 @@ class Event
   def trigger(event_type, data = {})
     data = base_data.merge(
       type: event_type,
-      occurred_at: Time.now.utc.iso8601(6),
-      data: data.map { |key, value| { key: key.to_s, value: value&.to_s } },
+      occurred_at: occurred_at(data),
+      data: data.map { |key, value| { key: key.to_s, value: formatted_value(value) } },
     )
     SendEventToDataWarehouseJob.perform_later(TABLE_NAME, data)
   rescue StandardError => e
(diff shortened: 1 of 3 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2789.london.cloudapps.digital>

david-mears-dfe:
Tested on staging. This task only creates data and does not destroy it - that will come later.

cesidio (approved):
makes sense to me

PR #2791, opened 2021-02-05
Add subscriptions incident recovery Rake task

csutter:
The last piece of the puzzle to recovering from the subscription
search criteria incident. This Rake task loads a JSON export of
subscription creation events, correlates them with records created
after the last good backup (subscriptions before that have already
been fixed) but before the incident (subscriptions created after
are not affected).

It requires a parameter of a local file (to be created on the
instance where the task is run) containing the event data:

```
rails location_fix[data.json]
```

Once the changes have been verified with the default dry run,
there is a second "commit" parameter to update subscriptions:

```
rails location_fix[data.json,true]
```

### DIFF
--- a/lib/tasks/location_fix.rake
+++ b/lib/tasks/location_fix.rake
@@ -1,0 +1,47 @@
+desc "Fix missing location search criteria"
+task :location_fix, %i[filename commit] => [:environment] do |_, args|
+  filename = args[:filename]
+  commit = (args[:commit] == "true")
+
+  Rails.logger.info("[DRY RUN]") unless commit
+
+  # Parse Bigquery JSON export into an array of hashes
+  bq_data = JSON.parse(File.read(filename))
+                .map { |item| item["data"] }
+                .map { |item| item.map { |subitem| [subitem["key"], subitem["value"]] }.to_h }
+
+  last_good_backup_time = Time.zone.parse("2021-02-02 05:00:00")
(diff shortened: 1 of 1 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2791.london.cloudapps.digital>

cesidio (approved):

lib/tasks/location_fix.rake
```diff
@@ -0,0 +1,47 @@
...
+    # Try and find a subscription in our mapping that matches the anonymised ID from the event
+    if affected_subscriptions.key?(anonymised_id)
+      subscription = affected_subscriptions[anonymised_id]
+
+      Rails.logger.info("Found subscription #{subscription.id} for anonymised ID #{anonymised_id}")
+      Rails.logger.info("Current search_criteria: #{subscription.search_criteria}")
+
+      # Do a dry run unless we explicitly want to go ahead
+      if commit
+        subscription.update(search_criteria: new_search_criteria)
```
david-mears-dfe, line 38: `update_column` iff you want to preserve the original `updated_at` - not sure if we do or not

david-mears-dfe (approved):

lib/tasks/location_fix.rake
csutter: Yeah, not too fussed about it (as we keep updating `updated_at` anyway, and IMHO this is a change worthy of an update 😅 )

cpjmcquillan (approved):

PR #2794, opened 2021-02-05
Fix for the remove_location_category_from_subscription_search_criteria task

josephhull676:
**This task needs to be tested.** 

Previous to these changes, the task did not check that location_category was present and location was blank before setting location to the value of location_category and deleting location_category. 

Because of this, all of the subscriptions that did not have a location_category in the search_criteria set location to nil - the return value of subscription.search_criteria.delete[location_category] is nil if location_category is not present.

Now the task has a clause that checks that location_category is present and location is blank before setting location_category to location and deleting location_category.

If location is not blank and the location_category is present, the location_category is deleted.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2794.london.cloudapps.digital>

david-mears-dfe:
https://ukgovernmentdfe.slack.com/archives/G0127R0KXQF/p1612547845022800

david-mears-dfe (approved):

PR #2800, opened 2021-02-08
Map searches in 'Kensington, London' to K&C borough

david-mears-dfe:
The 'Kensington, London' search query is suggested by autocomplete. We should map it to the name of the LocationPolygon record.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2800.london.cloudapps.digital>

cpjmcquillan (commented):
Should we? Kensington is distinctly different to Kensington & Chelsea - at least it is to me

PR #2809, opened 2021-02-08
Have StringAnonymiser not alter blank inputs

david-mears-dfe:
This is to make Events in BigQuery more readable - so that a reader can instantly see that an id is not present (nil), rather than remembering that xumor-boceg-dakuz-sulic-gukoz-rutas etc etc denotes a column that's empty.

An alternative would be to put the logic into the method trigger_feedback_provided_event. Which place is better? I prefer in the StringAnonymiser, but is there a reason not to put it there?

### DIFF
--- a/app/services/string_anonymiser.rb
+++ b/app/services/string_anonymiser.rb
@@ -11,8 +11,9 @@ class StringAnonymiser
   end
 
   def to_s
-    Digest::SHA256.bubblebabble(@raw_string)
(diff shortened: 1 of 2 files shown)


csutter (approved):
Sold 👍

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2809.london.cloudapps.digital>

PR #2814, opened 2021-02-09
Fix pagination according to comment in routes.rb

david-mears-dfe:
Fixes a production bug whereby pagination links are directed to `/teaching-jobs-in-?keyword=&page=2` (page not found). This fixed the issue locally.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -151,8 +151,8 @@ Rails.application.routes.draw do
   match "/500", to: "errors#internal_server_error", via: :all
 
   # If parameters are used that are the same as those in the search form, pagination with kaminari will break
-  match "teaching-jobs-in-:location",
-        to: "vacancies#index", as: :location, via: :get,
+  match "teaching-jobs-in-:location_facet",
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -151,8 +151,8 @@
   match "/500", to: "errors#internal_server_error", via: :all
 
   # If parameters are used that are the same as those in the search form, pagination with kaminari will break
-  match "teaching-jobs-in-:location",
-        to: "vacancies#index", as: :location, via: :get,
+  match "teaching-jobs-in-:location_facet",
+        to: "vacancies#index", as: :location_facet, via: :get,
```
cesidio, line 155:
```
  match "teaching-jobs-in-:location_facet",
        to: "vacancies#index", as: :location, via: :get,
        constraints: ->(request) { LocationPolygon.include?(request.params[:location_facet]) }
```

david-mears-dfe: Thanks - this solves it without creating a second bug!

PR #2816, opened 2021-02-09
Attempt to tame Bigquery export memory bloat

csutter:
This will make it run more slowly, which we don't really care about as
it's a background job, but hopefully it will stop biting off more than
it can chew.

- Reduce batch size to 100 (from default 500)
- Stop using multiple threads (which isn't all that efficient in Ruby
  anyway, and 4*500 AR objects being actively manipulated is too much
  for our worker memory apparently)


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2816.london.cloudapps.digital>

cesidio (approved):

PR #2818, opened 2021-02-09
Update VacancyFacets to use Rails cache

csutter:
- Change `VacancyFacets` to use Rails cache instead of manually
  interacting with Redis
- Make `VacancyFacets` always return 0 when caching is disabled so we
  don't keep hitting Algolia in scenarios where it's expensive
  (e.g. dev)
- Do not show facet if count is zero
- Simplify how `VacancyFacets` is used in controller/view
- Get rid of `RefreshVacancyFacets` job and remove it from the schedule
- Add `#compact` to computation of polygon boundaries in
  `LocationBuilder` to avoid issue with review apps not having any
  polygons imported
- Set `@vacancies` and `@stats` in search builder to an empty array if
  there are no results (rather than `nil`)
- Combine and improve homepage facets system specs


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2818.london.cloudapps.digital>

cesidio (approved):

cpjmcquillan (approved):

PR #2820, opened 2021-02-09
[TEVA-1757] Part n+1: Update Job Alert Feedback controller/form to rely on new Feedback table

david-mears-dfe:
## Jira ticket URL

## Changes in this PR:

Update Job Alert Feedback controller/form to rely on new Feedback table

- And use trigger_feedback_provided_event to create Events. This method uses
feedback_params, resulting in a slightly unconventional feedback_params
method on JobAlertFeedbacksController, since this is the only type of
feedback that has more than one action (they can be updated as well as
created).

- Update Event#format_value to keep floats and integers rather than
transform them to strings. The idea here is to save deconverting them
after they are sent to BigQuery.

- This PR includes a single test for links in job alert emails already sent,
in order to avoid a bug such as that fixed in
https://github.com/DFE-Digital/teaching-vacancies/pull/2737.

### DIFF
--- a/app/controllers/job_alert_feedbacks_controller.rb
+++ b/app/controllers/job_alert_feedbacks_controller.rb
@@ -3,41 +3,55 @@ class JobAlertFeedbacksController < ApplicationController
     # This action creates the JobAlertFeedback record.
     # This is because it is called from a link in the alert email. Such links can only perform
     # GET requests. HTTP forbids redirecting from GET to POST.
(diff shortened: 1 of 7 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2820.london.cloudapps.digital>

cesidio (approved):

PR #2821, opened 2021-02-09
Gentle initial refactor of `Search::SearchBuilder`

csutter:
- Move logic from being always run and assigned to instance variables
  on `initialize` to be in memoised instance methods (this means e.g.
  `wider_search_suggestions` is no longer auto-populated on requests
  where it is then never used, such as homepage facets)
- Rename `#any?` to `any_criteria_given?` to make its purpose more
  obvious

### DIFF
--- a/app/components/jobseekers/search_results/job_alerts_link_component.rb
+++ b/app/components/jobseekers/search_results/job_alerts_link_component.rb
@@ -6,6 +6,6 @@ class Jobseekers::SearchResults::JobAlertsLinkComponent < ViewComponent::Base
   end
 
   def render?
-    @vacancies_search.any? && !ReadOnlyFeature.enabled? && @count.positive?
+    @vacancies_search.any_criteria_given? && !ReadOnlyFeature.enabled? && @count.positive?
   end
 end
--- a/app/services/search/search_builder.rb
+++ b/app/services/search/search_builder.rb
@@ -2,9 +2,7 @@ class Search::SearchBuilder
   DEFAULT_HITS_PER_PAGE = 10
   DEFAULT_PAGE = 1
 
-  attr_reader :hits_per_page, :keyword, :location_search, :page, :params_hash,
-              :point_coordinates, :search_filters, :search_replica, :sort_by, :stats, :total_count, :vacancies,
-              :wider_search_suggestions
+  attr_reader :params_hash, :keyword, :page, :hits_per_page, :stats, :total_count, :vacancies
 
   def initialize(form_hash)
     @params_hash = form_hash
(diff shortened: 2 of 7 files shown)


app/components/jobseekers/search_results/job_alerts_link_component.rb
```diff
@@ -6,6 +6,6 @@ def initialize(vacancies_search:, count:, origin:)
   end
 
   def render?
-    @vacancies_search.any? && !ReadOnlyFeature.enabled? && @count.positive?
+    @vacancies_search.any_criteria_given? && !ReadOnlyFeature.enabled? && @count.positive?
```
david-mears-dfe, line 9: `filters` are slightly different, conceptually, from criteria - I'd say filters excludes keywords (for example) but criteria includes them. Suggest renaming this method to `any_filters?`

app/services/search/search_builder.rb
```diff
@@ -2,62 +2,65 @@ class Search::SearchBuilder
...
-    build_search_replica
     call_search
-    build_suggestions if @vacancies.empty? && any?
   end
 
   def only_active_to_hash
-    @params_hash.merge(keyword: @keyword).delete_if do |k, v|
-      v.blank? || (k.eql?(:radius) && @params_hash[:location].blank?) || k.eql?(:jobs_sort) || k.eql?(:page)
+    @params_hash.merge(keyword: @keyword).except(:jobs_sort, :page).reject do |k, v|
+      v.blank? || (k.eql?(:radius) && @params_hash[:location].blank?)
```
david-mears-dfe, line 18: Instead of `v.blank?`:

https://ruby-doc.org/core-2.4.0/Hash.html#method-i-compact

app/components/jobseekers/search_results/job_alerts_link_component.rb
csutter: This method also includes the keyword – I think its meaning is "did the user input anything significant or did they just hit 'Search'"

app/services/search/search_builder.rb
csutter: Good shout!

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2821.london.cloudapps.digital>

app/services/search/search_builder.rb
csutter: I take that back - we need to explicitly take blank (e.g. empty string) into account too here!

cpjmcquillan (approved):

cesidio (approved):

app/services/search/search_builder.rb
david-mears-dfe: Oh!

app/services/search/search_builder.rb
```diff
@@ -2,62 +2,65 @@ class Search::SearchBuilder
...
-    build_search_filters
-    build_search_replica
     call_search
-    build_suggestions if @vacancies.empty? && any?
   end
 
   def only_active_to_hash
-    @params_hash.merge(keyword: @keyword).delete_if do |k, v|
-      v.blank? || (k.eql?(:radius) && @params_hash[:location].blank?) || k.eql?(:jobs_sort) || k.eql?(:page)
+    @params_hash.merge(keyword: @keyword).except(:jobs_sort, :page).reject do |k, v|
```
cesidio, line 17: I would prefer:

```
@params_hash.merge(keyword: @keyword)
                          .except(:jobs_sort, :page)
                          .reject { |k, v| v.blank? || (k == :radius && @params_hash[:location].blank?) }
```

csutter: 👍 

To be honest having slept on it I'm also a bit confused as to why this is merging `keyword` back into `@params_hash`... where it came from in the first place!

cpjmcquillan: Context, there was some code for a business decision once that updated the keyword when a LocationPolygon was missing. E.g I search for London, but the polygon was missing, so London was added to the keyword. 

This was removed and I assume this slipped through somebody's cracks 😂

PR #2828, opened 2021-02-10
Remove auditing of searches

csutter:
We do not actually use this data, and the way it casually accesses the
raw Algolia answer is problematic.

Eventually we should have an event for this, but right now we can infer
this data from the default `visited` event anyway.


cesidio (approved):

cpjmcquillan (approved):
Hallelujah 🚀

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2828.london.cloudapps.digital>

PR #2832, opened 2021-02-10
Get rid of "stats" in `SearchBuilder`

csutter:
Instead of passing around and accessing an array of values, introduce
`SearchBuilder#page_from` and `#page_to` to return the number of the
first and last result shown on this page respectively.

These can be calculated trivially from the total count of results, the
current page, and the number of items shown per page, so don't need to
be calculated separately based on whether we search Algolia or the DB.


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2832.london.cloudapps.digital>

PR #2836, opened 2021-02-10
Rename all the search things

csutter:
In preparation for making them a bit less coupled.

- Rename `SearchBuilder` and `AlertBuilder` to `VacancySearch` and
  `AlertVacancySearch` as these don't really just build a search, they
  actually perform a search
- Rename `Search::AlgoliaSearchRequest` to `Search::Strategies::Algolia`
  and `Search::VacancyPaginator` to `Search::Strategies::Database` to
  clarify that they serve the same purpose (but do it differently)

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -7,7 +7,7 @@ class VacanciesController < ApplicationController
       @landing_page_translation = "#{params[:pretty]}.#{@landing_page.parameterize.underscore}"
     end
     @jobs_search_form = Jobseekers::SearchForm.new(algolia_search_params)
-    @vacancies_search = Search::SearchBuilder.new(@jobs_search_form.to_hash)
+    @vacancies_search = Search::VacancySearch.new(@jobs_search_form.to_hash)
     @jobs_search_form.jobs_sort = @vacancies_search.sort_by
     @vacancies = VacanciesPresenter.new(@vacancies_search.vacancies)
   end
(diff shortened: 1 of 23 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2836.london.cloudapps.digital>

alexbowen:
i like this way of thinking for the naming. i dont know if this is pedantic or what the thought behind it is but `Search::Strategies::Algolia` i would think should be `Search::Strategy::Algolia`

cpjmcquillan (approved):

csutter:
@alexbowen Ruby needs the folder names to be identical to the "namespace", otherwise it doesn't know how to autoload the class/module constants when they are referenced.

Having slept on it, I feel like there might still be a better name for these classes. Perhaps they'd be better off having "strategy" in the name, i.e. `AlgoliaStrategy` 😕

csutter:
Will have a think today and update them in a later PR maybe, so I can get this in.

alexbowen:
> @alexbowen Ruby needs the folder names to be identical to the "namespace", otherwise it doesn't know how to autoload the class/module constants when they are referenced.
> 
> Having slept on it, I feel like there might still be a better name for these classes. Perhaps they'd be better off having "strategy" in the name, i.e. `AlgoliaStrategy` 😕

ah yeah sure, the components example is a bad one as i guess the components example (which should be components/shared/thing.rb) - the components folder is resolved by the framework

PR #2838, opened 2021-02-10
missing test for searchable collection component

alexbowen:
no ticket, the spec for this component hadnt been added

### DIFF
--- a/spec/components/shared/searchable_collection_component_spec.rb
+++ b/spec/components/shared/searchable_collection_component_spec.rb
@@ -1,0 +1,78 @@
+require "rails_helper"
+
+RSpec.describe Shared::SearchableCollectionComponent, type: :component do
+  let(:form) { instance_double(GOVUKDesignSystemFormBuilder::FormBuilder) }
+
+  let(:collection) { [1, 2, 3, 4, 5].freeze }
+
+  before do
+    allow(form).to receive(:govuk_collection_radio_buttons)
+    allow(form).to receive(:govuk_collection_check_boxes)
+  end
+
+  let(:base) do
+    {
+      form: form,
+      attribute_name: :attributes,
+      collection: collection,
+      text_method: :first,
+      hint_method: :first,
+      value_method: :first,
+    }
+  end
+
+  context "when using radio button variant" do
+    let(:options) { { threshold: 10 } }
+    let(:radio_collection) do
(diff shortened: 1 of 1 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2838.london.cloudapps.digital>

spec/components/shared/searchable_collection_component_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  let(:collection) do
+    [1, 2, 3, 4, 5].freeze
+  end
+
+  before do
+    allow(form).to receive(:govuk_collection_radio_buttons)
+    allow(form).to receive(:govuk_collection_check_boxes)
+  end
+
+  context "Radio button searchable collection" do
```
david-mears-dfe, line 15: Our convention is that we either use `context "when(/with/without) something is the case"`, or `describe "feature/object/method"`. This comes from https://www.betterspecs.org/

spec/components/shared/searchable_collection_component_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+                          collection: collection,
+                          text_method: :first,
+                          hint_method: :first,
+                          value_method: :first,
+                          threshold: 10,
+                          scrollable: true).with_variant(:radiobutton)
+    end
+
+    describe "renders radio button variant" do
+      let!(:inline_component) { render_inline(radio_collection) }
```
david-mears-dfe, line 28: This line, and the duplicate of it, can be lifted up into the top `RSpec.describe` block. That then renders the `describe` lines 27 and 54 redundant.

david-mears-dfe (approved):
Pre-approved, but see comments.

spec/components/shared/searchable_collection_component_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+                          attribute_name: :attributes,
+                          collection: collection,
+                          text_method: :first,
+                          hint_method: :first,
+                          value_method: :first,
+                          threshold: 10,
+                          scrollable: true).with_variant(:radiobutton)
+    end
+
+    describe "renders radio button variant" do
```
cesidio, line 27: As a convention we should use describe only for methods. In this case I would use 
`it "renders radio button variant" do`

PR #2840, opened 2021-02-10
[TEVA-1978] Fix smoke test for production

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1978

## Changes in this PR:

When it runs in the scheduled workflow, the SMOKE_TEST_DOMAIN is an
empty string. We now use the production domain for this case.

### DIFF
--- a/spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
+++ b/spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
@@ -1,9 +1,15 @@
 require "capybara/rspec"
 require "i18n_helper"
 
+PRODUCTION_DOMAIN = "teaching-vacancies.service.gov.uk".freeze
+
 RSpec.describe "Page availability", js: true, smoke_test: true do
   context "Jobseeker visits vacancy page" do
-    let(:smoke_test_domain) { ENV.fetch("SMOKE_TEST_DOMAIN", "teaching-vacancies.service.gov.uk") }
+    let(:smoke_test_domain) do
+      smoke_test_domain = ENV.fetch("SMOKE_TEST_DOMAIN", "")
+      smoke_test_domain = PRODUCTION_DOMAIN if smoke_test_domain.empty?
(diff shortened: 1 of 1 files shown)


nickromneydfe (approved):
Looks good to me

spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
```diff
@@ -1,9 +1,15 @@
 require "capybara/rspec"
 require "i18n_helper"
 
+PRODUCTION_DOMAIN = "teaching-vacancies.service.gov.uk".freeze
+
 RSpec.describe "Page availability", js: true, smoke_test: true do
   context "Jobseeker visits vacancy page" do
-    let(:smoke_test_domain) { ENV.fetch("SMOKE_TEST_DOMAIN", "teaching-vacancies.service.gov.uk") }
+    let(:smoke_test_domain) do
+      smoke_test_domain = ENV.fetch("SMOKE_TEST_DOMAIN", "")
```
cpjmcquillan, line 9: This could all be

```ruby
let(:smoke_test_domain) { ENV.fetch("SMOKE_TEST_DOMAIN").presence ||  "teaching-vacancies.service.gov.uk" }
```

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2840.london.cloudapps.digital>

spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
saliceti: we tried it and .presence is only available in rails, so we would need to require it. is it worth it?
also it fails before .presence with `KeyError: key not found: "SMOKE_TEST_DOMAIN"`

saliceti: I'll merge this for now as we need smoke tests, let's discuss later a more elegant option

PR #2841, opened 2021-02-10
[TEVA-1757] The last step in refactoring existing feedback forms: unsubscribe feedback

david-mears-dfe:
## Jira ticket URL

## Changes in this PR:

- Update UnsubscribeFeedback code to rely on new Feedback table
- Refactor all feedback controllers to separate form-params from feedback-attributes. Attributes are used for two purposes: triggering events, and creating Feedbacks. These responsibilities have been separated out from the form params, which are used for validations.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -84,7 +84,7 @@ class ApplicationController < ActionController::Base
 
   def trigger_feedback_provided_event
     anonymise_these_params = %w[jobseeker_id publisher_id]
-    feedback_data = feedback_params.to_h.each_with_object({}) do |(key, value), params|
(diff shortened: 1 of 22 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2841.london.cloudapps.digital>

app/controllers/application_controller.rb
```diff
@@ -84,7 +84,7 @@ def request_event
 
   def trigger_feedback_provided_event
```
csutter, line 85: I reckon this might be better off in a concern that is included in the various feedback controllers rather than in `ApplicationController`

csutter (approved):
I dared wait until after lunch 🥗 

Looks good, although see comment about maybe moving into a concern.

david-mears-dfe:
I'll make a concern.

PR #2852, opened 2021-02-11
Fix and update failing spec

cpjmcquillan:
While dealing with the VCR/Algolia::WebMock mess I came across this failing spec when running 
`bundle exec rspec spec/system/jobseekers_can_sea*`

This PR fixes the spec and brings in some trendy page objects too.


alexbowen (approved):
this looks good. but can i suggest that before we go too gung-ho with page objects that i can give a bit of thought to how we use css selectors in them and document how we should use them? i mentioned it a couple of weeks ago i think. this is fine to be merged and actually is a perfect example for me to ponder over next few days. theres a few things here that might set a dangerous'ish precendent. well not dangerous but might lead to complications down the line...

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2852.london.cloudapps.digital>

cpjmcquillan:
> this looks good. but can i suggest that before we go too gung-ho with page objects that i can give a bit of thought to how we use css selectors in them and document how we should use them? i mentioned it a couple of weeks ago i think. this is fine to be merged and actually is a perfect example for me to ponder over next few days. theres a few things here that might set a dangerous'ish precendent. well not dangerous but might lead to complications down the line...

I'll merge it in but let's do some work early next week to try and decide a consistent approach 👍

PR #2856, opened 2021-02-11
[TEVA-1650] Update Geocoder Config to use Google Geocoding API

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1650

## Changes in this PR:

- I have changed the configuration of geocoder so that it uses the Google Geocoding API. However, I noticed that this change lead to searches for 'Ilford' returning coordinates for a place in Kentucky - not good. 

- I took advantage of the Geocoding API's component filtering to restrict searches to the UK.

- I also changed the country code in location_suggestion.rb to make it consistent with the code used in geocoding.rb

- Searches for Ilford now return the correct coordinates. I have also tested this on a number other locations.

Screenshot 2021-02-11 at 16 26 35


david-mears-dfe:
You can limit it to the UK: https://developers.google.com/maps/documentation/geocoding/overview#component-filtering

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2856.london.cloudapps.digital>

david-mears-dfe:
I added your england bounds to my saved places on google maps:

Screenshot 2021-02-11 at 17 16 04

The positions look good to me. You are including the Isles of Scilly which is always nice (though not so much this time of year).

david-mears-dfe:
An excellent opportunity to use the `+map` cheat code! This looks good to me, but I'd like someone else to review it too, since search is so mission-critical.

david-mears-dfe (approved):

david-mears-dfe:
After merging, it would be worth waiting for the deploy notification and checking with `+map` on production, since there is an opportunity for the api key configuration to be wrong (being different on different environments).

cpjmcquillan (approved):

david-mears-dfe:
If it's not too much bother, you could add the GB component thing I linked too, just to exclude the edge of France and part of Ireland.

cpjmcquillan:
> If it's not too much bother, you could add the GB component thing I linked too, just to exclude the edge of France and part of Ireland.

Looks like we can [here](https://github.com/alexreisner/geocoder/blob/61b4a19ee7c7f9e41fae092da85fc5e9b5a0c5b1/lib/geocoder/lookups/google.rb#L85)

config/initializers/geocoder.rb
```diff
@@ -1,6 +1,6 @@
 Geocoder.configure(
-  lookup: :uk_ordnance_survey_names,
-  api_key: ENV.fetch("ORDNANCE_SURVEY_API_KEY", nil),
+  lookup: :google,
+  api_key: ENV.fetch("GOOGLE_LOCATION_SEARCH_API_KEY", ""),
```
alexbowen, line 3: am assuming its ok that second argument is empty string here but was nil before?

alexbowen (approved):
small query, but lgtm

config/initializers/geocoder.rb
josephhull676: If the given name does not exist and neither default nor a block a provided an KeyError is raised.

> I think this is just to prevent an error if the env var isn't found? I'm not too sure it matters if it's an empty string or nil.

In google_api.rb empty strings are used so I just copied that.

```
GOOGLE_API_JSON_KEY = ENV.fetch("GOOGLE_API_JSON_KEY", "")
GOOGLE_ANALYTICS_PROFILE_ID = ENV.fetch("GOOGLE_ANALYTICS_PROFILE_ID", "")
GOOGLE_LOCATION_SEARCH_API_KEY = ENV.fetch("GOOGLE_LOCATION_SEARCH_API_KEY", "")
```

alexbowen: im sure its fine, just noticed thats all

cpjmcquillan (approved):

PR #2857, opened 2021-02-11
Delete old spec

david-mears-dfe:
This spec is never run, because its filename does not end _spec.rb.

The route tested does not exist. The tests don't look very valuable.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2857.london.cloudapps.digital>

cpjmcquillan (approved):
😂

PR #2862, opened 2021-02-12
Include search criteria info in unsubscribe/job alert feedback

david-mears-dfe:
This info is handy to include, partly because the search criteria may be edited
after the feedback is submitted, and partly for convenience.

Added matcher to check two json strings match which are not necessarily in the same order,
since hashes can vary their orders. Named by analogy with existing hash_including method.

### DIFF
--- a/app/controllers/jobseekers/job_alert_feedbacks_controller.rb
+++ b/app/controllers/jobseekers/job_alert_feedbacks_controller.rb
@@ -40,13 +40,16 @@ class Jobseekers::JobAlertFeedbacksController < ApplicationController
     # `trigger_feedback_provided_event`, which we use to create Events, relies on feedback_attributes.
     # This method allows events to be created for either type of action. Here felt like the best place for this logic,
     # since job alert feedback is the only feedback type that has >1 action.
-    attributes = case action_name
-                 when "new"
-                   job_alert_email_link_params
-                 when "update"
-                   further_feedback_form_params
-                 end
-    attributes.merge(feedback_type: "job_alert", subscription_id: @subscription.id)
+    case action_name
+    when "new"
+      feedback_attributes_base.merge(job_alert_email_link_params)
+    when "update"
+      feedback_attributes_base.merge(further_feedback_form_params)
+    end
+  end
+
+  def feedback_attributes_base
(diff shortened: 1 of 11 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2862.london.cloudapps.digital>

app/controllers/jobseekers/job_alert_feedbacks_controller.rb
```diff
@@ -40,13 +40,16 @@ def feedback_attributes
...
+    case action_name
+    when "new"
+      feedback_attributes_base.merge(job_alert_email_link_params)
+    when "update"
+      feedback_attributes_base.merge(further_feedback_form_params)
+    end
+  end
+
+  def feedback_attributes_base
+    { feedback_type: "job_alert", search_criteria: @subscription.search_criteria, subscription_id: @subscription.id }
```
cpjmcquillan, line 52: Given when we trigger subscription events we send a `subscription_identifier`, should we be sending that here instead of `subscription_id`?

david-mears-dfe: Oh yeah, let's do that

cpjmcquillan (approved):
I wonder if consistent naming of `anonymised_subscription_id` here and `subscription_identifier` in the subscription events triggered by `subscriptions_controller.rb` might be useful for @stevenleggdfe? Apart from that looks good to me

david-mears-dfe:
> I wonder if consistent naming of `anonymised_subscription_id` here and `subscription_identifier` in the subscription events triggered by `subscriptions_controller.rb` might be useful for @stevenleggdfe? Apart from that looks good to me

There are two patterns in use at the moment - (whatever)_identifier in the top-level hash and anonymised_(whatever)_id in the hash contained therein. I feel like this is the least bad option, short of renaming loads of stuff (which would complicate queries with needing to be backwards-compatible).

david-mears-dfe:
> There are two patterns in use at the moment - `(whatever)_identifier` in the top-level hash and `anonymised_(whatever)_id` in the hash contained therein. I feel like this is the least bad option, short of renaming loads of stuff (which would complicate queries with needing to be backwards-compatible).

This is wrong. I'll see what I can do without making the code ugly.

PR #2867, opened 2021-02-12
Add jobs search system spec

cpjmcquillan:
VCR records request/response to cassettes which can be replayed for testing purposes.

In order to ensure we can all use the same test files and avoid the need for app IDs or secrets in the pipeline, responses have been anonymised and use the placeholder environment variables. 

For now VCR has been implemented only for algolia search, and mainly for a single spec file. There is a generic algoliasearch cassette which can be used to stub a useful response for other specs. 

In future I imagine we may want to refine this approach, and also add cassettes to improve other system specs that call out to external API services (e.g Google Drive API).

## Changes in this PR
- Add VCR gem
- Update Algolia config for test environment 
- Add a jobs search system spec

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -1,7 +1,7 @@
 module Indexable
   extend ActiveSupport::Concern
 
-  INDEX_NAME = [Rails.env.test? ? nil : ENV.fetch("ALGOLIA_INDEX_PREFIX", nil), DOMAIN, Vacancy].compact.join("-").freeze
+  INDEX_NAME = [Rails.configuration.algolia_index_prefix, DOMAIN, Vacancy].compact.join("-").freeze
 
   included do
     include AlgoliaSearch
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -1,14 +1,14 @@
 module Indexable
   extend ActiveSupport::Concern
 
-  INDEX_NAME = [ENV["ALGOLIA_INDEX_PREFIX"], DOMAIN, Vacancy].compact.join("-").freeze
+  INDEX_NAME = [Rails.env.test? ? nil : ENV.fetch("ALGOLIA_INDEX_PREFIX", nil), DOMAIN, Vacancy].compact.join("-").freeze
 
   included do
     include AlgoliaSearch
 
     scope :unindexed, (-> { live.where(initially_indexed: false) })
 
-    algoliasearch index_name: INDEX_NAME, auto_index: true, auto_remove: true, if: :listed? do
+    algoliasearch index_name: INDEX_NAME, auto_index: !Rails.env.test?, auto_remove: !Rails.env.test?, if: :listed? do
       attributes :education_phases, :job_roles, :job_title, :parent_organisation_name, :salary, :subjects, :working_patterns, :_geoloc
 
       attribute :expires_at do
--- a/app/services/search/strategies/algolia.rb
+++ b/app/services/search/strategies/algolia.rb
@@ -15,11 +15,11 @@ class Search::Strategies::Algolia
     @vacancies ||= Vacancy.includes(organisation_vacancies: :organisation).search(@keyword, search_arguments)
   rescue Algolia::AlgoliaProtocolError => e
     Rollbar.error("Algolia search error", details: e, search_arguments: search_arguments)
-    @vacancies = nil
+    @vacancies = Vacancy.none
   end
 
   def total_count
-    return 0 unless vacancies
+    return 0 if vacancies.none?
 
     vacancies.raw_answer["nbHits"]
   end
--- a/config/initializers/algoliasearch.rb
+++ b/config/initializers/algoliasearch.rb
@@ -1,5 +1,5 @@
 AlgoliaSearch.configuration = {
-  application_id: ENV.fetch("ALGOLIA_APP_ID", "Placeholder"),
-  api_key: ENV.fetch("ALGOLIA_WRITE_API_KEY", "Placeholder"),
+  application_id: Rails.env.test? ? "Placeholder" : ENV.fetch("ALGOLIA_APP_ID", "Placeholder"),
+  api_key: Rails.env.test? ? "Placeholder" : ENV.fetch("ALGOLIA_WRITE_API_KEY", "Placeholder"),
   pagination_backend: :kaminari,
 }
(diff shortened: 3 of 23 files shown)


config/initializers/algoliasearch.rb
```diff
@@ -1,5 +1,5 @@
 AlgoliaSearch.configuration = {
-  application_id: ENV.fetch("ALGOLIA_APP_ID", "Placeholder"),
-  api_key: ENV.fetch("ALGOLIA_WRITE_API_KEY", "Placeholder"),
+  application_id: Rails.env.test? ? "Placeholder" : ENV.fetch("ALGOLIA_APP_ID", "Placeholder"),
```
alexbowen, line 2: why does this pass `"Placeholder"` as second argument for a none test environment?

db/seeds.rb
```diff
@@ -51,7 +51,7 @@
                               town: "Farnham",
                               county: "Not recorded",
                               postcode: "GU9 8UG",
-                              geolocation: "(51.2023732521965,0.814476304733643)")
+                              geolocation: "(51.2023732521965,-0.814476304733643)")
```
alexbowen, line 54: this intentional? coordinates for farnham look like they should be +0.8ish

cpjmcquillan: [Wrong](https://www.google.com/search?ei=JDUqYJWBBMSS8gL8xZaYCg&q=51.2023732521965%2C0.814476304733643) vs [correct](https://www.google.com/search?q=51.2023732521965%2C-0.814476304733643)

config/initializers/algoliasearch.rb
cpjmcquillan: Because it can't be nil. Although in theory there shouldn't be a non-test environment where it is nil

alexbowen: ok, just as joe was passing an empty string in his PR - i guess for the reason you state

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2867.london.cloudapps.digital>

app/services/search/strategies/algolia.rb
```diff
@@ -15,11 +15,11 @@ def vacancies
     @vacancies ||= Vacancy.includes(organisation_vacancies: :organisation).search(@keyword, search_arguments)
   rescue Algolia::AlgoliaProtocolError => e
     Rollbar.error("Algolia search error", details: e, search_arguments: search_arguments)
-    @vacancies = nil
+    @vacancies = Vacancy.none
```
csutter, line 18: 😍

app/models/concerns/indexable.rb
```diff
@@ -1,14 +1,14 @@
 module Indexable
   extend ActiveSupport::Concern
 
-  INDEX_NAME = [ENV["ALGOLIA_INDEX_PREFIX"], DOMAIN, Vacancy].compact.join("-").freeze
+  INDEX_NAME = [Rails.env.test? ? nil : ENV.fetch("ALGOLIA_INDEX_PREFIX", nil), DOMAIN, Vacancy].compact.join("-").freeze
```
csutter, line 4: This has been here for a while but at least the first bit feels like it should live somewhere in configuration instead rather than us testing whether we are in a specific environment deep within app code 😕 

How about a `config.algolia_index_prefix = ENV.fetch("ALGOLIA_INDEX_PREFIX", nil)` in `config/application.rb`, and a `config.algolia_index_prefix = nil` in `config/enviroments/test.rb`?

csutter (approved):
See comment about Algolia index config, other than that it looks great 👍

spec/system/jobseekers_can_search_from_home_page_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe "Searching on the home page" do
+RSpec.describe "Searching on the home page", vcr: { cassette_name: "algoliasearch" } do
```
cesidio, line 3: ❤️

cesidio (approved):

app/models/concerns/indexable.rb
cpjmcquillan: Yeah that sounds good to me

app/models/concerns/indexable.rb
```diff
@@ -1,14 +1,14 @@
...
-  INDEX_NAME = [ENV["ALGOLIA_INDEX_PREFIX"], DOMAIN, Vacancy].compact.join("-").freeze
+  INDEX_NAME = [Rails.configuration.algolia_index_prefix, DOMAIN, Vacancy].compact.join("-").freeze
 
   included do
     include AlgoliaSearch
 
     scope :unindexed, (-> { live.where(initially_indexed: false) })
 
-    algoliasearch index_name: INDEX_NAME, auto_index: true, auto_remove: true, if: :listed? do
+    algoliasearch index_name: INDEX_NAME, auto_index: !Rails.env.test?, auto_remove: !Rails.env.test?, if: :listed? do
```
david-mears-dfe, line 11: I wouldn't want to make the test behaviour different from the real behaviour without good reason. What's the reason?

cpjmcquillan: If we don't do this then we'll be making requests to Algolia automatically every time we create a vacancy and they won't have been recorded in VCR cassettes

spec/system/jobseekers_can_search_for_jobs_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    it "displays page 1 jobs", vcr: { cassette_name: "algoliasearch teacher page 1" } do
+      expect(jobs_page).to have_jobs(count: 2)
+      expect(jobs_page.stats).to have_content(strip_tags(I18n.t("jobs.number_of_results_html", first: 1, last: 2, count: 6)))
+    end
+
+    context "when navigating between pages" do
+      it "displays page 3 jobs", vcr: { cassette_name: "algoliasearch teacher page 3" } do
+        jobs_page.pagination.go_to("3")
+
+        expect(jobs_page).to have_jobs(count: 2)
```
david-mears-dfe, line 34: Is `have_jobs` something page objects provides?

app/models/concerns/indexable.rb
cpjmcquillan: Which means lots and lots and lots of errors

spec/system/jobseekers_can_search_for_jobs_spec.rb
cpjmcquillan: https://github.com/site-prism/site_prism#testing-for-the-existence-of-the-element-collection

The count part is capybara

PR #2868, opened 2021-02-15
Remove `AlertVacancySearch`

csutter:
This doesn't really do enough to warrant being a class of its own - it
just internally overrides some parameters to the Algolia search that we
can pass at the time of use instead.

- Move keyword generation when a subscription does not have a keyword in
  its search criteria from `AlertVacancySearch` to `Subscription`
- Allow fuzziness ("typo tolerance") to be specified as an argument to
  `VacancySearch`
- Make subscription use vanilla `VacancySearch` with appropriate params
- Delete now unneeded `AlertVacancySearch`
- Remove pointless limiting of already limited search results in
  `AlertEmail::Base#vacancies_for_subscription`


app/models/subscription.rb
```diff
@@ -35,7 +37,14 @@ def unsubscribe
...
   def vacancies_for_range(date_from, date_to)
-    Search::AlertVacancySearch.new(search_criteria.symbolize_keys.merge(from_date: date_from, to_date: date_to)).vacancies
+    criteria = search_criteria.symbolize_keys.merge(from_date: date_from, to_date: date_to, keyword: keyword)
+    Search::VacancySearch.new(criteria, per_page: MAXIMUM_RESULTS_PER_RUN, fuzzy: false).vacancies
+  end
+
+  def keyword
+    # Generate a meaningful keyword for this alert if one hasn't been specified as part of the search criteria
+    search_criteria["keyword"].presence ||
+      [search_criteria["subject"], search_criteria["job_title"]].reject(&:blank?).join(" ")
```
cpjmcquillan, line 47: How do you feel about doing this once for all subscriptions and removing the logic? We have something funky going on for old subscriptions with search criteria containing `newly_qualified_teacher: true` too

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2868.london.cloudapps.digital>

app/models/subscription.rb
csutter: It did strike me as something that we may as well do once and be done with it – but I figured maybe it's because we may want to tweak that logic later, and if we were to set it on subscription creation then we could no longer differentiate between subscriptions where the jobseeker specified a keyword and those where we generated one for them...

Not sure about the NQT though – where is that? 😕

cpjmcquillan: In the filters... 

It's not so much business logic but more like "we just don't want to break stuff so keep it working". It hasn't been tweaked in 9 months so I'm pretty sure we can wipe the slate clean. Keywords never existed alongside subject and job title as far as I'm aware - and I'm fairly certain we're not going back to subject and job title again

cpjmcquillan (approved):

PR #2869, opened 2021-02-15
Start reporting CSP violations again

csutter:
We had turned this off because the Rollbar gem generated a CSP violation
on every pageview due to it skipping the CSP nonce if `unsafe_inline`
was set (which we do for backwards compatibility), but this has been
fixed in 3.1.2 and there seem to be no spurious CSP violations left.

We'll continue running in report-only mode for the time being, but in
theory we should no longer receive any reports now! 🎉


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2869.london.cloudapps.digital>

PR #2883, opened 2021-02-16
Do not send arrays to BigQuery

david-mears-dfe:
Do not pass Arrays to BigQuery: otherwise, it will give the error 'Array specified for non repeated field' when the array has length of 1.

Having tested the task on a local production-like data dump, connected to staging BigQuery dataset, I will run the task on production after this is deployed.

### DIFF
--- a/app/services/event.rb
+++ b/app/services/event.rb
@@ -34,15 +34,13 @@ class Event
   end
 
   ##
-  # For json objects or hashes passed to Event#trigger as values in the `data` param, such as
-  # Subscription#search_criteria or Feedback#search_criteria, we should format these as json for BigQuery,
-  # rather than strings, for easier data manipulation by Performance Analysis.
-  # Floats, Integers and Arrays, such as job alert Feedbacks with a job_alert_vacancy_ids attribute,
-  # should remain as they are.
+  # For json objects or hashes passed to Event#trigger as values in the `data` param, such as Feedback#search_criteria,
+  # we should format these as json for BigQuery, rather than strings, for easier manipulation by Performance Analysis.
+  # Floats and Integers should remain as they are. Do not pass Arrays to BigQuery without converting them to string:
+  # otherwise, it will give the error "Array specified for non repeated field" when the array has length of 1.
   # @param [Object] value Any value in the data passed to the event.
   def formatted_value(value)
     return value if value.is_a?(Float) || value.is_a?(Integer)
-    return value.map { |item| formatted_value(item) } if value.is_a?(Array)
 
     value.respond_to?(:keys) ? value.to_json : value&.to_s
   end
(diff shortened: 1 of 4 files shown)


app/services/event.rb
```diff
@@ -34,15 +34,13 @@ def base_data
...
-  # rather than strings, for easier data manipulation by Performance Analysis.
-  # Floats, Integers and Arrays, such as job alert Feedbacks with a job_alert_vacancy_ids attribute,
-  # should remain as they are.
+  # For json objects or hashes passed to Event#trigger as values in the `data` param, such as Feedback#search_criteria,
+  # we should format these as json for BigQuery, rather than strings, for easier manipulation by Performance Analysis.
+  # Floats and Integers should remain as they are. Do not pass Arrays to BigQuery: otherwise, it will give the
+  # error "Array specified for non repeated field" when the array has length of 1.
   # @param [Object] value Any value in the data passed to the event.
   def formatted_value(value)
     return value if value.is_a?(Float) || value.is_a?(Integer)
```
cpjmcquillan, line 43: We can still pass arrays, but if the array is length 1 we should just pass the value. This stops us sending arrays altogether though doesn't it?

david-mears-dfe: Yes, but I reckon it would be confusing to be storing an attribute in multiple formats: both single values and stringified arrays.

david-mears-dfe: Compare: https://github.com/DFE-Digital/teaching-vacancies/blob/338d3b0a9553bbe5dec4dc0dd78bb1e6efac6f75/lib/export_tables_to_big_query.rb#L109

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2883.london.cloudapps.digital>

spec/services/event_spec.rb
```diff
@@ -10,7 +10,6 @@
         data: [
           { key: "foo", value: "Bar" },
           { key: "pi", value: 3.14 },
-          { key: "baz", value: [1, "string", 0.5] },
```
cpjmcquillan, line 13: Can we leave this

spec/services/event_spec.rb
```diff
@@ -19,7 +18,6 @@
         subject.trigger(:reticulated_splines,
                         foo: "Bar",
                         pi: 3.14,
-                        baz: [1, "string", 0.5],
```
cpjmcquillan, line 22: And replace this with `baz: [1, "string", 0.5].to_s` (whatever that is)?

app/services/event.rb
cpjmcquillan: Ok. Can you update the comment to say that Arrays are converted to strings rather than they're not sent?

cpjmcquillan (approved):

app/services/event.rb
david-mears-dfe: Yeah

spec/services/event_spec.rb
david-mears-dfe: yep

PR #2884, opened 2021-02-16
Persist generated subscription keywords

csutter:
We currently have code that generates a keyword for a subscription from
the subject and job_title in its search criteria, if the subscription
doesn't have a keyword to begin with. This used to live in a separate
`AlertSearchBuilder`, which we removed in a recent PR, and is now a
method in `Subscription`.

Newly created subscriptions no longer need this because keywords are now
generated in `CriteriaDeviser`, so we can just persist the generated
keywords against the existing "old" subscriptions, and then after get
rid of the `Subscription#keyword` method.

Also removes the now unneeded subscription location category task.


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2884.london.cloudapps.digital>

david-mears-dfe (approved):

PR #2888, opened 2021-02-16
Add an HTTP permissions policy

csutter:
Enable the default Rails permissions policy and make it more restrictive
(we are not using any of these APIs anytime soon).


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2888.london.cloudapps.digital>

PR #2891, opened 2021-02-16
[TEVA-1901] cookie consent ab test [DO NOT MERGE]

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1901


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2891.london.cloudapps.digital>

app/components/ab_test/cookies_banner_component/cookies_banner_component.html+gds.slim
```diff
@@ -0,0 +1,14 @@
+.govuk-cookie-banner.cookies-banner-component--gds role="region" aria-label = t("cookies_preferences.banner.aria.label")
+  .govuk-cookie-banner__message.govuk-width-container
+    .govuk-grid-row
+      .govuk-grid-column-two-thirds
+        h2.govuk-cookie-banner__heading.govuk-heading-m = "Cookies on Teaching Vacancies"
```
alexbowen, line 5: havent added translations, i think bit premature as this is a test

csutter: In which case you could get rid of outputting a static string:

```slim
h2.govuk-cookie-banner__heading.govuk-heading-m Cookies on Teaching Vacancies
```

app/controllers/concerns/ab_testable.rb
```diff
@@ -5,6 +5,10 @@ module AbTestable
     helper_method :ab_variant_for, :current_ab_variants
   end
 
+  def show_test_variant(name_of_test, variant_name)
+    Rails.env.production? && ab_variant_for(name_of_test) == variant_name
```
csutter, line 9: Why do we need to test for prod here?

app/controllers/application_controller.rb
```diff
@@ -35,6 +35,10 @@ def cookies_preference_set?
     cookies["consented-to-cookies"].present?
   end
 
+  def cookie_preferences_show?
+    controller_name != "cookies_preferences" && !cookies_preference_set?
```
csutter, line 39: I think it might be nicer to have this just test for `!cookies_preference_set?`, and then override this method to always return false in `CookiesPreferencesController` (rather than checking for the name of a controller).

app/controllers/concerns/ab_testable.rb
csutter: Also the method name kinda implies to me that this has a side effect, plus we use just `test` (without name) in other places in this file and elsewhere – how about `current_variant?(test, variant)` instead?

app/views/layouts/application.html.slim
```diff
@@ -20,11 +20,16 @@ html.govuk-template.app-html-class lang="en"
...
     = csrf_meta_tags
   body class=body_class data-vacancy_state=(@vacancy.present? ? @vacancy.state : "create")
+    - if show_test_variant(:"2021_02_cookie_consent_test", :modal) && cookie_preferences_show?
+      .modal-container
     = render partial: "shared/google_tag_manager_body" if cookies["consented-to-cookies"] == "yes"
     = javascript_tag(nonce: true) do
       | document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
     = render "shared/skip_links"
-    = render "layouts/cookies_banner" unless controller_name == "cookies_preferences" || cookies_preference_set?
+    = render(AbTest::CookiesBannerComponent.new(create_path: create_cookies_preferences_path(cookies_consent: "yes", **utm_parameters), preferences_path: cookies_preferences_path(**utm_parameters)).with_variant(:bottom)) if show_test_variant(:"2021_02_cookie_consent_test", :bottom) && !cookie_preferences_hidden?
```
csutter, line 29: Could we just do this in one go? e.g.:

```ruby
render(AbTest::CookiesBannerComponent.new(...).with_variant(ab_variant_for(:"2021_02_cookie_consent_test"))) if cookie_preferences_show?
```

csutter (changes requested):
Couple of thoughts that I'd like to see addressed, but looks good otherwise!

app/controllers/application_controller.rb
alexbowen: if thats the case ill remove this method. tbh didnt question really why `controller_name != "cookies_preferences"` was there, just a copy and paste from last abtest!

csutter: If it was in the application layout, it'll be because that's the only way the layout will know which controller it's rendered from. Here we're in the controller to begin with, so can do it more cleanly 🧹

app/controllers/concerns/ab_testable.rb
alexbowen: should more acurately be !Rails.env.test

app/views/layouts/application.html.slim
alexbowen: what i would have ideally like to do but the presence of another variant lower in the markup makes it tricky

csutter: Gotcha. Maybe another alternative could be:

```ruby
- unless cookie_preferences_hidden? || current_variant?(:"2021_02_cookie_consent_test", :top)
  = render(AbTest::CookiesBannerComponent.new(...).with_variant(ab_variant_for(:"2021_02_cookie_consent_test")))
```

app/controllers/concerns/ab_testable.rb
csutter: Do some of the tests expect the default variant and break if another one is selected?

I think a cleaner way in that case would be to configure the AB test to always show the default variant in test in `config/ab_tests.yml`:

```yaml
shared:
  # ...
test:
  # Force test environment to always show default cookie banner
  2021_02_cookie_consent_test:
    none: 0
    default: 1
    top: 0
    bottom: 0
    modal: 0
    gds: 0
```

csutter (approved):
Couple more comments then we're good to go 👍

app/controllers/cookies_preferences_controller.rb
```diff
@@ -16,6 +16,10 @@ def create
     end
   end
 
+  def cookies_preference_set?
+    true
```
cpjmcquillan, line 20: This can't be correct, can it?

cpjmcquillan (changes requested):
Cookie banner shouldn't be shown if preferences are set or on cookie preferences page itself

app/controllers/cookies_preferences_controller.rb
csutter: This overrides the method in `ApplicationController` and makes the banner not show on this controller

csutter:
> Cookie banner shouldn't be shown if preferences are set or on cookie preferences page itself

It doesn't for me? 😕

app/controllers/concerns/ab_testable.rb
```diff
@@ -5,8 +5,8 @@ module AbTestable
     helper_method :ab_variant_for, :current_ab_variants
   end
 
-  def show_test_variant(name_of_test, variant_name)
-    Rails.env.production? && ab_variant_for(name_of_test) == variant_name
+  def current_variant?(name_of_test, variant_name)
+    !Rails.env.test? && ab_variant_for(name_of_test) == variant_name
```
csutter, line 9: One final thing, I promise – can we take this out now? 😅

app/controllers/cookies_preferences_controller.rb
alexbowen: this makes the cookie banner never appear on the cookies page, to overide the general rule

app/controllers/concerns/ab_testable.rb
alexbowen: ah i forgot, its gone now

alexbowen:
@cpjmcquillan  the method i put in the c

> Cookie banner shouldn't be shown if preferences are set or on cookie preferences page itself

@cpjmcquillan closing this as i this is the implemented behaviour, just double checked

cpjmcquillan (approved):

alexbowen:
> Cookie banner shouldn't be shown if preferences are set or on cookie preferences page itself

@cpjmcquillan closing this as i this is the implemented behaviour, just double checked

app/controllers/cookies_preferences_controller.rb
cpjmcquillan: `Read the filename...`

PR #2910, opened 2021-02-18
Refactor email events to use data struct instead of base data

cpjmcquillan:
When we stream events to BigQuery, we stream them to a single table `events`. This table has a defined schema (essentially the base data from request events). 

We want to continue to stream events to a single table because it makes analytics easier. This means that email events cannot have their own base data, since these columns would not be defined in the schema. 

Instead email events have some default data which is sent alongside event data (triggered by the event) in the data struct to BigQuery.

This PR adds the concept of default data for events.

### DIFF
--- a/app/services/email_event.rb
+++ b/app/services/email_event.rb
@@ -14,12 +14,12 @@ class EmailEvent < Event
 
   attr_reader :notify_template, :email, :jobseeker, :publisher
 
-  def base_data
-    @base_data ||= super.merge(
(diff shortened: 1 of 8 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2910.london.cloudapps.digital>

david-mears-dfe (approved):
LGTM

cesidio (approved):

PR #2914, opened 2021-02-18
[TEVA 2076] Add initial automated a11y testing

csutter:
This uses the `axe` tool in a separate set of accessibility system specs
to identify violations of WCAG standards in what will eventually be a
representative sample of user journeys.

Fixes the following accessibility violations:
- Add `aria-hidden` to several purely decorative images (social media
  icons, GOV.UK crown in logotype)
- Filter and sort form on search results page should have distinct
  HTML IDs (they were clashing due to `as: ""` on the form)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,8 @@ group :development do
 end
 
 group :development, :test do
+  gem "axe-core-capybara"
+  gem "axe-core-rspec"
   gem "brakeman"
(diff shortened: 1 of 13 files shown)


alexbowen:
included in this is improvements/fixes for map a11y that was failing a11y test

david-mears-dfe (approved):

PR #2915, opened 2021-02-18
[TEVA-2071] implement card view component

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2071

- tidies up/corrects HTML used for cards
- new card component only added to jobseeker pages until component is finalised
- key/values used for card title/content, so some `_html` can be removed from locale.yml


app/components/shared/card_component/card_component.scss
```diff
@@ -0,0 +1,64 @@
...
+    margin-top: 0;
+    text-align: right;
+
+    .govuk-list {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.govuk-grid-column-two-thirds {
```
alexbowen, line 57: cant for the life of me remeber why this is here, so leaving as is

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,55 @@
...
+  class HeaderItemComponent < ViewComponent::Base
+    attr_accessor :label, :value
+
+    def initialize(value:, label: nil)
+      @label = label
+      @value = value
+    end
+
+    def call
+      (@label ? content_tag(:span, "#{@label}: ", { class: "govuk-!-font-weight-bold" }) : "") + @value
```
cesidio, line 24: I would simplify this with:
```
return value unless label

content_tag(:span, "#{label}: ", { class: "govuk-!-font-weight-bold" }) + value
```

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,55 @@
...
+  class BodyItemComponent < ViewComponent::Base
+    attr_accessor :label, :value
+
+    def initialize(value:, label: nil)
+      @label = label
+      @value = value
+    end
+
+    def call
+      (@label ? content_tag(:span, "#{@label}: ", { class: "govuk-!-font-weight-bold" }) : "") + @value
```
cesidio, line 37: if you have `attr_accessor :label, :value` you do not need the @ for @label and @value

alexbowen: i knew this

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2915.london.cloudapps.digital>

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,59 @@
...
+  renders_many :header_items, "HeaderItemComponent"
+
+  attr_accessor :html_attributes, :id
+
+  def initialize(id: nil, classes: [], html_attributes: {})
+    super(classes: classes, html_attributes: html_attributes)
+    @id = id
+  end
+
+  class HeaderItemComponent < ViewComponent::Base
```
david-mears-dfe: I don't have a problem with it, but why do you prefer to make these parts into component classes rather than methods?

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,41 @@
...
+  renders_many :header_items, "ItemComponent"
+
+  attr_accessor :html_attributes, :id
+
+  def initialize(id: nil, classes: [], html_attributes: {})
+    super(classes: classes, html_attributes: html_attributes)
+    @id = id
+  end
+
+  class ItemComponent < ViewComponent::Base
```
cpjmcquillan, line 15: When we've got components this concise, I think it makes sense to use [lambda components ](https://viewcomponent.org/#lambda-slots)

app/components/shared/card_component/card_component.html.slim
```diff
@@ -0,0 +1,16 @@
+dl.card-component id=id class=classes *html_attributes
+  .card-component__content
+    dt.card-component__header
+      ul.govuk-list
+        - header_items.each do |item|
+          li == render(item)
```
cpjmcquillan, line 6: You should be able to just do `li = item`

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,41 @@
+class Shared::CardComponent < GovukComponent::Base
+  include ViewComponent::SlotableV2
+
+  renders_many :action_items, "ActionItemComponent"
+  renders_many :body_items, "ItemComponent"
+  renders_many :header_items, "ItemComponent"
```
david-mears-dfe, line 6: Can you persuade me that we need two methods to access the same component?

app/components/shared/card_component.rb
```diff
@@ -0,0 +1,41 @@
+class Shared::CardComponent < GovukComponent::Base
+  include ViewComponent::SlotableV2
+
+  renders_many :action_items, "ActionItemComponent"
```
cpjmcquillan, line 4: I'd rather `renders_many :actions` with lamba components (and similar for body and header)

app/components/shared/card_component.rb
alexbowen: because the component does exactly the same thing but the items for each are rendered in different parts of the component

app/components/shared/card_component/card_component.html.slim
alexbowen: has to bee == or it doesnt render but yeah the render method not needed

david-mears-dfe (approved):
Tests look good. I think Connor makes a good point about the lambda slots. Either way, I approve.

PR #2923, opened 2021-02-19
Make seeding DB conditional on presence of vacancies

csutter:
We used to run review apps using a separate Rails environment of
"staging", which is no more. They now (sensibly) refuse to load the
seeds, so we need to relax the requirement of "never seed in production"
to "never seed in production when any vacancies exist".

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,4 +1,4 @@
-raise if Rails.env.production?
+raise "Aborting seeds - running in production with existing vacancies" if Rails.env.production? && Vacancy.any?
 
 require "faker"
 require "factory_bot_rails"


db/seeds.rb
```diff
@@ -1,4 +1,4 @@
-raise if Rails.env.production?
+raise "Aborting seeds - running in production with existing vacancies" if Rails.env.production? && Vacancy.any?
```
cesidio, line 1: or maybe check that `DOMAIN != PROD_DOMAIN` or something like that? I am ok with this too

cesidio (approved):

db/seeds.rb
csutter: I was thinking about that, or adding an `APP_ROLE` env var or something – but this strikes me as a good balance because it'll cover us without any extra effort even if we change domains etc 🤞

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2923.london.cloudapps.digital>

PR #2928, opened 2021-02-19
Make notification components less hostile to screen-readers

david-mears-dfe:
The issue which the html_attributes attribute tries to solve is when we
reuse the notification component as something other than a notification,
e.g. for an 'empty' box (such as 'You have no saved jobs'), or for an
info box (such as 'You can unsubscribe in such and such a way'), we don't
want to give it the html role of 'alert' or the tabindex of '-1' which
cause screen-readers to get to the non-notification before anything else
and mess with the tab-order.

This can be refactored in the very near future but I wanted to get this
piece of accessibility in first so that it can be retained when the
refactoring happens - but maybe doing things in the other order
(refactor then accessibilize) is better?

### DIFF
--- a/app/components/shared/notification_component.rb
+++ b/app/components/shared/notification_component.rb
@@ -4,14 +4,14 @@ class Shared::NotificationComponent < ViewComponent::Base
                  dismiss: true,
                  background: false,
                  alert: "warning",
-                 html_attributes: { role: "alert", tabindex: "-1" })
+                 html_attributes: nil)
     @content = content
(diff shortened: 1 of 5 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2928.london.cloudapps.digital>

cpjmcquillan (approved):
Let's merge this now and refactor them all soon. I have a few ideas for how we might do this and other similar things

PR #2929, opened 2021-02-19
Disable indexing jobs on Google on non-prod envs

cesidio:


spec/controllers/publishers/vacancies/application_controller_spec.rb
```diff
@@ -4,23 +4,17 @@
   describe "#update_google_index" do
     let!(:vacancy) { create(:vacancy) }
 
-    context "when in production" do
-      before(:each) do
-        allow(Rails).to receive(:env)
-          .and_return(ActiveSupport::StringInquirer.new("production"))
-      end
+    context "when DisableExpensiveJobs is not enabled" do
+      before(:each) { allow(DisableExpensiveJobs).to receive(:enabled?).and_return(false) }
```
cpjmcquillan, line 8: I'm pretty sure `before` is the same as `before(:each)`

cpjmcquillan (approved):
I'm happy with this approach - it seems cumbersome to have lots of different flags or a different approach until/unless we need that level of configurability

spec/controllers/publishers/vacancies/application_controller_spec.rb
cesidio: copy and 🍝

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2929.london.cloudapps.digital>

PR #2932, opened 2021-02-22
Improve location geocoding

cpjmcquillan:
The situation as I understand it...
- for some reason we were hitting this [usage limit](https://developers.google.com/maps/documentation/geocoding/usage-and-billing#other-usage-limits)
- we were caching no_match results 

This PR...
- catches `OVER_QUERY_LIMIT` errors from Google Geocoding API, and logs the error to Rollbar
- falls back to Ordnance Survey API
- only caches valid coordinates

I assume that Google calculates QPS (queries per second) based on Product SKU and not API key, so there wouldn't be a benefit to using separate API keys for maps, geocoding and places autocomplete. Does anybody have any ideas how we can confirm this assumption?

This should put us in a better position to know that something is going wrong, but it won't help us not hit the usage limit. If we want to tackle that issue we could think about...
- identifying and filtering out geocoding requests from bots
- adding random delays into geocoding requests
- pre-fetching popular geocoding requests on a scheduled basis once a month at sensible time intervals

We fallback to using Ordnance Survey whenever we receive `OVER_QUERY_LIMIT` errors from Google.

Please add any thoughts you might have!

_disclaimer: I don't like that we return `Geocoder::DEFAULT_STUB_COORDINATES` when `default_lookup` config is `:test`, but the fact we're specifically using several different lookups in the `coordinates` method means the default test lookup config does not work - suggestions welcome_


lib/geocoding.rb
```diff
@@ -8,15 +8,24 @@ def initialize(location)
   end
 
   def coordinates
+    geocoded_location = begin
+      Geocoder.coordinates(location, components: "country:uk")
```
david-mears-dfe, line 12: @josephhull676's original PR changed this to `components: "country:GB"` - I think GB is the correct country code.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2932.london.cloudapps.digital>

david-mears-dfe:
For information, we hit the usage limit on the first day we started using Google, and continued to hit the limit around 100-200 times an hour until we switched back to Ordnance Survey.

david-mears-dfe:
Perhaps we could fall back to Ordnance Survey geocoder if we hit the limit? (And then not cache it, I suppose.)

lib/geocoding.rb
david-mears-dfe: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#GB

cpjmcquillan: We're so special we have 2, but I can go with `gb` everywhere instead 👍

cpjmcquillan:
> Perhaps we could fall back to Ordnance Survey geocoder if we hit the limit? (And then not cache it, I suppose.)

Good idea, let me see what I can do

cesidio (approved):

PR #2934, opened 2021-02-22
[SPIKE] Use CurrentAttributes and AR callbacks for events

csutter:
# DO NOT MERGE – just for discussion

A tiny spike to test feasibility of using `CurrentAttributes` to handle
entity CRUD events in the model layer more cleanly.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2934.london.cloudapps.digital>

cpjmcquillan:
`ActiveSupport::CurrentAttributes` seems a bit scary, but I like the idea.

I think if we can get a nice config for allowed parameters this could be a relatively straightforward approach 👍

PR #2937, opened 2021-02-22
[TEVA-1842] Delete old schoolgroupmemberships

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1842

## Changes in this PR:

- Delete SchoolGroupMemberships which are no longer provided in import data (and which therefore shouldn't exist).
- Do this by:
1. marking all memberships as 'do_not_delete: false' before running any import tasks
2. marking the memberships which are the result of the current task with do_not_delete: true
3. and then deleting any left over memberships as old.

## Next steps:

- [ ] Test out the task by running it on production-like data
- [ ] Kill any left-over sidekiq jobs which are obsoleted but still hanging around in the queue (import_school_data, import_trust_data)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2937.london.cloudapps.digital>

cesidio:
Wouldn't adding a simple flag `do_not_delete` set to false to all records before starting the task be easier to implement? records that should be kept will set the the `do_not_delete` flag to true and at the end of the task, all records with the flag still false would be deleted.

lib/organisation_import/import_organisation_data.rb
```diff
@@ -4,11 +4,33 @@
...
+        send(method, row)
+      end
+    end
+    File.delete(csv_file_location)
+  end
+
   def create_organisation(row)
     organisation = convert_to_organisation(row)
-    organisation.save
+    organisation&.save
```
cpjmcquillan, line 33: Is it worth flagging when schools aren't present in our database at this point?

spec/lib/organisation_import/import_school_data_spec.rb
```diff
@@ -1,6 +1,7 @@
 require "rails_helper"
 
 RSpec.describe ImportSchoolData do
+  let(:import_task_run_id) { "bb762405-3163-4fe5-b951-8bcaa10d82ad" }
```
cpjmcquillan, line 4: Don't need this anymore do you?

spec/rails_helper.rb
```diff
@@ -92,7 +92,8 @@
   config.cassette_library_dir = "spec/fixtures/vcr"
   config.hook_into :webmock
   config.configure_rspec_metadata!
-  config.ignore_hosts "localhost", "127.0.0.1"
+  config.ignore_localhost = true
```
cpjmcquillan, line 95: `127.0.0.1` was part of our webmock config and I did have some failing tests without it - maybe the move to request specs has ironed that out!

cpjmcquillan (approved):
LGTM, one thing you could do is use a vcr cassette for the import job (useful in itself but also useful to get more members of the team familiar with how it works)!

david-mears-dfe:
> LGTM, one thing you could do is use a vcr cassette for the import job (useful in itself but also useful to get more members of the team familiar with how it works)!

Looked into it, but the response comes back as a binary, so we can't break it down into a small cassette - so the test would take the same length of time as the task!

spec/rails_helper.rb
david-mears-dfe: It hasn't, as we still use webmock in system specs.

lib/organisation_import/import_organisation_data.rb
david-mears-dfe: This is because of 
```ruby
return unless data_is_for_multi_academy_trust?(row)
```
returning nil.

app/jobs/import_organisation_data_job.rb
```diff
@@ -0,0 +1,15 @@
...
+class ImportOrganisationDataJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    return if DisableExpensiveJobs.enabled?
+
+    ImportOrganisationData.mark_all_school_group_memberships_to_be_deleted!
+    ImportSchoolData.new.run!
+    ImportTrustData.new.run!
+    ImportOrganisationData.delete_all_old_school_group_memberships!
```
cesidio, line 13: I would call it `delete_merked_old_school_group_memberships`

cesidio (approved):

lib/organisation_import/import_organisation_data.rb
```diff
@@ -4,11 +4,33 @@
...
+  def run!
+    csv_metadata.each { |metadata| import_data(metadata) }
+  end
+
+  def self.mark_all_school_group_memberships_to_be_deleted!
+    SchoolGroupMembership.update_all(do_not_delete: false)
+  end
+
+  def self.delete_marked_school_group_memberships!
+    SchoolGroupMembership.where(do_not_delete: false).delete_all
```
cesidio, line 16: I would add a check here:
```
return if SchoolGroupMembership.where(do_not_delete: false).count > SOME_THRESHOLD
```
if we have too many school_group_memberships to delete, maybe the task did not run correctly

cpjmcquillan:
> > LGTM, one thing you could do is use a vcr cassette for the import job (useful in itself but also useful to get more members of the team familiar with how it works)!
> 
> Looked into it, but the response comes back as a binary, so we can't break it down into a small cassette - so the test would take the same length of time as the task!

Fair enough!

PR #2940, opened 2021-02-23
Clamp Algolia expiry filter to start of hour

csutter:
This is the least effort way of making Algolia queries a bit easier to
cache. Because we need to filter on `expires_at_timestamp` at query
time, we can't cache any given query because the next one will have a
different timestamp.

### DIFF
--- a/app/services/search/filters_builder.rb
+++ b/app/services/search/filters_builder.rb
@@ -59,7 +59,7 @@ class Search::FiltersBuilder
   end
 
   def expired_now_filter
-    Time.current.to_time.to_i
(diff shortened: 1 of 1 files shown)


cpjmcquillan:
Have we thought about getting rid of the filter but reindexing (they get removed) expired vacancies instead?

Either way I think it's preferable to list a job in the results for up to an hour _after_ it expires though - but maybe we aren't best placed to decide.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2940.london.cloudapps.digital>

csutter:
Decided not to go ahead 🚮

PR #2944, opened 2021-02-23
Add `enforce_local_authority_allowlist` config

csutter:
We do not want to enforce the LA access allowlist in all environments -
`dev`, and `review` should allow logging in as any LA, but `staging` and
`production` should enforce it. Rather than checking for environment as
before, this makes this configuration explicit.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2944.london.cloudapps.digital>

terraform/workspace-variables/staging_app_env.yml
```diff
@@ -6,6 +6,7 @@ DFE_SIGN_IN_REDIRECT_URL: https://staging.teaching-vacancies.service.gov.uk/auth
 DFE_SIGN_IN_URL: https://pp-api.signin.education.gov.uk
 DISABLE_EXPENSIVE_JOBS: true
 DOMAIN: staging.teaching-vacancies.service.gov.uk
+ENFORCE_LOCAL_AUTHORITY_ALLOWLIST: false
```
cpjmcquillan, line 9: I think it should be `true` here in the interests of keeping staging as close to production as reasonably possible

cpjmcquillan (approved):

PR #2953, opened 2021-02-24
Re-enable arbitrary LA access on staging

csutter:
In the long term, we want to use staging purely within our CI pipeline.
However, various parts of the product team rely on having access to
staging for UR and demo purposes at the moment, and until we have found
an alternative solution we need to stop enforcing the LA allowlist on
staging.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2953.london.cloudapps.digital>

cpjmcquillan (approved):
😩

PR #2955, opened 2021-02-24
Actually use the cache when geocoding...

cpjmcquillan:
- Actually use the cache rather than making requests every time
- Only cache valid results
- Log when no match results are returned
- Log when Google Geocoding API responds with `OVER_QUERY_LIMIT`


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2955.london.cloudapps.digital>

csutter (approved):

PR #2956, opened 2021-02-24
Limit API to only return live vacancies

csutter:
Our API has a fair share of issues, and we need to take a step back and
rethink it in the future.

Until then, we need to limit vacancies being returned to just be live
vacancies as existing users are crawling hundreds of pages of expired
vacancies which is causing issues, and potential new users don't want
to deal with this overload.

### DIFF
--- a/app/controllers/api/vacancies_controller.rb
+++ b/app/controllers/api/vacancies_controller.rb
@@ -1,12 +1,11 @@
 class Api::VacanciesController < Api::ApplicationController
   before_action :verify_json_request, only: %i[show index]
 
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):
Previously we've spoken about either
- turning it off
- adding authentication
so we can see who actually uses it (complains) and then either get rid of the API or actually make it useful. 

At the minute it exists but receives zero attention or care

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2956.london.cloudapps.digital>

PR #2982, opened 2021-03-01
[TEVA-1758] Add route to jobseeker job applications index page

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1758

## Changes in this PR
- Consistent option order in routes (only/controller)
- Use resource for jobseeker job_application new/create routes (prevents clash + makes more sense)
- Add index route for jobseeker job_applications

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,17 +27,17 @@ Rails.application.routes.draw do
     end
 
     constraints(-> { JobseekerApplicationsFeature.enabled? }) do
-      scope path: ":job_id" do
-        resources :job_applications, only: %i[new create]
-      end
-
-      resources :job_applications, only: [] do
+      resources :job_applications, only: %i[index] do
         resources :build, only: %i[show update], controller: "job_applications/build" do
           resources :details, only: %i[new create edit update destroy], controller: "job_applications/details"
         end
         get :review
         post :submit
         resource :feedback, only: %i[create], controller: "job_applications/feedbacks"
+      end
+
+      scope path: ":job_id" do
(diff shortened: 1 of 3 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2982.london.cloudapps.digital>

config/routes.rb
```diff
@@ -27,18 +27,18 @@
...
         resources :build, only: %i[show update], controller: "job_applications/build" do
           resources :details, only: %i[new create edit update destroy], controller: "job_applications/details"
         end
         get :review
         post :submit
         resource :feedback, only: %i[create], controller: "job_applications/feedbacks"
       end
+
+      scope path: ":job_id" do
+        resource :job_application, only: %i[new create]
```
josephhull676, line 40: Why the change to resource (singular) and job_application (singular)?

cpjmcquillan: > Use resource for jobseeker job_application new/create routes (prevents clash + makes more sense)
It prevents a clash between the create action for a job application and the new route. And it makes more sense because there's never going to be more than one job application for a particular jobseeker for a particular job

david-mears-dfe: My guess: We're in the namespace job_ids within the namespace for jobseekers, and there can only be one job application per job per jobseeker.

david-mears-dfe (approved):

config/routes.rb
cpjmcquillan: I should probably do the same for saved jobs as well, but I'll make another micro PR for that

PR #2989, opened 2021-03-02
Refactor card component

cpjmcquillan:
- Refactor card component (this is necessary in order to be less restrictive about what goes in the header/body/action slots)
- Implement use
- Fix bug where account header email didn't break on small screens
- Use `h2` tag for account header and `h1` for page heading
- Add yarn to `.tool-versions` for people that use `asdf`


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2989.london.cloudapps.digital>

cesidio (approved):

david-mears-dfe (commented):
You're removing all uses of the `sort-inline` class, so it's probably worth getting rid of the class too.

PR #2990, opened 2021-03-02
Allow DB to be recreated on review apps

csutter:
Now that the review apps run in `production`, the app start command
including `rails db:schema:load` fails from the second deploy onwards
(as by default Rails won't allow destructive actions to be performed on
an existing production database).

This means one of the two instances (the db initialisation only runs on
the first CF instance) will refuse to come up starting from the second
deploy.

- Set `DISABLE_DATABASE_ENVIRONMENT_CHECK` in review apps to allow
  destructive actions to take place on their databases


cesidio (approved):

cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2990.london.cloudapps.digital>

PR #2993, opened 2021-03-02
WIP: [TEVA-2033] Update Job listing "Apply for job" CTA based on application status

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2033

## Changes in this PR:

- Added helpers to change the link_path and link_text of the BannerLinkComponent depending on whether a jobseeker has applied and the status of the application. 
- Added conditional and helper so that the 'save this job for later' link is only rendered if a jobseeker is not signed in or hasn't applied for the job (if they are signed in)
- Updated translations in related specs

### Jobseeker signed out

Screenshot 2021-03-02 at 16 07 01

### Jobseeker signed in and hasn't applied

Screenshot 2021-03-02 at 16 08 27

#### Application Started

Screenshot 2021-03-02 at 16 10 49

#### Application Complete

Screenshot 2021-03-02 at 16 19 44

### DIFF
--- a/app/helpers/jobseekers/job_applications_helper.rb
+++ b/app/helpers/jobseekers/job_applications_helper.rb
@@ -1,0 +1,33 @@
+module Jobseekers::JobApplicationsHelper
+  def application_link_path
+    if job_application.nil?
+      new_jobseekers_job_job_application_path(@vacancy.id)
+    elsif jobseeker_application_status == "submitted"
+      jobseekers_job_application_path(job_application)
+    elsif jobseeker_application_status == "draft"
+      jobseekers_job_application_review_path(job_application)
+    end
+  end
+
+  def application_link_text
+    if jobseeker_application_status.nil?
+      "apply"
+    elsif jobseeker_application_status == "submitted"
+      "submitted"
+    elsif jobseeker_application_status == "draft"
+      "draft"
+    end
+  end
+
+  def job_application
+    current_jobseeker&.job_applications&.find_by(vacancy_id: @vacancy.id)
+  end
+
+  def jobseeker_application_status
+    current_jobseeker&.job_applications&.find_by(vacancy_id: @vacancy.id)&.status
+  end
+
+  def jobseeker_has_applied?
+    current_jobseeker&.job_applications&.find_by(vacancy_id: @vacancy.id)&.present?
+  end
+end
--- a/app/views/shared/vacancy/_jobseeker_view.html.slim
+++ b/app/views/shared/vacancy/_jobseeker_view.html.slim
@@ -21,8 +21,8 @@
           = render(Shared::BannerLinkComponent.new(icon_class: "apply",
                                                    link_id: "apply-job-application",
                                                    link_method: :get,
-                                                   link_path: new_jobseekers_job_job_application_path(@vacancy.id),
-                                                   link_text: t("jobseekers.job_applications.apply")))
+                                                   link_path: application_link_path,
+                                                   link_text: t("jobseekers.job_applications.banner_links.#{application_link_text}")))
 
       .govuk-grid-column-one-third
         / The "origin" param is used both for the back button on the subsequent page and for performance analysis.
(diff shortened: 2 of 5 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-2993.london.cloudapps.digital>

config/locales/jobseekers.yml
```diff
@@ -42,6 +42,10 @@ en:
       applying_for_the_role_heading: Applying for the role
       applying_for_the_role_paragraph: Please complete the online application.
       current_step: Step %{current} of %{total}
+      banner_links:
+        apply: Apply for this job
+        draft: Started - continue application
```
david-mears-dfe, line 47: These hyphens look a bit stubby to me in the rendered page. Could you try them with en dashes and em dashes and see which you (or Chris) prefer?

https://en.wikipedia.org/wiki/Dash#Common_dashes_and_their_Unicode_code-points

app/helpers/jobseekers/job_applications_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    if jobseeker_application_status.nil?
+      "apply"
+    elsif jobseeker_application_status == "submitted"
+      "submitted"
+    else
+      "draft"
+    end
+  end
+
+  def jobseeker_has_not_signed_in_or_applied?
```
david-mears-dfe, line 22: This decides whether we show the 'save job' button, right?

I _think_ we only need to check if they have applied, not whether they're signed in.

app/helpers/jobseekers/job_applications_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    current_jobseeker.nil? || current_jobseeker.job_applications.find_by({ vacancy_id: @vacancy.id }).nil?
```
david-mears-dfe, line 23: Rather than passing a hash, you can pass it as an explicit parameter:

```ruby
find_by(vacancy_id: @vacancy.id)
```

app/helpers/jobseekers/job_applications_helper.rb
david-mears-dfe: Also, might it be more readable to write this method as a positive question: `unless jobseeker_has_applied?` vs. `if jobseeker_has_not_applied`?

app/helpers/jobseekers/job_applications_helper.rb
```diff
@@ -0,0 +1,33 @@
+module Jobseekers::JobApplicationsHelper
+  def application_link_path
+    if jobseeker_application_status.nil?
+      new_jobseekers_job_job_application_path(@vacancy.id)
```
david-mears-dfe, line 4: Correct me if I'm very much mistaken, but I think Rails is clever enough that you can pass in the vacancy object itself, and it will work out the `id`.

david-mears-dfe: Same for the other paths in this method.

app/views/shared/vacancy/_jobseeker_view.html.slim
```diff
@@ -21,8 +21,8 @@
           = render(Shared::BannerLinkComponent.new(icon_class: "apply",
                                                    link_id: "apply-job-application",
                                                    link_method: :get,
-                                                   link_path: new_jobseekers_job_job_application_path(@vacancy.id),
-                                                   link_text: t("jobseekers.job_applications.apply")))
```
david-mears-dfe, line 25: Is this translation now unused? If so, can you delete it from the translation file please?

josephhull676: I believe it's still used in the vacancy details component for a govuk_button_to

app/helpers/jobseekers/job_applications_helper.rb
david-mears-dfe: I was wrong in this case - it tries to use the vacancy slug rather than the id.

PR #3020, opened 2021-03-04
[TEVA-2141] fix padding on all page layouts

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2141

this fixes 2 things

- the padding on all pages this is now not done with a controller overide but just using default containers and the only padding is applied to all page titles. basically a load of weird margin overides for controllers have been removed and this feels like right thing to do as it was too brittle
- The notification component causes complications with background colours and so this has been refactored including use of a renamed `dashboard` -> `full-width-mast` mixin that is used only for this full width with a background
- removes pointless `shared/components/notification` partial
- markup slightly altered in notifications so the dismiss link is positioned correctly (its actually teva-2145 in same PR)

### DIFF
--- a/app/frontend/src/styles/application.scss
+++ b/app/frontend/src/styles/application.scss
@@ -18,7 +18,7 @@ $govuk-image-url-function: frontend-image-url;
 
 @import 'application/accordion';
 @import 'application/banners';
-@import 'application/dashboard';
+@import 'application/full-width-mast';
 @import 'application/form';
 @import 'application/icons';
 @import 'application/modal';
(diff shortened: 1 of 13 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3020.london.cloudapps.digital>

app/views/layouts/application.html.slim
```diff
@@ -61,11 +61,12 @@ html.govuk-template.app-html-class lang="en"
 
       main.govuk-main-wrapper#main-content role="main"
         - unless flash.empty?
```
david-mears-dfe, line 63: Can we get rid of this line? If `flash` is empty, does `flash.each` error, or do nothing?

app/views/layouts/application.html.slim
```diff
@@ -61,11 +61,12 @@ html.govuk-template.app-html-class lang="en"
...
-          .govuk-main-wrapper__notification
-            .govuk-width-container
-              = render "shared/components/notification"
```
david-mears-dfe, line 66: Now that this `_notification` partial is only used by one file (`publishers/vacancies/documents/destroy.html.slim`), we can update `destroy.html.slim`, and delete the `_notification` partial.

david-mears-dfe (commented):
I've left some comments but not an approval as this is billed as a WIP.

cpjmcquillan:
Notification backgrounds (colour and size) still seems a bit dodgy

alexbowen:
i found the job alert tab background for notification was incorrect else i can see a problem @cpjmcquillan

app/frontend/src/styles/application.scss
```diff
@@ -66,21 +66,18 @@ body {
...
-
-        > .govuk-grid-row {
-          margin-top: govuk-spacing(4);
-        }
-      }
+    .content-wrapper {
+      background-color: govuk-colour('white');
+    }
+
+    h1,
```
cpjmcquillan, line 78: I think the wrapper should have a margin at the top since h1 isn't always the top element in the page

cpjmcquillan (approved):
Small comment and there are still some issues with background colour and notifications, but let's get this merged to sort out the padding

PR #3024, opened 2021-03-04
Upgrade README.md on chromedriver

david-mears-dfe:
This is because I had the following issue:

> brew update --cask chromedriver

> Error: This command updates brew itself, and does not take formula names.
> Use brew upgrade --cask chromedriver instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,6 @@ which chromedriver
 xattr -d com.apple.quarantine /path/to/chromedriver
 ```
 
-If this has no effect, you may need to try another path, for example: `/usr/local/bin/chromedriver`
-
 ### AWS credentials, MFA, and role profiles
 
 When onboarded, you will be provided with an AWS user. You can use it to access the AWS console at:
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ brew install --cask chromedriver
 
 To update
 ```bash
-brew update --cask chromedriver
+brew upgrade --cask chromedriver
 ```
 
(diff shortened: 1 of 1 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3024.london.cloudapps.digital>

README.md
```diff
@@ -69,6 +69,8 @@ which chromedriver
 xattr -d com.apple.quarantine /path/to/chromedriver
 ```
 
+If this has no effect, you may need to try another path, for example: `/usr/local/bin/chromedriver`
```
csutter, line 72: 😕  `/path/to` is just a placeholder. Not a big Homebrew fan in general, but I know I'm in a total minority amongst Mac users and we assume it elsewhere in our READMEs, so let's just replace the placeholder with `/usr/local/bin`?

david-mears-dfe: I realise that /path/to is a placeholder, but when I replaced it with the output of `which chromedriver`, there was no effect.

david-mears-dfe:
```bash
$ which chromedriver
/Users/daivdmears/.local/bin/chromedriver
```

csutter: Ooh interesting – how did you end up with two Chromedrivers! And even more interestingly, how is this one before the Homebrew one in the `PATH`, yet the Homebrew one is the one used by your tests 🧐

david-mears-dfe: Maybe by being silly. I'll undo this part of the changes to the README anyway.

csutter (approved):

PR #3028, opened 2021-03-05
[TEVA-1784] Create expiry page if submitting application too late

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-1784

## Changes in this PR:

Redirect to expiry page if applying to expired job. When attempting to submit the application form, this before action will redirect to a page explaining that the job listing has expired.

## Screenshots of UI changes:

Screenshot 2021-03-05 at 18 05 00

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -1,6 +1,5 @@
 class Jobseekers::JobApplicationsController < Jobseekers::BaseController
   before_action :redirect_if_job_application_exists, only: %i[new create]
-  before_action :redirect_if_vacancy_has_expired, only: %i[new create review submit]
 
   helper_method :job_application, :review_form, :vacancy
 
@@ -15,6 +14,8 @@ class Jobseekers::JobApplicationsController < Jobseekers::BaseController
   end
 
   def submit
+    return redirect_to expired_jobseekers_job_job_application_path(vacancy.id) unless vacancy.listed?
+
     if params[:commit] == t("buttons.save_as_draft")
       redirect_to jobseeker_root_path, notice: "Application saved as draft"
     elsif review_form.valid?
@@ -67,7 +68,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::BaseController
 
   def vacancy
     @vacancy ||= if params[:job_id].present?
-                   Vacancy.find(params[:job_id])
+                   Vacancy.live.find(params[:job_id])
                  else
                    job_application.vacancy
                  end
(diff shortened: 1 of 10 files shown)


config/locales/jobseekers.yml
```diff
@@ -58,10 +58,10 @@ en:
             When you are attending an interview, you might benefit from extra support if you're disabled have a mental
             health condition or educational needs. You can ask for the support that you need now. The school can then
             make adjustments so you can attend an interview.
-          title: Ask for support if you have a disability – Application
+          title: Ask for support if you have a disability — Application
```
david-mears-dfe, line 61: hyphen to em-dash

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3028.london.cloudapps.digital>

cpjmcquillan (changes requested):
I have quite significant reservations about this approach - notably whether a user should just be chucked out mid application. I might have agonised over this application for weeks but now - not only have I missed the deadline - I cannot see or re-use anything I've previously been doing. That's a real killer. As a user I'd expect a cleaner experience. It's highly unlikely (in my opinion) somebody completes a job application in a linear, single session. 

I think it's sensible for now to prevent a user from submitting an application that has passed the application deadline, but we shouldn't do any more than that without thinking as a team about how we can best design for this scenario (and hopefully doing some research and testing with users).

david-mears-dfe:
> I have quite significant...

Updated.

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -66,7 +68,7 @@ def review_form_params
 
   def vacancy
     @vacancy ||= if params[:job_id].present?
-                   Vacancy.live.find(params[:job_id])
+                   Vacancy.find(params[:job_id])
```
cpjmcquillan, line 71: This change would mean I can start an application for an expired job

david-mears-dfe: You're right - I forgot to get rid of this after making the expired check only happen on the final submit.

spec/system/jobseekers_can_complete_a_job_application_spec.rb
```diff
@@ -4,15 +4,17 @@
...
 
   before do
     allow(JobseekerApplicationsFeature).to receive(:enabled?).and_return(true)
     login_as(jobseeker, scope: :jobseeker)
   end
 
-  it "allows jobseekers to complete application and go to review page" do
-    visit jobseekers_job_application_build_path(job_application, :personal_details)
+  it "allows jobseekers to complete an application and go to review page" do
+    visit job_path(vacancy)
```
cpjmcquillan, line 14: We have `jobseekers_can_start_a_job_application_spec.rb`, so I'd prefer to leave this as it is unless you can compel me otherwise!

cpjmcquillan (approved):

cpjmcquillan (approved):

PR #3041, opened 2021-03-09
use GDS correctly in auto-loaded component assets

alexbowen:
recreating earlier PR as it wasnt complete and was reverted. main issue was only the shared components were being included (as thats what i was working with elsewhere) so there were no styles for stuff like this sticky element

Screenshot 2021-03-09 at 14 55 50

`component.js` does a recursive search of the components folder and adds all js and scss files to the Webpack pipeline


cpjmcquillan (commented):
Aren't we now just importing everything twice?

Also are we sure the benefits of `components.js` outweigh the negatives? (Personally I'm not)

alexbowen:
@cpjmcquillan would help if i removed the original imports (done) but no we are not importing stuff twice. the idea is that we dont have to maintain a hardcoded list of dependencies for components application js/scss which people seemed to be in favour of when i explained yesterday

app/components/README.md
```diff
@@ -17,9 +17,9 @@ The `_component` postfix is integral to how view components work and is essentia
 `= render(MyWidgetComponent.new()`
 
 ### SCSS usage
-import in styles/application.scss
+all component JS and SASS is added to the webpack pipeline using the script `frontend/src/components.js`
 
-`@import 'my_widget_component/my_widget_component';`
+For application SASS to take advantage of GovUK frontend, you should import the relevant part of the GovUK frontend library. Settings/Tools/Helpers/Base all output no CSS, so they should be explicitly imported to the component SASS and will not bloat the CSS bundle. Also components should use the GovUK mixins not variables. e.g `govuk-colour("mid-grey")` NOT `$govuk-border-colour`
```
david-mears-dfe, line 22: Why do we use mixins over variables?

cpjmcquillan:
> @cpjmcquillan would help if i removed the original imports (done) so no we are not importing stuff twice. the idea is that we dont have to maintain a hardcoded list of dependencies for components application js/scss which people seemed to be in favour of when i explained yesterday. this is also in the view component documentation (Sidecar assets). the only downside i see is that we are assuming that we use all the components we have implemented. i think that is a reasonable assumption

To me the downsides are 
- Remembering in component scss files to import govuk stuff
- Not being able to use the nice variables in component scss files (`$govuk-border-colour` is more useful to me than `govuk-colour("mid-grey")`)

The benefits are
- Don't need to add import line to `application.js`
- Don't need to add import line to `application.scss`

To me those two benefits aren't that cumbersome that's all

alexbowen:
Ok maybe I can do something about the variables. But regarding remembering to import stuff in the component rather than at app level. I believe a component (or a least the library) should be explicit about its dependencies. But maybe I can do.something here also so it is at the library (our components level) rather than application

cpjmcquillan (approved):
🚢

PR #3044, opened 2021-03-09
Rename controller to disambiguate from job applications

david-mears-dfe:
We are soon going to introduce a JobApplicationController in this folder.

As with the jobseekers controllers, let's rename the Publishers::Vacancies::ApplicationController to a BaseController, to be consistent with the jobseekers side, and in order to disambiguate from the JobApplicationController and avoid confusion when navigating the controller folders.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3044.london.cloudapps.digital>

csutter (approved):

PR #3058, opened 2021-03-11
Improve "save as draft" for job applications

csutter:
- Redirect to the "My applications" page instead of "saved jobs"
- Consistently use draft saved translation
- Consistently use success flash message


cpjmcquillan:
Can we change the root path to the applications dashboard if the feature is enabled instead?

cpjmcquillan (approved):

csutter:
> Can we change the root path to the applications dashboard if the feature is enabled instead?

Had a look and it would be a bit ugly to try and do this in the routes - given that we quite likely to want to go to the list of applications when you save as draft (regardless of what the jobseeker root path might point to in the future) are you happy for this to go in as is? And then we can change the jobseeker root path after we go live with applications.

PR #3064, opened 2021-03-11
Use Devise current_publisher method

cesidio:


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3064.london.cloudapps.digital>

app/controllers/application_controller.rb
```diff
@@ -81,7 +81,7 @@ def recaptcha_is_invalid?(model)
   end
 
   def request_event
-    RequestEvent.new(request, response, session, current_jobseeker, current_publisher_oid)
+    RequestEvent.new(request, response, session, current_jobseeker, current_publisher&.oid)
```
cpjmcquillan, line 84: Can we refactor the `RequestEvent` to take the current publisher the same way it takes the current jobseeker?

spec/components/shared/navbar_component_spec.rb
```diff
@@ -5,6 +5,7 @@
 
   before do
     allow(controller).to receive(:jobseeker_signed_in?).and_return(true)
+    allow(controller).to receive(:publisher_signed_in?).and_return(false)
```
cpjmcquillan, line 8: Why is this needed?

spec/system/hiring_staff_have_to_accept_terms_spec.rb
```diff
@@ -66,6 +67,10 @@
   end
 
   context "the user has accepted the terms and conditions" do
+    before do
+      stub_publishers_auth(urn: school.urn, oid: oid)
```
cpjmcquillan, line 71: Now we're using Devise, do we need to have this helper method at all? Could we just use devise `login_as`?

cpjmcquillan (approved):
Couple of comments that you might want to address

spec/system/hiring_staff_have_to_accept_terms_spec.rb
cesidio: we still need to associate a publisher with an organisation, that is the next refactoring that I am doing

spec/components/shared/navbar_component_spec.rb
cesidio: because otherwise we have errors with devise as it sets things that are not accessible from view components

spec/system/hiring_staff_have_to_accept_terms_spec.rb
cpjmcquillan: Yeah I just realised that a second ago on my branch 😂

PR #3065, opened 2021-03-12
sign in pages correct padding

alexbowen:
no ticket

### DIFF
--- a/app/frontend/src/styles/application.scss
+++ b/app/frontend/src/styles/application.scss
@@ -90,6 +90,16 @@ body {
     }
   }
 
+  &[class*='_sessions_new'] {
+    .govuk-main-wrapper {
+      background-color: govuk-colour('white');
+
+      .content-wrapper {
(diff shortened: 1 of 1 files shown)


cpjmcquillan (changes requested):
By default the padding/margin needs to be correct because that is the majority of pages. It should be overridden by the full width mast rather than the other way round

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3065.london.cloudapps.digital>

alexbowen:
> By default the padding/margin needs to be correct because that is the majority of pages. It should be overridden by the full width mast rather than the other way round

but this element that the full-width-mast is applied to is inside the content-wrapper so it cant overide, the overide has to be applied to content-wrapper

cpjmcquillan (approved):
Approved so the PR can be merged, but can you put the ticket back into todo since this doesn't fix the issue please?

PR #3077, opened 2021-03-15
Improve password manager experience

csutter:
c.f. https://web.dev/change-password-url/

- Add "well-known" URL for changing passwords
- Add autocomplete attributes to help password managers understand what
  fields are for


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3077.london.cloudapps.digital>

cpjmcquillan (approved):
Awesome 👍.

In the past we(I)'ve thought about using [haveibeenpwned](haveibeenpwned.com/API/v3) for flagging weak/breached passwords to users, but I think it's a bit of a tricky UX problem to get right. I've noticed more recently some browsers are doing something like that now though!

PR #3080, opened 2021-03-15
Add `rack-attack` gem and configure basic throttling

csutter:
We want to limit the amount of requests that a given user can make to
our service in any given time period.

- Make `rack-attack` use Rails's `remote_ip` (as we are behind
  CloudFront)
- Only allow 10 requests per 5 seconds from the same IP address (should
  cover us for most scenarios and allow some leeway for NAT users)
- Only allow 5 requests per minute for logging in or resetting password
  for any one jobseeker user
- Only allow 3 requests per minute for publisher fallback authentication
  for any one publisher user


cesidio (approved):

ddippolito (approved):
🚀

ddippolito:
We probably need to revisit this in the short future for API users

cpjmcquillan (approved):
So if it takes me 5 failed sign in attempts in a minute before I remember my password (how likely is this to happen?), the user gets a blank email error message. Do we want to add any other notification to the user that indicates they should try again in a minute?

PR #3085, opened 2021-03-16
[TEVA-1780] Notification component preview

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1780

These are the basic files that will work when the design system templates are introduced in seperate PR. they will be finessed a bit in that PR also

At the moment they dont do anything

### DIFF
--- a/app/components/shared/notification_component_preview.rb
+++ b/app/components/shared/notification_component_preview.rb
@@ -1,0 +1,16 @@
+class Shared::NotificationComponentPreview < ViewComponent::Preview
+  layout "preview"
+
+  attr_accessor :variants, :form
+
+  def initialize
+    @variants = [
+      { id: "success", name: "Success", title: "Success alert", description: "Sometimes dismissable." },
+      { id: "notice", name: "Notice", title: "General information alert", description: "Sometimes dismissable." },
+      { id: "danger", name: "Danger", title: "Warnings to user", description: "Never dismissable. This is a warning message or something potentially bad for the user." },
+      { id: "empty", name: "Empty", title: "Inline emphasis", description: "Basic inline emphasis." },
+    ]
+
+    @form = Shared::NotificationComponentPreview::PreviewOptionsForm
+  end
+end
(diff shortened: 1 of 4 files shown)


app/components/shared/notification_component_preview.rb
```diff
@@ -0,0 +1,16 @@
+class Shared::NotificationComponentPreview < ViewComponent::Preview
+  layout "preview"
```
david-mears-dfe, line 2: I can't easily find a mention of the 'layout' keyword in ViewComponent documentation. What does it do?

app/components/shared/notification_component_preview/preview.html.slim
```diff
@@ -0,0 +1,61 @@
...
+            a.govuk-tabs__tab href="#slim"
+              = t("heading.templates.slim")
+
+        .govuk-tabs__panel id="slim"
+          table.govuk-table
+            tbody.govuk-table__body
+              tr.govuk-table__row
+                td.govuk-table__cell
+                  pre
+                    code = "render(Shared::NotificationComponent.new(\n\t\ttitle: '#{@preview_form.title}',\n\t\tbody: '#{@preview_form.body}'\n\t},\n\tvariant: '#{@preview_form.variant}',\n\tbackground: #{@preview_form.background},\n\tdismiss: #{@preview_form.dismiss}\n))"
```
david-mears-dfe, line 30: Not sure if there is a nice way to do multiline strings in ruby in slim

david-mears-dfe (approved):

app/components/shared/notification_component_preview.rb
alexbowen: it actually something i will define in preview component config rather than at this level but its the layout template in design system all that stuff will be in different PR

app/components/shared/notification_component_preview/preview.html.slim
alexbowen: will be nicer way to do it it im sure, its a bit hardcoded for the purposes of this first preview - for the moment

PR #3087, opened 2021-03-16
[TEVA-2123] ProcessStepsComponent uses Job Application on the Application Form

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2123

## Changes in this PR:

- Added methods to retrieve the completed step number and the process status for both job applications and vacancies
- Pass job application into the component when rendered in a view related to the application form

### Before

Screenshot 2021-03-16 at 16 17 42

### After

Screenshot 2021-03-16 at 16 15 33

### DIFF
--- a/app/components/shared/process_steps_component.rb
+++ b/app/components/shared/process_steps_component.rb
@@ -6,11 +6,25 @@ class Shared::ProcessStepsComponent < ViewComponent::Base
   end
 
   def render?
-    @process.blank? || %w[create review].include?(@process.state)
+    @process.blank? || %w[create draft review].include?(process_state)
   end
 
   def current_step_number
     @service.current_step_number
+  end
+
+  def completed_step_number
+    if @process.is_a?(JobApplication)
+      @process.completed_steps.map { |step| @service.steps[step.to_sym][:number] }.max
+    else
+      @process.completed_step
+    end
+  end
+
+  def process_state
+    return @process.status if @process.is_a?(JobApplication)
+
+    @process.state
   end
 
   def steps_to_display
--- a/spec/components/shared/process_steps_component_spec.rb
+++ b/spec/components/shared/process_steps_component_spec.rb
@@ -1,100 +1,155 @@
 require "rails_helper"
 
 RSpec.describe Shared::ProcessStepsComponent, type: :component do
-  let(:vacancy) { create(:vacancy, completed_step: completed_step) }
-  let(:completed_step) { 0 }
-  let(:current_step_number) { 1 }
-  let(:current_organisation) { build(:school_group) }
   let(:process_title) { "Process title" }
-  let(:steps) do
-    {
-      job_location: { number: 1, title: I18n.t("jobs.job_location") },
-      schools: { number: 1, title: I18n.t("jobs.job_location") },
-      job_details: { number: 2, title: I18n.t("jobs.job_details") },
-      pay_package: { number: 3, title: I18n.t("jobs.pay_package") },
-      important_dates: { number: 4, title: I18n.t("jobs.important_dates") },
-      supporting_documents: { number: 5, title: I18n.t("jobs.supporting_documents") },
-      documents: { number: 5, title: I18n.t("jobs.supporting_documents") },
-      applying_for_the_job: { number: 6, title: I18n.t("jobs.applying_for_the_job") },
-      job_summary: { number: 7, title: I18n.t("jobs.job_summary") },
-      review: { number: 8, title: I18n.t("jobs.review_heading") },
-    }.freeze
-  end
 
-  let(:steps_adjust) { current_organisation.is_a?(SchoolGroup) ? 0 : 1 }
+  context "when the component is built for a vacancy" do
+    let(:completed_step) { 0 }
+    let(:current_step_number) { 1 }
+    let(:vacancy) { create(:vacancy, completed_step: completed_step) }
+    let(:current_organisation) { build(:school_group) }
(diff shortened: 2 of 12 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3087.london.cloudapps.digital>

app/components/shared/process_steps_component.rb
```diff
@@ -6,13 +6,27 @@ def initialize(process:, service:, title:)
...
+  def completed_step_number
+    if @process.is_a?(JobApplication)
+      @process.completed_steps.map { |step| @service.steps[step.to_sym][:number] }.max
+    else
+      @process.completed_step
+    end
+  end
+
+  def process_state
+    return @process.status if @process.is_a?(JobApplication)
```
david-mears-dfe, line 25: This component used to be tied too closely to Vacancy, so that it only worked when used with Vacancies.

With the changes in this file, it becomes tied to both Vacancy and JobApplication. This is a step in the wrong direction: it's become less abstract.

Ideally, it should know almost nothing about the implementation of the things it's consuming. Imagine that we are planning to share this ProcessSteps component across government so that any website can use it, without requiring Vacancy or JobApplication - because that is in fact what we hope to do to our best components!

Ideally, the component would be 'abstract' in that you can plug anything in (a vacancy, a job application, a petition form, an ice cream order form) and, as long as the thing you plug in is plug-shaped (i.e. is able to answer to a few particular methods specified by the component), the component can render the steps.

For example, we could say that anything that gets plugged in to the socket of ProcessStepsComponent should be able to answer to just one out of `status` or `state`. An ice cream order form could have a state, and then we'd be able to plug that in, without writing a series of lines like

```ruby
return @process.stage if @process.is_a?(IceCream)
return @process.point_we_got_up_to if @process.is_a?(PersonalProtectiveEquipmentOrder)
return @process.step if @process.is_a?(HowToPaintCourse)
```

Probably we want to change JobApplication's `status` column to be called `state` to match Vacancy (Vacancy already has one of each).

spec/components/shared/process_steps_component_spec.rb
```diff
@@ -1,100 +1,155 @@
...
+
+    it "renders the supporting documents step" do
+      expect(rendered_component).to include(I18n.t(".personal_statement.heading"))
+    end
+
+    it "renders the application details step" do
+      expect(rendered_component).to include(I18n.t(".references.heading"))
+    end
+
+    it "renders the job summary step" do
```
david-mears-dfe, line 143: equal opportunities

spec/components/shared/process_steps_component_spec.rb
```diff
@@ -1,100 +1,155 @@
...
+      expect(rendered_component).to include(I18n.t(".equal_opportunities.heading"))
+    end
+
+    it "renders the job summary step" do
```
david-mears-dfe, line 147: ask for support

spec/components/shared/process_steps_component_spec.rb
```diff
@@ -1,100 +1,155 @@
...
+        references: { number: 6, title: I18n.t(".references.heading") },
+        equal_opportunities: { number: 7, title: I18n.t(".equal_opportunities.heading") },
+        ask_for_support: { number: 8, title: I18n.t(".ask_for_support.heading") },
+        declarations: { number: 9, title: I18n.t(".declarations.heading") },
+      }.freeze
+    end
+
+    let!(:inline_component) { render_inline(described_class.new(process: job_application, service: ProcessSteps.new({ steps: steps, adjust: 0, step: :personal_details }), title: process_title)) }
+
+    it "renders the job details step" do
```
david-mears-dfe, line 123: These description strings need to be updated, e.g. this is personal details.

app/components/shared/process_steps_component.rb
david-mears-dfe: Another example: you could make ProcessStepsComponent more abstract by moving this logic into a method that lives on JobApplication. Then ProcessStepsComponent just needs to call that method on the `@process`.

```ruby
if @process.is_a?(JobApplication)
   @process.completed_steps.map { |step| @service.steps[step.to_sym][:number] }.max
```

app/components/shared/process_steps_component.rb
```diff
@@ -6,13 +6,27 @@ def initialize(process:, service:, title:)
   end
 
   def render?
-    @process.blank? || %w[create review].include?(@process.state)
+    @process.blank? || %w[create draft review].include?(process_state)
```
david-mears-dfe, line 9: I think the abstract way to do this would be to make the process classes (Vacancy and JobApplication) responsible for knowing what states are renderable: it'll be a different list for every possible process. E.g.:

```ruby
class IceCreamOrder
  def states_where_the_process_steps_component_should_render
    %w[in_queue choosing_toppings adding_sprinkles]
  end
end
```

app/components/shared/process_steps_component.rb
cpjmcquillan: I'm was happy to let this fly at the minute since I was going to refactor it soon anyway - so I will approve but feel free to request changes if you want this refactoring to happen right away!

david-mears-dfe: I don't mind who does it, but perhaps it would be an interesting job for @josephhull676?

cpjmcquillan: Maybe we can pair on it @josephhull676?

PR #3093, opened 2021-03-17
Add environment banner to non-production apps

csutter:
This will help us distinguish between our myriad non-production apps at
a glance, and make it clear to internal users that they are not on
production for peace of mind when performing destructive actions.

- Add `APP_ROLE` environment variable to distinguish what we call an
  "environment" for operational purposes from Rails environment (which
  is always set to `production` outside of local development)
- Add `Rails.configuration.app_role` to allow for easy checking of
  current app role
- Set `APP_ROLE` for all existing environments
- Create a spiffy environment banner that is displayed when the app
  role is a) set and b) not "production"
- Include environment banner in global application layout

Screenshot 2021-03-17 at 11 37 49


app/components/shared/environment_banner_component.rb
```diff
@@ -0,0 +1,8 @@
+class Shared::EnvironmentBannerComponent < ViewComponent::Base
```
cpjmcquillan, line 1: How do you feel about extending `GovukComponent::Base`?

cpjmcquillan (approved):
💥  do you want to add a spec?

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3093.london.cloudapps.digital>

app/components/shared/environment_banner_component.rb
csutter: Sure! This was just copypasta'ed from another component 😅

spec/components/shared/environment_banner_component_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    context "when app_role is production" do
+      let(:app_role) { "production" }
+
+      it "does not render" do
+        expect(subject.render?).to eq(false)
+      end
+    end
+
+    context "when app_role is unknown" do
```
cpjmcquillan, line 17: if we do a `render_inline` somewhere then we can do things like `expect(subject).to be_rendered` vice versa, but I'm happy enough if you just want to test the method itself instead

PR #3099, opened 2021-03-17
Refactor govuk-main-wrapper usage to ensure all pages have correct padding

cpjmcquillan:
This fixes the padding issue. 

It removes notification wrapper backgrounds. I think this is the lesser of two evils and doesn't look particularly bad anyway.

If we really want to we can try and address that a later date, but given we've got lots of notification work/refactoring on the horizon, I'm not sure if it is worth it at this point in time anyway.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3099.london.cloudapps.digital>

cesidio (approved):

PR #3102, opened 2021-03-17
Display the right number of vacancies awaiting feedback.

david-mears-dfe:
For school groups, 'organisation.vacancies' will only retrieve vacancies with a location of 'central_office', rather than all vacancies at the organisation.

Previously, the number of jobs displayed in the 'awaiting feedback' tab and in the flash notice about giving feedback were different.

This uses a controller spec because that is the easiest (and probably best) way to test an instance variable assignment.

copied from Slack:
> I've used a controller spec to test that the instance variable is assigned to what it should be assigned to. I prefer that to request specs because I don't think request specs can access instance variable assignments (except by testing the response body, which is effectively equivalent to a system spec at that point). Alternatively, if your hatred for controller specs is sufficiently high, I can write a vacancies_awaiting_feedback_count method on Organisation, and test that.

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -47,7 +47,7 @@ class Publishers::OmniauthCallbacksController < Devise::OmniauthCallbacksControl
   end
 
   def organisation_from_request
-    # https://github.com/david-mears-dfe/login.dfe.public-api/blob/patch-1/README.md#organisation-categories
+    # https://github.com/DFE-Digital/login.dfe.public-api#how-do-ids-map-to-categories-and-types
     case auth_hash.dig("extra", "raw_info", "organisation", "category", "id")
     when "001"
       School.find_by!(urn: auth_hash.dig("extra", "raw_info", "organisation", "urn"))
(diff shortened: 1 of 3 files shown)


david-mears-dfe:
I'm going to redo the test as a request spec.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3102.london.cloudapps.digital>

app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -47,7 +47,7 @@ def id_token
   end
 
   def organisation_from_request
-    # https://github.com/david-mears-dfe/login.dfe.public-api/blob/patch-1/README.md#organisation-categories
+    # https://github.com/DFE-Digital/login.dfe.public-api#how-do-ids-map-to-categories-and-types
```
david-mears-dfe, line 50: (My forked change was accepted into their README)

cesidio (changes requested):
we should not create anymore controller specs and have request specs instead

david-mears-dfe:
> we should not create anymore controller specs and have request specs instead

I don't think request specs can check variable assignment, as I said in the description. So I will create a method on vacancies_awaiting_feedback_count method on Organisation, and test that.

cesidio:
Having had a better think about the problem here, and having discussed about it with Christian, I think that the best solution  is not to test this specific thing as we have system specs that cover the happy path. I would just have the minimum change that fix the problem.

cesidio (approved):

PR #3105, opened 2021-03-18
Report and show error for OmniAuth failures

csutter:
We currently just silently swallow OmniAuth failures and redirect the
user to the homepage - we need to surface the errors and show a
meaningful message to publishers.

- Ship OmniAuth errors to Rollbar
- Redirect to login page on failure (instead of homepage)
- Show an error message to the user on failure

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -17,7 +17,11 @@ class Publishers::OmniauthCallbacksController < Devise::OmniauthCallbacksControl
   end
 
   def failure
-    redirect_to root_path
+    omniauth_error = request.respond_to?(:get_header) ? request.get_header("omniauth.error") : request.env["omniauth.error"]
+    Rollbar.error(omniauth_error, strategy: failed_strategy.name)
+    Rails.logger.error("DSI failure - strategy: #{failed_strategy.name}, reason: #{omniauth_error.inspect}")
+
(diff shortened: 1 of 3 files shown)


app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -17,7 +17,11 @@ def dfe
   end
 
   def failure
-    redirect_to root_path
+    omniauth_error = request.respond_to?(:get_header) ? request.get_header("omniauth.error") : request.env["omniauth.error"]
+    Rollbar.error(omniauth_error, strategy: failed_strategy.name)
+    Rails.logger.error("DSI failure - strategy: #{failed_strategy.name}, reason: #{omniauth_error.inspect}")
+
+    redirect_to new_publisher_session_path, flash: { warning: t(".message") }
```
cpjmcquillan, line 24: When you redirect like this does it mean you can get a warning flash without having it in `add_flash_types`?

cpjmcquillan (approved):

app/controllers/publishers/omniauth_callbacks_controller.rb
csutter: It's in `add_flash_types` now that Alex has added it (as of a few minutes ago).

The problem is that `redirect_to` will only accept either `alert` or `notice`, not any of the custom types we have defined, so we need to nest it under the `flash` key. 😿

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3105.london.cloudapps.digital>

PR #3106, opened 2021-03-18
[TEVA-1780] Design system

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-1780

Changes in this PR:
- this introduces design system controller/templates and partials
- notification preview tweaks to align with above

go to `/components/shared/previews/notification_component/preview` to see whats below

Screenshot 2021-03-20 at 01 47 54


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3106.london.cloudapps.digital>

alexbowen:
spoke to christian and he agreed we can tackle rspec tests seperatly for previews

david-mears-dfe:
It seems that the code says 'icon: false' even though the components have icons. Is there any way to ensure that the code displayed is always the same as the code that is being rendered into a component?

Screenshot 2021-03-22 at 13 52 40

david-mears-dfe:
Pipe dream: allow me to edit the code live on the page and see the result!

david-mears-dfe:
I clicked the 'icon' checkmark. Is `icon: ["true"]` really the correct syntax rather than `icon: true`?
Screenshot 2021-03-22 at 13 56 18

david-mears-dfe:
> Is there any way to ensure that the code displayed is always the same as the code that is being rendered into a component?

We could use this: https://github.com/banister/method_source

alexbowen:
@david-mears-dfe some variants the icon is always present (as per requirements) so the code is actually correct. maybe the option should not be in the code snippet at all in these cases. some varaints are 'never dissmissible' so the same kind of thing applies. regarding the ['true'] thing yes need to fix that

alexbowen:
> > Is there any way to ensure that the code displayed is always the same as the code that is being rendered into a component?
> 
> We could use this: https://github.com/banister/method_source

nice. can investigate this for sure...

david-mears-dfe:
> > We could use this: https://github.com/banister/method_source
> 
> nice. can investigate this for sure...

Probably also worth looking at how GDS formbuilder do it. In this PR I only had to change one line and it changed both the rendered output and the displayed code. The maintainer `@peteryates` might be able to give us pointers.

david-mears-dfe:
Is it legal to have red/green links? I wonder how that plays with links being marked as visited/focused.

app/components/shared/previews/notification_component_preview.rb
```diff
@@ -0,0 +1,21 @@
+class Shared::Previews::NotificationComponentPreview < ViewComponent::Preview
+  layout "design_system"
```
david-mears-dfe, line 2: I still wonder what this line does and how!

app/components/shared/previews/notification_component_preview.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def self.component_name
+    "notification"
```
david-mears-dfe, line 5: Alternatively,

```ruby
compent_class.to_s.underscore.humanize.split("/").second.split(" ").first
# or if we get rid of the 'shared' folder:
compent_class.to_s.underscore.humanize.split(" ").first
```

david-mears-dfe: But that's probably premature optimization

app/components/shared/previews/notification_component_preview/options_form.rb
```diff
@@ -0,0 +1,17 @@
...
+  attr_reader :background, :dismiss, :icon
+
+  def initialize(params = {})
+    preview_criteria = if params[:shared_previews_notification_component_preview_options_form].present?
+                         JSON.parse(params[:shared_previews_notification_component_preview_options_form].to_json).symbolize_keys
+                       else
+                         {}
+                       end
+
+    @icon = preview_criteria[:icon] ? preview_criteria[:icon].first : false
```
david-mears-dfe, line 13: It might be preferable to use `.present?` - more readable for Ruby devs

app/components/shared/previews/notification_component_preview/preview.html.slim
```diff
@@ -0,0 +1,17 @@
+= form_for @preview_form, url: preview_view_component_path("#{@preview.preview_name}/#{@example_name}"), html: { method: "get" } do |f|
```
david-mears-dfe, line 1: Does this `f` get used before the last line of this file? If not we can move this line down there.

app/components/shared/previews/notification_component_preview.rb
david-mears-dfe: ("component" is gsubbed in `app/views/design_system/_nav_side.html.slim` but could equally well do so here)

david-mears-dfe (approved):

alexbowen:
> Is it legal to have red/green links? I wonder how that plays with links being marked as visited/focused.

very good point

app/components/shared/previews/notification_component_preview.rb
alexbowen: no i think its a simple change that removes some hardcodedness, and thats usually a good thing

PR #3110, opened 2021-03-19
Restrict flash types shown in application layout

csutter:
Devise adds a `timedout` flash key to the flash when a user's session
times out (in addition to a friendly message). We don't want to show
this (or in fact any other random thing in the flash), so restrict the
keys shown.

Unfortunately the Rails flash hash isn't a proper hash (so we can't do
a simple `Hash#slice`) and is keyed with strings, so this isn't the
prettiest way of doing it but it does the job.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2218


cpjmcquillan (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3110.london.cloudapps.digital>

cesidio (approved):

PR #3116, opened 2021-03-19
[TEVA-2217] remove mini-css-extract-plugin

alexbowen:
removing this solves the issue of why the cache busting string had gone missing on application css bundle. Why? i do not know. But i have a feeling its a DXW hangover why its there.

but removing this plugin makes no difference to bundle sizes or dev experience as far as i can tell. i do want to check the source maps on the preview env this creates however


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3116.london.cloudapps.digital>

cesidio (approved):

PR #3120, opened 2021-03-22
Prevent jobseekers submitting applications for expired jobs

cpjmcquillan:
https://ukgovernmentdfe.slack.com/archives/C01BB5D318W/p1616168038044800

## Changes in this PR
Rather than allowing a jobseeker to submit an application for an expired job, we prevent them doing this and render some useful and informative (pending) content. 

Ideally we will notify users (e.g emails, notifications, notification_banners) X days before the job expires, but designs for these scenarios are pending and will be outlined in tickets TBC.


app/views/jobseekers/job_applications/review.html.slim
```diff
@@ -4,6 +4,11 @@
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
+    / TODO: Confirm content
+    - unless vacancy.listed?
+      = govuk_notification_banner title: "Deadline passed" do
+        p.govuk-body The deadline for this job listing has passed.
```
alexbowen, line 10: in locale.yml? i guess you might be waiting to confirm copy as comment says

spec/requests/jobseekers/job_applications_spec.rb
```diff
@@ -35,6 +56,81 @@
...
+      it "raises an error and does not submit the job application or send email" do
+        assert_emails 0 do
+          expect { post jobseekers_job_application_submit_path(job_application.id), params: params }
+            .to not_change { job_application.reload.status }
+            .and raise_error(ActionController::RoutingError, /non-draft/)
+        end
+      end
+    end
+
+    context "when the commit is `Save as draft`" do
```
alexbowen, line 96: personally i find this wording a bit confusing as to what a 'commit' is and hence what this describes the test is supposed to do

alexbowen (approved):
pending the copy being finalised this LGTM

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3120.london.cloudapps.digital>

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -13,7 +13,8 @@ def create
   end
 
   def submit
-    return redirect_to expired_jobseekers_job_job_application_path(vacancy.id) unless vacancy.listed?
+    raise ActionController::RoutingError, "Cannot submit application for non-listed job" unless vacancy.listed?
```
david-mears-dfe, line 16: This line won't be run when a vacancy is not listed, because the previous line returns.

david-mears-dfe (approved):
One confusion, otherwise approved.

spec/requests/jobseekers/job_applications_spec.rb
cpjmcquillan: I added the word `param` which should clarify things

app/controllers/jobseekers/job_applications_controller.rb
cpjmcquillan: Yeah you've lost me here 😂

david-mears-dfe: I didn't notice that the return line was removed, not added!

david-mears-dfe: Need to adjust contrast on my monitor...

PR #3121, opened 2021-03-22
Make the timeline component care less about dates

cpjmcquillan:


app/components/timeline_component/timeline_component.scss
```diff
@@ -42,7 +42,7 @@
         width: 5px;
       }
 
-      &__key {
+      .timeline-component__key {
```
alexbowen, line 45: why remove the &__key notation?

app/components/timeline_component/timeline_component.scss
```diff
@@ -57,7 +57,7 @@
         }
       }
 
-      &__value {
+      .timeline-component__value {
```
alexbowen, line 60: and here

app/components/timeline_component/timeline_component.scss
alexbowen: & inside sass modules/components or whatever makes the code more flexible if the wrapper naming were to change

cpjmcquillan: because it wouldn't make sense anymore. The class isn't `timeline-component__item__key`

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3121.london.cloudapps.digital>

app/components/timeline_component/timeline_component.scss
alexbowen: the class and/or structure of this sass is incorrect then, but willing to let it go as that happened somewhere along the way before this PR

alexbowen (approved):
nice to make this a bit more abstract

PR #3131, opened 2021-03-23
[TEVA-2141] Add validation to LA publisher preferences

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2141

## Screenshots of UI changes:
Screenshot 2021-03-23 at 11 22 24

### DIFF
--- a/app/controllers/publishers/publisher_preferences_controller.rb
+++ b/app/controllers/publishers/publisher_preferences_controller.rb
@@ -1,12 +1,21 @@
 class Publishers::PublisherPreferencesController < Publishers::BaseController
+  before_action :strip_empty_publisher_preference_checkboxes, only: %i[create update]
+
   def new
     @publisher_preference = PublisherPreference.new
   end
 
   def create
-    publisher_preference = PublisherPreference.create(publisher: current_publisher, organisation: current_organisation)
-    set_local_authority_publisher_schools(publisher_preference)
-    redirect_to organisation_path
+    @publisher_preference = PublisherPreference.new(publisher: current_publisher, organisation: current_organisation)
+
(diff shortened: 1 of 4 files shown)


config/locales/publishers.yml
```diff
@@ -94,6 +94,7 @@ en:
         page_title: Set up your Teaching Vacancies account
       form:
         schools_in: Schools in %{organisation} local authority
```
david-mears-dfe, line 96: Wrong PR, but: would it be simpler to update the school import task so that local authorities have 'local authority' added in the name attribute? That way we don't have to duplicate translations or adapt translations depending on what type of organisation they are being used for. E.g. will 'publisher preferences' always be used by local authorities?

cesidio: I wouldn't alter what we get from GIAS

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3131.london.cloudapps.digital>

app/controllers/publishers/publisher_preferences_controller.rb
```diff
@@ -1,44 +1,58 @@
...
     end
     redirect_to jobs_with_type_organisation_path(params[:publisher_preference][:jobs_type])
   end
 
   private
 
-  def set_local_authority_publisher_schools(publisher_preference)
-    strip_empty_checkboxes(%i[school_ids], :publisher_preference)
-    publisher_preference.local_authority_publisher_schools.delete_all
+  def set_local_authority_publisher_schools
```
cpjmcquillan, line 41: maybe these methods could be called `create_...`

cpjmcquillan (approved):

PR #3133, opened 2021-03-23
Limit host Faker URLs to example.com

csutter:
These sometimes generate URLs with existing spammy websites which isn't
a good look (especially as we use them to generate fake data for review
apps). This way we end up with URLs that are guaranteed to not exist.


cpjmcquillan (approved):

cesidio (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3133.london.cloudapps.digital>

PR #3134, opened 2021-03-23
[TEVA-2001] Redirect NQT Job Alert

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2001

## Changes in this PR:

- sign-up-for-NQT-job-alerts now routes to subscriptions#new
- If the user reaches the create a job alert form via /sign-up-for-NQT-job-alerts the old copy from the nqt job alert is displayed and the "Suitable for NQTs" checkbox is checked. 
- session[:subscription_origin] set to "/sign-up-for-NQT-job-alerts" if the user reaches the form via /sign-up-for-NQT-job-alerts
- Removed nqt job alerts

## Comments

- No idea if I've gone about this the right way but it seems to work
- It'd probably be helpful if the job role accordion section was expanded when "Suitable for NQTs" is pre-checked. Not sure how I could achieve this though...


config/routes.rb
```diff
@@ -103,8 +103,8 @@
     resources :unsubscribe_feedbacks, only: %i[new create], controller: "jobseekers/unsubscribe_feedbacks"
   end
 
-  get "sign-up-for-NQT-job-alerts", to: "nqt_job_alerts#new", as: "nqt_job_alerts"
-  post "sign-up-for-NQT-job-alerts", to: "nqt_job_alerts#create", as: "new_nqt_job_alert"
+  get "sign-up-for-NQT-job-alerts", to: "subscriptions#new", as: "nqt_job_alerts", defaults: { origin: "/sign-up-for-NQT-job-alerts" }
```
cpjmcquillan, line 106: Rather than adding the origin param can we add `nqt_suitable` to the job role parameters?

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3134.london.cloudapps.digital>

config/routes.rb
josephhull676: Not too sure what you mean. What job role parameters?

cpjmcquillan: The job role part of the search criteria parameter

spec/system/jobseekers_can_subscribe_to_an_nqt_job_alert_spec.rb
```diff
@@ -16,10 +16,10 @@
       expect(message_delivery).to receive(:deliver_later)
```
david-mears-dfe, line 16: You didn't write line 14, but it should probably be a `let`.
```ruby
message_delivery = instance_double(ActionMailer::MessageDelivery)
```

app/controllers/subscriptions_controller.rb
```diff
@@ -102,7 +102,12 @@ def email
     params.permit(:email)
   end
 
+  def nqt_job_alert?
+    request.original_fullpath == "/sign-up-for-NQT-job-alerts"
```
david-mears-dfe, line 106: Is it possible to get this string programmatically, e.g. is it `nqt_job_alerts_url`?

david-mears-dfe: Not sure this logic deserves to be its own method - could live inside `origin_param` method as it's simple and not reused.

app/views/shared/search/_keyword.html.slim
```diff
@@ -1,4 +1,11 @@
-= f.govuk_text_field :keyword,
-  label: { text: t("jobs.filters.keyword"), size: "s" },
-  placeholder: t("jobs.filters.keyword_hint"),
-  width: filters ? "full" : "two-thirds"
+- if @origin == "/sign-up-for-NQT-job-alerts"
```
david-mears-dfe, line 1: This string is duplicated in several places so I'd like to find a way to get it programmatically.

david-mears-dfe: Is it possible to use fewer lines here by putting the `if` inside the hint text? Not sure what happens if you pass `nil` or `""` to `hint: { text: }`.

```ruby
hint: { text: (@origin == "/sign-up-for-NQT-job-alerts" ? t("jobs.filters.keyword_nqt_hint") : nil },
```

app/controllers/subscriptions_controller.rb
david-mears-dfe: Alternatively you can make this method a `helper_method` so it's available in the view, and then that prevents the string duplication on line 1 of `app/views/shared/search/_keyword.html.slim`.

josephhull676: So I'd create a file called `subscriptions_helper.rb` then include that file int the controller so I can use it?

josephhull676: If you have a look at the vacancies controller on line 2 it declare a controller method (job application) as a helper method. 

To avoid creating a new file for only one method. I'll copy this in the subscriptions controller and then I can use nqt_job_alert? in the view too. Is that ok with you?

david-mears-dfe: Yes - that's the smoothest way to create a helper_method for a specific controller:

```ruby
class ControllerName
  helper_method :my_method
  
  def my_method
    "Jim's the name, being Jim's the game"
  end
end
```

david-mears-dfe (approved):

spec/system/jobseekers_can_subscribe_to_an_nqt_job_alert_spec.rb
```diff
@@ -6,20 +6,21 @@
   end
 
   describe "A jobseeker" do
```
cpjmcquillan, line 8: I think this whole system spec file should be deleted now

app/controllers/subscriptions_controller.rb
```diff
@@ -103,6 +103,7 @@ def email
   end
 
   def origin_param
+    params[:origin] = nqt_job_alerts_path if helpers.nqt_job_alert?
```
cpjmcquillan, line 106: I'm a bit confused why we're doing this?

csutter (changes requested):
Just had a look at this and I'd rather we didn't test for URLs (even with the path helpers now).

Maybe a nicer way – given you already add a parameter to `defaults` in the routes – would be to add another like so:
```
get "sign-up-for-NQT-job-alerts", to: "subscriptions#new", as: "nqt_job_alerts", defaults: { search_criteria: { job_roles: ["nqt_suitable"] }, nqt_job_alert: true }
```

And then add something like `@nqt_job_alert = params[:nqt_job_alert]` to `SubscriptionsController#new` and test for `@nqt_job_alert` instead of having a `nqt_job_alert?` helper? Open for better naming ideas too 😃

spec/system/jobseekers_can_subscribe_to_an_nqt_job_alert_spec.rb
josephhull676: No need to test this journey at all?

app/views/shared/search/_keyword.html.slim
```diff
@@ -1,4 +1,5 @@
 = f.govuk_text_field :keyword,
   label: { text: t("jobs.filters.keyword"), size: "s" },
-  placeholder: t("jobs.filters.keyword_hint"),
+  hint: { text: (@nqt_job_alert == true ? t("jobs.filters.keyword_nqt_hint") : nil) },
```
csutter, line 3: You don't need `== true`, just `@nqt_job_alert ? ___ : ___` is enough to test if it's truthy 💡

csutter (approved):
Great stuff - much simpler and DRYer now 🙌

spec/system/jobseekers_can_subscribe_to_an_nqt_job_alert_spec.rb
cpjmcquillan: As long as there is a request spec we're all good I think.

This journey is the same as creating a job alert from a search - which already exists

spec/requests/subscriptions_spec.rb
```diff
@@ -20,6 +20,16 @@
       end
     end
 
+    context "when hit via the nqt job alerts url" do
```
cpjmcquillan, line 23: I'd like to see a request spec that tests the redirect

PR #3135, opened 2021-03-23
Always say 'local authority' after an LA's name

david-mears-dfe:
This is the lesser evil for now, until we do the refactoring that properly separates LAs from Trusts and removes the logic that
checks for the group_type.


cesidio (approved):

cpjmcquillan (approved):

david-mears-dfe:
https://teaching-vacancies-review-pr-3135.london.cloudapps.digital/

PR #3136, opened 2021-03-23
Ignore most common CSP violations

csutter:
Most of the CSP violations that currently come through on Rollbar fall
into one of the following categories:
- Misbehaving browser plugins
- Facebook mobile app injecting JS into in-app browser

We don't want to drown in CSP violation errors on Rollbar, so start
ignoring these (and possibly others in the future).

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -33,7 +33,9 @@ class ErrorsController < ApplicationController
   end
 
   def csp_violation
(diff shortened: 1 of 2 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3136.london.cloudapps.digital>

cpjmcquillan (approved):

PR #3139, opened 2021-03-23
Bugfix for publisher viewing job listing from dashboard with applications feature enabled

david-mears-dfe:
Previously, there was an error because `job_application` is undefined in the Publishers::OrganisationsController.

### DIFF
--- a/app/views/shared/vacancy/_jobseeker_view.html.slim
+++ b/app/views/shared/vacancy/_jobseeker_view.html.slim
@@ -17,15 +17,15 @@
     .govuk-grid-row
       - if JobseekerApplicationsFeature.enabled? && @vacancy.apply_through_teaching_vacancies == "yes" && @vacancy.expires_at.future?
         .govuk-grid-column-one-third
-          - case job_application&.status
-          - when "submitted"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.submitted"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
-          - when "draft"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.draft"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
+          - if defined?(job_application) && ["withdrawn", nil].exclude?(job_application&.status)
+            - if job_application.status == "draft"
+              = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.draft"),
+                href: jobseekers_job_application_path(job_application),
+                icon: "green-tick"
+            - else
+              = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.submitted"),
+                href: jobseekers_job_application_path(job_application),
+                icon: "green-tick"
           - else
             = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.apply"),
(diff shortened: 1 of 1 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3139.london.cloudapps.digital>

app/views/shared/vacancy/_jobseeker_view.html.slim
```diff
@@ -17,13 +17,8 @@
...
         .govuk-grid-column-one-third
-          - case job_application&.status
-          - when "submitted"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.submitted"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
-          - when "draft"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.draft"),
+          - if defined?(job_application) && job_application.present?
+            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.#{job_application.status}"),
```
cpjmcquillan, line 21: A job application can be more statuses than `submitted` and `draft` so we should think about what we say for each. I reckon that we probably want to distinguish between `draft` and `not_draft` rather than being that granular though

cpjmcquillan:
Maybe we should display different banner buttons when a publisher is signed in? The actions aren't particularly valid in that scenario

app/views/shared/vacancy/_jobseeker_view.html.slim
```diff
@@ -17,15 +17,15 @@
...
-          - when "submitted"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.submitted"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
-          - when "draft"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.draft"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
+          - if defined?(job_application) && ["withdrawn", nil].exclude?(job_application&.status)
+            - if job_application.status == "draft"
```
cpjmcquillan, line 21: If `job_application` is defined, then we can just do `if job_application.draft?`

app/views/shared/vacancy/_jobseeker_view.html.slim
```diff
@@ -17,15 +17,15 @@
...
-          - case job_application&.status
-          - when "submitted"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.submitted"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
-          - when "draft"
-            = render BannerButtonComponent.new text: t("jobseekers.job_applications.banner_links.draft"),
-              href: jobseekers_job_application_path(job_application),
-              icon: "green-tick"
+          - if defined?(job_application) && ["withdrawn", nil].exclude?(job_application&.status)
```
cpjmcquillan, line 20: If `job_application` is defined can the status ever be `nil`? If it can't can we just do `&& !job_application.withdrawn?`?

cpjmcquillan (approved):

app/views/shared/vacancy/_jobseeker_view.html.slim
david-mears-dfe: It can be `nil`, yeah: when they have not started an application

PR #3152, opened 2021-03-24
Jobseekers cannot view draft applications

cpjmcquillan:
Previously they could (misplaced link or via the path), but they shouldn't (it doesn't make sense/UX).

So now (even by path), they can't (raises error).


david-mears-dfe (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3152.london.cloudapps.digital>

PR #3156, opened 2021-03-24
Jobseekers can withdraw an application

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1953

## Changes in this PR
- Jobseekers can withdraw submitted or shortlisted job applications

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -1,7 +1,7 @@
 class Jobseekers::JobApplicationsController < Jobseekers::BaseController
   before_action :redirect_if_job_application_exists, only: %i[new create]
 
-  helper_method :job_application, :review_form, :vacancy
+  helper_method :job_application, :review_form, :vacancy, :withdraw_form
 
   def new
     request_event.trigger(:vacancy_apply_clicked, vacancy_id: vacancy.id)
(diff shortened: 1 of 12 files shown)


app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -43,6 +43,27 @@ def destroy
...
+    raise ActionController::RoutingError, "Cannot withdraw non-submitted or non-shortlisted application" unless
+      job_application.status.in?(%w[shortlisted submitted])
+  end
+
+  def withdraw
+    raise ActionController::RoutingError, "Cannot withdraw non-submitted or non-shortlisted application" unless
+      job_application.status.in?(%w[shortlisted submitted])
+
+    if params[:commit] == t("buttons.cancel")
+      return redirect_to jobseekers_job_applications_path if URI(params[:origin]).path == jobseekers_job_applications_path
```
alexbowen, line 56: there is a line of code after the return?

cpjmcquillan: there's an
```ruby
if
```
after it

spec/system/jobseekers_can_withdraw_a_job_application_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  it "submits the form, renders error, then withdraws the application" do
+    click_on I18n.t("buttons.withdraw_application")
+
+    expect(page).to have_content("There is a problem")
+
+    choose "jobseekers-job-application-withdraw-form-withdraw-reason-other-field"
+
+    expect { click_on I18n.t("buttons.withdraw_application") }
```
cesidio, line 22: Not entirely sure, but if we test record creation in request specs, we can probably test only the interface in system spec, things that are on the screen.

cesidio (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3156.london.cloudapps.digital>

spec/system/jobseekers_can_withdraw_a_job_application_spec.rb
cpjmcquillan: Yeah I'm with you on that actually. 

I'd like to keep system specs as clean and simple as possible across the whole application 👍

PR #3157, opened 2021-03-24
Update tests for jobseeker view of job show page

david-mears-dfe:
Previously, the spec only tested the following scenario:
1. Load the job show page
2. Become logged in
3. Click a button etc etc

We also need to test the more common scenario:
1. Be logged in already
2. Load the job show page when logged in
3. Click a button etc etc

This lets us test, among other things, that the correct CTA is displayed on the banner button, depending on the job application status (draft etc).

We should continue to test the 'load page first' scenario, because we want to display specific flash messages in these cases.


david-mears-dfe:
https://teaching-vacancies-review-pr-3157.london.cloudapps.digital/

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3157.london.cloudapps.digital>

alexbowen (approved):

PR #3166, opened 2021-03-26
Only use Rails subgems we actually need

csutter:
This came out of a subgem of Rails (`activestorage`) using a dependency
that was yanked, blocking us from bundling our dependencies. It should
make our app leaner and reduce the surface area of these sorts of things
happening a tiny bit.

We still need to do something about `govuk-date-fields` because that is
still pulling in _all_ of Rails, but this is a first step.

- Split out individual gems in Gemfile
- Update `govuk-components` to new version that doesn't depend on all of
  Rails
- Update `mimemagic` to new version
- Update Dockerfile to install `shared-mime-info` (new `mimemagic`
  runtime dependency)


cpjmcquillan:
Are we even using `gov-uk-date-fields` at all?

csutter:
> Are we even using gov-uk-date-fields at all?
Sadly yes, although my thinking is we should get rid of them as the form builder supports date fields anyway 🤷‍♀️ 

We just probably never got around to it?

cpjmcquillan:
> > Are we even using gov-uk-date-fields at all?
> > Sadly yes, although my thinking is we should get rid of them as the form builder supports date fields anyway 🤷‍♀️
> 
> We just probably never got around to it?

Ah ok I can see it now. Yeah let's get rid ASAP

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3166.london.cloudapps.digital>

felixtheflex (approved):

PR #3167, opened 2021-03-26
Remove gem we don't need or use

cpjmcquillan:
Mimemagic related clean up. 

We don't actually use/need this gem, and it's a bit of legacy code we didn't notice before it bit us in the ...

There might be some next steps around cleaning up dates anyway but that lives to fight another day.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,6 @@ gem "google-apis-indexing_v3"
 gem "google-cloud-bigquery"
 gem "google_drive", require: false
 gem "govuk-components"
-gem "gov_uk_date_fields"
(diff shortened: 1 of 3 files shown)


cesidio (approved):
niiiiice

csutter:
🎉 

Only thing I can think of: Does this impact our error messages in any way?

david-mears-dfe (approved):

cpjmcquillan:
> 🎉
> 
> Only thing I can think of: Does this impact our error messages in any way?

No, all good everywhere I checked (happy path creating a job with all manner of date errors)

PR #3171, opened 2021-03-26
[TEVA-2046] implement publisher application list page

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2046

Screenshot 2021-03-26 at 14 00 16

Screenshot 2021-04-01 at 00 43 44

Screenshot 2021-04-05 at 23 10 50

### DIFF
--- a/app/controllers/publishers/vacancies_controller.rb
+++ b/app/controllers/publishers/vacancies_controller.rb
@@ -1,7 +1,9 @@
 class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   include Publishers::Wizardable
 
-  before_action :set_vacancy, only: %i[destroy edit preview review show summary]
+  helper_method :applications, :sort, :sort_form
+
+  before_action :set_vacancy, only: %i[destroy edit preview review show summary job_applications]
   before_action :redirect_if_published, only: %i[preview review]
   before_action :redirect_unless_permitted, only: %i[preview summary]
   before_action :devise_job_alert_search_criteria, only: %i[show preview]
@@ -60,6 +62,18 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   end
 
   private
+
+  def applications
+    @applications ||= @vacancy.job_applications.not_draft.order(sort.column => sort.order)
+  end
+
+  def sort
+    @sort ||= Publishers::JobApplicationSort.new.update(column: params[:sort_column])
+  end
+
+  def sort_form
+    @sort_form ||= SortForm.new(sort.column)
+  end
 
   def devise_job_alert_search_criteria
     @devised_job_alert_search_criteria = Search::CriteriaDeviser.new(@vacancy).criteria
--- a/app/helpers/job_application_helper.rb
+++ b/app/helpers/job_application_helper.rb
@@ -1,5 +1,5 @@
 module JobApplicationHelper
-  PUBLISHER_STATUS_MAPPINGS = { submitted: "review", unsuccessful: "rejected" }.freeze
+  PUBLISHER_STATUS_MAPPINGS = { shortlisted: "shortlisted", submitted: "review", unsuccessful: "rejected", withdrawn: "withdrawn" }.freeze
 
   JOB_APPLICATION_STATUS_TAG_COLOURS = {
     draft: "pink", submitted: "blue", shortlisted: "green", unsuccessful: "red", withdrawn: "grey"
@@ -15,5 +15,9 @@ module JobApplicationHelper
     govuk_tag text: PUBLISHER_STATUS_MAPPINGS[status.to_sym],
               colour: JOB_APPLICATION_STATUS_TAG_COLOURS[status.to_sym],
               classes: "govuk-!-margin-bottom-2"
+  end
+
+  def status_tag_colour(status)
+    JOB_APPLICATION_STATUS_TAG_COLOURS[status]
   end
 end
--- a/app/views/publishers/vacancies/job_applications.html.slim
+++ b/app/views/publishers/vacancies/job_applications.html.slim
@@ -1,0 +1,81 @@
+- content_for :page_title_prefix, current_organisation.name
+
+= render BannerComponent.new do
+  - if @vacancy.expires_on.past?
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.expired.tab_heading")}": jobs_with_type_organisation_path(:expired), "#{@vacancy.job_title}": organisation_job_job_applications_path(@vacancy.id) }
+  - else
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.published.tab_heading")}": organisation_path, "#{@vacancy.job_title}": organisation_job_job_applications_path(@vacancy.id) }
+  .govuk-grid-row
+    .govuk-grid-column-full.application-header
+      - if current_organisation.is_a?(SchoolGroup)
+        h2.govuk-caption-l = current_organisation.name
+        h1.govuk-heading-l
+          span.govuk-heading-l class="govuk-!-font-weight-regular" = t(".heading")
+          = @vacancy.job_title
+      - else
+        h2.govuk-caption-l = t(".heading")
+        h1.govuk-heading-l = @vacancy.job_title
+
+  .govuk-grid-row
+    .govuk-grid-column-one-quarter
+      .vacancy-deadline class=("vacancy-deadline--passed" if @vacancy.expires_on.past?)
+        h4.govuk-body.govuk-caption-s class="govuk-!-font-weight-bold govuk-!-margin-bottom-0"
+          = @vacancy.expires_on.past? ? t(".deadline.after") : t(".deadline.before")
+        span.govuk-body = expiry_date_and_time(@vacancy)
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:submitted)}"
+        .govuk-heading-l = applications.submitted.count
+        h3.govuk-caption-s = t(".review")
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:shortlisted)}"
+        .govuk-heading-l = applications.shortlisted.count
+        h3.govuk-caption-s = t(".shortlisted")
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:unsuccessful)}"
+        .govuk-heading-l = applications.unsuccessful.count
+        h3.govuk-caption-s = t(".rejected")
+
+.govuk-grid-row
+  .govuk-grid-column-one-third
+    = form_for sort_form, as: "", url: organisation_job_job_applications_path(@vacancy.id), method: :get, data: { "auto-submit": true, "hide-submit": true } do |f|
+      = f.govuk_collection_select :sort_column,
+        sort,
+        :column,
+        :display_name,
+        label: { text: t("jobs.sort_by.label"), size: "s", class: "govuk-label inline govuk-!-margin-right-2" },
+        selected: :last
+
+  .govuk-grid-column-two-thirds.vacancy-actions
+    - unless @vacancy.expires_on.past?
+      = govuk_link_to(t(".buttons.edit"), edit_organisation_job_path(@vacancy.id), class: "govuk-button--secondary", button: true)
+
+    - if @vacancy.expires_on.past?
+      = govuk_link_to(t(".buttons.relist"), "/", class: "govuk-button--secondary", button: true)
+
+    = govuk_link_to(t(".buttons.copy"), organisation_job_copy_path(@vacancy.id), class: "govuk-button--secondary", button: true)
+
+    - unless @vacancy.expires_on.past?
+      = govuk_link_to(t(".buttons.end"), organisation_job_end_listing_path(@vacancy.id), class: "govuk-button--secondary", button: true)
+
+.govuk-grid-row
+  .govuk-grid-column-full
+    - if applications.any?
+      - applications.each do |application|
+        = render Shared::CardComponent.new(classes: "application-#{application.status}") do |card|
+          - card.header do
+            = tag.div(govuk_link_to("#{application.first_name} #{application.last_name}", organisation_job_job_application_path(@vacancy.id, application.id)))
+
+          - card.body do
+            = tag.div(publisher_job_application_status_tag(application.status))
+            = tag.div(card.labelled_item(t(".received"), application.submitted_at))
+
+          - if application.shortlisted? || application.submitted?
+            - card.actions do
+              - if application.submitted?
+                = tag.div(govuk_link_to(t(".actions.shortlist"), organisation_job_job_application_shortlist_path(@vacancy.id, application.id)))
+              = tag.div(govuk_link_to(t(".actions.reject"), organisation_job_job_application_reject_path(@vacancy.id, application.id)))
+    - else
+      = render EmptySectionComponent.new title: t(".no_applicants")
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,33 +1,13 @@
 {
   "ignored_warnings": [
     {
-      "warning_type": "Unscoped Find",
-      "warning_code": 82,
-      "fingerprint": "0154c62f733be2e251649d6eb9d004013037c8c82b10c73704efe4012f2620ba",
-      "check_name": "UnscopedFind",
-      "message": "Unscoped call to `EmergencyLoginKey#find`",
-      "file": "app/controllers/publishers/sign_in/email/sessions_controller.rb",
-      "line": 70,
-      "link": "https://brakemanscanner.org/docs/warning_types/unscoped_find/",
-      "code": "EmergencyLoginKey.find(params[:login_key])",
-      "render_path": null,
-      "location": {
-        "type": "method",
-        "class": "Publishers::SignIn::Email::SessionsController",
-        "method": "get_key"
-      },
-      "user_input": "params[:login_key]",
-      "confidence": "Weak",
-      "note": "Intended behaviour - we do not have a user available at this point"
-    },
-    {
       "warning_type": "Authentication",
       "warning_code": 101,
       "fingerprint": "1d94226b7d2099d5fbdfc82158f8100f98d20a42e468bc779bf98c2ac77b7c05",
       "check_name": "Secrets",
       "message": "Hardcoded value for `NOTIFY_JOBSEEKER_RESET_PASSWORD_TEMPLATE` in source code",
       "file": "config/initializers/notifications.rb",
-      "line": 5,
+      "line": 8,
       "link": "https://brakemanscanner.org/docs/warning_types/authentication/",
       "code": null,
       "render_path": null,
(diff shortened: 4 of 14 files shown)


app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,6 +59,13 @@ def summary
     @feedback_form = Publishers::JobListing::FeedbackForm.new
   end
 
+  def applications
+    @vacancy = VacancyPresenter.new(@vacancy)
+    @sort = Publishers::VacancySort.new(current_organisation, @selected_type).update(column: params[:sort_column])
```
alexbowen, line 64: i realise this sort isnt the right sort, its just to see the select in UI for now

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3171.london.cloudapps.digital>

app/views/publishers/vacancies/applications.html.slim
```diff
@@ -0,0 +1,72 @@
...
+        :column,
+        :display_name,
+        label: { text: t("jobs.sort_by.label"), size: "s", class: "govuk-label inline govuk-!-margin-right-2" }
+
+  .govuk-grid-column-two-thirds.vacancy-actions
+    - unless deadline_passed?(@vacancy.expires_at)
+      = govuk_link_to(t("publishers.vacancies.job_applications.buttons.edit"), edit_organisation_job_url(@vacancy.id), class: "govuk-button--secondary", button: true)
+
+    - if deadline_passed?(@vacancy.expires_at)
+      = govuk_link_to(t("publishers.vacancies.job_applications.buttons.relist"), "/", class: "govuk-button--secondary", button: true)
```
alexbowen, line 46: was it decided where the relist link should go to?

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,25 @@ def summary
     @feedback_form = Publishers::JobListing::FeedbackForm.new
   end
 
+  def applications
+    @vacancy = VacancyPresenter.new(@vacancy)
+    @application_statuses = @vacancy.job_applications.group_by { |k| k[:status] }.transform_values(&:count)
```
david-mears-dfe, line 66: It would save 1 reloading of the applications from the database if you call your `vacancy_applications` method here instead of `@vacancy.job_applications`.

app/views/publishers/vacancies/applications.html.slim
```diff
@@ -0,0 +1,72 @@
+- content_for :page_title_prefix, current_organisation.name
+
+= render BannerComponent.new do
+  = govuk_breadcrumbs breadcrumbs: { "#{t("breadcrumbs.active_jobs")}": root_path, "#{@vacancy.job_title}": root_path }
```
david-mears-dfe, line 4: Should both these breadcrumbs really point to the same path?

alexbowen: no, good spot and i havent tested it either

alexbowen: in fact the second shouldnt be a link at all

alexbowen: according to figma anyway, ill run it past chris

app/helpers/job_application_helper.rb
```diff
@@ -16,4 +16,12 @@ def publisher_job_application_status_tag(status)
               colour: JOB_APPLICATION_STATUS_TAG_COLOURS[status.to_sym],
               classes: "govuk-!-margin-bottom-2"
   end
+
+  def status_tag_colour(status)
+    JOB_APPLICATION_STATUS_TAG_COLOURS[status.to_sym]
+  end
+
+  def deadline_passed?(date)
```
david-mears-dfe, line 24: Does the `expired` scope on Vacancy not meet your needs? (That returns _published_ vacancies that have a past expiry_date)

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,75 @@
...
+  .govuk-grid-row
+    .govuk-grid-column-one-quarter
+      .application-deadline class=("application-deadline--passed" if deadline_passed?(@vacancy.expires_at))
+        h4.govuk-body.govuk-caption-s class="govuk-!-font-weight-bold govuk-!-margin-bottom-0"
+          = deadline_passed?(@vacancy.expires_at) ? t("publishers.vacancies.job_applications.applicants.deadline.after") : t("publishers.vacancies.job_applications.applicants.deadline.before")
+        span.govuk-body = @vacancy.expires_at
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:submitted)}"
+        .govuk-heading-l = @application_statuses["submitted"] || 0
```
david-mears-dfe, line 22: Another way of writing this would be `.to_i` since `nil.to_i` returns 0.

david-mears-dfe: and `.to_i` could go in the controller.

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,75 @@
+- content_for :page_title_prefix, current_organisation.name
+
+= render BannerComponent.new do
+  - if deadline_passed?(@vacancy.expires_at)
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.expired.tab_heading")}": jobs_with_type_organisation_url(:expired), "#{@vacancy.job_title}": organisation_job_job_applications_url(@vacancy.id) }
+  - else
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.published.tab_heading")}": organisation_url, "#{@vacancy.job_title}": organisation_job_job_applications_url(@vacancy.id) }
+  .govuk-grid-row
+    .govuk-grid-column-full
+      h2.govuk-caption-l = t("publishers.vacancies.job_applications.applicants.heading")
```
david-mears-dfe, line 10: I don't think you need the first `publishers.vacancies.job_applications` - they are implied by the file location.

david-mears-dfe: applies across several translations

david-mears-dfe: For my taste, "applicants" is unnecessary

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,25 @@ def summary
...
   end
 
+  def job_applications
+    @vacancy = VacancyPresenter.new(@vacancy)
+    @application_statuses = vacancy_applications.group_by { |k| k[:status] }.transform_values(&:count)
+  end
+
   private
 
+  def vacancy_applications
```
david-mears-dfe, line 71: Could we omit applications with draft status at this stage, to make certain that they won't be displayed anywhere if a bug is introduced in the view?

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,75 @@
...
+      - vacancy_applications.each do |application|
+        = render Shared::CardComponent.new(classes: "application-#{application.status}") do |card|
+          - card.header do
+            = tag.div(govuk_link_to("#{application.first_name} #{application.last_name}", organisation_job_job_application_url(@vacancy.id, application.id)))
+
+          - card.body do
+            = tag.div(publisher_job_application_status_tag(application.status))
+            = tag.div(card.labelled_item(t("publishers.vacancies.job_applications.applicants.received"), application.submitted_at))
+
+          - if application.status == "shortlisted" || application.status == "submitted"
```
david-mears-dfe, line 68: `if application.shortlisted? || application.submitted?`

app/controllers/publishers/vacancies_controller.rb
david-mears-dfe: I couldn't get into the review app so was unable to check for myself that this doesn't display draft applications.

spec/system/hiring_staff_can_manage_vacancy_applications_spec.rb
```diff
@@ -0,0 +1,205 @@
+require "rails_helper"
+
+RSpec.describe "Hiring staff can manage vacancy applications" do
```
david-mears-dfe, line 3: Could you add a test to make sure that no draft application is shown?

alexbowen: one or two other tests i need to add, this is one didnt cross my mind

app/controllers/publishers/vacancies_controller.rb
alexbowen: review app gh action was failing this aft for some reason

app/helpers/job_application_helper.rb
alexbowen: not too sure what you mean here, i dont see how expired scope helps for what this is needed for. ill pick this up with you tues

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,25 @@ def summary
...
+    @vacancy_applications ||= @vacancy.job_applications.order("#{sort.column} #{sort.order}").reject { |a| a[:status] == "draft" }
```
cpjmcquillan, line 72:
```ruby
vacancy.job_applications.not_draft.order(sort.column => sort.order)
```
We can scope by job application status, and we only need to do an interpolated order when we want to sort by something on a joined table (e.g `vacancies.publish_on`)

app/helpers/job_application_helper.rb
cpjmcquillan: I'm not sure what the need for this method is either?

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,74 @@
+- content_for :page_title_prefix, current_organisation.name
+
+= render BannerComponent.new do
+  - if deadline_passed?(@vacancy.expires_at)
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.expired.tab_heading")}": jobs_with_type_organisation_url(:expired), "#{@vacancy.job_title}": organisation_job_job_applications_url(@vacancy.id) }
```
cpjmcquillan, line 5: We should be using path helpers rather than urls

spec/system/hiring_staff_can_manage_vacancy_applications_spec.rb
```diff
@@ -0,0 +1,218 @@
...
```
cpjmcquillan, line 3: I would rather we said `Publishers can manage job applications for a vacancy` for two reasons
- Publishers is the name of the model 
- Vacancy applications makes me want to cry 😂

db/seeds.rb
```diff
@@ -277,3 +277,13 @@
                   id: "6683c564-15e6-41af-ab44-7adf125f4c84",
                   jobseeker: jobseeker,
                   vacancy: physics_job)
+
+FactoryBot.create(:job_application, :status_shortlisted,
+                  id: "6683c564-15e6-41af-ab44-7adf125f4c85",
+                  jobseeker: jobseeker,
+                  vacancy: physics_job)
+
+FactoryBot.create(:job_application, :status_shortlisted,
```
cpjmcquillan, line 286: I'd err against creating seeds that don't make sense. A jobseeker can't apply for the same job twice, let alone be shortlisted twice for the same role

spec/system/hiring_staff_can_manage_vacancy_applications_spec.rb
cpjmcquillan: Also, this seems like an awful lot of examples when we want system specs to be happy path

config/locales/publishers.yml
```diff
@@ -160,18 +160,38 @@ en:
...
           alert: The candidate will be sent an email telling them that they have been shortlisted.
           heading: Shortlist applicant
           page_title: Shortlist applicant
+        shortlisted: Shortlisted applicants
         show:
           page_title: Application
           status_heading: Applicant status
           timeline: Timeline
+        sort_by:
+          applicant_surname: applicant surname (A to Z)
```
cpjmcquillan, line 193: We should say `last name` because that's what the field is called everywhere else

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,74 @@
...
+          - card.header do
+            = tag.div(govuk_link_to("#{application.first_name} #{application.last_name}", organisation_job_job_application_url(@vacancy.id, application.id)))
+
+          - card.body do
+            = tag.div(publisher_job_application_status_tag(application.status))
+            = tag.div(card.labelled_item(t(".received"), application.submitted_at))
+
+          - if application.shortlisted? || application.submitted?
+            - card.actions do
+              - if application.status == "submitted"
```
cpjmcquillan, line 70: `application.submitted?`

cpjmcquillan, line 70: `application.submitted?`

app/controllers/publishers/vacancies_controller.rb
cpjmcquillan, line 71: Can this be called `job_application` like we do elsewhere?

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,74 @@
...
+
+  .govuk-grid-row
+    .govuk-grid-column-one-quarter
+      .application-deadline class=("application-deadline--passed" if deadline_passed?(@vacancy.expires_at))
+        h4.govuk-body.govuk-caption-s class="govuk-!-font-weight-bold govuk-!-margin-bottom-0"
+          = deadline_passed?(@vacancy.expires_at) ? t(".deadline.after") : t(".deadline.before")
+        span.govuk-body = expiry_date_and_time(@vacancy)
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:submitted)}"
```
cpjmcquillan, line 21: Do we need a helper method for this? 

If we do, should we pass in a symbol and convert it to a string or should we pass in a string in the first place?

db/seeds.rb
alexbowen: sure was for my benefit only

app/controllers/publishers/vacancies_controller.rb
alexbowen: ill call it applications as job_applications is the name of the route method?

config/brakeman.ignore
```diff
@@ -56,6 +36,26 @@
       "confidence": "Weak",
       "note": "Sort columns and orders come from an allowed list of values, so SQL injection risk can be ignored"
     },
+    {
```
cpjmcquillan, line 39: You shouldn't need this now

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,25 @@ def summary
     @feedback_form = Publishers::JobListing::FeedbackForm.new
   end
 
+  def job_applications
+    @vacancy = VacancyPresenter.new(@vacancy)
```
cpjmcquillan, line 65: Do we need the presenter?

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,25 @@ def summary
...
+    @application_statuses = applications.group_by { |k| k[:status] }.transform_values(&:count)
```
cpjmcquillan, line 66: Can we live without this instance variable? In the template we might be able to do `job_applications.submitted.count` for example instead

app/controllers/publishers/vacancies_controller.rb
david-mears-dfe: Maybe it's better the other way round: the route called 'applications' and the helper 'job_applications'? I don't know. Just in service of nice-looking urls.

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +61,22 @@ def summary
     @feedback_form = Publishers::JobListing::FeedbackForm.new
   end
 
+  def job_applications; end
```
cpjmcquillan, line 64: You don't need this now. As long as the template exists, rails will know what to do

app/views/publishers/vacancies/job_applications.html.slim
```diff
@@ -0,0 +1,74 @@
...
+        .govuk-heading-l = applications.submitted.count
+        h3.govuk-caption-s = t(".review")
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour("shortlisted")}"
+        .govuk-heading-l = applications.shortlisted.count
+        h3.govuk-caption-s = t(".shortlisted")
+
+    .govuk-grid-column-one-quarter
+      .application-count.govuk-tag class="govuk-tag--#{status_tag_colour("unsuccessful")}"
```
cpjmcquillan, line 31: Sorry my comment before didn't read the method properly. Passing in `:unsuccessful` is ok, we just don't need to to `:unsuccessful.sym` in the method, because it's already a symbol

app/controllers/publishers/vacancies_controller.rb
alexbowen: i agree about nice looking url, ive been round and round in circles with this and it didnt work for some reason which i cant even remeber! ill try again

config/locales/publishers.yml
```diff
@@ -160,18 +160,38 @@ en:
...
           alert: The candidate will be sent an email telling them that they have been shortlisted.
           heading: Shortlist applicant
           page_title: Shortlist applicant
+        shortlisted: Shortlisted applicants
         show:
           page_title: Application
           status_heading: Applicant status
           timeline: Timeline
+        sort_by:
+          applicant_last_name: applicant surname (A to Z)
```
cpjmcquillan, line 193: This still says `surname`

david-mears-dfe (approved):

config/locales/publishers.yml
alexbowen: yeah thtas what it says in figma, ill change to surname pending chris saying its cool too

cpjmcquillan (changes requested):
Requesting changes because it's otherwise approved but it doesn't actually work and there are related gaps in testing

app/helpers/job_application_helper.rb
```diff
@@ -16,4 +16,8 @@ def publisher_job_application_status_tag(status)
               colour: JOB_APPLICATION_STATUS_TAG_COLOURS[status.to_sym],
               classes: "govuk-!-margin-bottom-2"
   end
+
+  def status_tag_colour(status)
+    JOB_APPLICATION_STATUS_TAG_COLOURS[status.to_sym]
```
cpjmcquillan, line 21: You can remove the `.to_sym`

spec/system/publishers_can_manage_job_applications_for_a_vacancy.rb
```diff
@@ -0,0 +1,218 @@
...
+
+  context "when a vacancy has expired and it has applications" do
+    let(:vacancy) { create(:vacancy, :expired) }
+    let!(:job_application_submitted) { create(:job_application, :status_submitted, vacancy: vacancy) }
+    let!(:job_application_shortlisted) { create(:job_application, :status_shortlisted, vacancy: vacancy) }
+    let!(:job_application_unsuccessful) { create(:job_application, :status_unsuccessful, vacancy: vacancy) }
+    let!(:job_application_draft) { create(:job_application, :status_draft, vacancy: vacancy) }
+
+    before do
+      vacancy.organisation_vacancies.create(organisation: organisation)
```
cesidio, line 17: This could be part of line 10:
```
let(:vacancy) { create(:vacancy, :expired, organisation_vacancies_attributes: [{ organisation: organisation }]) }
```

cpjmcquillan (approved):
I think the test could be slimmed down a bit, and the quality gate won't pass until coverage issues are addressed, but I'm approving to remove my big "requested changes" ❌

PR #3178, opened 2021-03-26
alter paths of previews to be more intuitive

alexbowen:
no ticket

makes path bit cleaner e.g `components/notification_component/preview`

### DIFF
--- a/app/controllers/previews_controller.rb
+++ b/app/controllers/previews_controller.rb
@@ -1,4 +1,4 @@
-class PreviewController < ApplicationController
+class PreviewsController < ApplicationController
   layout "design_system"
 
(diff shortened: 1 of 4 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3178.london.cloudapps.digital>

app/controllers/previews_controller.rb
```diff
@@ -1,4 +1,4 @@
-class PreviewController < ApplicationController
+class PreviewsController < ApplicationController
```
alexbowen, line 1: renamed as the convention seems to be to use plural

david-mears-dfe (approved):

PR #3186, opened 2021-03-29
Filters component refactor

alexbowen:
- update tests and code to use shared examples
- remove custom accordion and use GDS component instead
- move out of shared namespace
- add preview component

Screenshot 2021-03-29 at 09 48 01

Screenshot 2021-03-29 at 09 50 21

todo (in another PR)
- look into accordion.scss as we shouldnt need this file, any styles we do need should prob live elsewhere
- remove dependency of moj-filter
- there is complexity can be removed from this component as there are options that we dont use


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3186.london.cloudapps.digital>

csutter (approved):

PR #3191, opened 2021-03-30
Revert "Detect MIME type for uploads by inspecting content"

csutter:
Reverts DFE-Digital/teaching-vacancies#3174

Turns out `mimemagic` is not smart enough in our case to tell Word documents and Zip files apart 😞 

Reverting this until we find a way to improve on this or come up with a better solution.


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3191.london.cloudapps.digital>

cpjmcquillan (approved):

PR #3193, opened 2021-03-30
[TEVA-1747] Qualifications step: Part 1

david-mears-dfe:
## Jira ticket URL

## Part 1 of 3

Mears Syndicated says: _Your convenience is our pleasure. We are committed to helping you by splitting our work into multiple PRs._

Step 1 (here) introduces the basic forms.

Step 2 will adapt the DetailComponent and ReviewComponent so that the user can see what qualifications they've got, on the review page (and on the main qualifications step)

Step 3 will allow some forms to gain or lose extra fieldsets on demand (to add a new subject).

# Screenshots

<details>
  <summary>Main qualifications step page (Click to expand!)</summary>
  
localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_build_qualifications (1)
</details>

<details>
  <summary>Choose qualification category</summary>
  
localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_qualifications_submit_category
</details>

<details>
  <summary>Qualification category type 1: common secondary school quals</summary>
  
  TODO in a later PR: sort out the inline error message so that it displays nicely

localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_qualifications_category=gcse
</details>

<details>
  <summary>Qualification category type 2: other secondary school quals</summary>
  
    TODO in a later PR: sort out the inline error message so that it displays nicely

localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_qualifications_category=other_secondary
</details>

<details>
  <summary>Qualification category type 3: graduate quals</summary>

localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_qualifications_category=undergraduate
</details>

<details>
  <summary>Qualification category type 4: any other quals</summary>

localhost_3000_jobseekers_job_applications_f104fda8-f13d-4d00-9cc7-b2b80354dea7_qualifications_category=other
</details>


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3193.london.cloudapps.digital>

david-mears-dfe:
Brakeman:
```
== Warnings ==

Confidence: Medium
Category: Remote Code Execution
Check: UnsafeReflection
Message: Unsafe reflection method `constantize` called with parameter value
File: app/controllers/jobseekers/job_applications/qualifications_controller.rb
Line: 84
```

```ruby
def form_class
    name = if %w[select_category submit_category].include?(action_name)
             "CategoryForm"
           else
             case params[:category]
             when "gcse", "a_level", "as_level"
               "Secondary::CommonForm"
             when "other_secondary"
               "Secondary::OtherForm"
             when "undergraduate", "postgraduate"
               "DegreeForm"
             when "other"
               "OtherForm"
             end
           end
    "Jobseekers::JobApplication::Details::Qualifications::#{name}".constantize
  end
```

This is fine, right?

cpjmcquillan (commented):
Two thoughts I have so far...

Do we really need so many different qualification forms? How would it look like with a single form that applied different validations depending on the qualification category?

Also, how unreasonable is it to add extra subjects to a qualification without any JS at all?

david-mears-dfe:
> Do we really need so many different qualification forms? How would it look like with a single form that applied different validations depending on the qualification category?

I can try this out.

> Also, how unreasonable is it to add extra subjects to a qualification without any JS at all?

I'd be more comfortable writing this as I'm no JS whiz, but will chat with Chris.

cpjmcquillan:
> > Do we really need so many different qualification forms? How would it look like with a single form that applied different validations depending on the qualification category?
> 
> I can try this out.
> 
> > Also, how unreasonable is it to add extra subjects to a qualification without any JS at all?
> 
> I'd be more comfortable writing this as I'm no JS whiz, but will chat with Chris.

I doubt the end user will notice the difference but if we were really concerned it could also be a nice place to try out https://hotwire.dev/ if you fancy it

david-mears-dfe:
> Do we really need so many different qualification forms? How would it look like with a single form that applied different validations depending on the qualification category?

I've remembered why I split the forms out now. It's because we need different labels and error messages for each form. Once I'd de-duplicated all of the text shared between forms, there were 6 error messages and 6 labels that broke the mould. I'm happy with this as the price for having text that makes sense in the context of different types of qualification, but maybe there are further ways to decrease the code complexity.

cpjmcquillan:
> > Do we really need so many different qualification forms? How would it look like with a single form that applied different validations depending on the qualification category?
> 
> I've remembered why I split the forms out now. It's because we need different labels and error messages for each form. Once I'd de-duplicated all of the text shared between forms, there were 6 error messages and 6 labels that broke the mould. I'm happy with this as the price for having text that makes sense in the context of different types of qualification, but maybe there are further ways to decrease the code complexity.

Initial thought is maybe we can add some label/error message methods to the form model (I've thought about this before when we've got school/mat/la funk going on). That might just move the complexity elsewhere, but it might be an acceptable pattern for scenarios like this

david-mears-dfe:
I stuck with the multiple forms for the reason above. I agree with Christian (on Slack):

> Contrarian point – I don’t have lots of context but I kinda like the small and single responsibility multiple forms

app/models/job_application.rb
```diff
@@ -6,6 +6,8 @@ class JobApplication < ApplicationRecord
   array_enum completed_steps: {
     personal_details: 0,
     professional_status: 1,
+    qualifications: 2,
+    professional_status: 1,
```
cesidio: this line is duplicated of line 8

app/views/jobseekers/job_applications/build/qualifications.html.slim
```diff
@@ -1,13 +1,14 @@
 - content_for :page_title_prefix, t(".title")
 
-= render BannerComponent.new do
+= render BannerComponent.new doe
```
cesidio: doe?

app/controllers/concerns/jobseekers/wizardable.rb
```diff
@@ -7,6 +7,10 @@ def professional_status_fields
     %i[qualified_teacher_status qualified_teacher_status_year qualified_teacher_status_details statutory_induction_complete]
   end
 
+  def qualifications_fields
+    []
```
alexbowen, line 11: is this a part 1 thing that will have the array filled up in subsequent parts?

config/locales/forms.yml
```diff
@@ -317,6 +321,36 @@ en:
         reason_for_leaving: Reason for leaving
         salary: Salary scale or level
         subjects: Subjects and key stages taught
+      jobseekers_job_application_details_qualifications_category_form:
+        category_options:
+          gcse: GCSE
```
alexbowen, line 326: alhpabetical order

app/frontend/src/styles/application/form.scss
```diff
@@ -67,6 +67,21 @@
   }
 }
 
+.inline-form-group {
+  display: inline-block;
+
+  &.inline-grade {
```
alexbowen, line 73: i would either name these classes more generically or move them to a scss file specific to what they are specific too

app/frontend/src/styles/application/form.scss
```diff
@@ -67,6 +67,21 @@
...
+    $width: govuk-grid-width(one-third);
+    $padding-left: govuk-spacing(3);
+    padding-left: $padding-left;
+    width: calc(#{$width} - #{$padding-left});
```
alexbowen, line 77: not sure exactly the use case of this calculation but can you use flex(box) to acheive this?

alexbowen: not that using calc is wrong but i try to use them as a last resort

cesidio (approved):

app/models/qualification.rb
```diff
@@ -0,0 +1,11 @@
+class Qualification < ApplicationRecord
+  belongs_to :job_application
+
+  enum category: { gcse: 0, as_level: 1, a_level: 2, other_secondary: 3, undergraduate: 4, postgraduate: 5, other: 6 }
+
+  def name
+    return read_attribute(:name) if read_attribute(:name).present? || other? || other_secondary?
```
cpjmcquillan, line 7: I'm a bit confused by what's going on here, is there something I'm missing?

app/controllers/jobseekers/job_applications/qualifications_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+  def back_path
+    @back_path ||= jobseekers_job_application_build_path(job_application, :qualifications)
+  end
+
+  def qualification
+    @qualification ||= job_application.qualifications.find(params[:id])
+  end
+
+  def job_application
+    @job_application ||= current_jobseeker.job_applications.find(params[:job_application_id])
```
cpjmcquillan, line 96: This should be scoped to draft job applications only

app/controllers/jobseekers/job_applications/qualifications_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  def destroy
+    qualification.destroy
+    redirect_to back_path, success: t(".success")
+  end
+
+  private
+
+  def form
+    @form ||= action_name == "select_category" ? form_class.new : form_class.new(qualification_params)
```
cpjmcquillan, line 44: Should this use the `form_attributes` method?

```ruby
@form ||= form_class.new(form_attributes)
```

and update the case...when to be 
```ruby
when "new", "select_category"
  {}
```

cpjmcquillan (commented):
Can you add a request spec to this PR please?

Also would like a system spec like `jobseekers_can_add_qualifications_to_their_job_application` - but I expect you might have been saving this for a subsequent PR?

app/models/qualification.rb
david-mears-dfe: I created this `name` method a bit early, will use it next PR, but the idea is that

- Some types of qualification form don't have a 'name' field (GCSE etc) and others do have a 'name' (non-standard qualifications).
- We want to display the name as part of the DetailsComponent/ReviewComponent.
- If the name column is nil in the database (e.g. it's a GCSE), fall back to using the category instead when 'name' is called.
- Now any qualification can have 'name' called and reliably return something useful to display.

app/controllers/concerns/jobseekers/wizardable.rb
david-mears-dfe: No - it's like the reference_fields later in this file. These steps don't have any fields, but they need this method in order to be step-shaped for the wizard.

config/locales/forms.yml
david-mears-dfe: The order in this file is determined by the order that we want them to be displayed on the form.

PR #3213, opened 2021-04-06
Jobseekers can quick apply for jobs

cpjmcquillan:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2037

## Changes in this PR
- Add a quick apply service
- Jobseekers can quick apply for a job, reusing data from the most recently submitted job application

### DIFF
--- a/spec/requests/jobseekers/job_applications_spec.rb
+++ b/spec/requests/jobseekers/job_applications_spec.rb
@@ -33,6 +33,16 @@ RSpec.describe "Job applications" do
           expect(get(new_jobseekers_job_job_application_path(vacancy.id))).to redirect_to(jobseekers_job_applications_path)
         end
       end
+
+      context "when a non-draft job application already exists" do
+        let!(:job_application) { create(:job_application, :status_submitted, jobseeker: jobseeker, vacancy: vacancy) }
+        let(:new_vacancy) { create(:vacancy, organisation_vacancies_attributes: [{ organisation: build(:school) }]) }
+
+        it "redirects to `new_quick_apply_jobseekers_job_job_application_path`" do
+          expect(get(new_jobseekers_job_job_application_path(new_vacancy.id)))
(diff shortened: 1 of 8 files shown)


cpjmcquillan:
The job applications controller is starting to hit that limit of "This feels kinda big and maybe we should split some of it out", so feel free to think about/comment on that and I might prep a PR for doing that work shortly

spec/requests/jobseekers/job_applications_spec.rb
```diff
@@ -33,6 +33,16 @@
           expect(get(new_jobseekers_job_job_application_path(vacancy.id))).to redirect_to(jobseekers_job_applications_path)
         end
       end
+
+      context "when a non-draft job application already exists" do
+        let!(:job_application) { create(:job_application, :status_submitted, jobseeker: jobseeker, vacancy: vacancy) }
+        let(:new_vacancy) { create(:vacancy, organisation_vacancies_attributes: [{ organisation: build(:school) }]) }
+
+        it "redirects to `new_quick_apply_jobseekers_job_job_application_path`" do
```
alexbowen, line 41: rather than the path in rails speak, would it be better to describe in english where its redirecting to ?

csutter (approved):

spec/requests/jobseekers/job_applications_spec.rb
cpjmcquillan: Hmm. My thinking is... controllers and request specs don't care much about humans or english, but happy path (system) specs, care very much about humans and english https://github.com/DFE-Digital/teaching-vacancies/blob/2437b7bf2d60685c4ab0cdbd61e9b22726c02a47/spec/system/jobseekers_can_quick_apply_for_a_vacancy_spec.rb#L15

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3213.london.cloudapps.digital>

PR #3221, opened 2021-04-07
[teva-2093] Update documentation concurrency deployment management

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-teva-2093

## Changes in this PR:

Update documentation on how we currently manage multiple merges

### DIFF
--- a/documentation/deployments.md
+++ b/documentation/deployments.md
@@ -49,6 +49,13 @@ When a PR is approved and merged into `master` branch, the GitHub actions workfl
 - Calls the [deploy_app.yml](/.github/workflows/deploy_app.yml) workflow to use Terraform to update the `web` and `worker` apps to use the new Docker image, and apply any changes to the `production` environment
 - Sends a Slack notification to the `#twd_tv_dev` channel
 
+### Merge and concurrency deployment management
+When there are multiple merges, this could lead to race conditions. To mitigate this, the `turnstyle` action prevents two or more instances of the same workflow from running concurrently.
+
+Furthermore, to help with workflow code reuse, we trigger a separate deployment workflow via the [workflow_dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) action. We must pass the `ref` (branch) or pull request `ref` to locate the workflow.
+
+To block the calling workflow until the triggered workflow is completed, we use `action-wait-for-check`. This checks and waits on a `sha` (commit) instead of `ref` (branch), which is a moving target.
+
 ### Build and deploy to an environment - Makefile
 
 This builds and deploys a Docker image from local code, then updates the environment to use that image


saliceti (approved):

documentation/deployments.md
```diff
@@ -49,6 +49,13 @@ When a PR is approved and merged into `master` branch, the GitHub actions workfl
 - Calls the [deploy_app.yml](/.github/workflows/deploy_app.yml) workflow to use Terraform to update the `web` and `worker` apps to use the new Docker image, and apply any changes to the `production` environment
 - Sends a Slack notification to the `#twd_tv_dev` channel
 
+### Merge and concurrency deployment management
+When there are multiple merges, this could lead to raise conditions. To mitigate this, the `turnstyle` action prevents two or more instances of the same workflow from running concurrently.
```
david-mears-dfe, line 53: "race conditions"

documentation/deployments.md
```diff
@@ -49,6 +49,13 @@ When a PR is approved and merged into `master` branch, the GitHub actions workfl
...
+
+Furthermore, to help with code reuse we trigger a separate deployment workflow via the workflow dispatch action. We must pass the `ref` (branch) or pull request `ref` to locate the workflow.
```
david-mears-dfe, line 55: What does the workflow dispatch action do? I'm not sure I understand how it helps with code reuse.

To be consistent, let's say `workflow_dispatch` (with backticks `) rather than 'workflow dispatch'.

documentation/deployments.md
```diff
@@ -49,6 +49,13 @@ When a PR is approved and merged into `master` branch, the GitHub actions workfl
...
+
+To block the calling workflow until the triggered workflow is completed, we utilise `action-wait-for-check`. This checks and waits on a `sha` (commit) instead of `ref` (branch), which is a moving target.
```
david-mears-dfe, line 57: Can we say 'use' instead of 'utilise' as it's easier to understand?

documentation/deployments.md
david-mears-dfe: Maybe 'avoid' instead of 'mitigate'? It's a more common word.

documentation/deployments.md
felixtheflex: workflow_dispatch: allows manual trigger of workflows via the UI and programtically.

The deployment workflows all share\reuse `deploy_app` workflow, which contains the terraform code logic for deploying `sha` commits to the environment(s).

david-mears-dfe: > The deployment workflows all share\reuse deploy_app workflow, which contains the terraform code logic for deploying sha commits to the environment(s).

This is helpful context - may be worth adding in brackets?

david-mears-dfe: Unless you think this is likely to change, which would make the documentation out of date.

saliceti: @david-mears-dfe we made it more generic with `workflow code reuse` as it applies to terraform, but also smoke tests or any reusable workflow

PR #3222, opened 2021-04-07
fix mobile cookie banner button width

alexbowen:
no ticket

Screenshot 2021-04-07 at 14 44 22

Screenshot 2021-04-12 at 02 07 00


cpjmcquillan (commented):
Hmm. We should be using `.govuk-button-group` (I think) in the template instead https://design-system.service.gov.uk/components/button/#grouping-buttons

alexbowen:
> .govuk-button-group

doesnt work this instance (at least not without out adding custom styling anyway) with one of the buttons being a govuk_button_to as it needs to send POST request

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3222.london.cloudapps.digital>

david-mears-dfe:
Does this affect any ongoing A B tests?

alexbowen:
> Does this affect any ongoing A B tests?

i dont beleive so. the cookie banner ab test isnt being run at the moment

alexbowen:
> Does this affect any ongoing A B tests?

actually you are kind of right to raise this, the visual regression tests for sign in AB test needed updating and i have included the new snapshots in this PR

alexbowen:
> Hmm. We should be using `.govuk-button-group` in the template instead https://design-system.service.gov.uk/components/button/#grouping-buttons

ok, i used the button group as best i think you can in this instance, still requires a few styles as it doesnt work 100% out of the box. but its probably a bit neater i agree

cpjmcquillan (approved):

PR #3227, opened 2021-04-07
Add caption and rename job_application_steps to steps

cpjmcquillan:


app/views/jobseekers/job_applications/build/_caption.html.slim
```diff
@@ -0,0 +1 @@
+.govuk-caption-m = t("jobseekers.job_applications.current_step", current: JobApplication.completed_steps[step] + 1, total: JobApplication.completed_steps.values.last + 2)
```
david-mears-dfe, line 1: I don't agree that the current step is always the one after the last completed step.

1. Start new application
2. Click 'save as draft' on first step
3. Continue the application
4. You are taken to the review step and can leap into any step (all incomplete)

david-mears-dfe, line 1: I don't think this is your fault but while we're here, it's really weird that our captions say "Step X of 10" but the steps component on the right only goes up to 9. (It doesn't include the 'review' step.)

david-mears-dfe: Aha - Figma says the error is in the steps component, not in the caption.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3227.london.cloudapps.digital>

app/views/jobseekers/job_applications/build/_caption.html.slim
david-mears-dfe: I'll make a little ticket to address that

david-mears-dfe: Slash I'll ask if Joe wants to do it

david-mears-dfe (approved):

app/views/jobseekers/job_applications/build/_caption.html.slim
cpjmcquillan: Haha I missed that on Joe's PR, slack him and see if he wants to just update it in the morning

PR #3236, opened 2021-04-08
Update local_authorities_extra_schools.yml

milimaldedfe:
Added all of Kent LA's schools outside their boundary (all in this [spreadsheet](https://docs.google.com/spreadsheets/d/1AgMGFm39eN0kolLBB8rKqjCk_JLzMAkmMFhfksZRKm4/edit#gid=1219364607))

Please check that I have followed the correct format :) NB two URNs are only 5 digits long but I checked these on GIAS and that is what's there so I'm keeping them. But it's possible they're not correct, and they might cause tests to fail? Any advice appreciated!


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3236.london.cloudapps.digital>

cesidio (approved):

PR #3238, opened 2021-04-08
[TEVA-2054] Update Review Page Confirmation Checkboxes and Add Related Error Messages

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2054

## Changes in this PR:

- Updated checkbox labels
- Added description for second checkbox
- Added custom error messages to be displayed when checkboxes not... checked

### Before

Screenshot 2021-04-08 at 17 28 51

### After

Screenshot 2021-04-08 at 17 29 12

### Error Messages 

Screenshot 2021-04-08 at 17 29 35

Screenshot 2021-04-08 at 17 29 51

### DIFF
--- a/app/views/jobseekers/job_applications/review.html.slim
+++ b/app/views/jobseekers/job_applications/review.html.slim
@@ -29,13 +29,24 @@
         li = render "jobseekers/job_applications/review/ask_for_support", allow_edit: true
         li = render "jobseekers/job_applications/review/declarations", allow_edit: true
 
-      h3.govuk-heading-l Confirmation
+      h3.govuk-heading-m = t(".confirmation.heading")
 
       = f.govuk_check_boxes_fieldset :confirm_data_accurate, multiple: false, legend: nil do
-        = f.govuk_check_box :confirm_data_accurate, 1, 0, multiple: false, link_errors: true, label: { text: "Confirm data accurate" }
+        = f.govuk_check_box :confirm_data_accurate, 1, 0, multiple: false, link_errors: true
+
+      h3.govuk-heading-m = t(".confirmation.how_we_use_your_data.heading")
+
+      p.govuk-body = t(".confirmation.how_we_use_your_data.description")
+
+      ul.govuk-list.govuk-list--bullet
+        li = t(".confirmation.how_we_use_your_data.list.dfe")
+        li = t(".confirmation.how_we_use_your_data.list.listing_organisation")
+        li = t(".confirmation.how_we_use_your_data.list.employer")
+
+      p.govuk-body = t(".confirmation.how_we_use_your_data.privacy_policy.description_html", link_to: govuk_link_to(t(".confirmation.how_we_use_your_data.privacy_policy.link_text"), page_path("privacy-policy")))
 
       = f.govuk_check_boxes_fieldset :confirm_data_usage, multiple: false, legend: nil do
-        = f.govuk_check_box :confirm_data_usage, 1, 0, multiple: false, link_errors: true, label: { text: "Confirm data usage" }
+        = f.govuk_check_box :confirm_data_usage, 1, 0, multiple: false, link_errors: true
 
       - if vacancy.listed?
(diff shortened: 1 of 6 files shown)


app/views/jobseekers/job_applications/review.html.slim
```diff
@@ -28,13 +28,21 @@
         li = render "jobseekers/job_applications/review/ask_for_support", allow_edit: true
         li = render "jobseekers/job_applications/review/declarations", allow_edit: true
 
-      h3.govuk-heading-l Confirmation
+      h3.govuk-heading-m = t(".confirmation.heading")
 
       = f.govuk_check_boxes_fieldset :confirm_data_accurate, multiple: false, legend: nil do
-        = f.govuk_check_box :confirm_data_accurate, 1, 0, multiple: false, link_errors: true, label: { text: "Confirm data accurate" }
+        = f.govuk_check_box :confirm_data_accurate, 1, 0, multiple: false, link_errors: true, label: { text: t(".confirmation.checkbox_label.confirm_data_accurate") }
```
david-mears-dfe, line 34: Instead of having a label in the view, you can add a label to the `label` section of `forms.yml`. It'll be something like adding `jobseekers_job_application_review_form` underneath `label`.

david-mears-dfe: Same for confirm_data_usage

config/locales/jobseekers.yml
```diff
@@ -237,6 +237,23 @@ en:
...
+          heading: Confirmation
+          how_we_use_your_data:
+            description: "When you submit your application, your data will be shared with:"
+            heading: How we use your data
+            list_html: |
+              <ul class="govuk-list govuk-list--bullet">
+                <li>the Department for Education</li>
+                <li>the school, trust or local authority which posted the job listing</li>
+                <li>the school or schools the job is at</li>
+              </ul>
```
david-mears-dfe, line 253: It makes me uncomfortable to write HTML directly in this file - I think localizations should focus on text, and views should handle markup.

We're not reusing this text anywhere else, so I suggest just putting it directly into the slim view.

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3238.london.cloudapps.digital>

config/locales/jobseekers.yml
alexbowen: yes seeing this in the locale file makes me shudder

josephhull676: There's a few more examples in en.yml and publishers.yml if you fancy wincing some more.

david-mears-dfe (approved):

app/views/jobseekers/job_applications/review.html.slim
cpjmcquillan: As David points out, these label translations shouldn't be here and we should use the default lookup in the formbuilder to find them

cpjmcquillan (changes requested):

spec/system/jobseekers_can_give_application_feedback_spec.rb
```diff
@@ -14,8 +14,8 @@
   it "allows jobseekers to give job application feedback after submitting the application" do
     visit jobseekers_job_application_review_path(job_application)
 
-    check "Confirm data accurate"
-    check "Confirm data usage"
+    check "I confirm that the above information is accurate and complete"
```
cpjmcquillan, line 17: This could use the translation

spec/system/jobseekers_can_submit_a_job_application_spec.rb
```diff
@@ -20,8 +22,8 @@
       click_on I18n.t("buttons.submit_application")
       expect(page).to have_content("There is a problem")
 
-      check "Confirm data accurate"
-      check "Confirm data usage"
+      check confirm_data_accurate_checkbox_label
```
cpjmcquillan, line 25: And so could these

cpjmcquillan (approved):

PR #3240, opened 2021-04-09
Use a boolean field for "Apply through TV"

csutter:
This makes more sense as a boolean field (for SQL values of "boolean",
i.e. true/false/NULL).

- Add `enable_job_applications` boolean field (to be more agnostic of
  our service name) and migrate existing data
- Remove old `apply_through_teaching_vacancies` field and change all
  usages

### DIFF
--- a/app/views/publishers/vacancies/build/applying_for_the_job.html.slim
+++ b/app/views/publishers/vacancies/build/applying_for_the_job.html.slim
@@ -16,11 +16,11 @@
         = render "publishers/vacancies/vacancy_form_partials/hidden_state_input", f: f
 
         - if JobseekerApplicationsFeature.enabled?
-          = f.govuk_radio_buttons_fieldset :apply_through_teaching_vacancies do
-            = f.govuk_radio_button :apply_through_teaching_vacancies, :yes, link_errors: true do
+          = f.govuk_radio_buttons_fieldset :enable_job_applications do
+            = f.govuk_radio_button :enable_job_applications, true, link_errors: true do
               = f.govuk_text_area :personal_statement_guidance, label: { size: "s" }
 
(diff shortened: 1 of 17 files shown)


app/views/publishers/vacancies/build/applying_for_the_job.html.slim
```diff
@@ -16,11 +16,11 @@
...
 
         - if JobseekerApplicationsFeature.enabled?
-          = f.govuk_radio_buttons_fieldset :apply_through_teaching_vacancies do
-            = f.govuk_radio_button :apply_through_teaching_vacancies, :yes, link_errors: true do
+          = f.govuk_radio_buttons_fieldset :enable_job_applications do
+            = f.govuk_radio_button :enable_job_applications, true, link_errors: true do
               = f.govuk_text_area :personal_statement_guidance, label: { size: "s" }
 
-            = f.govuk_radio_button :apply_through_teaching_vacancies, :no do
+            = f.govuk_radio_button :enable_job_applications, "false" do
```
cesidio, line 23: why "false" is a string here?

cesidio (approved):

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3240.london.cloudapps.digital>

app/views/publishers/vacancies/build/applying_for_the_job.html.slim
csutter: It wasn't working with literal `false` – I _think_ because it's falsey the form builder doesn't pick it up (not a problem for `true` 😕 

Another option would be "0" and "1", but I think I prefer it this way even if it's a bit ugly...

PR #3242, opened 2021-04-09
Prohibit changing enable_job_applications once listed

csutter:
To save ourselves several headaches, we've decided that for MAP we will
not allow publishers to change whether or not a vacancy can be applied
for through TV after it has gone live.

- Change "Applying for the job" step to not show the radio button with
  collapsible subfields when the vacancy is listed, and instead just
  show the subfields relevant to the original selection
- Add validation to ensure `enable_job_applications` cannot be changed
  when vacancy state is listed (e.g. by a user manipulating the hidden
  field)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2368

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -60,6 +60,7 @@ class Vacancy < ApplicationRecord
   paginates_per 10
 
   validates :slug, presence: true
+  validate :enable_job_applications_cannot_be_changed_once_listed
 
   before_save :on_expired_vacancy_feedback_submitted_update_stats_updated_at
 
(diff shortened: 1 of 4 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3242.london.cloudapps.digital>

app/models/vacancy.rb
```diff
@@ -126,4 +127,10 @@ def on_expired_vacancy_feedback_submitted_update_stats_updated_at
 
     self.stats_updated_at = Time.current
   end
+
+  def enable_job_applications_cannot_be_changed_once_listed
+    return unless persisted? && listed? && enable_job_applications_changed?
```
david-mears-dfe, line 132: Nice rhyme

david-mears-dfe (approved):

app/models/vacancy.rb
csutter: Now I'm thinking whether there is a word for "change" that rhymes with -isted 🤔

david-mears-dfe: twisted!

PR #3243, opened 2021-04-09
remove semantically incorrect aside tag

alexbowen:
no ticket - slipped through when i resolved rebase conflict incorrectly. part of the a11y work

the reason this is a problem is that the tags header, main, footer, aside (maybe one or two more) should be direct children of the body element. this is their intended use as per the HTML specification. To use the aside tag correctly it would mean bringing it into application.html template. this would require significant refactoring and i deem that out of scope for the value it would bring/effort involved in the immediate future


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3243.london.cloudapps.digital>

cpjmcquillan (approved):

PR #3252, opened 2021-04-12
[Teva 2087] review app block destruction until creation has completed

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2087

When a PR is closed soon after it has been created, it leads to a race condition between destroy.yml and deploy_app.yml.  
destroy.yml is likely to complete first, but runs against non-existent infrastructure because it is still being spun up.  
By the time deploy_app.yml spins up the infrastructure, there is no automated way of destroying the infrastructure subsequently.


felixtheflex:
close - pr

PR #3266, opened 2021-04-14
[TEVA-1912] Create anonymised report and reset equal opportunities on job application submission

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-1912

## Changes in this PR
When a job application is submitted, it fills in the equal opportunities report of the vacancy (mostly counters on the possible values) and resets equal opportunities attributes on the job application.

### DIFF
--- a/app/models/job_application.rb
+++ b/app/models/job_application.rb
@@ -1,5 +1,6 @@
 class JobApplication < ApplicationRecord
   before_save :update_status_timestamp, if: :will_save_change_to_status?
+  before_save :anonymise_report, if: :will_save_change_to_status?
 
   extend ArrayEnum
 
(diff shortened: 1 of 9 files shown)


app/models/equal_opportunity_report.rb
```diff
@@ -0,0 +1,3 @@
+class EqualOpportunityReport < ApplicationRecord
+  belongs_to :job_application
```
cpjmcquillan, line 2: Shouldn't this be this instead?
```ruby
belongs_to :vacancy
```

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3266.london.cloudapps.digital>

app/models/job_application.rb
```diff
@@ -44,4 +45,35 @@ class JobApplication < ApplicationRecord
...
+      end
+    end
+    report.increment(:total_submissions)
+    report.save
+  end
+
+  def anonymise_equal_opportunities_attributes
+    Jobseekers::JobApplication::EqualOpportunitiesForm::ATTRIBUTES.each do |attr|
+      attr_value = public_send(attr)
+      self[attr] = StringAnonymiser.new(attr_value)
```
csutter, line 76: When we say "anonymise" here, what we want to do is set these to `nil` – it would be too easy to reconstruct if we just run them through the `StringAnonymiser` hashing (because there are only a handful of options to begin with).

csutter (approved):
See comment about setting to nil instead of using `StringAnonymiser`.

Otherwise in terms of strategy it looks good to me 👍

PR #3271, opened 2021-04-14
[TEVA-1747] Qualifications: Part 2

david-mears-dfe:
## Jira ticket URL

## Changes

Adds the DetailsComponent and ReviewComponent for qualifications.

Complicated by the fact that grouping of qualifications into a single box happens when the qualifications share a name (e.g. GCSE), school, and year.

Minor fixes included:

- Display nothing (rather than mangled yaml) when professional_status is an empty string
on the review page

- Correct erroneous usage of 'many' instead of 'other' in I18n pluralization

### Qualifications step

Screenshot 2021-04-14 at 16 09 27
Screenshot 2021-04-14 at 16 09 50
Screenshot 2021-04-14 at 16 09 41

### Review page

Screenshot 2021-04-19 at 18 25 58

## Next

The ability to add / edit / delete multiple secondary school qualifications in one form will come in a later PR.

### DIFF
--- a/app/controllers/jobseekers/job_applications/qualifications_controller.rb
+++ b/app/controllers/jobseekers/job_applications/qualifications_controller.rb
@@ -1,4 +1,6 @@
 class Jobseekers::JobApplications::QualificationsController < Jobseekers::BaseController
+  include QualificationFormConcerns
+
   helper_method :back_path, :category, :form, :job_application, :qualification,
                 :submit_text
 
@@ -29,14 +31,15 @@ class Jobseekers::JobApplications::QualificationsController < Jobseekers::BaseCo
   end
 
   def destroy
-    qualification.destroy
-    redirect_to back_path, success: t(".success")
+    count = qualifications.count
+    qualifications.each(&:destroy)
+    redirect_to back_path, success: t(".success", count: count)
   end
 
   private
 
   def form
-    @form ||= form_class.new(form_attributes)
+    @form ||= form_class(category).new(form_attributes)
   end
 
   def form_attributes
@@ -53,32 +56,13 @@ class Jobseekers::JobApplications::QualificationsController < Jobseekers::BaseCo
   end
 
   def qualification_params
-    form_param_key = form_class.to_s.underscore.tr("/", "_").to_sym
     case action_name
     when "new", "select_category", "submit_category"
-      (params[form_param_key] || params).permit(:category)
+      (params[qualification_form_param_key(category)] || params).permit(:category)
     when "create", "edit", "update"
-      params.require(form_param_key)
+      params.require(qualification_form_param_key(category))
             .permit(:category, :finished_studying, :finished_studying_details, :grade, :institution, :name, :subject, :year)
     end
-  end
-
-  def form_class
-    name = if %w[select_category submit_category].include?(action_name)
-             "CategoryForm"
-           else
-             case category
-             when "gcse", "a_level", "as_level"
-               "Secondary::CommonForm"
-             when "other_secondary"
-               "Secondary::OtherForm"
-             when "undergraduate", "postgraduate"
-               "DegreeForm"
-             when "other"
-               "OtherForm"
-             end
-           end
-    "Jobseekers::JobApplication::Details::Qualifications::#{name}".constantize
   end
 
   def category
@@ -101,7 +85,11 @@ class Jobseekers::JobApplications::QualificationsController < Jobseekers::BaseCo
     @qualification ||= job_application.qualifications.find(params[:id])
   end
 
+  def qualifications
+    @qualifications ||= [job_application.qualifications.find(params[:ids])].flatten
+  end
+
   def submit_text
-    category.in?(%w[undergraduate postgraduate other]) ? t("buttons.save_qualification.one") : t("buttons.save_qualification.many")
+    category.in?(%w[undergraduate postgraduate other]) ? t("buttons.save_qualification.one") : t("buttons.save_qualification.other")
   end
 end
--- a/app/models/qualification.rb
+++ b/app/models/qualification.rb
@@ -1,9 +1,13 @@
 class Qualification < ApplicationRecord
+  include ActionView::Helpers::SanitizeHelper
+
   belongs_to :job_application
+
+  SECONDARY_QUALIFICATIONS = %w[gcse as_level a_level other_secondary].freeze
 
   enum category: { gcse: 0, as_level: 1, a_level: 2, other_secondary: 3, undergraduate: 4, postgraduate: 5, other: 6 }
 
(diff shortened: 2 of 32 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3271.london.cloudapps.digital>

bencmitchell:
When I have added multiple subjects to a qualification with the same school and year I was taken to a Page not found page. I assume this is because the multiple qualifications PR is still outstanding but just want to double check.
image

config/routes.rb
```diff
@@ -30,9 +30,12 @@
       resources :job_applications, only: %i[index show destroy] do
         resources :build, only: %i[show update], controller: "job_applications/build"
         resources :employments, only: %i[new create edit update destroy], controller: "job_applications/employments"
-        resources :qualifications, only: %i[new create edit update destroy], controller: "job_applications/qualifications" do
-          get :select_category, on: :collection
-          post :submit_category, on: :collection
+        resources :qualifications, only: %i[new create edit update], controller: "job_applications/qualifications" do
+          collection do
+            delete :destroy, as: :destroy
```
cpjmcquillan, line 35: What's the reason for the destroy route not being in the array on line 33?

app/models/qualification.rb
```diff
@@ -21,4 +26,24 @@ def remove_irrelevant_data
       self.year = nil
     end
   end
+
+  def finished_studying_option_with_details
```
cpjmcquillan, line 30: This feels like it probably shouldn't be in the model

app/models/qualification.rb
```diff
@@ -1,17 +1,22 @@
...
 
     I18n.t("helpers.label.jobseekers_job_application_details_qualifications_category_form.category_options.#{category}")
   end
 
-  def remove_irrelevant_data
+  def remove_inapplicable_data
+    # When `finished_studying` changes, remove answers questions that no longer apply, so that they aren't displayed
     return if finished_studying.nil?
 
     if finished_studying?
```
cpjmcquillan, line 22: I think you can lose the `self` references here and do `finished_studying_details = ""` etc instead

app/views/jobseekers/job_applications/build/qualifications.html.slim
```diff
@@ -14,14 +14,39 @@
...
     - if job_application.qualifications.any?
-      / = render Jobseekers::JobApplications::DetailComponent.with_collection(job_application.references.includes(:job_application), title_attribute: "name", info_to_display: qualifications_info)
+      - job_application.qualification_groups.each do |group|
+        = render DetailComponent.new title: group.first.title_for_group do |detail|
+          - detail.body do
+            = govuk_summary_list classes: "govuk-!-margin-bottom-0" do |c|
+              - if group.first.secondary?
+                - c.slot(:row,
+                  key: t("jobseekers.job_applications.qualifications.subjects_and_grades"),
+                  value: group.map { |qual| sanitize([qual.subject, qual.grade].join(" – ")) }.join("</br>").html_safe)
```
cpjmcquillan, line 24: Rather than `</br>` elements, can you use `safe_join` to render them in spans/divs?

app/views/jobseekers/job_applications/review/_professional_status.html.slim
```diff
@@ -9,7 +9,7 @@
     = govuk_summary_list do |component|
       - component.slot(:row,
         key: t("helpers.legend.jobseekers_job_application_professional_status_form.qualified_teacher_status"),
-        value: safe_join([tag.div(t("helpers.label.jobseekers_job_application_professional_status_form.qualified_teacher_status_options.#{job_application.qualified_teacher_status}"), class: "govuk-body"),
+        value: safe_join([tag.div(job_application.readable_qualified_teacher_status, class: "govuk-body"),
```
cpjmcquillan, line 12: Should we be passing anything to `value` at all if qualified_teacher_status is nil?

config/routes.rb
david-mears-dfe: That would create a route for destroying a single thing and expect a single id; this one will take as a param a list of records to destroy.

app/models/qualification.rb
david-mears-dfe: Agreed. Presenter?

app/views/jobseekers/job_applications/review/_professional_status.html.slim
david-mears-dfe: We should display the row, but I suppose I should pass a nil to value.

app/models/qualification.rb
cpjmcquillan: Hmmm. I still don't think we've ironed out presenter/viewcomponent/helper (s) and how we want to use each of them. 

Do you want to drop a message in the engineering chat about it?

app/models/qualification.rb
david-mears-dfe: It appears the `self` is required

app/models/qualification.rb
david-mears-dfe: Got rid of method

app/controllers/jobseekers/job_applications/qualifications_controller.rb
```diff
@@ -101,7 +85,11 @@ def qualification
     @qualification ||= job_application.qualifications.find(params[:id])
   end
 
+  def qualifications
+    @qualifications ||= [job_application.qualifications.find(params[:ids])].flatten
```
cesidio, line 89: what is the difference between:
`[job_application.qualifications.find(params[:ids])].flatten`
and just
`job_application.qualifications.find(params[:ids])`

david-mears-dfe: This happened because, when I passed in a single id, `find` returned a qualification, not an array, which broke things like `.count`.

I will fix this by passing the single id in an array, since `.find([id])` returns an array `[qualification]`.

david-mears-dfe: Changes pushed

cesidio: ah OK, now I get it. if you want to always be a collection, you can do:
`job_application.qualifications.where(id: params[:ids])`
that will always return a collection if you pass a string or an arrays, and later you can do:
`qualifications.destroy_all` 
instead of 
`qualifications.each(&:destroy)`

cesidio (approved):

app/controllers/jobseekers/job_applications/qualifications_controller.rb
david-mears-dfe: Will fix in next PR.

PR #3281, opened 2021-04-16
Prohibit LAs from creating applyable vacancies

csutter:
This removes the ability from publishers signed in as a local authority
to create vacancies that can be applied for through the service.

- Pass current organisation through to the vacancy forms
- Make the "applying for the job" view show only the fields for the
  existing value (as we would for a live vacancy), which is the online
  application link (or exceptionally, if a school itself created the
  vacancy and set it to be applyable, the personal statement, which the
  LA should still be allowed to edit in that case)
- Add a callback to `ApplyingForTheJobForm` that sets the
  `enable_job_applications` to `false` if the current organisation is
  an LA (so that a valid value is set despite the field not being there
  in the form)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2367


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3281.london.cloudapps.digital>

cpjmcquillan (approved):

PR #3287, opened 2021-04-18
ensure stylelint covers all sub-folders and fix violations

alexbowen:
no ticket

noticed in a PR the other day that something passed linting when it shouldnt have. This was due some SASS not being checked as the glob wasnt covering subfolders. This PR addresses that and fixes lint violations surfaced.

caveat - i have inline ignored:

- cookie abtest sass - once the abtest has finished only one variant will remain and hence a lot of styles will be removed and ill fix linting at that point as i would rather not touch it while the test is active
- process steps sass - these styles were largley written pre my time on project and at 600 lines of code i am not inclined to fix lint problems when this will disappear shortly anyway  to be replaced by @josephhull676 far more succint steps component

### DIFF
--- a/app/components/ab_test/cookies_banner_component/cookies_banner_component.scss
+++ b/app/components/ab_test/cookies_banner_component/cookies_banner_component.scss
@@ -1,5 +1,6 @@
 @import 'base_component';
 
+/* stylelint-disable */
 .cookies-banner-component {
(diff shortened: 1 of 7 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3287.london.cloudapps.digital>

csutter (approved):

PR #3289, opened 2021-04-18
[TEVA-2058] Print a job application

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2058

Front end code for printing application page (or potentially any page). Adds print button JS functionality and basic print stylesheet for printing application via this button or standard ctrl p way. 

Screenshot 2021-04-19 at 21 05 16

i also took the liberty while i was in JS/no JS territory of fixing a few things in other places:

- hiding elements that are for JS behaviour and make no sense being there if JS is not available for whatever reason
- some non JS positional styling was a little bit broken

### DIFF
--- a/app/components/filters_component/filters_component.html.slim
+++ b/app/components/filters_component/filters_component.html.slim
@@ -12,7 +12,7 @@
           button.govuk-link.govuk-link--no-visited-state.filters-component--show-mobile-open id="return-to-results"
             = t("shared.filter_group.return_to_results")
         - if options[:close_all]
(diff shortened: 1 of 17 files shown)


martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3289.london.cloudapps.digital>

csutter (approved):
> There is no actual application however to print on this page and i am unsure if addressing that is in scope of this ticket?

Addressing that, no, but I'm guessing the ticket's not "done" until it can be revisited once there is meaningful content on the page 🎫

alexbowen:
yes @josephhull676 is doing the application view content so i wont resolve this until i can get that code in this branch. there will be be additions to the print stylesheet as a consequence

PR #3295, opened 2021-04-19
Update cutoff date in Gdrive deletion task

csutter:
We need to prune even more old documents until we have a long-term
solution in place. This updates the deletion task to delete an extra
month's worth of documents.


cpjmcquillan (approved):
😰

martin-bangoura:
Review app deployed to <https://teaching-vacancies-review-pr-3295.london.cloudapps.digital>

PR #3331, opened 2021-04-21
[TEVA-2269] Remove secondary button cancel

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2269

## Changes in this PR:
We are removing the secondary button `Cancel` and rely on users using the browser back button to navigate back


cpjmcquillan (approved):
😍 

All the specs you removed that said `when the commit param is "Cancel"`, can you remove the context that says `when the commit param is "Something else"` but keep those specs?

PR #3333, opened 2021-04-21
Remove IDs from seed

cesidio:
## Changes in this PR:
Not sure why we were creating records with specific IDs, maybe because we had a shared algolia app so that we did not need to reindex the data


david-mears-dfe:
The ids make spec/mailers/previews/job_application_preview.rb work.

PR #3336, opened 2021-04-22
Updates to copy on sign in page when authentication fallback is switc…

RoseMackworth:
…hed on

Jira ticket: https://dfedigital.atlassian.net/browse/TEVA-2306?atlOrigin=eyJpIjoiZWVhZTVhNTczZjYwNGU0MDg0MzE3NTM1YTZmNjI5NGQiLCJwIjoiaiJ9

I've only done the half of this ticket (changing the copy in the warning box), as David advised that a developer should do the other updates in the ticket.

### DIFF
--- a/config/locales/publishers.yml
+++ b/config/locales/publishers.yml
@@ -88,7 +88,9 @@ en:
           new:
             title: Sign in to Teaching Vacancies
             heading: Temporary login
-            body: DfE Sign In is experiencing problems. Please sign in using your email address.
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
This PR was useful to do as an example of how to edit files, but I think it's cleaner to have a developer do the whole ticket, as we will need to put a line break in the paragraph, which is not possible to do by editing the `.yml` file.

PR #3341, opened 2021-04-22
[TEVA 2436] align status tags for application sections to right

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2436

below is where this has got to. i havent time tonight to do any more. i think need to discuss my comment r.e full width...

i dont feel like the default classes on this component are right either, need to sort that out

Desktop
Screenshot 2021-04-26 at 11 06 32

Mobile
Screenshot 2021-04-22 at 20 37 06

tablet
Screenshot 2021-04-26 at 10 58 00

### DIFF
--- a/app/components/review_component.rb
+++ b/app/components/review_component.rb
@@ -28,13 +28,23 @@ class ReviewComponent < GovukComponent::Base
     end
 
     def call
-      tag.h2(class: classes, **html_attributes) { safe_join([title, edit_link, content].compact) }
+      tag.div(class: classes, **html_attributes) do
+        safe_join([
+          tag.div(class: "review-component__heading__title") do
+            safe_join([
+              tag.h2(class: "govuk-heading-m") { title },
+              edit_link,
+            ].compact)
+          end,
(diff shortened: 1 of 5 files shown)


app/components/review_component.rb
```diff
@@ -28,13 +28,23 @@ def initialize(title:, text: nil, href: nil, classes: [], html_attributes: {})
...
+        safe_join([
+          tag.div(class: "review-component__heading__title") do
+            safe_join([
+              tag.h2(class: "govuk-heading-m") { title },
+              edit_link,
+            ].compact)
+          end,
+          tag.div(class: "review-component__heading__status") { content },
+        ].compact)
+      end
```
alexbowen, line 41: im sure there is a neater to way to write this maybe?

alexbowen:
sonarcloud fail doesnt seem to be to do with this PR but i have rebased against master latest

josephhull676 (approved):
Not sure about the question you left re: neatness, but looks good to me :)

Maybe worth asking someone directly about this?

PR #3357, opened 2021-04-26
Update build step call to action and redirect behaviour for job applications

cpjmcquillan:
## Changes in this PR
- Always redirect to review page when updating subsequent job application
- Update build step primary call to action based on the same criteria

## Product sign off
- [x] Looks good!

### DIFF
--- a/app/controllers/jobseekers/job_applications/build_controller.rb
+++ b/app/controllers/jobseekers/job_applications/build_controller.rb
@@ -6,24 +6,15 @@ class Jobseekers::JobApplications::BuildController < Jobseekers::BaseController
   steps :personal_details, :professional_status, :qualifications, :employment_history, :personal_statement, :references,
         :equal_opportunities, :ask_for_support, :declarations
 
-  helper_method :back_path, :form, :job_application, :qualification_form_param_key, :vacancy
+  helper_method :back_path, :form, :job_application, :qualification_form_param_key, :redirect_to_review?, :vacancy
 
   def show
     render_wizard
   end
 
   def update
-    if params[:commit] == t("buttons.save_and_come_back")
-      job_application.update(update_params)
-      redirect_to jobseekers_job_applications_path, success: t("messages.jobseekers.job_applications.saved")
-    elsif form.valid?
-      if referrer_is_finish_wizard_path?
-        job_application.update(completed_update_params)
-        redirect_to finish_wizard_path, success: t("messages.jobseekers.job_applications.saved")
-      else
-        job_application.assign_attributes(completed_update_params)
-        render_wizard job_application
-      end
+    if save_and_come_back? || form.valid?
+      update_job_application
     else
       render_wizard
     end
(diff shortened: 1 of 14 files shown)


app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -17,7 +17,7 @@ def update
       job_application.update(update_params)
       redirect_to jobseekers_job_applications_path, success: t("messages.jobseekers.job_applications.saved")
     elsif form.valid?
-      if referrer_is_finish_wizard_path?
+      if redirect_to_review?
```
alexbowen, line 20: is it worth  breaking this into a few methods? as soon as i see nested if/else/elseif type things it makes me think breaking it down worth thinking about

app/views/jobseekers/job_applications/build/ask_for_support.html.slim
```diff
@@ -28,7 +28,7 @@
           = f.govuk_text_area :support_needed_details, rows: 10
         = f.govuk_radio_button :support_needed, :no
 
-      = f.govuk_submit current_jobseeker.job_applications.not_draft.none? ? t("buttons.save_and_continue") : t("buttons.save_changes") do
+      = f.govuk_submit redirect_to_review? ? t("buttons.save") : t("buttons.save_and_continue") do
```
alexbowen, line 31: this condition is repeated a lot, can it be put in helper method?

cpjmcquillan: Unless we decide 

> Ternary operators are evil!!!

I'd be a bit averse to adding a helper method, just because I don't think it would make much difference either way. 

But if we do want to decide that as a team then I am really happy to do it!

app/controllers/jobseekers/job_applications/build_controller.rb
cpjmcquillan: I agree, but I just hadn't dedicated any brain time to thinking about doing it any differently. Let me have a think

app/views/jobseekers/job_applications/build/ask_for_support.html.slim
alexbowen: ternary operators are fine by me (unless people nest them, that is evil. and not what you are doing here). but you have 5+ instances of the same ternary operator and if they were in one place that is one peice of code to potentially maintain and im thinking DRY. its no biggie. just an observation and trying to feedback on PR

alexbowen: i like ternary operators, i think they read well.

cpjmcquillan: Are you thinking something like...

```suggestion
= f.govuk_submit some_button_text_helper do
```

app/controllers/jobseekers/job_applications/build_controller.rb
cpjmcquillan: What do you think of those changes?

app/views/jobseekers/job_applications/build/ask_for_support.html.slim
alexbowen: pretty much

app/controllers/jobseekers/job_applications/build_controller.rb
alexbowen: yeah i find it little easier to understand the flow of whats going on. potentially easier to unit test but realise its a controller so you wouldnt go there neccessarily

app/views/jobseekers/job_applications/build/ask_for_support.html.slim
alexbowen: as much as i think ternary operators are far nicer than a 5+ line if/else, a well named method is even better

cpjmcquillan: Let me know what you think.

I'm generally a bit wary of helper methods - particularly for something this simple. But I'm happy to go with this if you prefer

alexbowen: horses for courses, i prefer it

alexbowen (approved):

cesidio (approved):

PR #3359, opened 2021-04-26
[TEVA-1747] Qualifications step 3 of 3

david-mears-dfe:
## Jira ticket URL

## Changes in this PR:

The final step of 3.

Adds the ability to add/edit/delete multiple qualifications in the same page (for secondary school qualifications).

JS tests are WIP, and will (if necessary) be moved to a separate PR in order to merge this for review purposes.

## Next steps:

- [x] Product review


bencmitchell:
Add another subject looks good and works as expected. 
image
Can go in and edit, and delete as a group.

All looks good to me.

spec/system/jobseekers_can_add_qualifications_to_their_job_application_spec.rb
```diff
@@ -11,66 +11,78 @@
...
-      validates_step_complete(button: I18n.t("buttons.save_qualification.other"))
-      fill_in_secondary_qualification
-      # TODO: fill_in_another_secondary_qualification
-      click_on I18n.t("buttons.save_qualification.other")
-      expect(current_path).to eq(jobseekers_job_application_build_path(job_application, :qualifications))
-      expect(page).to have_content("Welsh Baccalaureates")
-      expect(page).to have_content("Happy Rainbows School for High Achievers")
-      expect(page).to have_content("Science – 5")
-      expect(page).to have_content("2020")
+    context "when JavaScript is required", js: true do
```
david-mears-dfe, line 52: This is because of https://ukgovernmentdfe.slack.com/archives/G0127R0KXQF/p1619434351013200

david-mears-dfe: Setting a spec to `js: true` seems make the radio button invisible.

app/frontend/src/styles/application/jobseekers/subjects-and-grades.scss
```diff
@@ -0,0 +1,57 @@
+#subjects-and-grades {
```
alexbowen, line 1: far better to use class selectors for css styling. actually far better to avoid ids for this kind of thing in general. what you have done here is pretty specific to this use case (nothing particuarly wrong with that in context of TV application), but i always like or at least try to think when im writing a front end 'components' how can i write this so it can potentially used multiple times on the same page, hence avoiding ids apart from pretty specific reasons

app/frontend/src/styles/application/govuk-error-summary.scss
```diff
@@ -0,0 +1,8 @@
+.govuk-error-summary__list {
```
alexbowen, line 1: bit nit picky but put classes on things and use that selector to apply styles. ok its pretty obvious what an li inside a selector called `__list` is but its slightly less clear for the a below. using classes for stling is a good habit to be in, thats why i mention it

app/views/jobseekers/job_applications/qualifications/edit.html.slim
```diff
@@ -1,14 +1,14 @@
...
 
-      = form_for form, url: jobseekers_job_application_qualification_path(job_application, qualification), method: :patch do |f|
-        = f.govuk_error_summary link_base_errors_to: :subject
+      = form_for form, url: jobseekers_job_application_qualifications_path(job_application, ids: qualifications.pluck(:id)), method: :patch do |f|
+        = f.govuk_error_summary
 
         = render "fields", f: f
 
-        = f.govuk_submit submit_text
+        = f.govuk_submit t("buttons.save_qualification.#{secondary? ? 'other' : 'one'}")
```
alexbowen, line 14: is this a slight misuse of other/one? im guessing you are saying that you can have many gcse's/ a levels from a school but one degree from a place? i guess most people with degrees do but that doesnt have to be the case?

spec/system/jobseekers_can_add_qualifications_to_their_job_application_spec.rb
```diff
@@ -11,66 +11,78 @@
...
-      validates_step_complete(button: I18n.t("buttons.save_qualification.one"))
-      fill_in_undergraduate_degree
-      click_on I18n.t("buttons.save_qualification.one")
-      expect(current_path).to eq(jobseekers_job_application_build_path(job_application, :qualifications))
-      expect(page).to have_content(I18n.t("buttons.add_another_qualification"))
-      expect(page).to have_content("Undergraduate degree")
-      expect(page).to have_content("University of Life")
-      expect(page).not_to have_content("Subjects and grades")
-      expect(page).not_to have_content("School, college, or other organisation")
+    context "when JavaScript is not required" do
```
alexbowen, line 14: i think required should be available?

spec/system/jobseekers_can_add_qualifications_to_their_job_application_spec.rb
alexbowen: yeah would be good to get the bottom of this as per slack thread earlier

alexbowen: but that doesnt have to be in the scope ofthis PR

app/frontend/src/styles/application/jobseekers/subjects-and-grades.scss
alexbowen: even if its unlikely you are going to use it on same page, that mantra is only going to make the code nicer. we'll go to town a bit more on that in pairing we talked about. but id like to see the id here go away tbh

app/views/jobseekers/job_applications/qualifications/edit.html.slim
david-mears-dfe: The design forces you to only be able to add degrees (and similar things) one at a time; in cases like that, I'll use the singular here. There isn't really anything we can pass to `count:` - I could pass a `1` or a `2` but that feels quite arbitrary and obscures my intention.

app/frontend/src/styles/application/govuk-error-summary.scss
```diff
@@ -0,0 +1,8 @@
...
+  li {
```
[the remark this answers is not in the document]
david-mears-dfe: I don't think I can apply classes to these `li`s because they exist inside the govuk-error-summary, which is html from a gem, not from any of our templates.

app/views/jobseekers/job_applications/qualifications/edit.html.slim
alexbowen: ok fair enough, kind of makes sense

spec/system/jobseekers_can_add_qualifications_to_their_job_application_spec.rb
david-mears-dfe: I separated the tests for journeys that require JS from journeys that don't require JS (these tests are for different journeys).

app/frontend/src/styles/application/govuk-error-summary.scss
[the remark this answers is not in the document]
alexbowen: oh i didnt appeciate it was gov component. to me seems odd they wouldnt put classes on everything contained but maybe they came to same conclusion i mentioned above thats its pretty obvious whats going to be inside something called `__list`. id still say the `<a>` is a bit more debatable but its part of the thing so cool

spec/system/jobseekers_can_add_qualifications_to_their_job_application_spec.rb
alexbowen: id say that a journey a user is going on doesnt require or not require JS but the journey maybe different depending... thats why available makes more sense to me. really not a big deal tho, up to you

david-mears-dfe: The reason I want to separate them is that the journey for adding (eg) a degree doesn't need JS, which means I can test the radio buttons, but the journey for adding many secondary school qualifications does involve JS, which means I can't test the radio buttons

app/views/jobseekers/job_applications/qualifications/fields/_secondary_school.html.slim
```diff
@@ -1,8 +1,23 @@
...
+          form_group: { classes: "govuk-!-width-one-half" }
+        = f.govuk_text_field "grade#{index + 1}".to_sym,
+          label: { text: t("helpers.label.jobseekers_job_application_details_qualifications_shared_labels.grade") },
+          aria: { required: index.zero? },
+          form_group: { classes: "govuk-!-width-one-quarter" }
+        - unless index.zero?
+          = govuk_link_to t("buttons.delete_subject"), "#", id: "delete_#{index + 1}", class: "js-action delete-button govuk-!-margin-bottom-6 govuk-!-padding-bottom-2"
 
-= f.govuk_text_field :institution, label: { size: "s" }, aria: { required: true }, width: "three-quarters"
+.add-subject
```
david-mears-dfe, line 15: accidental, will remove

app/frontend/src/application/jobseekers/manageQualifications.test.js
```diff
@@ -0,0 +1,60 @@
...
+  document.body.innerHTML = `<div id="${FIELDSET_ID}">
+  <div class="${ROW_CLASS}">
+    <label for="subject1">Subject 1</label>
+    <input class="govuk-input" id="s1" value="Economics 101">
+  </div>
+  <div class="${ROW_CLASS}"></div>
+  <a id="${SUBJECT_LINK_ID}" href="#">Add subject</a>
+</div>`;
+
+  // beforeEach(() => {
```
alexbowen, line 20: remove coomented, beforeEach not needed

alexbowen:
would be good to get the JS tested as much as possible and id like to refactor 1) renumber mehtod refactor. 2) the delete button should really be treated exactly the same of the rest of the row (inputs), i cant see benefit atm of it being treated/added diffently to other elements in arrow

spec/form_models/jobseekers/job_application/details/qualifications/secondary/common_form_spec.rb
```diff
@@ -8,7 +10,77 @@
 
   it_behaves_like "validates year format"
 
-  describe "#subject_and_grade_correspond?" do
-    # TODO: Add test when functionality is finished
+  context "when there is only one subject or grade" do
+    let(:params) { { subject1: "Clarinet" } }
```
cesidio, line 14: love this subject :)

app/controllers/concerns/qualification_form_concerns.rb
```diff
@@ -22,4 +22,6 @@ def form_class(category)
            end
     "Jobseekers::JobApplication::Details::Qualifications::#{name}".constantize
   end
+
+  alias form_param_key qualification_form_param_key
```
david-mears-dfe, line 26: note to self: move this to QualificationsController.

david-mears-dfe:
> 2) the delete button should really be treated exactly the same of the rest of the row (inputs), i cant see benefit atm of it being treated/added diffently to other elements in arrow

The reason this happens currently is that adding subjects works by copying the first row (the only row that's guaranteed to exist); and this row does not contain a delete button.

app/controllers/jobseekers/base_controller.rb
```diff
@@ -11,4 +11,9 @@ def storable_location?
   def store_jobseeker_location!
     store_location_for(:jobseeker, request.fullpath)
   end
+
+  def update_in_progress_steps!(step)
```
cpjmcquillan, line 15: What's the reason we need to do this?

app/controllers/jobseekers/job_applications/qualifications_controller.rb
```diff
@@ -49,28 +50,52 @@ def form_attributes
...
+      (qualifications + built_qualifications).map.with_index { |qualification, index|
+        attributes.merge!(qualification.slice(:subject, :grade).transform_keys { |key| key + (index + 1).to_s })
+      }.uniq
+      attributes.merge(qualification_params.to_h)
     when "create", "update", "submit_category"
       qualification_params
     end
   end
 
+  def built_qualifications
```
cpjmcquillan, line 63: I'm trying quite hard to understand this code, but it doesn't feel very readable and seems like quite a complex method

david-mears-dfe: I'll split it in two and give those methods understandable names.

app/controllers/jobseekers/base_controller.rb
david-mears-dfe: Otherwise, the review page will claim that these details steps are 'not started'.

app/controllers/jobseekers/job_applications/qualifications_controller.rb
david-mears-dfe: (Refactored)

cpjmcquillan (commented):
I think that a lot of my apprehension stems from the complexity of this entire functionality for our first attempt, and this is borne out in the code being complex and that makes it complex to review. 

But we're here now, so I guess we need to buckle up and do the best we can!

I've been trying to wrap my head around things off and on for a couple of days, but I will endeavour to approve/add some constructive comments before I do anything else today.

cpjmcquillan:
How do we feel about not adding multiple qualifications at once? As far as I can tell this is where much of the complexity stems from. This would allow us to get the feature out the door, and then iterate on the code/feature from a better starting point.

If we can hold fire on that JS functionality (and reduce the associated backend complexity), how does the code look then?

Interested to hear people's thoughts. And as I type this I see @csutter has sent a slack message in a similar vein 😂

david-mears-dfe:
> How do we feel about not adding multiple qualifications at once?

I believe this would increase user annoyance with the whole feature significantly. Imagine having to fill in 10 form steps to enter your 10 GCSEs, instead of just 1. On a bad day that might trigger me to give up applying completely.

cpjmcquillan:
> > How do we feel about not adding multiple qualifications at once?
> 
> I believe this would increase user annoyance with the whole feature significantly. Imagine having to fill in 10 form steps to enter your 10 GCSEs, instead of just 1. On a bad day that might trigger me to give up applying completely.

What if you add 1 GCSE 10 times but it feels like you added 10 GCSEs once? Is there a trade off we can manage between code complexity and responsiveness of the user experience?

david-mears-dfe:
We're going to merge in https://github.com/DFE-Digital/teaching-vacancies/pull/3379 even though there are outstanding comments on it that should be addressed, in order to keep the code together.

app/controllers/jobseekers/base_controller.rb
```diff
@@ -11,4 +11,9 @@ def storable_location?
...
+    job_application.in_progress_steps = job_application.in_progress_steps.append(step.to_s).uniq
```
david-mears-dfe, line 16: unless it's completed

david-mears-dfe:
This approach was superseded by different PRs.

PR #3367, opened 2021-04-27
Reduce jobseeker session timeout to 8 hours

csutter:
Now that we will start handling more personal data, improve security by
decreasing the amount of time jobseekers stay logged in to their account
without activity.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2529


david-mears-dfe:
Just checking - this doesn't have side-effects for hiring staff sessions, does it? As we use Devise for them too

csutter:
> Just checking - this doesn't have side-effects for hiring staff sessions, does it? As we use Devise for them too

We override it for hiring staff here: https://github.com/DFE-Digital/teaching-vacancies/blob/master/app/models/publisher.rb#L14 👍

cesidio (approved):

PR #3371, opened 2021-04-27
[TEVA-2461] Changes to Applicant Counting

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2461

## Changes in this PR:

- Removed text related to applications for scheduled jobs
- The number of applications shown now only counts submitted applications
- Added link to publisher job applications index page for jobs that have not received any applications or have only received withdrawn applications

## Comments:

- Have not done 1 and 1a (in the acceptance criteria) as this had already been done.

### No Applicants / Only Withdrawn Applicants

Screenshot 2021-04-27 at 17 06 17

### Scheduled Jobs

Screenshot 2021-04-27 at 16 13 48

### DIFF
--- a/app/components/publishers/vacancies_component.rb
+++ b/app/components/publishers/vacancies_component.rb
@@ -46,13 +46,17 @@ class Publishers::VacanciesComponent < ViewComponent::Base
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    if vacancy.job_applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: vacancy.job_applications.count),
+    after_submission_job_applications = vacancy.job_applications.after_submission
+
+    if after_submission_job_applications.any?
+      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: after_submission_job_applications.count),
                            organisation_job_job_applications_path(vacancy.id),
                            class: "govuk-link--no-visited-state")
       tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), link))
-    elsif vacancy.job_applications.none?
-      text = tag.span(I18n.t("jobs.manage.view_applicants", count: 0))
+    elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.none?
+      text = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: 0),
+                           organisation_job_job_applications_path(vacancy.id),
+                           class: "govuk-link--no-visited-state")
       tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), text))
     end
   end
@@ -79,7 +83,7 @@ class Publishers::VacanciesComponent < ViewComponent::Base
   end
 
   def include_job_applications?
-    organisation.group_type != "local_authority" && @selected_type.in?(%w[published pending expired])
+    organisation.group_type != "local_authority" && @selected_type.in?(%w[published expired])
   end
 
   def selected_scope
--- a/app/components/publishers/vacancies_component/vacancies_component.html.slim
+++ b/app/components/publishers/vacancies_component/vacancies_component.html.slim
@@ -61,7 +61,6 @@
                     = view_applicants(vacancy, card)
                     = tag.div(card.labelled_item(t("jobs.manage.closing_date"), vacancy.application_deadline))
                   - when "pending"
-                    = view_applicants(vacancy, card)
                     = tag.div(card.labelled_item(t("jobs.publication_date"), vacancy.publish_on))
                     = tag.div(card.labelled_item(t("jobs.manage.closing_date"), vacancy.application_deadline))
                   - when "expired"
--- a/app/models/job_application.rb
+++ b/app/models/job_application.rb
@@ -39,6 +39,7 @@ class JobApplication < ApplicationRecord
   has_many :references, dependent: :destroy
 
   scope :submitted_yesterday, -> { submitted.where("DATE(submitted_at) = ?", Date.yesterday) }
+  scope :after_submission, -> { where(status: %w[submitted reviewed shortlisted unsuccessful]) }
 
   def submit!
     submitted!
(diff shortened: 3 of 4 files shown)


app/components/publishers/vacancies_component.rb
```diff
@@ -46,13 +46,15 @@ def view_applicants(vacancy, card)
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    if vacancy.job_applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: vacancy.job_applications.count),
+    if vacancy.job_applications.submitted.any?
```
cpjmcquillan, line 49: The ticket doesn't make this clear, but the count should include all job applications that don't have a status of draft or withdrawn

app/components/publishers/vacancies_component/vacancies_component.html.slim
```diff
@@ -31,7 +31,7 @@
 
               = f.hidden_field :jobs_type, value: @selected_type
 
-      .moj-filter-layout__content class=grid_column_class
+      .moj-filter-layout__content class="grid_column_class"
```
cpjmcquillan, line 34: Why is this change here?

app/components/publishers/vacancies_component.rb
```diff
@@ -46,13 +46,15 @@ def view_applicants(vacancy, card)
...
+      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: vacancy.job_applications.submitted.count),
                            organisation_job_job_applications_path(vacancy.id),
                            class: "govuk-link--no-visited-state")
       tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), link))
-    elsif vacancy.job_applications.none?
-      text = tag.span(I18n.t("jobs.manage.view_applicants", count: 0))
+    elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.none?
```
david-mears-dfe, line 54: This seems wrong - if there are 55 submitted applications, and 1 withdrawn application, this will evaluate to true, and it will say '0 applications'.

app/components/publishers/vacancies_component/vacancies_component.html.slim
josephhull676: I just happened to come across it whilst I was poking around in the file. The class name should be wrapped in double quotes, no?

app/components/publishers/vacancies_component.rb
josephhull676: `if vacancy.job_applications.submitted.any?` comes first, so wouldn't it display the number of submitted applications if there are any?

app/components/publishers/vacancies_component/vacancies_component.html.slim
david-mears-dfe: This is not a class name, it's a method - I presume, otherwise it would be throwing errors.

app/components/publishers/vacancies_component.rb
david-mears-dfe: I see, OK.

david-mears-dfe: What is the difference between `elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.none?` and `else`?

app/components/publishers/vacancies_component.rb
```diff
@@ -100,4 +102,8 @@ def set_organisation_options
       OpenStruct.new({ id: organisation.id, name: "Trust head office", label: "Trust head office (#{count})" }),
     )
   end
+
+  def post_submission_job_applications(job_applications)
```
david-mears-dfe, line 106: This could be a scope - see the scopes in vacancy.rb for examples.

david-mears-dfe (commented):
What do you think, does this need any additional tests in vacancies_component_spec.rb?

app/components/publishers/vacancies_component.rb
```diff
@@ -46,13 +46,15 @@ def view_applicants(vacancy, card)
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    if vacancy.job_applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: vacancy.job_applications.count),
+    if post_submission_job_applications(vacancy.job_applications).present?
+      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: post_submission_job_applications(vacancy.job_applications).count),
```
david-mears-dfe, line 50: Minor performance improvement:

If you use an instance variable, you can avoid having to run the `post_submission_job_applications` database query twice.

That instance variable could be set in a controller helper_method or in the controller action itself.

app/components/publishers/vacancies_component.rb
```diff
@@ -43,16 +43,19 @@ def no_jobs_text
   end
 
   def view_applicants(vacancy, card)
-    return unless vacancy.enable_job_applications?
-    return unless include_job_applications?
+    return unless vacancy.enable_job_applications? && include_job_applications?
```
cesidio, line 46: The reason I made this in 2 lines is because it is more readable. unless with an and or an or is more complicated to reason in your mind

josephhull676: I felt guilty about adding another line so I put this on to one line to compensate for it lol.

I don't think the one line is more readable, so I'll change it back to what it was.

cesidio: It would be more readable if it was an `if`

app/components/publishers/vacancies_component.rb
josephhull676: `  enum status: { draft: 0, submitted: 1, reviewed: 2, shortlisted: 3, unsuccessful: 4, withdrawn: 5 }`

I guess nothing lol. Will replace with an else.

app/components/publishers/vacancies_component.rb
```diff
@@ -46,13 +46,17 @@ def view_applicants(vacancy, card)
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    if vacancy.job_applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: vacancy.job_applications.count),
+    after_submission_job_applications = vacancy.job_applications.after_submission
+
+    if after_submission_job_applications.present?
```
cesidio, line 51: when it's a collection, I prefer `any?` to `present?`, it reads better

cesidio (approved):

PR #3378, opened 2021-04-28
[TEVA-2417] Job alert email uids

cpjmcquillan:
## Jira ticket
https://dfedigital.atlassian.net/browse/TEVA-2417

## Changes in this PR
- Send a uid with all email events
- Set utm source to the uid in the job alert email links

## Food for thought
Is the uid unique enough?

## Product sign off
- [x] Looks good!

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -2,15 +2,31 @@ class ApplicationMailer < Mail::Notify::Mailer
   helper NotifyViewHelper
   helper OrganisationHelper
 
+  helper_method :uid, :utm_campaign
+
   after_action :trigger_email_event
 
   private
+
+  attr_reader :template, :to
+
+  def trigger_email_event
+    email_event.trigger(email_event_type, email_event_data)
+  end
 
   def email_event_data
     {}
   end
 
-  def trigger_email_event
-    email_event.trigger("#{email_event_prefix}_#{action_name}".to_sym, email_event_data)
+  def email_event_type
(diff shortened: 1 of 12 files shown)


app/mailers/application_mailer.rb
```diff
@@ -2,15 +2,23 @@ class ApplicationMailer < Mail::Notify::Mailer
...
+
   def email_event_data
     {}
   end
 
-  def trigger_email_event
-    email_event.trigger("#{email_event_prefix}_#{action_name}".to_sym, email_event_data)
+  def uid
+    # Time in milliseconds since epoch. We're assuming that 2 emails don't get sent in the same millisecond.
+    # If this assumption isn't safe, then we should encode some other details (e.g email_event_type) to ensure the uid is unique.
```
stevenleggdfe, line 21: @cpjmcquillan I have this nasty suspicion this might not be a safe assumption. We send a *lot* of these out each day!

stevenleggdfe: I think a combination of email address, frequency and date should always be unique to an email.

stevenleggdfe:
@cpjmcquillan theoretically I think it would be nice to get rid of utm_campaign, but I think it's useful because it's a standard way to indicate where traffic comes from across the internet at large. So it plugs in nicely to our GA data, for example. And it provides a quick way to cut the amount of data we're processing in the events table down before joining it on to email data, which saves CPU hours.

cesidio (approved):

PR #3394, opened 2021-05-04
Refactor qualifications step

csutter:
> PRing for initial feedback, to get a review app, and to go through in our session this afternoon – still needs some tests and perhaps a bit of refactoring.

Our treatment of secondary qualifications as individual qualifications
that are only summarised at display level has proven a bit of a struggle
when it came to allowing users to add/edit them. This introduces a
sub-level of "qualification results" (of which a qualification can have
many), and finishes implementation of the qualifications step for
secondary qualifications (pending some extra tests and refactoring).

- Create `QualificationResult` as an associated model of `Qualification`
  to represent subjects/grades for secondary qualifications
- Add `CommonForm#qualification_results_attributes=` to allow
  use of `fields_for` in the views
- Make use of `accepts_nested_attributes_for` to allow `Qualification`
  to handle persisting nested `QualificationResult`s once the form
  object validations have passed
- Add handling of qualification results in "Quick Apply"
- Remove handling of "grouped" qualifications everywhere (now that each
  qualification is a top-level entity)

### DIFF
--- a/app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
+++ b/app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
@@ -1,14 +1,50 @@
-class Jobseekers::JobApplication::Details::Qualifications::Secondary::CommonForm < Jobseekers::JobApplication::Details::Qualifications::QualificationForm
-  validates :institution, :year, presence: true
-  validate :subject_and_grade_correspond?
-  validates :year, format: { with: /\A\d{4}\z/.freeze }, if: -> { year.present? }
+module Jobseekers::JobApplication::Details::Qualifications::Secondary
+  class CommonForm < ::Jobseekers::JobApplication::Details::Qualifications::QualificationForm
+    MAXIMUM_NUMBER_OF_RESULTS = 6
 
-  def subject_and_grade_correspond?
-    return if subject.present? && grade.present?
+    validates :institution, :year, presence: true
+    validates :year, format: { with: /\A\d{4}\z/.freeze }, if: -> { year.present? }
+    validate :at_least_one_qualification_result
+    validate :all_qualifications_valid
 
-    errors.add(:subject, I18n.t("qualification_errors.subject_and_grade_correspond.false"))
-    # Empty string: highlight grade field with red, but rely on the user reading the error on subject.
-    # The resulting empty anchor tag in the error summary correctly links to the field.
-    errors.add(:grade, "")
+    def initialize(attributes = nil)
+      super(attributes)
+      pad_qualification_results
+    end
+
+    def qualification_results_attributes=(attrs)
+      @qualification_results ||= []
+      attrs.each do |_idx, qualification_params|
+        @qualification_results.push(QualificationResultForm.new(qualification_params))
+      end
+    end
+
+    private
+
+    def at_least_one_qualification_result
+      return if qualification_results.reject(&:empty?).any?
+
+      errors.add(:base, :at_least_one_result_required)
+
+      # Force first results form to validate so the first set of fields shows an error state on the form
+      qualification_results.first.valid?
+    end
+
+    def all_qualifications_valid
+      qualification_results.reject(&:empty?).each do |qualification|
+        next if qualification.valid?
+
+        errors.add(:base, :incomplete_result)
+      end
+    end
+
+    def pad_qualification_results
+      # Ensures the number of QualificationResults present in the form are the highest of:
+      #   - however many are already there (in case we lower the maximum number in the future), or
+      #   - the maximum number we want to have
+      # by padding the qualification results with empty objects
+      @qualification_results ||= []
+      @qualification_results += [QualificationResultForm.new] * [MAXIMUM_NUMBER_OF_RESULTS - qualification_results.count, 0].max
+    end
   end
 end
--- a/app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
+++ b/app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
@@ -5,7 +5,7 @@ module Jobseekers::JobApplication::Details::Qualifications::Secondary
     validates :institution, :year, presence: true
     validates :year, format: { with: /\A\d{4}\z/.freeze }, if: -> { year.present? }
     validate :at_least_one_qualification_result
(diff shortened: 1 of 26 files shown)


db/migrate/20210504084958_create_qualification_results.rb
```diff
@@ -0,0 +1,11 @@
+class CreateQualificationResults < ActiveRecord::Migration[6.1]
+  def change
+    create_table :qualification_results, id: :uuid do |t|
+      t.belongs_to :qualification, null: false, foreign_key: true, type: :uuid
+      t.string :subject, null: false
```
cpjmcquillan, line 5: Do we want to `default: ""`? We spoke about it before (I can't remember where 😂 ) and I don't think we reached a consensus, but we've been doing it ever since

app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
```diff
@@ -1,14 +1,50 @@
...
-    errors.add(:subject, I18n.t("qualification_errors.subject_and_grade_correspond.false"))
-    # Empty string: highlight grade field with red, but rely on the user reading the error on subject.
-    # The resulting empty anchor tag in the error summary correctly links to the field.
-    errors.add(:grade, "")
+module Jobseekers::JobApplication::Details::Qualifications::Secondary
+  class CommonForm < ::Jobseekers::JobApplication::Details::Qualifications::QualificationForm
+    MAXIMUM_NUMBER_OF_RESULTS = 6
+
+    validates :institution, :year, presence: true
+    validates :year, format: { with: /\A\d{4}\z/.freeze }, if: -> { year.present? }
```
cpjmcquillan, line 6: This probably isn't in the acceptance criteria, but can we check if we should validate the year isn't in the future? (This came up for the QTS year question eventually)

cpjmcquillan (commented):
Looks like a reasonable approach to me!

cesidio (approved):

alexbowen (approved):

db/migrate/20210504084958_create_qualification_results.rb
csutter: Hmmm 🤔  I don't _think_ so, given that having one of these records doesn't make sense without both of these fields being present.

app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
csutter: 👍  Have added a note-to-self!

db/migrate/20210504084958_create_qualification_results.rb
david-mears-dfe: Do non-secondary qualifications also have QualificationResults? If so, they need to sometimes have grades without subjects.

app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
david-mears-dfe: We allow people to list qualifications they have not finished (there is a column or two for this, finished_studying and finsished_studying_details), so they ought to be able to list qualifications that end in a future year (within limits?).

db/migrate/20210504084958_create_qualification_results.rb
csutter: For simplicity, they do not - non-secondary qualifications keep the subject and the grade on the `Qualification` record

david-mears-dfe:
Can I check I understand this approach:

- A form for secondary qualifications is related to a single Qualification and 1 or more QualificationResults
- A form for other qualifications (e.g. degree) is related to a single Qualification and 0 QualificationResults

Is that right?

csutter:
> A form for secondary qualifications is related to a single Qualification and 1 or more QualificationResults
> A form for other qualifications (e.g. degree) is related to a single Qualification and 0 QualificationResults
Yes to both ✅ 

The forms do the validation for the presence of these right now, I might add some extra validations on the `Qualification` model itself just to make sure the data integrity doesn't get messed up by anything else by accident.

app/form_models/jobseekers/job_application/details/qualifications/secondary/common_form.rb
david-mears-dfe: Actually, it's not so simple.

When we ask if a qual is finished, they can only enter a year if they said "Yes" - i.e. they have finished.

Some forms do not ask if a qual is finished (secondary quals).

Therefore, on some forms it makes sense to validate that the year is this year or earlier - if it's future, they should have selected 'No'.

Screenshot 2021-05-04 at 17 31 06

david-mears-dfe:
Screenshot 2021-05-04 at 17 35 08

This is the DetailsComponent, on the Quals step.

csutter:
Urgh. Good spot! 👀

david-mears-dfe:
Screenshot 2021-05-04 at 17 39 04

Was a conscious decision made to highlight individual fields, rather than highlight the fieldset (as per Figma/previous PR)? There is a problem with the implementation of this: clicking the base-error "Enter all subjects and grades" does not focus any field on the form - it looks for something called 'subject-field' that isn't present.

app/models/qualification.rb
```diff
@@ -32,10 +32,10 @@ def remove_inapplicable_data
   def display_attributes
     if secondary?
       %w[institution year]
+    elsif finished_studying?
+      %w[subject institution grade year]
     else
-      %w[subject institution].tap do |attrs|
-        attrs << %i[grade year] if finished_studying?
-      end
+      %w[subject institution]
```
david-mears-dfe, line 38: Subject is an optional field

csutter: Hmm, this was there before though – I think I forgot a "don't show this if" somewhere else 😞

csutter:
> Was a conscious decision made to highlight individual fields, rather than highlight the fieldset (as per Figma/previous PR)? There is an accessibility problem with the implementation of this: clicking the base-error "Enter all subjects and grades" does not focus any field on the form - it looks for something called 'subject-field' that isn't present.

Not a conscious decision, but a temporary intentional omission for the first pass at it. I definitely want to get this right before we release, and it's the next thing on my list.

It will require some massaging of the GOV.UK Form Builder though, so I wanted to keep this PR a bit smaller.

PR #3395, opened 2021-05-04
[TEVA-2488] Fix flaky spec

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2488

Not sure why but when using a randomly generated string, we had a slightly different version of it persisted, with words like `et` omitted.


alexbowen (approved):
not sure why this fixes this but willing to approve if it does! could you maybe write a sentence in description why?

PR #3398, opened 2021-05-05
[TEVA-2572] disable map mouse scroll to stop font requests

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2572

disabling mouse scroll feature stops the requests for fonts that 404. whilst doesnt find the root cause my feeling that it is more important to stop the 404 requests and noisy logging in papertrail

### DIFF
--- a/app/frontend/src/application/jobseekers/map.js
+++ b/app/frontend/src/application/jobseekers/map.js
@@ -49,6 +49,7 @@ window.initMap = () => {
 
     const map = new google.maps.Map(document.getElementById('map'), {
       zoom: 14,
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):
Reproduced on production, and unable to reproduce in the review app ✅ 

But genuinely... wtf 😂

alexbowen:
> Reproduced on production, and unable to reproduce in the review app ✅
> 
> But genuinely... wtf 😂

personally i find map mouse zoom very irritating anyway so 🤷

PR #3400, opened 2021-05-05
open govuk link in new tab

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2551


cpjmcquillan (approved):

david-mears-dfe:
GDS guidance is that `(opens in new tab)` should be part of the link itself.

PR #3405, opened 2021-05-05
[TEVA-2187] update terms conditions copy

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2187


cpjmcquillan (approved):

milimaldedfe:
All good. Think one check has failed though, not sure about that!

PR #3411, opened 2021-05-06
Refactor qualifications

david-mears-dfe:
Some small improvements left over from https://github.com/DFE-Digital/teaching-vacancies/pull/3359

- submit_text is unnecessary as a method - not worth the bother
- We no longer test the button text in request specs (as we got rid of cancel/back buttons), so it doesn't need its own line
- A forgotten width change which makes the name field align with the other fields


cpjmcquillan (approved):

csutter (approved):
💯

PR #3413, opened 2021-05-07
[TEVA-2566] Stream job application data to BigQuery (part 1)

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2566

## Changes in this PR:

Discussed these fields with Performance Analysis.

The 'params' field on the notifications table does not contain the message of the notification - it contains entire records (e.g. vacancy, job_application), so I am excluding it from the streamed data.

I still need to:

- Watch the data being received at the other end, and check that it looks correct

- In a separate PR, add code to export equal_opportunities_reports data. As this data is sensitive, and could be deduced from matching up timestamps of the updated equal_opportunities_reports and job_applications, we will need to handle it separately and differently, rather than sending equal_opportunities_reports to BigQuery whenever the report is updated (which is the same thing as 'whenever a job application is submitted'). So it will look something like triggering an event when a job closes; that event will only export the record if the job has more than [threshold] applications.

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -1,3 +1,4 @@
+# This file lists all the attributes to be exported, including those to be anonymised.
 shared:
   documents:
     - id
@@ -9,6 +10,19 @@ shared:
     - created_at
     - updated_at
     - vacancy_id
+  employments:
+    - id
+    - organisation
+    - job_title
+    - salary
+    - subjects
+    - current_role
+    - main_duties
+    - started_on
+    - ended_on
+    - job_application_id
+    - created_at
+    - updated_at
   feedbacks:
     - id
     - created_at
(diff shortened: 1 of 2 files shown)


config/analytics.yml
```diff
@@ -9,6 +10,19 @@ shared:
...
+    - salary
+    - subjects
+    - current_role
+    - main_duties
+    - started_on
+    - ended_on
+    - job_application_id
+    - created_at
+    - updated_at
   feedbacks:
```
stevenleggdfe, line 26: @david-mears-dfe are there any new fields in feedbacks introduced along with applications?

david-mears-dfe: There is, but we already accounted for it: 'application_id'.

PR #3415, opened 2021-05-07
[TEVA-2550] Remove Validation from Ask For Support Form and Add Callback to Job Application

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2550

## Comments

- Tried to do the resetting in the form model but couldn't figure out a way to make it work so put it in the job application model. Any suggestions welcome!

### DIFF
--- a/app/models/job_application.rb
+++ b/app/models/job_application.rb
@@ -1,6 +1,7 @@
 class JobApplication < ApplicationRecord
   before_save :update_status_timestamp, if: :will_save_change_to_status?
   before_save :anonymise_report, if: :will_save_change_to_status?
+  before_save :reset_support_needed_details
 
   extend ArrayEnum
 
@@ -83,5 +84,9 @@ class JobApplication < ApplicationRecord
 
   def reset_equal_opportunities_attributes
     Jobseekers::JobApplication::EqualOpportunitiesForm::ATTRIBUTES.each { |attr| self[attr] = "" }
+  end
+
+  def reset_support_needed_details
+    self[:support_needed_details] = "" if support_needed == "no"
   end
 end
--- a/spec/models/job_application_spec.rb
+++ b/spec/models/job_application_spec.rb
@@ -24,6 +24,17 @@ RSpec.describe JobApplication do
     end
   end
 
+  context "when setting support needed to 'no'" do
+    subject { create(:job_application) }
+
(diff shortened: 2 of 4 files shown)


app/models/job_application.rb
```diff
@@ -84,4 +85,8 @@ def fill_in_report
   def reset_equal_opportunities_attributes
     Jobseekers::JobApplication::EqualOpportunitiesForm::ATTRIBUTES.each { |attr| self[attr] = "" }
   end
+
+  def reset_support_needed_details
+    self[:support_needed_details] = "" if support_needed == "no"
```
cpjmcquillan, line 90: can you do this?
```ruby
support_needed_details = "" if support_needed == "no"
```

cpjmcquillan (commented):
Do you want to add a model spec perhaps?

josephhull676:
> Do you want to add a model spec perhaps?

Yeah was just doing this!

app/models/job_application.rb
cpjmcquillan: Shame

spec/models/job_application_spec.rb
```diff
@@ -24,6 +24,18 @@
     end
   end
 
+  context "when setting support needed to 'no'" do
+    subject { create(:job_application) }
```
cesidio, line 28: put some explicit value here for `support_needed_details` so it is clear that it is going to be blanked

spec/models/job_application_spec.rb
```diff
@@ -24,6 +24,18 @@
...
+
+    before do
+      subject.update(support_needed: "no")
+    end
+
+    it "resets support needed details" do
+      expect(subject.support_needed_details).to eq("")
```
cpjmcquillan, line 35: What do you think about if we did something like this instead?

```ruby
before { subject.update(support_needed: "no", support_needed_details: "something secret") }

it "..."
  expect(subject.support_needed).to eq "no"
  expect(subject.support_needed_details).to be_blank
end
```

spec/models/job_application_spec.rb
josephhull676: Sorry yeah that's obvious isn't it lol

cpjmcquillan (approved):

david-mears-dfe (approved):

PR #3417, opened 2021-05-07
Create S3 bucket for supporting documents

csutter:
This creates an S3 bucket (and IAM user and credentials) for storing
supporting documents for vacancies, and attaches the credentials to
the PaaS app.

- Create an S3 bucket per environment to store supporting documents
- Create an IAM user and access key for the app to access the bucket
- Merge bucket name and credentials into app environment
- Change service abbreviation to lowercase (not used anywhere else,
  and S3 doesn't support uppercase characters in bucket names)

## Jira ticket URL

## Changes in this PR:

## Next steps:

### DIFF
--- a/terraform/app/modules/paas/main.tf
+++ b/terraform/app/modules/paas/main.tf
@@ -25,6 +25,52 @@ resource "cloudfoundry_user_provided_service" "papertrail" {
   syslog_drain_url = var.papertrail_url
 }
 
+resource "aws_s3_bucket" "documents_s3_bucket" {
+  bucket = local.documents_s3_bucket_name
+}
+
+resource "aws_s3_bucket_public_access_block" "documents_s3_bucket_block" {
+  bucket = aws_s3_bucket.documents_s3_bucket.id
+
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
+resource "aws_iam_policy" "documents_s3_bucket_policy" {
+  name   = "${local.documents_s3_bucket_name}-policy"
+  path   = "/attachment_buckets_policies/"
+  policy = data.aws_iam_policy_document.documents_s3_bucket_policy_document.json
+}
+
+data "aws_iam_policy_document" "documents_s3_bucket_policy_document" {
+  statement {
+    actions   = ["s3:ListAllMyBuckets"]
+    resources = ["arn:aws:s3:::*"]
+    effect    = "Allow"
+  }
+  statement {
+    actions   = ["s3:*"]
+    resources = [aws_s3_bucket.documents_s3_bucket.arn]
+    effect    = "Allow"
+  }
+}
+
+resource "aws_iam_user" "documents_s3_bucket_user" {
+  name = "${local.documents_s3_bucket_name}-user"
+  path = "/attachment_buckets_users/"
+}
+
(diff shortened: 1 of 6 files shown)


terraform/app/modules/paas/main.tf
```diff
@@ -25,6 +25,52 @@ resource "cloudfoundry_user_provided_service" "papertrail" {
...
+  policy = data.aws_iam_policy_document.documents_s3_bucket_policy_document.json
+}
+
+data "aws_iam_policy_document" "documents_s3_bucket_policy_document" {
+  statement {
+    actions   = ["s3:ListAllMyBuckets"]
+    resources = ["arn:aws:s3:::*"]
+    effect    = "Allow"
+  }
+  statement {
```
saliceti, line 53: I undertstand this is for the app top read and write objects inside the bucket. Then the resources should be: `"${aws_s3_bucket.documents_s3_bucket.arn}/*"`

Also, can we restrict it to:
```
        "s3:PutObject",
        "s3:GetObject",
        "s3:DeleteObject"
```

terraform/app/modules/paas/main.tf
```diff
@@ -25,6 +25,52 @@ resource "cloudfoundry_user_provided_service" "papertrail" {
...
+
+resource "aws_iam_policy" "documents_s3_bucket_policy" {
+  name   = "${local.documents_s3_bucket_name}-policy"
+  path   = "/attachment_buckets_policies/"
+  policy = data.aws_iam_policy_document.documents_s3_bucket_policy_document.json
+}
+
+data "aws_iam_policy_document" "documents_s3_bucket_policy_document" {
+  statement {
+    actions   = ["s3:ListAllMyBuckets"]
```
saliceti, line 49: Shall we add `s3:ListBucket` to be able to list objects in the bucket?
See [ListBucket IAM permission](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html#amazons3-actions-as-permissions) and [ListObjectsV2 command](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html)

csutter: I _think_ we won't need to list objects in the bucket, so happy to not include it (`GetObject` is enough). I think `ListAllMyBuckets` is required because Terraform fails otherwise.

terraform/app/modules/paas/main.tf
```diff
@@ -25,6 +25,56 @@ resource "cloudfoundry_user_provided_service" "papertrail" {
...
```
felixtheflex, line 49: do we need ListAllMyBuckets....., this is primarily used for console access and not Cli? Presumably, there is a only one bucket in scope...?

saliceti (approved):

terraform/app/modules/paas/main.tf
```diff
@@ -25,6 +25,51 @@ resource "cloudfoundry_user_provided_service" "papertrail" {
   syslog_drain_url = var.papertrail_url
 }
 
+resource "aws_s3_bucket" "documents_s3_bucket" {
+  bucket = local.documents_s3_bucket_name
+}
+
+resource "aws_s3_bucket_public_access_block" "documents_s3_bucket_block" {
```
felixtheflex, line 32: not sure if we need this...? by default, a bucket is blocked to the public any?

felixtheflex (approved):

PR #3431, opened 2021-05-10
Correct timestamps

david-mears-dfe:
There was a bug where the notifications were displaying hour:hour instead of hour:minute for the time.

GDS style guide recommends 2:30pm over 2:30 pm so I have fixed this generally

### DIFF
--- a/app/helpers/dates_helper.rb
+++ b/app/helpers/dates_helper.rb
@@ -14,7 +14,17 @@ module DatesHelper
   def format_time(time)
     return "" if time.nil?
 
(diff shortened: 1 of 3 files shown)


app/helpers/dates_helper.rb
```diff
@@ -14,7 +14,17 @@ def format_date(date, format = :default)
   def format_time(time)
     return "" if time.nil?
```
david-mears-dfe, line 16:
```suggestion
# Testing out adding a suggestion
```

PR #3437, opened 2021-05-11
[TEVA-2566] Send equal opportunities report publication events to BigQuery

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2566

## Changes in this PR:

As this data is sensitive, and could be deduced from matching up timestamps of the updated equal_opportunities_reports and job_applications, we will need to handle it separately and differently than we do normal entity create/update/destroy events, rather than sending equal opportunities reports to BigQuery whenever the report is updated (which is the same thing as 'whenever a job application is submitted').

A new job runs every night after midnight. This works by triggering an event for every job listing that expired yesterday, but only if the listing has more than [threshold] applications, to preserve anonymity in the streamed events db.

## Next steps:

- [x] Write tests

- [ ] Check the data coming out of the other end of the pipeline

### DIFF
--- a/app/components/publishers/vacancies_component.rb
+++ b/app/components/publishers/vacancies_component.rb
@@ -46,10 +46,10 @@ class Publishers::VacanciesComponent < ViewComponent::Base
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    after_submission_job_applications = vacancy.job_applications.after_submission
+    applications = vacancy.job_applications.where(status: %w[submitted reviewed shortlisted unsuccessful])
 
-    if after_submission_job_applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: after_submission_job_applications.count),
+    if applications.any?
+      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: applications.count),
                            organisation_job_job_applications_path(vacancy.id),
                            class: "govuk-link--no-visited-state")
       tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), link))
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -51,14 +51,15 @@ class Vacancy < ApplicationRecord
 
   scope :active, (-> { where(status: %i[published draft]) })
   scope :applicable, (-> { where("expires_at >= ?", Time.current) })
-  scope :expires_within_data_access_period, (-> { where("expires_at >= ?", Time.current - DATA_ACCESS_PERIOD_FOR_PUBLISHERS) })
-  scope :expired, (-> { published.where("expires_at < ?", Time.current) })
   scope :awaiting_feedback, (-> { expired.where(listed_elsewhere: nil, hired_status: nil) })
-  scope :listed, (-> { published.where("publish_on <= ?", Date.current) })
-  scope :pending, (-> { published.where("publish_on > ?", Date.current) })
-  scope :live, (-> { listed.applicable })
-  scope :published_on_count, (->(date) { published.where(publish_on: date.all_day).count })
+  scope :expired, (-> { published.where("expires_at < ?", Time.current) })
+  scope :expired_yesterday, (-> { where("expires_at BETWEEN ? AND ?", Time.zone.yesterday.midnight, Date.current.midnight) })
+  scope :expires_within_data_access_period, (-> { where("expires_at >= ?", Time.current - DATA_ACCESS_PERIOD_FOR_PUBLISHERS) })
   scope :in_organisation_ids, (->(ids) { joins(:organisation_vacancies).where(organisation_vacancies: { organisation_id: ids }).distinct })
+  scope :listed, (-> { published.where("publish_on <= ?", Date.current) })
+  scope :live, (-> { listed.applicable })
+  scope :pending, (-> { published.where("publish_on > ?", Date.current) })
+  scope :published_on_count, (->(date) { published.where(publish_on: date.all_day).count })
 
   paginates_per 10
 
@@ -66,6 +67,8 @@ class Vacancy < ApplicationRecord
   validate :enable_job_applications_cannot_be_changed_once_listed
 
   before_save :on_expired_vacancy_feedback_submitted_update_stats_updated_at
+
+  EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD = 5
 
   def organisation
     organisation_vacancies.first&.organisation
@@ -109,6 +112,10 @@ class Vacancy < ApplicationRecord
 
   def education_phases
     organisations.map(&:readable_phases).flatten.uniq
+  end
+
+  def publish_equal_opportunities_report?
+    job_applications.after_submission.count >= EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD
   end
 
   private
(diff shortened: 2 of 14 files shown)


app/components/publishers/vacancies_component.rb
```diff
@@ -46,10 +46,10 @@ def view_applicants(vacancy, card)
     return unless vacancy.enable_job_applications?
     return unless include_job_applications?
 
-    after_submission_job_applications = vacancy.job_applications.after_submission
+    applications = vacancy.job_applications.where(status: %w[submitted reviewed shortlisted unsuccessful])
```
david-mears-dfe, line 49: This does not change the pre-existing logic of this method. I changed the definition of the `after_submission` scope in this PR to include withdrawn applications, which had been deliberately excluded here.

app/models/vacancy.rb
```diff
@@ -51,13 +51,14 @@ class Vacancy < ApplicationRecord
...
   scope :awaiting_feedback, (-> { expired.where(listed_elsewhere: nil, hired_status: nil) })
+  scope :expired, (-> { published.where("expires_at < ?", Time.current) })
+  scope :expired_yesterday, (-> { where("expires_at BETWEEN ? AND ?", Time.zone.yesterday.midnight, Date.current.midnight) })
+  scope :in_organisation_ids, (->(ids) { joins(:organisation_vacancies).where(organisation_vacancies: { organisation_id: ids }).distinct })
   scope :listed, (-> { published.where("publish_on <= ?", Date.current) })
-  scope :pending, (-> { published.where("publish_on > ?", Date.current) })
   scope :live, (-> { listed.applicable })
+  scope :pending, (-> { published.where("publish_on > ?", Date.current) })
   scope :published_on_count, (->(date) { published.where(publish_on: date.all_day).count })
-  scope :in_organisation_ids, (->(ids) { joins(:organisation_vacancies).where(organisation_vacancies: { organisation_id: ids }).distinct })
```
david-mears-dfe, line 60: The above confusion of + and - is just
- Alphabetize
- Add `expired_yesterday` scope

config/analytics.yml
```diff
@@ -23,6 +23,43 @@ shared:
     - job_application_id
     - created_at
     - updated_at
+  equal_opportunities:
+    - id
+    - vacancy_id
```
david-mears-dfe, line 28: @stevenleggdfe , I've just had the thought that we could improve code complexity while trading off quality for quantity in the analytics data, by obscuring the link between the equal opportunities reports and the vacancies.

Currently we only send you equal opportunities reports if there are >=5 applications. If we omit to send the timestamps and _id columns then it'll be next to difficult to associate the reports with the specific applicants, and you will get equal opportunities data for vacancies with <5 applicants (which might be the majority of vacancies - certainly for shortage subjects).

david-mears-dfe: The timestamps are sure to be the sticking point here - the data is pretty useless without seeing how it changes over time.

david-mears-dfe: Possibly its also essential to link the data to specific vacancies so we can do things like 'how many people of X type apply to jobs of Y type'?

stevenleggdfe: @david-mears-dfe I think that last point is key - we'll want to be able to identify particular types of schools or vacancies where there might be an issue. So I think we need to keep the relationship between vacancy and equal opportunities report. Sorry!

stevenleggdfe: Also @david-mears-dfe a side thought (probably not for this ticket!) - is there a reason why this table is stored in denormalised form in the database? Oddly, one of the first things I'll end up doing is normalising it!

david-mears-dfe: What would it look like if it was normalized?

david-mears-dfe: E.g. a table of possible genders and an id on the equal_opportunities_report 'gender' column that just points to a row in the genders table?

stevenleggdfe: id, vacancy_id, characteristic, number_of_applications instead of a separate field for each characteristic

spec/jobs/stream_equal_opportunities_report_publication_job_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+RSpec.describe StreamEqualOpportunitiesReportPublicationJob do
+  subject(:job) { described_class.perform_later }
+  let(:vacancy) { create(:vacancy) }
+  let(:publish_report?) { true }
+  let(:event_data) { {} }
+
+  before do
+    vacancy.create_equal_opportunities_report
+    allow(Vacancy).to receive_message_chain(:expired_yesterday, :listed).and_return([vacancy])
+    allow(vacancy).to receive(:publish_equal_opportunities_report?).and_return(publish_report?)
```
cpjmcquillan, line 12: what's the harm in not stubbing these things?

david-mears-dfe: If these methods break I want the test that fails to be the most useful test: the unit tests for those methods. This StreamEqualOpportunitiesReportPublicationJob shouldn't know about/depend on the internals of Vacancy.

cpjmcquillan: Hmmm this feels a bit philosophical. Generally I feel wary of stubbing things too much, besides external services. 

I'd be interested to hear the rest of the team's thoughts on it!

david-mears-dfe: @dfe-teaching-vacancies How do you feel about this?

alexbowen: correct me if i am wrong but i have viewed all tests apart from system specs as unit tests and to be true unit tests id say they should be stubbed. confused by connors comments a little `what's the harm in not stubbing these things?` -> `Generally I feel wary of stubbing things too much` as these arent services? i could be misunderstanding tho!

alexbowen: that comment probably doesnt help much - making tests effective trumps being purist imo

spec/factories/vacancies.rb
```diff
@@ -112,6 +112,16 @@
       expires_at { Faker::Time.backward(days: 6) }
     end
 
+    trait :expired_yesterday do
+      expires_on { Time.zone.yesterday.midday }
```
alexbowen, line 116: should this be same range as expired_tomorrow i.e `expires_on { Time.zone.yesterday.end_of_day }` ?

alexbowen (approved):
this looks neat and ramped up my knowledge going over it. put a query on a time range else ill approve although you might want a connor/cesi approval too ;)

cpjmcquillan (approved):
I don't think we should stub things so heavily in the spec, but I'm not going to gatekeep the code. 

We should agree an approach as a team though.

config/analytics.yml
david-mears: I'm not sure I see the difference between that and what we have.

david-mears: I see.

david-mears: @cesidio Is there a reason for us to have one column per characteristic rather than a string column called 'characteristic'?

david-mears-dfe:
I ended up taking Christian's lead on the stubbing question. https://ukgovernmentdfe.slack.com/archives/G0127R0KXQF/p1620898555003800

> I think what gives me pause in this particular instance is that this spec is setting up a factory-created "real" vacancy, but then stubbing it

PR #3440, opened 2021-05-11
Fix signing message when auth fallback is off

cesidio:


config/locales/publishers.yml
```diff
@@ -30,7 +30,8 @@ en:
           page.</li>
           </ul>
         sign_in:
-          description: Enter your email below and we will send you a link that will take you straight into your account.
+          description: You must sign in to your Teaching Vacancies account to list a job.
+          description_fallback: Enter your email below and we will send you a link that will take you straight into your account.
```
david-mears-dfe, line 34: Alternatively scope this to e.g.
```yml
sign_in:
  email:
    sessions:
      new:
        description:
```

cesidio: We are going to simply that. let's have a quick fix for now and later we will tidy up that view

PR #3446, opened 2021-05-12
[TEVA-2592] Rename job summary field as job advert

cpjmcquillan:
## Jira ticket
https://dfedigital.atlassian.net/browse/TEVA-2592

## Changes in this PR
- Rename job summary and update references

## Product sign off
- [x] Looks good!


config/locales/jobs.yml
```diff
@@ -242,6 +242,7 @@ en:
     how_to_apply: Application instructions
 
     job_summary: Job summary
```
cesidio, line 244: I guess you can remove job_summary on line 244

cesidio (approved):

config/locales/jobs.yml
cpjmcquillan: unfortunately not! We label it as "Job summay" for jobseekers 🤷

bencmitchell:
Job advert heading where required - all good 
image
Still "job summary" on the listing itself so this is good to go 👍

PR #3450, opened 2021-05-12
Set enable_job_applications to false on all published vacancies where nil

david-mears-dfe:
This was causing validation errors when updating already published vacancies with a nil enable_job_applications value.

https://ukgovernmentdfe.slack.com/archives/C016H89QSCA/p1620822285026500

These errors should not be possible for published records. For published vacancies we disallow changing the value of the field.

Screenshot 2021-05-12 at 13 57 32

Screenshot 2021-05-12 at 13 24 07

### DIFF
--- a/lib/tasks/data.rake
+++ b/lib/tasks/data.rake
@@ -1,3 +1,13 @@
+namespace :bugfix do
+  desc "Set all published vacancies' enable_job_applications attribute to false"
+  task set_enable_job_applications_false: :environment do
+    Vacancy.published.where(enable_job_applications: nil).in_batches.each_record do |vacancy|
+      vacancy.enable_job_applications = false
+      vacancy.save
+    end
+  end
+end
+
 namespace :algolia do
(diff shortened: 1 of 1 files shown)


csutter (approved):
The title gave me cause for concern (it's only published pre-launch), but the code does the right thing 👍

david-mears-dfe:
> The title gave me cause for concern (it's only published pre-launch), but the code does the right thing 👍

Updated title!

cpjmcquillan (commented):
How about a task instead of a migration? 

I know we have lots already (and I wrote some of them probably), but referencing models and attributes in migrations can cause a bit of pain down the line sometimes

david-mears-dfe:
> How about a task instead of a migration?
> 
> I know we have lots already (and I wrote some of them probably), but referencing models and attributes in migrations can cause a bit of pain down the line sometimes

Task'd

cpjmcquillan (approved):
Nice one!

PR #3452, opened 2021-05-12
Change homepage copy

RoseMackworth:
Change homepage copy to include that users can now apply for jobs as well as search for them


cpjmcquillan (approved):
🚀

PR #3464, opened 2021-05-13
[TEVA-2129] Add A:B Test for Job Alert Account Creation CTA

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2129

## Changes in this PR:

- Added variants to ab_tests.yml
- Added variants to confirm page
- Updated _jobseeker_account_prompt partial to reflect changes in designs

## Designs

https://www.figma.com/file/8GxfMTTqBsg1Cco7ffgViQ/create-account?node-id=0%3A1

## Variants

- `bottom`
- `right_blue`
- `right_grey`

### Bottom

https://user-images.githubusercontent.com/30624173/118146813-e0d83000-b406-11eb-978f-4da300cee32d.mov

### Right Blue 

https://user-images.githubusercontent.com/30624173/118146952-02391c00-b407-11eb-85c5-d6fed751eaf0.mov

### Right Grey

https://user-images.githubusercontent.com/30624173/118146981-09602a00-b407-11eb-943c-176d43c5c9ab.mov

### DIFF
--- a/app/frontend/src/styles/application/signin.scss
+++ b/app/frontend/src/styles/application/signin.scss
@@ -10,3 +10,53 @@
     color: govuk-colour('white');
   }
 }
+
+.job-alert-signin {
+  @include govuk-responsive-padding(5);
+  margin-bottom: govuk-spacing(5);
+
+  &__submit {
+    width: 100%;
+  }
+
+  &__prompt {
+    text-align: center;
+  }
+
+  .wrap-email {
+    word-break: break-all;
+  }
+
+  input[type="submit"] {
+    white-space: normal;
+    word-break: break-word;
+  }
+
+  &--blue {
+    background-color: govuk-colour('blue');
+
+    .govuk-link:not(:focus) {
+      color: govuk-colour('white');
+    }
+
+    * {
+      color: govuk-colour('white');
+    }
+
+    input[type="password"] {
+      color: govuk-colour('black');
+    }
+  }
+
+  &--grey {
+    background-color: govuk-colour('light-grey');
+
+    .govuk-link:not(:focus) {
+      color: govuk-colour('black');
+    }
+
+    input[type="submit"] {
+      color: govuk-colour('white');
+    }
+  }
+}
--- a/app/views/subscriptions/_jobseeker_account_prompt.html.slim
+++ b/app/views/subscriptions/_jobseeker_account_prompt.html.slim
@@ -4,10 +4,11 @@
     p.govuk-body = t(".body.sign_in")
 
     = form_for @jobseeker, url: jobseeker_session_path, method: :post do |f|
-      = f.govuk_email_field :email,
-        label: { text: t("helpers.label.jobseekers_sign_in_form.email"), size: "s" },
-        width: "two-thirds",
-        value: @jobseeker.email
+      .govuk-heading-s class="govuk-!-margin-bottom-2"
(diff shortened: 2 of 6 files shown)


alexbowen:
second what @david-mears-dfe  said the other day, like your habit of putting videos in PR! immediate observation on the bottom two it feels like the right hand column gets too thin before it snaps to other layout?

app/frontend/src/styles/application/signin.scss
```diff
@@ -10,3 +10,48 @@
     color: govuk-colour('white');
   }
 }
+
+.job-alert-signin {
+  @include govuk-responsive-padding(5);
+  margin-bottom: govuk-spacing(5);
+
+  &__submit {
+    width: 100%
```
alexbowen, line 19: ;

app/frontend/src/styles/application/signin.scss
```diff
@@ -10,3 +10,48 @@
...
+  }
+
+  .wrap-email {
+    word-break: break-all;
+  }
+
+  &--blue {
+    background-color: govuk-colour('blue');
+
+    .govuk-link:not(:focus) {
```
alexbowen, line 33: does this do anything when line 37 follows?

alexbowen: i guess so link isnt blue when it has :visited psuedo class, am i correct? if so 👍

alexbowen: (or the purple colour i think it is for gov links that would look very similar to background)

alexbowen:
a rebase will possibly get rid of that sonar cloud warning

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -23,7 +23,10 @@
   end
 
   describe "job alert confirmation page" do
+    let(:job_alert_account_creation_prompt_variant) { :bottom }
```
alexbowen, line 26: you can avoid changing tests by just adding a test block to ab_test.yml with one variant set to 1, if i understand why you have changed these

app/frontend/src/styles/application/signin.scss
josephhull676: Without it the link is blue. 

Screenshot 2021-05-14 at 13 43 45

I don't think this has anything to do with a :visited psuedo class? It seems, for some reason that the style from the govuk-components gem (in _links.scss I think?) appears to override the style from signing.scss

Is that correct and if so, why is this / could this be the case?

alexbowen: hmmm ok, if thats the case and the * rule below does make it white its because the rule isnt specific enough, if it was it would overide it (remember the specificty order we went through many months ago in prep for your exam. if its the case its a slight misuse of this selector but it is making it more specific. it doesnt matter too much for the purposes of this abtest, just letting you know. i have a closer look when i finished my lunch if theres a obvious alternative. and if you want me to explain what i meant about visited i can as that was the only proper reason i could think of to use it - i say 'proper' theres no right or wrong, but you get to pick up on possible little code smells like this...

alexbowen: dont neccessarily need to change this for this PR, just pointing it out

alexbowen (approved):

PR #3465, opened 2021-05-13
Remove hiring staff access to applications after 1 year

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2315

## Changes in this PR:

This implements the minimal code required for the feature, because we will not need the feature until a year from now.

If we want to at the time, we can spruce it up and, for example, tell the user why they can't view the application. Another future approach could be to remove the 'view application' link.

But for now we can save time by doing the minimal change that makes us compliant.

## Screenshots

Page that lists applications:

Screenshot 2021-05-13 at 17 48 28

Click on the application and see:

Screenshot 2021-05-13 at 17 48 50

### DIFF
--- a/app/controllers/publishers/vacancies/job_applications_controller.rb
+++ b/app/controllers/publishers/vacancies/job_applications_controller.rb
@@ -65,6 +65,10 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
   end
 
   def vacancy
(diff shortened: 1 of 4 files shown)


david-mears-dfe:
Conversation https://ukgovernmentdfe.slack.com/archives/C020RU6LX1T/p1620908134011900

PR #3471, opened 2021-05-14
Remove candidate specification fields from vacancies

cpjmcquillan:
These fields haven't been used in over a year now, so the columns should be dropped.


cesidio (approved):
🔥

PR #3472, opened 2021-05-14
[TEVA-2315] Remove publishers' access to application data after 1 year

david-mears-dfe:
## Slack conversation

https://ukgovernmentdfe.slack.com/archives/C020RU6LX1T/p1620908134011900

Related to https://dfedigital.atlassian.net/browse/TEVA-2315

## Changes in this PR:

This implements the minimal code required for the feature - does not include communicating anything to publishers using notifications/banners/emails.

We do this by scoping notifications to those which have been created in the past year, and job applications to those belonging to a vacancy that expired in the past year.

## Screenshots of UI changes:

In the vacancies dashboard:

Screenshot 2021-05-14 at 13 02 36

After clicking the link to view applications:

Screenshot 2021-05-14 at 13 02 45

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -5,6 +5,8 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { EventContext.trigger_event(:entity_updated, event_data) unless saved_change_to_attribute?(:last_activity_at) }
   after_destroy { EventContext.trigger_event(:entity_destroyed, event_data) }
 
+  DATA_ACCESS_PERIOD_FOR_PUBLISHERS = 1.year.freeze
+
   private
 
   def event_data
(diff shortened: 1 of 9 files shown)


cpjmcquillan (approved):
Nice. Can we align the names of the scopes since they check similar things? `within` seems more intuitive to me

cpjmcquillan (approved):
💥

app/models/application_record.rb
```diff
@@ -5,6 +5,8 @@ class ApplicationRecord < ActiveRecord::Base
   after_update { EventContext.trigger_event(:entity_updated, event_data) unless saved_change_to_attribute?(:last_activity_at) }
   after_destroy { EventContext.trigger_event(:entity_destroyed, event_data) }
 
+  DATA_RETENTION_PERIOD_FOR_PUBLISHERS = 1.year.freeze
```
david-mears-dfe, line 8: query the name of 'retention' - we are retaining the data.

csutter (approved):
Like I mentioned on the call, I'd maybe rename it `visibility` instead of `retention` to make it more semantically correct.

Other than that - 👍

PR #3475, opened 2021-05-14
[TEVA-2622] smoke test slack notification on failure

felixtheflex:
## Jira ticket URL

### DIFF
--- a/spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
+++ b/spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
@@ -2,7 +2,6 @@ require "capybara/rspec"
 require "i18n_helper"
 require "yaml"
 
-SERVICE_DOMAIN = "teaching-vacancies.service.gov.uk".freeze
 CLOUDAPPS_DOMAIN = "london.cloudapps.digital".freeze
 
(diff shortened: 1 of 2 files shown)


saliceti (changes requested):
We can probably simplufy the logic in the workflow and not change the smoke test

spec/smoke_tests/jobseekers_can_view_homepage_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   context "Jobseeker visits vacancy page" do
     let(:smoke_test_domain) do
-      paas_environment = !ENV.include?("PAAS_ENVIRONMENT") || ENV["PAAS_ENVIRONMENT"].empty? ? "production" : ENV["PAAS_ENVIRONMENT"]
+      paas_environment = ENV["PAAS_ENVIRONMENT"]
```
csutter, line 16: If `PAAS_ENVIRONMENT` should always be set now, it might be worth changing `ENV["PAAS_ENVIRONMENT"]` to `ENV.fetch("PAAS_ENVIRONMENT")` - this will raise an exception if it's not there and it might be easier to figure out when something goes wrong.

csutter (approved):

saliceti (approved):

PR #3476, opened 2021-05-14
Remove site_prism gem and page objects

cpjmcquillan:
We briefly experimented with the idea of using site_prism and page objects for system specs, but we found it added more overhead (both cognitively and to the code) than it was worth.


cesidio (approved):
🔥

PR #3174, opened 2021-03-26
Detect MIME type for uploads by inspecting content

csutter:
We currently rely on the browser reporting file types correctly when
uploading supporting documents for vacancies.

This changes it to detect MIME type based on content instead. It
requires the infamous `mimemagic` gem, so developers will need the
requisite libraries installed on their machine.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2271

PR #3484, opened 2021-05-17
Use deep MIME type inspection to validate uploads

csutter:
> This is attempt number 2 at #3174, this time including a workaround for
> Office 2007+ documents not being recognised, and a spec to save us from
> regressions 🙇 

We currently rely on the browser reporting file types correctly when
uploading supporting documents for vacancies.

This changes it to detect MIME type based on content instead. It
requires the infamous mimemagic gem, so developers will need the
requisite libraries installed on their machine.

- Add custom MIME types for Office 2007+ documents (this has recently
  been removed from the `mimemagic` gem, but the database we use does
  not have them in it)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2271


config/initializers/mimemagic_office_files.rb
```diff
@@ -0,0 +1,11 @@
+# Add Office 2007+ file types to MimeMagic
```
cpjmcquillan, line 1: Is this something we'll need to keep an eye on/keep up to date?

cpjmcquillan (approved):

config/initializers/mimemagic_office_files.rb
csutter: Hopefully not - this data has been static in the gem (before being removed) for a while.

If anything, it might eventually find its way into the MIME database on Alpine and we can get rid of it entirely 🤞

PR #3491, opened 2021-05-17
use gds print hide class where possible

alexbowen:
no ticket

use govuk-!-display-none-print class where possible for print application. hasnt been used on shared code i.e components

### DIFF
--- a/app/components/review_component.rb
+++ b/app/components/review_component.rb
@@ -48,7 +48,7 @@ class ReviewComponent < GovukComponent::Base
     end
 
     def edit_link
-      govuk_link_to text, href, aria: { label: "#{text} #{title}" } if text && href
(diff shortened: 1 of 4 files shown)


app/components/review_component.rb
```diff
@@ -48,7 +48,7 @@ def default_classes
     end
 
     def edit_link
-      govuk_link_to text, href, aria: { label: "#{text} #{title}" } if text && href
+      govuk_link_to text, href, aria: { label: "#{text} #{title}" }, classes: "govuk-!-display-none-print" if text && href
```
cpjmcquillan, line 51: Why is this one here?

cpjmcquillan: Well I can see why it's here once I scrolled down lol, but why was this rule in place to begin with?

cpjmcquillan (approved):

PR #3492, opened 2021-05-18
[TEVA-2642] fix layout of dashboard header

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2642

### DIFF
--- a/app/components/shared/dashboard_component/dashboard_component.scss
+++ b/app/components/shared/dashboard_component/dashboard_component.scss
@@ -9,23 +9,17 @@
 
   @include govuk-media-query ($from: desktop) {
     .govuk-width-container {
-      position: relative;
+      display: flex;
     }
 
     &__button {
-      bottom: govuk-spacing(3);
-      margin-top: 0;
-      position: absolute;
-      right: 0;
+      flex-grow: 1;
+      text-align: right;
     }
   }
 
   h2 {
     word-wrap: break-word;
-  }
-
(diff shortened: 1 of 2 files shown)


app/components/shared/dashboard_component/dashboard_component.scss
```diff
@@ -9,25 +9,19 @@
...
+      display: flex;
     }
 
     &__button {
-      bottom: govuk-spacing(3);
-      margin-top: 0;
-      position: absolute;
-      right: 0;
+      flex-grow: 1;
+      text-align: right;
```
cpjmcquillan, line 17: can you add a `margin-bottom` to the button?

alexbowen: i just pushed a different chnage for this

alexbowen: removing the top margin below brings this into alignment

cpjmcquillan: I think on the template there are margin top/bottom 0 classes, so might need to change those

alexbowen: they are needed

alexbowen: the margin top one can be removed and i removed it

cpjmcquillan (approved):

PR #3501, opened 2021-05-19
Dates refactor - Part 1

cpjmcquillan:
Rethink how we validate dates that use multiparameter attribute assignment.

## Context
- Dates are submitted as multiparameter attributes. 
- We use form models (multiparameter attribute assignment is part of ActiveRecord still, despite several efforts to migrate the API to ActiveModel). 
- This means we can't reliably use date validation gems like `validates_timeliness`

## Changes in this PR
- Update date validator to use more options
- Refactor extend deadline and job application employment forms

### DIFF
--- a/app/form_models/publishers/job_listing/extend_deadline_form.rb
+++ b/app/form_models/publishers/job_listing/extend_deadline_form.rb
@@ -1,65 +1,38 @@
 class Publishers::JobListing::ExtendDeadlineForm
   include ActiveModel::Model
   include ActiveRecord::AttributeAssignment
+  include DateAttributeAssignment
 
   EXPIRY_TIME_OPTIONS = %w[9:00 12:00 17:00 23:59].freeze
 
-  attr_accessor :expires_on, :expiry_time, :starts_on, :starts_asap, :previous_deadline
+  attr_accessor :expiry_time, :starts_asap, :previous_deadline
+  attr_reader :expires_at, :starts_on
 
-  validates :expires_on, presence: true, date: true
-  validates :starts_on, date: true, if: proc { starts_on.present? && starts_asap == "0" }
+  validates :expires_at, date: { on_or_after: :now, after: :previous_deadline }
   validates :expiry_time, inclusion: { in: EXPIRY_TIME_OPTIONS }
-
-  validate :expires_at_extended
-  validate :expires_at_in_future
-  validate :starts_on_in_future, if: proc { starts_on.present? && starts_asap == "0" }
-  validate :starts_on_after_expires_at, if: proc { starts_on.present? && starts_asap == "0" }
+  validates :starts_on, date: { on_or_after: :today, after: :expires_at }, allow_blank: true,
+                        if: proc { starts_asap == "0" }
   validate :starts_on_and_starts_asap_not_present
 
   def attributes_to_save
     {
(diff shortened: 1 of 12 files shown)


cesidio (approved):

app/views/publishers/vacancies/extend_deadline/show.html.slim
```diff
@@ -24,7 +24,7 @@
           = f.govuk_date_field :starts_on
 
           .clear-form__checkbox
-            = f.govuk_check_box :starts_asap, true, 0, multiple: false, link_errors: false
+            = f.govuk_check_box :starts_asap, "true", 0, multiple: false, link_errors: false
```
alexbowen, line 27: true => "true" seems odd, sure you have your reasons?

cpjmcquillan: If you go to that page now, check starts_asap but don't enter an expiry date (so validation fails) and submit the form, the starts_asap won't be checked, but this fixes that

app/form_models/publishers/job_listing/extend_deadline_form.rb
```diff
@@ -1,65 +1,38 @@
...
 
   EXPIRY_TIME_OPTIONS = %w[9:00 12:00 17:00 23:59].freeze
 
-  attr_accessor :expires_on, :expiry_time, :starts_on, :starts_asap, :previous_deadline
+  attr_accessor :expiry_time, :starts_asap, :previous_deadline
+  attr_reader :expires_at, :starts_on
 
-  validates :expires_on, presence: true, date: true
-  validates :starts_on, date: true, if: proc { starts_on.present? && starts_asap == "0" }
+  validates :expires_at, date: { on_or_after: :now, after: :previous_deadline }
```
david-mears-dfe, line 11: What defines `:now`?

app/validators/date_validator.rb
```diff
@@ -1,9 +1,36 @@
...
+
+      if DEFAULT_CHECK_VALUES.key?(restriction_option)
+        value_to_compare = DEFAULT_CHECK_VALUES[restriction_option]
+      elsif record.respond_to?(restriction_option)
+        value_to_compare = record.send(restriction_option)
+      end
+
+      next if value_to_compare.blank? || value_to_compare.is_a?(Hash)
+
+      raise ArgumentError, "give me something to work with!!" unless value_to_compare
```
david-mears-dfe, line 31: If value_to_compare is blank, this line will never be read, because `next` would have been triggered on line 29, right?

david-mears-dfe (approved):

app/validators/date_validator.rb
cpjmcquillan: Correct you are, good spot. 

Theres a conflict here between when a date is nil because it couldn't be cast as a date (e.g `validates :starts_on, date: { after: :expires_at }` when `expires_at` is invalid/incomplete/blank which means we don't want to do the validation, and when we pass in something that isn't valid (i.e it isn't in `DEFAULT_CHECK_VALUES` or the record doesn't `respond_to` it which means we want to raise an error.

cpjmcquillan: *updated

what do you think of that?

PR #3502, opened 2021-05-19
Change supporting document download to S3

csutter:


csutter:
WIP PR has diverged from master in the meantime, better to just try again.

PR #3505, opened 2021-05-20
reduce size of application printout

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2609

reducing font-size actually makes neglibible difference to the size of the printout and it is margins and padding that needed tweeking as much. so this ticket has become a bit more geared to the size generally rather than just font. the total size has come down by about 25% because of this PR

its a very longwinded process tweeking print stylesheet like this with no dev tools at your disposal!

before page 1
Screenshot 2021-05-20 at 15 51 51

after page 1
Screenshot 2021-05-20 at 15 51 13

### DIFF
--- a/app/frontend/src/styles/application/publishers/print-application.scss
+++ b/app/frontend/src/styles/application/publishers/print-application.scss
@@ -1,5 +1,49 @@
 @media print {
   .publishers_vacancies_job_applications_show {
+
+    .review-component__body {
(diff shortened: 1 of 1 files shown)


david-mears-dfe (approved):
Looks very much improved!

bencmitchell:
This looks much better to me!

PR #3509, opened 2021-05-21
Remove state column from vacancies

cpjmcquillan:
This field initially arose from a performance analysis need, and ended up being used for things it shouldn't have.
This need no longer exists, so we should remove the field.
This PR does that, but prioritises maintaining functionality over code quality.

This is part of a wider refactor around the create a job journey.

## Changes in this PR
- Remove vacancy state


cesidio (approved):
🔥

PR #3510, opened 2021-05-21
Refactor Active Storage migration into a job

csutter:
Our nice-and-simple approach didn't work, so this makes it a bit more
"enterprise".

- Move migration of documents into a job, and change the original Rake
  task to simply kick off one job per document
- Add integrity checks in the job to make sure we've actually
  successfully migrated the document
- Add a new Rake task `verify_migration` to verify documents and
  supporting documents match after migration
- Add `EventContext#suppress_events!` to make sure we don't fire events
  when all we're doing is programatically updating something

### DIFF
--- a/app/models/event_context.rb
+++ b/app/models/event_context.rb
@@ -1,8 +1,19 @@
 class EventContext < ActiveSupport::CurrentAttributes
   attribute :request_event
+  attribute :events_suppressed
+
+  # Stop sending events in the current context for the duration of a block
+  # (useful for situations where records are updated programatically, e.g. in background jobs,
+  # and we don't want to send events)
+  def suppress_events
+    self.events_suppressed = true
+    yield
+    self.events_suppressed = false
+  end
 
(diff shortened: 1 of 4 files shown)


app/models/event_context.rb
```diff
@@ -1,8 +1,16 @@
 class EventContext < ActiveSupport::CurrentAttributes
   attribute :request_event
+  attribute :events_suppressed
+
+  # Stop sending events in the current context (useful for e.g. background jobs where records are
+  # updated programatically)
+  def suppress_events!
+    self.events_suppressed = true
```
david-mears-dfe, line 8: How does it know only to suppress events in the background job?

csutter: The background job (or anywhere else that you would want to not send events) explicitly calls `suppress_events!`.

But I've found an issue with it that I'm about to fix that will also make this clearer (hopefully) 🤞

cpjmcquillan (approved):

PR #3513, opened 2021-05-21
[TEVA-2639]-Deploy to Production using tagging

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2639

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,7 +17,8 @@ jobs:
 
   build:
     name: build image
-
+    outputs:
+      build-tag: ${{steps.tag_version.outputs.new_tag}}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -98,6 +98,12 @@ jobs:
       run: |
         echo "TAG=${{ github.sha }}" >> $GITHUB_ENV
 
+    - name: Bump version and push tag
```
saliceti, line 101: If this is inside the matrix, it will generate a new tag for each environment. Can it be moved to the build job and the tag provided as output?

saliceti (approved):

PR #3524, opened 2021-05-24
[TEVA-2375] Use Steps Component in Job Listing Journey

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2375

## Changes in this PR:

- Added steps partial to render the steps component on multiple pages along the job listing journey
- Removed process steps component


app/services/process_steps.rb
```diff
@@ -1,4 +1,6 @@
 class ProcessSteps
+  attr_reader :steps
```
cpjmcquillan, line 2: Can you get rid of this service?

josephhull676: Are you asking me to remove it or are you asking for my opinion? lol

cpjmcquillan: Both 😂 . 

I expected that it would get deleted, so I'm curious whether we wanted to and we couldn't (and the why), or whether we hadn't thought about it!

app/views/publishers/vacancies/build/_steps.html.slim
```diff
@@ -0,0 +1,10 @@
+- if current_organisation.is_a?(School)
+  = render StepsComponent.new title: t("publishers.vacancies.review.steps"), classes: "govuk-!-margin-top-6" do |component|
+    - school_vacancy_steps(process_steps).each do |vacancy_step, step_number|
+      - component.step(label: t("publishers.vacancies.build.#{vacancy_step}.step_title"), current: (step_number == process_steps.current_step_number), completed: vacancy_step_completed?(vacancy, step_number))
```
alexbowen, line 4: how is this even working/tests passing with this commented out?

alexbowen: sorry this is github formatting, ignore this!

app/services/process_steps.rb
david-mears-dfe: Looks like not only the ProcessStepsComponent depends on it, but also VacancyFormPageHeadingComponent.

david-mears-dfe: Oh, and StepsComponent uses it too.

cpjmcquillan: The StepsComponent doesn't use it? 

I don't think the fact we use it in the heading component means that it needs to exist. A service to subtract a number seems a bit overkill. 

We could have a simple helper method instead if we didn't want to have the `current_organisation.school? ? number : number + 1` logic in a template.

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -3,7 +3,7 @@
 class Publishers::Vacancies::BaseController < Publishers::BaseController
   include Publishers::Wizardable
 
-  helper_method :process_steps, :step_current, :steps_adjust, :vacancy
+  helper_method :process_steps, :step_current, :steps_adjust, :steps_config, :vacancy
```
david-mears-dfe, line 6: I can't see where this new helper_method gets used. Is it for the methods you added to VacanciesHelper?

josephhull676: Yes!

app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -15,7 +15,7 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
                               application/vnd.openxmlformats-officedocument.presentationml.presentation ].freeze
   FILE_SIZE_LIMIT = 10.megabytes
 
-  helper_method :form
+  helper_method :form, :step
```
david-mears-dfe, line 18: This makes a private method public, as far as I can tell. Maybe that private method should be public now?

Again, I can't see where it's being used.

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -37,6 +37,7 @@ def review
     end
 
     @vacancy = VacancyPresenter.new(vacancy)
+    @step = :review
```
david-mears-dfe, line 40: It looks like the `step_current` method takes care of this:

```ruby
def step_current 
  defined?(step) ? step : :review
end
```

So this line can be rewritten:

https://github.com/DFE-Digital/teaching-vacancies/blob/c97d938b16d4bbc441086c00225f8e77caf95aee/app/views/publishers/vacancies/build/_steps.html.slim#L5

app/views/publishers/vacancies/build/_steps.html.slim
```diff
@@ -0,0 +1,10 @@
+- if current_organisation.is_a?(School)
```
david-mears-dfe, line 1: Is the only difference between these two branches `school_vacancy_steps` vs `school_group_vacancy_steps`? It seems like too much repetition.

app/views/publishers/vacancies/review.html.slim
```diff
@@ -41,4 +41,4 @@
       = govuk_link_to t("buttons.back_to_manage_jobs"), jobs_with_type_organisation_path("draft", from_review: @vacancy.id), class: "govuk-link--no-visited-state"
 
     .govuk-grid-column-one-third
-      = render(Shared::ProcessStepsComponent.new(process: @vacancy, service: process_steps, title: t("publishers.vacancies.build.process_title")))
+        = render "publishers/vacancies/build/steps"
```
david-mears-dfe, line 44: Extra indent?

app/controllers/publishers/vacancies/base_controller.rb
david-mears-dfe: I think it makes more sense to pass in `steps_config` as an argument to the VacanciesHelper methods (similar to the other helper methods), because then (a) it's clear what info the helpers depend on, and (b) the vacancies helper module should be able to slot in anywhere (in other words, it should be 'modular'), without dragging along a hidden dependency. One hidden dependency is tolerable, but if it gets more over time it will become 'spaghetti code' that's difficult to change without causing more changes elsewhere.

app/controllers/publishers/vacancies/documents_controller.rb
josephhull676: I think it's a hangover from another approach I took. I'll remove it.

app/helpers/vacancies_helper.rb
```diff
@@ -90,4 +90,17 @@ def vacancy_school_visits_hint(vacancy)
...
+  def vacancy_step_completed?(vacancy, step_number)
+    return false if vacancy.completed_step.nil?
+
+    step_number <= vacancy.completed_step
+  end
+
+  def steps_to_display(steps)
+    steps_to_remove = current_organisation.is_a?(School) ? %i[supporting_documents review] : %i[supporting_documents schools review]
+
+    steps.transform_values { |step_details| step_details[:number] - steps_adjust }
```
david-mears-dfe, line 103: I think steps_adjust should be an argument passed in too, for the same reason as https://github.com/DFE-Digital/teaching-vacancies/pull/3524/files#r639838273

app/helpers/vacancies_helper.rb
```diff
@@ -90,4 +90,17 @@ def vacancy_school_visits_hint(vacancy)
...
+         .reject { |step_name, step_number| steps_to_remove.include?(step_name) || step_number.zero? }
```
david-mears-dfe, line 104: I believe you can more simply do

```ruby
.except(steps_to_remove).reject { |step_name, step_number| step_number.zero? }
```

app/views/publishers/vacancies/build/_steps.html.slim
```diff
@@ -0,0 +1,4 @@
+= render StepsComponent.new title: t("publishers.vacancies.review.steps"), classes: "govuk-!-margin-top-6" do |component|
+  - steps_to_display(steps_config).each do |vacancy_step, step_number|
+    - component.step(label: t("publishers.vacancies.build.#{vacancy_step}.step_title"), current: (vacancy_step == step_current), completed: vacancy_step_completed?(vacancy, step_number))
+  - component.step(label: t("publishers.vacancies.review.heading"), current: step_current == :review, completed: false)
```
david-mears-dfe, line 4: `completed: false`

Is this partial shown if we go to edit an already published job listing?

In that case should the review step be shown as complete?

david-mears-dfe (commented):
Optionally (I am being demanding):

You could turn the steps partial into a view component called something like `VacancyStepsComponent`, because we like view components, because they are easy to unit test. It would only need the arguments `steps_config` and `current_step`. The methods you've added to the vacancies helper would live inside the view component (and so can `steps_adjust` I think), and that would improve the organisation of the code (because you'd have collected together the methods responsible for a single thing).

app/helpers/vacancies_helper.rb
josephhull676: Ah that's much cleaner!

josephhull676:
> Optionally (I am being demanding):
> 
> You could turn the steps partial into a view component called something like `VacancyStepsComponent`, because we like view components, because they are easy to unit test. It would only need the arguments `steps_config` and `current_step`. The methods you've added to the vacancies helper would live inside the view component (and so can `steps_adjust` I think), and that would improve the organisation of the code (because you'd have collected together the methods responsible for a single thing).

Perhaps this could be done in another PR?

app/views/publishers/vacancies/build/_steps.html.slim
josephhull676: It is shown if we edit an already published job listing, but it shouldn't be. I'll change that now.

cesidio (approved):

PR #3529, opened 2021-05-25
[TEVA-2674] Show dashboard link when there are draft applications

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2674

### Before
Screenshot 2021-05-25 at 11 40 33

### After
Screenshot 2021-05-25 at 11 40 08

### DIFF
--- a/app/components/publishers/vacancies_component.rb
+++ b/app/components/publishers/vacancies_component.rb
@@ -48,17 +48,10 @@ class Publishers::VacanciesComponent < ViewComponent::Base
 
     applications = vacancy.job_applications.where(status: %w[submitted reviewed shortlisted unsuccessful])
 
-    if applications.any?
-      link = govuk_link_to(I18n.t("jobs.manage.view_applicants", count: applications.count),
-                           organisation_job_job_applications_path(vacancy.id),
-                           class: "govuk-link--no-visited-state")
-      tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), link))
-    elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.none?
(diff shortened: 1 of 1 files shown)


app/components/publishers/vacancies_component.rb
```diff
@@ -53,7 +53,7 @@ def view_applicants(vacancy, card)
                            organisation_job_job_applications_path(vacancy.id),
                            class: "govuk-link--no-visited-state")
       tag.div(card.labelled_item(I18n.t("jobs.manage.applications"), link))
-    elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.none?
+    elsif vacancy.job_applications.withdrawn.any? || vacancy.job_applications.draft.any? || vacancy.job_applications.none?
```
cpjmcquillan, line 56: It's only the count that's different isn't it? Could we rewrite the method to make it a bit simpler?

david-mears-dfe: I'm confused by it - why isn't it equivalent to `else`?

cesidio: code is simplified now

app/components/publishers/vacancies_component.rb
```diff
@@ -47,18 +47,12 @@ def view_applicants(vacancy, card)
     return unless include_job_applications?
 
     applications = vacancy.job_applications.where(status: %w[submitted reviewed shortlisted unsuccessful])
```
david-mears-dfe, line 49: Could you do the `.count` on this line to save space?

cpjmcquillan (approved):

PR #3530, opened 2021-05-25
[TEVA-2682] Allow `&` character to be in title and salary field

cesidio:
## Jira ticket
https://dfedigital.atlassian.net/browse/TEVA-2682

## Screenshots of UI changes:
image
image

### DIFF
--- a/app/form_models/publishers/job_listing/copy_vacancy_form.rb
+++ b/app/form_models/publishers/job_listing/copy_vacancy_form.rb
@@ -10,7 +10,8 @@ class Publishers::JobListing::CopyVacancyForm < Publishers::JobListing::Importan
   private
 
   def job_title_has_no_tags?
-    return if job_title == sanitize(job_title, tags: [])
+    job_title_without_escaped_characters = job_title.delete("&")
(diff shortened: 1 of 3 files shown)


david-mears-dfe (commented):
Did you try this? From the documentation of `sanitize`:

```
 To set the default allowed tags or attributes across your application:  
  # In config/application.rb
  config.action_view.sanitized_allowed_tags = ['strong', 'em', 'a']   
  config.action_view.sanitized_allowed_attributes = ['href', 'title']
  ```

cesidio:
> Did you try this? From the documentation of `sanitize`:
> 
> ```
>  To set the default allowed tags or attributes across your application:  
>   # In config/application.rb
>   config.action_view.sanitized_allowed_tags = ['strong', 'em', 'a']   
>   config.action_view.sanitized_allowed_attributes = ['href', 'title']
> ```

The problems is not with tags or attributes, the problem is with the special character `&` that gets escaped. All we are doing here is to check that the content does not have tags in order to present the error. We are not stripping or sanitising the actual content

david-mears-dfe (approved):

PR #3536, opened 2021-05-26
Add to qualification system specs

david-mears-dfe:
Covers the editing of qualification results:
- Editing a shared attribute like an institution
- Editing an individual result
- Removing an individual result by leaving the fields blank


app/controllers/jobseekers/job_applications/qualifications_controller.rb
```diff
@@ -49,7 +49,7 @@ def form_attributes
     when "edit"
       qualification
         .slice(:category, :finished_studying, :finished_studying_details, :grade, :institution, :name, :subject, :year, :qualification_results)
-        .reject { |_, v| v.blank? }
+        .reject { |_, v| v.blank? && v != false }
```
david-mears-dfe, line 52: `false.blank?` returns `true` so the edit pages were missing the `finished_studying` attribute.

cesidio (approved):

PR #3538, opened 2021-05-26
[TEVA-2600] add trix editor for create vacancy textfields

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2600

edit/create vacancy
Screenshot 2021-05-28 at 13 18 36

vacancy listing
Screenshot 2021-05-28 at 13 07 45

(mili requsted the section title are bigger so they standout from user inputed bold text)


cpjmcquillan (commented):
I think we index a truncated job advert field in algolia. We should probably strip out all the tags when we do that.

But looking awesome!

Done 👍

david-mears-dfe:
Is it possible to limit the use of headings, so that they don't overshadow our own headings?

alexbowen:
> Is it possible to limit the use of headings, so that they don't compete with/outdo our own headings?

not really to limit them. i could remove the text size change button though might be good idea. but to a degree going to have to trust people not to do stuff to make their listing look shite!

PR #3560, opened 2021-06-02
Update README.md

david-mears-dfe:
I'm using Connor's old laptop now and going through the onboarding process.


cesidio (approved):
I am OK with adding this, the only thing is that there are a lot of things (too many in fact) that needs to be sorted out when installing tools. We say that the app needs Postgres, but installing it and running it properly requires some knowledge and many instructions that change and need to be figured out. I leave this to you

PR #3564, opened 2021-06-02
Re-order and update README

david-mears-dfe:
AWS credentials don't work until you run `bundle` so I moved that earlier.

Added suggestions for using homebrew to install postgres.

GIAS import is no longer required since orgs are created by seeding the db, so I reordered and edited the import data section.


david-mears-dfe:
Actually something weird is happening - when I try to sign in locally DSI forwards me to dev.teaching-vacancies rather than localhost. So I might be wrong about the GIAS import task.

cesidio (approved):

PR #3565, opened 2021-06-02
Remove job application feature flag

cesidio:


Gemfile.lock
```diff
@@ -337,8 +337,6 @@ GEM
     multi_json (1.15.0)
     multi_xml (0.6.0)
     multipart-post (2.1.1)
-    net-http-persistent (4.0.1)
```
alexbowen, line 340: just checking this is intentional? i know you were playing about with gemfile earlier?

cesidio: it is intentional, even though it is not relevant to this PR, it should have been part of the previous one when I updated the gems. No point in doing a PR only for that

alexbowen (approved):

PR #3568, opened 2021-06-03
[TTEVA-2071] use gds bottom sticky variant of cookies banner

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2071

- cookie banner variant chosen from abtest taken forward
- cookie banner added to skiplinks
- refactoring to make use of the govuk component (exact same markup was used for the test however) and also to allow us to test this like we do other components


app/components/cookies_banner_component.rb
```diff
@@ -0,0 +1,21 @@
+class CookiesBannerComponent < GovukComponent::Base
```
david-mears-dfe, line 1: Is this file just the deleted `app/components/ab_test/cookies_banner_component.rb` but moved?

alexbowen: yes and a few stuctural bits added for default component tests (html_attrs for e.g)

david-mears-dfe: I thought most IDEs were capable of recording when files are renamed/moved. It makes git history easier to understand. I think you can also use `git mv hello.html directory` instead of `mv hello.html directory` to make git history show that it's a moved file rather than two files that were deleted/added.

No worries for this PR but it is a nice-to-have for the future.

app/components/cookies_banner_component.rb
```diff
@@ -0,0 +1,21 @@
...
+    super(classes: classes, html_attributes: html_attributes)
+
+    @create_path = create_path
+    @reject_path = reject_path
+    @preferences_path = preferences_path
+  end
+
+  private
+
+  def default_classes
```
david-mears-dfe, line 18: Where is default_classes used?

app/components/cookies_banner_component.rb
alexbowen: you are absolutely right and I even mentioned this not do long ago in an eng workshop. It did cross my mind just after I created Pr but I came to conclusion the original only had 2 updates for abyest and in fact weren't losing isting anything useful in term of history

app/components/cookies_banner_component.rb
alexbowen: It is something that extending from gov component gives and just classes that are applied to component container

david-mears-dfe (approved):

PR #3575, opened 2021-06-04
Update monitoring to alert at >10% failed requests/5 mins

david-mears-dfe:
We found that 5% became very noisy, and made us blind to monitoring alerts. Even in relative terms, one or two failed requests at times make the difference between above 5% and below.

https://ukgovernmentdfe.slack.com/archives/CP987RP6J/p1622117155040100

### DIFF
--- a/terraform/monitoring/config/alert.rules.yml
+++ b/terraform/monitoring/config/alert.rules.yml
@@ -3,11 +3,11 @@ groups:
   rules:
   - alert: ProdRequestsFailuresElevated
     # Condition for alerting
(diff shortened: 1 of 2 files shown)


david-mears-dfe:
This has a failing test which will be addressed next week
https://ukgovernmentdfe.slack.com/archives/CP987RP6J/p1622811050020100?thread_ts=1622117155.040100&cid=CP987RP6J

alexbowen (approved):

felixtheflex (approved):

PR #3586, opened 2021-06-08
Remove modal document delete dialogue

csutter:
We don't *need* to get confirmation from a publisher before removing
a supporting document from a vacancy. This allows us to get rid of
some nasty legacy JS code and just use a vanilla `link_to`.

In the future, we might want to improve on this even further and
remove the need for UJS by using a button.


cesidio (approved):
🔥

alexbowen (approved):

PR #3588, opened 2021-06-08
Update Brakeman and previously blocked dependencies

csutter:
Brakeman had a bug prior to 5.0.2 where it misparsed the version of
`loofah`. Now that's fixed, we can bump our dependencies that failed
upgrading earlier.


alexbowen:
So am i right with this change that we could update those component and formbuilder gems? (they were both failing because of the loofah dependency)

alexbowen (approved):

csutter:
Yes - although the component gem is causing an unrelated issue with deprecation warnings that I will need to look into 😞

PR #3602, opened 2021-06-09
[TEVA-2709] Increase fountainhead/action-wait-for-check timeout to 840

felixtheflex:
Following recent changes, we observed that most often than not,
the fountainhead/action-wait-for-check actions timeout while waiting
for the corresponding Deploy App to Environment to complete. Hence the
need for the increase in timeout.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2709

## Changes in this PR:

.github/workflows/build_and_deploy.yml

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -178,7 +178,7 @@ jobs:
         token: ${{ secrets.GIT_HUB_SERVICE_ACCOUNT_TOKEN }}
         checkName: Deploy ${{ env.DOCKER_IMAGE_TAG }} to ${{ env.ENVIRONMENT }} # Job name within triggered workflow
         ref: ${{ env.BUILD_GIT_TAG }}
-        timeoutSeconds: 720
+        timeoutSeconds: 1200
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -178,7 +178,7 @@ jobs:
         token: ${{ secrets.GIT_HUB_SERVICE_ACCOUNT_TOKEN }}
         checkName: Deploy ${{ env.DOCKER_IMAGE_TAG }} to ${{ env.ENVIRONMENT }} # Job name within triggered workflow
         ref: ${{ env.BUILD_GIT_TAG }}
-        timeoutSeconds: 720
+        timeoutSeconds: 840
```
saliceti, line 181: Could we bump it even higher to guarantee success? Like 1200 (20 min)

saliceti: Also could you fix the task id? Noticed it's still called "wait_for_deploy_app_staging"

saliceti (approved):

PR #3607, opened 2021-06-09
refactor of filters front end code (removing moj dependency)

alexbowen:
no ticket

@csutter bought up the MOJ frontend dependency we have. And off the back of it this PR has given me great pleasure

## Changes in this PR:

- it turns out MOJ was purely for a specific part of the filters UI and it has been on the radar to remove the dependency for sometime and this PR removes it. removes ~70k of unminified CSS that we were using for 2 or 3 style rules.
- talked to @jesseyuen about a few filters related bits and i have refactored a heading that was being used as a button on mobile. now it is a button on mobile and h3 on desktop.
- tightened up the filters JS unit tests so the tests are a bit more tied to to the script it is testing in terms of HTML selectors used otherwise false positives could creep in.

there is no before and after as they should be exactly the same. the one area changed as described above follows

(ignore the showing x results of ... part which looks nasty. again mentioned this to @jesseyuen but i will adress that in separate PR as its not really anything to do with filters)

### Before
Screenshot 2021-06-09 at 15 35 04

Screenshot 2021-06-09 at 15 40 48

### After
Screenshot 2021-06-09 at 15 56 15

Screenshot 2021-06-09 at 15 40 39

### DIFF
--- a/app/components/filters_component/filters_component.scss
+++ b/app/components/filters_component/filters_component.scss
@@ -32,6 +32,7 @@
 
         .filters-component--show-mobile-open {
           display: block;
+          top: 6px;
         }
 
         .filters-component__groups {
@@ -46,17 +47,25 @@
           .filters-component__heading-applied {
             display: none;
           }
+
+          .govuk-heading-s {
+            display: block;
+          }
+
+          .govuk-button {
+            display: none;
+          }
         }
       }
     }
   }
 
-  .govuk-link {
+  .filters-button {
     background: none;
     border: 0;
     color: $govuk-link-colour;
     cursor: pointer;
-    font-size: 1rem;
+    font-size: 0.9rem;
     padding: 0;
     position: absolute;
     right: 0;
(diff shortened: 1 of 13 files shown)


csutter (approved):
Nice one! 👏

alexbowen:
thanks @csutter think there some specs that still use some moj selectors ill look at now...

jesseyuen:
I'd like to suggest a copy change here @alexbowen. The new filter button doesn't actually do the filtering so I'm suggesting we use `Show filters (1 applied)` instead of the `Filter results` which suggests the button does the filtering.

alexbowen:
> I'd like to suggest a copy change here @alexbowen. The new filter button doesn't actually do the filtering so I'm suggesting we use `Show filters (1 applied)` instead of the `Filter results` which suggests the button does the filtering.

i put screenshot i did make that change @jesseyuen

app/components/filters_component/filters_component.scss
```diff
@@ -225,13 +257,13 @@
 }
 
 // misc (for dashboard)
-.moj-filter-sidebar {
-  &.moj-filter-sidebar__hidden {
+.moj-remove-filter-sidebar {
```
alexbowen, line 260: forgot about these too..

PR #3609, opened 2021-06-09
Anonymise vacancy ids in request events

david-mears-dfe:
In order to match up the events with records, we need to use the anonymised version.

Also:

Allow click_events to have further event data (from a click_event_data param)
to include in the event. The use case here is to match up a 'save vacancy' click
event with the vacancy in question. We have to explicitly permit such parameters
in order to convert the click_event_data param into a hash for consumption by
Event.

### DIFF
--- a/spec/requests/documents_spec.rb
+++ b/spec/requests/documents_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe "Documents" do
     it "triggers a `vacancy_document_downloaded` event" do
       expect { get document_path(document) }
         .to have_triggered_event(:vacancy_document_downloaded)
-        .and_data(vacancy_id: vacancy.id, document_id: document.id, filename: document.name)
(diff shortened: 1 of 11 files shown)


spec/requests/documents_spec.rb
```diff
@@ -14,7 +14,7 @@
     it "triggers a `vacancy_document_downloaded` event" do
       expect { get document_path(document) }
         .to have_triggered_event(:vacancy_document_downloaded)
-        .and_data(vacancy_id: vacancy.id, document_id: document.id, filename: document.name)
+        .and_data(vacancy_id: anything, document_id: anything, filename: document.name)
```
csutter, line 17: I think `anonymised_form_of(vacancy.id)` should do the trick here!

csutter (approved):
Looks good - although you can make the `anything`s more specific I think (see comment) 👍

cesidio (approved):

PR #3614, opened 2021-06-10
refactor search header to fix the broken mobile look

alexbowen:
no ticket

- this fixes the slightly broken look of the search header on mobile after chat with @jesseyuen
- also noticed a little visual defect on publisher dashboard as a  consequence of work i did yesterday and this is fixed here too
- mixed in with this are some sass lint fixes as i noticed that STILL the lint tool wasnt covering everything -sorry should have been another PR but here it is anyway

before

before-desktop
before2

after

after-desktop
after


app/frontend/src/styles/application/jobseekers/search-header.scss
```diff
@@ -0,0 +1,33 @@
+.search-results__header {
+  display: flex;
+  flex-direction: column;
+
+  @include govuk-media-query($from: tablet) {
+    flex-direction: row;
+  }
+
+  &-stats {
```
david-mears-dfe, line 9: That's an interesting language feature I wasn't aware of (joining a class name with `&` and `-`).

david-mears-dfe (approved):

app/frontend/src/styles/application/jobseekers/search-header.scss
alexbowen: can explain exactly what it does if you want - the &. the - is not a feature its part of the class name

PR #3634, opened 2021-06-14
[TEVA-2556] Refactor the conversion of Devise flash variants

david-mears-dfe:
# Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2556

# Changes in this PR:

Previously, we used methods such as 'replace_devise_notice_flash_with_success' and 'replace_devise_alert_flash_with_notice' on particular actions in order to convert flash messages originating from Devise into the variant we wanted (alert, notice, success variants).

This was unexpected, from the point of view of Devise's internals, and resulted in [odd behaviour](https://dfedigital.atlassian.net/browse/TEVA-2556) such as flash messages persisting for one extra page load. Devise expected certain messages to be stored in certain flash variants.

Therefore, instead of manipulating the variant of the flash itself, I have added some logic that alters the 'variant' parameter passed to FlashComponent, mapping all Devise alert flashes to pass 'notice' as the variant to FlashComponent, and
mapping 'notice' to 'success'. This preserves the same variants as we were already using. Any messages that are not from Devise are not altered.

# Screenshots of UI changes:

## The odd behaviour no longer happens:

https://user-images.githubusercontent.com/60350599/121948218-dcc56800-cd4e-11eb-9251-4b85987ecf83.mov

## The variants of other Devise messages remain correct:

Screenshot 2021-06-14 at 19 56 08

## The variants of non-Devise messages remain correct:

Screenshot 2021-06-14 at 20 08 59

Screenshot 2021-06-14 at 20 08 21

### DIFF
--- a/app/controllers/jobseekers/unlocks_controller.rb
+++ b/app/controllers/jobseekers/unlocks_controller.rb
@@ -1,3 +1,1 @@
-class Jobseekers::UnlocksController < Devise::UnlocksController
-  after_action :replace_devise_notice_flash_with_success!, only: %i[show]
-end
+class Jobseekers::UnlocksController < Devise::UnlocksController; end
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -53,7 +53,7 @@ html.govuk-template.app-html-class lang="en"
 
     main.govuk-width-container#main-content role="main"
       - flash.each do |variant, message|
-        = render FlashComponent.new variant: variant, message: message
+        = render FlashComponent.new(variant: convert_devise_flash_variant(variant, message), message: message)
 
       = yield
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,7 @@ class ApplicationController < ActionController::Base
   helper_method :cookies_preference_set?, :referred_from_jobs_path?, :utm_parameters, :current_variant?
 
   include Publishers::AuthenticationConcerns
+  include DeviseFlashConcerns
   include AbTestable
 
   def check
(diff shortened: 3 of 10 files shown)


app/controllers/jobseekers/unlocks_controller.rb
```diff
@@ -1,3 +1 @@
-class Jobseekers::UnlocksController < Devise::UnlocksController
-  after_action :replace_devise_notice_flash_with_success!, only: %i[show]
-end
+class Jobseekers::UnlocksController < Devise::UnlocksController; end
```
csutter, line 1: We can get rid of this entirely 🚮

csutter (approved):

app/controllers/jobseekers/unlocks_controller.rb
david-mears-dfe: I thought we could, but it resulted in some tests failing with "can't find the unlocks controller"... I guess because we were using `jobseeker_unlocks_path`, created by routes:

https://github.com/DFE-Digital/teaching-vacancies/blob/6392539ed212e593b3b482fc1ccca18d792b73e4/config/routes.rb#L18

I'll investigate what we can change here.

app/views/layouts/application.html.slim
```diff
@@ -53,7 +53,7 @@ html.govuk-template.app-html-class lang="en"
 
     main.govuk-width-container#main-content role="main"
       - flash.each do |variant, message|
-        = render FlashComponent.new variant: variant, message: message
+        = render FlashComponent.new(variant: convert_devise_flash_variant(variant, message), message: message)
```
cesidio, line 56: I like it!

app/controllers/jobseekers/unlocks_controller.rb
csutter: Ah. You should be able to remove the custom route, and then it'll pick up the default Devise one!

david-mears-dfe: It seems that we need to keep our custom controller in order to use our custom view.

PR #3641, opened 2021-06-16
Refactor time formatting

david-mears-dfe:
Use the same method everywhere, instead of duplicating formatting.

### DIFF
--- a/app/helpers/dates_helper.rb
+++ b/app/helpers/dates_helper.rb
@@ -1,16 +1,20 @@
 module DatesHelper
-  class FormatDateError < RuntimeError; end
+  def format_time_to_datetime_at(time)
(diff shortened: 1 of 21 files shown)


app/helpers/dates_helper.rb
```diff
@@ -1,16 +1,20 @@
 module DatesHelper
-  class FormatDateError < RuntimeError; end
+  def format_time_to_datetime_at(time)
+    return "" if time.nil?
```
cesidio, line 3: does it have to return an empty string?

I would return nil and write this line:

`return unless time`

But that is just a personal preference

cesidio (approved):

cesidio:
Nice work! thanks for refactoring this

PR #3642, opened 2021-06-16
[TEVA-2070] refactor navbar component to govuk standard

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2070

LAST COMPONENT!

## Changes in this PR:

- refactored component to extend govuk and add shared example tests
- fixed inconsistency between publisher/jobseeker signed in state and where the sign out link is placed (see below screen shots, ~im going to confirm this with jesse tomorrow~ he's good with this)
- tests added for publisher signed in state

before

Screenshot 2021-06-16 at 20 24 35
Screenshot 2021-06-16 at 20 24 15

after

Screenshot 2021-06-16 at 20 25 41
Screenshot 2021-06-16 at 20 25 07

### DIFF
--- a/app/components/navbar_component.rb
+++ b/app/components/navbar_component.rb
@@ -1,8 +1,23 @@
-class NavbarComponent < ViewComponent::Base
+class NavbarComponent < GovukComponent::Base
   delegate :active_link_class, to: :helpers
-  delegate :current_organisation, to: :helpers
-  delegate :current_publisher, to: :helpers
-  delegate :organisation_type_basic, to: :helpers
-  delegate :jobseeker_signed_in?, to: :helpers
-  delegate :publisher_signed_in?, to: :helpers
+
+  def initialize(classes: [], html_attributes: {})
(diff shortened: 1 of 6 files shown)


spec/system/navigation_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "rails_helper"
+
+RSpec.describe "Main navivation for users to sign in and out" do
+  let(:jobseeker) { create(:jobseeker) }
+  let(:publisher) { create(:publisher) }
```
cesidio, line 5: a published should be linked to an organisation, even though it is not relevant for this spec. But if in the future if you need to add anything that is relevant to the organisation, it will not work correctly

cesidio:
I am not 100% sure this should not be view component. the benefits of a view component are that you can test it better, in isolation and it is quicker. The system spec that you have written is much slower than the view component spec. I agree that the view component spec was not covering some cases and needed to be improved, but I still think it should be a view component

cesidio (approved):

PR #3648, opened 2021-06-18
remove unused npm rollbar dependency

alexbowen:
loosely related to

https://dfedigital.atlassian.net/browse/TEVA-2153

does fix client side rollbar logging with fallbacks for Rollbar not available and test scenarios

doesnt address whats actually being logged, but id like to see what starts being logged before making decisions about that. whats being logged might well give some clues for some JS related production 'defects' that have existed in rollbar dashboard for some time

i also added log to textarea formatting as i just noticed something in rollbar that needs investigating

everything from Js logging is logged under production/development etc in Rollbar dashboard same as serverside logs

### After

Screenshot 2021-06-18 at 10 57 53

### DIFF
--- a/app/frontend/src/application/publishers/textareaFormat.js
+++ b/app/frontend/src/application/publishers/textareaFormat.js
@@ -1,9 +1,21 @@
+import logger from '../../lib/logging';
+
 document.addEventListener('trix-initialize', (e) => {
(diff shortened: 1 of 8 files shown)


csutter (approved):
🔥

PR #3649, opened 2021-06-18
add alt attr to header logo

alexbowen:
only small issue outstanding on ticket but this creates noise in SEM dashboard as its on every page. plus images should have an alt tag!

https://dfedigital.atlassian.net/browse/TEVA-2153


csutter (approved):
👑

PR #3650, opened 2021-06-18
[Teva-2735] Add /attachment* behaviour pattern to cloudfront 

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2735

## Changes in this PR:

Add /attachment* behaviour pattern to cloudfront

Also revamped the custom_error_response using dynamic block as
it is a repeatable nested block. This makes the code easier to manage

### DIFF
--- a/terraform/app/modules/cloudfront/variables.tf
+++ b/terraform/app/modules/cloudfront/variables.tf
@@ -47,4 +47,12 @@ locals {
   cloudfront_aliases_cnames                              = [for zone in var.route53_zones : "${var.route53_cname_record}.${zone}"]
   cloudfront_aliases                                     = concat(var.route53_a_records, local.cloudfront_aliases_cnames)
   cloudfront_viewer_certificate_minimum_protocol_version = "TLSv1.2_2018"
+  cloudfront_cached_paths                                = ["/packs/*", "/attachment*"]
+  cloudfront_custom_response = {
+    404 = { ttl = "10" },
+    500 = { ttl = "60", response_code = "500", page_path = "${var.offline_bucket_origin_path}/index.html" },
+    502 = { ttl = "60", response_code = "502", page_path = "${var.offline_bucket_origin_path}/index.html" },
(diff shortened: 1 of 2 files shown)


terraform/app/modules/cloudfront/variables.tf
```diff
@@ -47,4 +47,10 @@ locals {
   cloudfront_aliases_cnames                              = [for zone in var.route53_zones : "${var.route53_cname_record}.${zone}"]
   cloudfront_aliases                                     = concat(var.route53_a_records, local.cloudfront_aliases_cnames)
   cloudfront_viewer_certificate_minimum_protocol_version = "TLSv1.2_2018"
+  cloudfront_path_pattern                                = { pattern_1 = { path = "/packs/*" }, pattern_2 = { path = "/attachment*" } }
```
saliceti, line 50: Since there is only 1 value for each iteration, we could use a simple list for cloudfront_path_pattern:
```
cloudfront_path_pattern = ["/packs/*", "/attachment*"]
```

saliceti: Also these are specifically for cached content, so a more explicit variable name may be used such as `cloudfront_cached_paths`

terraform/app/modules/cloudfront/variables.tf
```diff
@@ -47,4 +47,10 @@ locals {
...
+  cloudfront_custom_response = { response_1 = { er_code = "404", ttl = "10", response_code = "404", page_path = "${var.offline_bucket_origin_path}/index.html" },
+    response_2 = { er_code = "500", ttl = "60", response_code = "500", page_path = "${var.offline_bucket_origin_path}/index.html" },
```
saliceti, line 52: `"${var.offline_bucket_origin_path}/index.html"` is always the same, so it could be removed from the map and hard coded in the dynamic block

saliceti (approved):

PR #3652, opened 2021-06-18
Add Job Application Email Address to Analytics ymls

josephhull676:
## Changes in this PR:

- Added the email address associated with job applications to both yml files so they're sent to big query

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -118,6 +118,7 @@ shared:
     - in_progress_steps
     - reviewed_at
     - country
+    - email_address
   jobseekers:
     - id
     - email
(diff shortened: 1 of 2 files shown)


david-mears (approved):

config/analytics.yml
```diff
@@ -118,6 +118,7 @@ shared:
     - in_progress_steps
     - reviewed_at
     - country
+    - email_address
```
alexbowen, line 121: if i were being pedantic, i would say alphabetical order but you have been, its what was there before so dont worry...

alexbowen (approved):

config/analytics.yml
josephhull676: Yeah typically I would but I saw the rest weren't, so just stuck it on at the end!

alexbowen: fair enough

david-mears: Is it perhaps following the order in schema.db? If so that makes it easier to compare the two

PR #3657, opened 2021-06-21
Upgrade application to Ruby 3.0

csutter:
- Bump Ruby from 2.7.3 to 3.0.1 across the board
- Bump Bundler from 2.2.16 to 2.2.20
- Monkey-patch ancient Algolia Kaminari integration to fix deprecated
  implicit kwargs behaviour
- Fix a handful of other places that rely on implicit kwargs behaviour


cesidio (approved):
❤️

PR #3658, opened 2021-06-22
Add sensitive fields to parameter filter

csutter:


cesidio (approved):

david-mears (commented):
National insurance number is pretty sensitive too. Not sure where the line gets drawn between this short list and the giant list in analytics_pii.yml

csutter:
> National insurance number is pretty sensitive too. Not sure where the line gets drawn between this short list and the giant list in analytics_pii.yml

Good point, added that in. These are things we _never_ want to log, even for debugging purposes (as opposed to the `analytics_pii`, which is general PII that is fine on logging/auditing systems but not on our analytics platform).

PR #3661, opened 2021-06-22
Configure default cloudfront behaviour to use non caching policy

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2736

## Changes in this PR:
1 - “Configure default cloudfront behaviour to use non caching policy” it’s to simplify the code and stop using the legacy cache behavior

2 - for “Change Cloudfront cached path for attachment” it’s to make it more accurate and also more consistent with /packs/*

### DIFF
--- a/terraform/app/modules/cloudfront/main.tf
+++ b/terraform/app/modules/cloudfront/main.tf
@@ -29,7 +29,7 @@ resource "aws_cloudfront_distribution" "default" {
   enabled = true
   aliases = local.cloudfront_aliases
 
-    ordered_cache_behavior {
+  ordered_cache_behavior {
     allowed_methods  = ["GET", "HEAD"]
     cached_methods   = ["GET", "HEAD"]
     target_origin_id = "${var.service_name}-${var.environment}-offline"
--- a/terraform/app/modules/cloudfront/main.tf
+++ b/terraform/app/modules/cloudfront/main.tf
@@ -29,27 +29,7 @@ resource "aws_cloudfront_distribution" "default" {
   enabled = true
   aliases = local.cloudfront_aliases
 
-  default_cache_behavior {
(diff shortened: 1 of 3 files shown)


terraform/app/modules/cloudfront/main.tf
```diff
@@ -30,18 +30,11 @@ resource "aws_cloudfront_distribution" "default" {
...
-
-      cookies {
-        forward = "all"
-      }
-    }
+    allowed_methods          = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods           = ["GET", "HEAD"]
+    target_origin_id         = "${var.service_name}-${var.environment}-default-origin"
+    cache_policy_id          = data.aws_cloudfront_cache_policy.Managed-CachingDisabled.id
+    origin_request_policy_id = data.aws_cloudfront_origin_request_policy.Managed-AllViewer.id
```
saliceti, line 37: The comment below doesn't make sense anymore and can be removed

terraform/app/modules/cloudfront/main.tf
```diff
@@ -30,18 +30,11 @@ resource "aws_cloudfront_distribution" "default" {
...
-
-    forwarded_values {
-      query_string = true
-      headers      = var.default_header_list
-
-      cookies {
-        forward = "all"
-      }
-    }
+    allowed_methods          = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
```
saliceti, line 33: Could you move the `default_cache_behavior` block after the other `ordered_cache_behavior` blocks? This would show the processing order explicitly and would match the UI visually

saliceti (approved):

PR #3665, opened 2021-06-23
Raise Error when Boundary Retrieval Unsuccessful

josephhull676:
## Changes in this PR:

- Created custom error class.
- Error now raised when the ArcGis API does not respond with a 200.

### DIFF
--- a/app/services/import_polygons.rb
+++ b/app/services/import_polygons.rb
@@ -3,6 +3,8 @@ class ImportPolygons
   BUFFER_DISTANCES_IN_MILES = [5, 10, 15, 20, 25].freeze
   URL_MAXIMUM_LENGTH = 33_000
 
+  class ArcgisResponseError < StandardError; end
+
   include DistanceHelper
 
   attr_reader :location_name, :api_location_type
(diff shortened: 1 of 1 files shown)


app/services/import_polygons.rb
```diff
@@ -14,6 +26,8 @@ def initialize(api_location_type:)
   def call
     response = JSON.parse(HTTParty.get(LOCATION_POLYGON_SETTINGS[api_location_type][:boundary_api]))
 
+    raise ArcgisResponseError.new(response.dig("error", "code")) unless response.dig("error", "code") == 200
```
david-mears-dfe, line 29: This could be a little DRYer as you repeat `response.dig("error", "code")`.

app/services/import_polygons.rb
```diff
@@ -3,6 +3,18 @@ class ImportPolygons
   BUFFER_DISTANCES_IN_MILES = [5, 10, 15, 20, 25].freeze
   URL_MAXIMUM_LENGTH = 33_000
 
+  class ArcgisResponseError < StandardError
+    attr_reader :response_code
```
david-mears-dfe, line 7: Arguably this attr_reader should be a private method as you aren't trying to make it public. You can put `attr_reader` below a `private`.

josephhull676: I've made some changes. Let me know what you think :)

cesidio (approved):

PR #3667, opened 2021-06-24
Fix feedbacks with empty job_application_id

cesidio:
## Changes in this PR:
`application_id` field on feedbacks is redundant as we have `job_application_id`


david-mears-dfe (approved):
👍

PR #3670, opened 2021-06-24
[TEVA-2151] test canonial rel links on duplicate pages

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2151

## Changes in this PR:

- mark duplicate content job pages as <link rel="canonical" />
- testing this on two vacancies that SEM rush as flagged with errors because of this error. then can think about making this change for all vacancy listing pages

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -1,6 +1,10 @@
 class VacanciesController < ApplicationController
   helper_method :job_application
 
+  attr_accessor :canonical_href
+
+  SEO_CANONICAL_TEST_VACANCIES = %w[c9f9a45f-1f42-4686-9515-311bc169af29 ef21f5d2-4226-4cf8-b8d2-3155e149b0b2].freeze
+
   def index
     set_map_display
     params[:location] = params[:location_facet] if params[:location_facet]
@@ -22,6 +26,8 @@ class VacanciesController < ApplicationController
       return render "/errors/trashed_vacancy_found", status: :not_found
     end
 
+    canonical
+
     return redirect_to(job_path(vacancy), status: :moved_permanently) if old_vacancy_path?(vacancy)
 
     @saved_job = current_jobseeker&.saved_jobs&.find_by(vacancy_id: vacancy.id)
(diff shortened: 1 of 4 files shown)


app/views/layouts/application.html.slim
```diff
@@ -8,6 +8,10 @@ html.govuk-template.app-html-class lang="en"
     title #{content_for :page_title_prefix} — #{t("app.title")}
     meta content="width=device-width, initial-scale=1" name="viewport"
     meta content="blue" name="theme-color"
+
+    - if canonical_href
+      link rel="canonical" href="#{canonical_href}"
```
csutter, line 13: There's a [content area for links in the application layout](https://github.com/DFE-Digital/teaching-vacancies/blob/master/app/views/layouts/application.html.slim#L17). You can use something like:
```slim
- content_for(:head_links) do
  link rel="canonical" href="#{canonical_href}"
```
in `views/vacancies/show.html.slim`, that way you do not need to a) edit the layout, and b) add `canonical_href` to `ApplicationController`: https://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

app/controllers/vacancies_controller.rb
```diff
@@ -1,6 +1,10 @@
 class VacanciesController < ApplicationController
   helper_method :job_application
 
+  attr_accessor :canonical_href
```
csutter, line 4: It doesn't quite make sense to have an attr_accessor on a controller. `canonical` could be a helper method, and I think it might actually be better off in `VacanciesHelper`.

app/controllers/vacancies_controller.rb
```diff
@@ -34,6 +40,16 @@ def show
 
   private
 
+  UUID_REGEX = /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
+
+  def canonical
+    return unless SEO_CANONICAL_TEST_VACANCIES.include?(vacancy.id)
+
+    return unless UUID_REGEX.match(id) || !job_path(vacancy).end_with?(vacancy.slug)
```
csutter, line 48: Having slept on it, I'm not sure how I feel about this - are we overthinking this? Given that we redirect _any request_ to the jobs path to `job_url(vacancy)`, I reckon this could just be returning `job_url(vacancy)` in any case - for now just for the two test vacancies.

csutter: i.e. in `VacanciesHelper`:

```ruby
def canonical_url(vacancy)
  return nil unless SEO_CANONICAL_TEST_VACANCIES.include?(vacancy.id)
  
  job_url(vacancy)
end
```

csutter (changes requested):
Some Rails comments, and a thought on the approach in general now that I've slept on it. Let me know what you think!

PR #3693, opened 2021-06-30
docker security scan

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2761

## Changes in this PR:-

Introduce docker security scanning


sfawcett123 (commented):
You have chosen to NOT fail build, which I accept is the best way.

So what I would have done is create a new JOB for anchore and that job 'NEEDS' build,  
you deploy should also 'NEED' build, so the new Job and the Deploy job will run in parallel and you wont add time to your pipeline

sfawcett123 (approved):

PR #3697, opened 2021-07-01
[TEVA-2739] Update search input

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2739

## Changes in this PR:

As the html is shared across three pages, this PR doesn't restrict its scope to the homepage (I allow side-effects on other pages but did not deliberately try to implement the desired updates for these pages.)

This PR cannot be merged until the work to reorganise the radius input is done, as the styling of the radius input is now bad. This work is in progress at the time of writing: https://dfedigital.atlassian.net/browse/TEVA-2646

In this commit, I tried to create a consistent naming convention where files scoped under, or named, 'filters', refer to checkbox filters, while 'search' refers to the basic text inputs and the radius. This seems like a good place to split because the checkboxes are subtracted as a whole 'module' on the home page and included in other cases.

Other renaming of interest: it was unclear to me why 'filters' was the name of the boolean determining whether search inputs were displayed with full or 2/3 width, so I renamed it to 'wide'.

Later, we might want to consider combining all the shared/search partials into a single partial, so that we only have to write 'wide: true' once instead of three times for each page with search inputs.

### After

Screenshot 2021-07-08 at 16 33 47
Screenshot 2021-07-08 at 16 34 08
Screenshot 2021-07-08 at 16 34 21
Screenshot 2021-07-08 at 16 34 30
Screenshot 2021-07-08 at 16 34 44
Screenshot 2021-07-08 at 16 35 01

## Next steps:

- [ ] Product/UX sign off


david-mears-dfe:
https://teaching-vacancies-review-pr-3697.london.cloudapps.digital/

alexbowen (approved):

PR #3698, opened 2021-07-02
Get vacancy view count from BigQuery

csutter:
- Add basic initial code to get vacancy view count data from BigQuery
  (to be refactored when we have multiple things we need to retrieve)
- Change BigQuery dataset configuration to use Rails configuration
  instead of fetching from `ENV` throughout the code

### DIFF
--- a/app/services/publishers/vacancy_stats.rb
+++ b/app/services/publishers/vacancy_stats.rb
@@ -1,0 +1,39 @@
+class Publishers::VacancyStats
+  CACHE_DURATION = 6.hours
+  TABLE_NAME = "vacancies_published".freeze
+
+  def initialize(vacancy)
+    @vacancy = vacancy
+  end
+
+  def number_of_unique_views
+    query_single_field(:number_of_unique_vacancy_views)
+  end
+
+  private
+
+  attr_reader :vacancy
+
+  def query_single_field(field)
+    Rails.cache.fetch([field, vacancy.id], expires_in: CACHE_DURATION, skip_nil: true) do
+      # publish_on added to make query more efficient as the BigQuery table is partitioned by it
+      sql = <<~SQL
+        SELECT #{field}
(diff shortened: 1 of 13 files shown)


spec/models/vacancy_spec.rb
```diff
@@ -406,4 +406,19 @@
...
   end
+
+  describe "#bigquery_view_count" do
+    subject { build_stubbed(:vacancy) }
+    let(:response) { [{ number_of_unique_vacancy_views: 42 }] }
+    let(:big_query) { double("Bigquery") }
+
+    before do
+      allow(Google::Cloud::Bigquery).to receive(:new).and_return(big_query)
+      allow(big_query).to receive(:query).with(/SELECT number_of_unique_vacancy_views.*WHERE id="#{StringAnonymiser.new(subject.id)}"/m).and_return(response)
```
alexbowen, line 417: should this regex allow for multiple 'WHERE' - i mean where... and... actually im confused why .with() is regex at all?

csutter: Basically it being a regex implies "I want the query method to be called with a string that matches this regex".

I've not added the `AND` in because that's mostly a query optimisation on our end and it would still work without, and I didn't want to tie the test too closely to the SQL - but maybe I'll change it to do a proper string match to be more precise.

app/services/publishers/vacancy_stats.rb
```diff
@@ -0,0 +1,37 @@
+class Publishers::VacancyStats
+  CACHE_DURATION = 6.hours
+  TABLE_NAME = "vacancies_published".freeze
+
+  def initialize(vacancy)
+    @vacancy = vacancy
+  end
+
+  def number_of_unique_views = query_single_field(:number_of_unique_vacancy_views)
```
cesidio, line 9: nice! end-less methods, haven't used them yet!

cesidio (approved):

PR #3699, opened 2021-07-02
[TEVA-2443] Add dependent: :destroy to has_many :organisations

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2443

## Changes in this PR:
When we update organisations using nested attributes, we lose the destroy callback that send events to BigQuery. This does it manually and callbacks are not lost


csutter (approved):
🎉

PR #3701, opened 2021-07-02
Remove legacy view/more info click tracking

csutter:
This is now handled entirely in our BigQuery events stack, and will
soon be surfaced to publishers through the new `VacancyStats`.

We no longer need to track this data separately.


cesidio (approved):
🔥

PR #3707, opened 2021-07-06
Add radius to all location searches

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2646

## Changes in this PR:

- Removed javascript responsible for enabling an disabling the radius field
- Changed radius distances 
- LocationPolygons' polygons and buffers now updated if the polygons has changed or we have changed the radius distances
- Searches using location polygons will now retrieve buffer boundary using the radius given by the user
- Updated content rendered by the heading component
- Updated specs

## Comments 

- Because of the changes in this PR, polygons are not actually used at all, except for calculating the buffer polygons. I have another branch locally where I have removed buffers from LocationPolygon. Instead I store the buffer polygons in the polygons attribute. I intend on creating a PR for it once / if this is approved.


app/components/jobseekers/search_results/heading_component.rb
```diff
@@ -4,25 +4,26 @@ def initialize(vacancies_search:)
...
     @polygon_boundaries = @vacancies_search.location_search.polygon_boundaries
+    @radius = @vacancies_search.search_criteria[:radius]
     @total_count = @vacancies_search.total_count
   end
 
   def heading
     if @keyword.present? && @polygon_boundaries.present? && @vacancies_search.location_search.buffer_radius.nil?
       I18n.t("jobs.search_result_heading.keyword_location_polygon_html",
-             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count)
+             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count, radius: @radius)
```
david-mears-dfe, line 14: But the condition says radius is nil?

app/components/jobseekers/search_results/heading_component.rb
```diff
@@ -4,25 +4,26 @@ def initialize(vacancies_search:)
...
       I18n.t("jobs.search_result_heading.keyword_location_html",
-             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count)
+             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count, radius: @radius)
     elsif @keyword.present?
       I18n.t("jobs.search_result_heading.keyword_html",
              jobs_count: number_with_delimiter(@total_count), keyword: @keyword, count: @total_count)
     elsif @polygon_boundaries.present? && @vacancies_search.location_search.buffer_radius.nil?
       I18n.t("jobs.search_result_heading.location_polygon_html",
-             jobs_count: number_with_delimiter(@total_count), location: @location, count: @total_count)
+             jobs_count: number_with_delimiter(@total_count), location: @location, count: @total_count, radius: @radius)
```
david-mears-dfe, line 23: > But the condition says radius is nil?
+ 1

app/services/import_polygons.rb
```diff
@@ -53,8 +53,8 @@ def call
       # when we change a location's location-type mapping.
       location_polygon.update(location_type: LOCATIONS_MAPPED_TO_HUMAN_FRIENDLY_TYPES[location_name])
 
-      # Skip buffers API call if the points have not changed since last time we used them to calculate the buffers.
-      location_polygon.update(polygons: polygons_hash, buffers: get_buffers(polygons)) unless polygons_hash == location_polygon.polygons
+      # Skip buffers API call if the points and the buffer radii have not changed since last time we used them to calculate the buffers.
```
david-mears-dfe, line 56: To be more precise, it's the radius options

app/services/import_polygons.rb
```diff
@@ -1,6 +1,6 @@
 class ImportPolygons
   BUFFER_API_URL = "https://tasks.arcgisonline.com/arcgis/rest/services/Geometry/GeometryServer/buffer?".freeze
-  BUFFER_DISTANCES_IN_MILES = [5, 10, 15, 20, 25].freeze
+  BUFFER_DISTANCES_IN_MILES = Search::RadiusSuggestionsBuilder::RADIUS_OPTIONS
```
david-mears-dfe, line 3: Nice DRYing. Can we just use the RADIUS_OPTIONS directly instead of assigning it to another constant?

app/services/search/location_builder.rb
```diff
@@ -44,7 +44,7 @@ def initialize_polygon_boundaries
 
     @polygon_boundaries = []
     locations.compact.each do |location|
-      polygons = buffer_radius.present? ? location.buffers[buffer_radius] : location.polygons["polygons"]
+      polygons = buffer_radius.present? ? location.buffers[buffer_radius] : location.buffers[radius.to_s]
```
david-mears-dfe, line 47: We might want to rethink our separation of 'buffer_radius' and 'radius'. It looks like `buffer_radius` is what we call the radius when it comes from wider search suggestions after a no-result search; and I don't think we want to do anything differently in the search using the `buffer_radius` parameter compared to using the `radius` parameter (so, they 'mean' the same thing).

If that's correct, I'm happy for that refactor to be a later PR.

config/locales/jobs.yml
```diff
@@ -71,35 +71,43 @@ en:
           <span class='govuk-!-font-weight-bold text-wrap-apostrophe'>%{keyword}</span>
       location_html:
         one: >-
-          <span class='govuk-!-font-weight-bold'>%{count}</span> job found near
+          <span class="govuk-!-font-weight-bold">%{count}</span> job found within
+          <span class='govuk-!-font-weight-bold'>%{radius}</span> miles of
```
david-mears-dfe, line 75: Again, happy for this to be a separate PR: shall we distinguish between "1 mile" and "n mile *s*"? We can use the `pluralize` method on the string 'miles', where `mile.pluralize(1)` gives 'mile' and `mile.pluralize(2)` gives 'miles'.

spec/components/jobseekers/search_results/heading_component_spec.rb
```diff
@@ -8,6 +8,7 @@
   let(:location) { "London" }
   let(:polygon_boundaries) { [[51.0, 51.0]] }
   let(:count) { 10 }
+  let(:radius) { 1 }
```
david-mears-dfe, line 11: In this spec we are testing the cases where radius and buffer_radius are both present, but we never test the cases where only one of them is present, which is the more likely scenario.

spec/services/import_polygons_spec.rb
```diff
@@ -92,42 +92,100 @@
...
-        let(:original_buffers) { { "5" => "This is a hash that won't be updated" } }
+      before do
+        create(:location_polygon,
+               name: location_name,
+               location_type: api_location_type,
+               polygons: polygons,
+               buffers: original_buffers)
+      end
+
+      context "when the buffer radii are different from the last time the task was run" do
```
david-mears-dfe, line 131: Let's be more precise: "when the buffer radius options are different"

spec/services/import_polygons_spec.rb
```diff
@@ -92,42 +92,100 @@
...
+          end
           subject.call
         end
 
+        it "imports buffers" do
+          expect(imported_polygon.buffers).to eq(new_buffers)
+        end
+      end
+
+      context "when the buffer radii are the same as the last time the task was run" do
```
david-mears-dfe, line 164: Same here

spec/services/import_polygons_spec.rb
```diff
@@ -92,42 +92,100 @@
...
+        it "skips the buffers API call and returns the current buffers" do
+          expect(subject).to_not receive(:get_buffers)
+          expect(imported_polygon.buffers).to eq(original_buffers)
+          subject.call
+        end
+      end
+
+      context "when the points in the response are the same as the last time the task was run" do
         it "skips the buffers API call and returns the current buffers" do
+          expect(subject).to_not receive(:get_buffers)
```
david-mears-dfe, line 174: (y)

spec/services/search/location_builder_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   let(:location) { nil }
   let(:point_location) { "SW1A 1AA" }
-  let(:radius) { nil }
+  let(:radius) { 5 }
```
david-mears-dfe, line 16: For the tests of `buffer_radius`, this is making a `Search::LocationBuilder` that has both a radius and a buffer_radius parameter, which is a scenario we don't expect to happen.

app/services/search/location_builder.rb
josephhull676: I've actually removed buffer_radius in another branch!

david-mears-dfe: In that case don't worry about my comments about the tests needing to test the cases when only one of radius and buffer_radius is present, as those tests will be short-lived.

app/components/jobseekers/search_results/heading_component.rb
```diff
@@ -4,25 +4,26 @@ def initialize(vacancies_search:)
...
+    @radius = @vacancies_search.search_criteria[:buffer_radius] || @vacancies_search.search_criteria[:radius]
     @total_count = @vacancies_search.total_count
   end
 
   def heading
-    if @keyword.present? && @polygon_boundaries.present? && @vacancies_search.location_search.buffer_radius.nil?
+    if @keyword.present? && @polygon_boundaries.present?
       I18n.t("jobs.search_result_heading.keyword_location_polygon_html",
-             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count)
+             jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count, radius: "#{@radius} #{I18n.t('jobs.search_result_heading.unit_of_length').pluralize(@radius.to_i)}")
```
david-mears-dfe, line 14: This way looks cleaner - combining the radius and the units in the translation instead of here in the component:

```ruby
jobs_count: number_with_delimiter(@total_count), location: @location, keyword: @keyword, count: @total_count, radius: @radius, units: I18n.t('jobs.search_result_heading.unit_of_length').pluralize(@radius.to_i))
```

david-mears-dfe (commented):
Am I right in thinking that the updated import polygons task needs to be run before the other changes can be deployed?

josephhull676:
> Am I right in thinking that the updated import polygons task needs to be run before the other changes can be deployed?

Yeah, I'd guess so otherwise we'd offer radii which we don't have the buffer poylgons for, right?

How should we do that? Separate PR with the changes to the import polygons task and the changes to the radius_suggestions_builder (which has the new radius options)?

app/services/import_polygons.rb
```diff
@@ -68,7 +67,7 @@ def location_in_scope?
 
   def get_buffers(polygons)
     buffers = {}
-    BUFFER_DISTANCES_IN_MILES.each do |distance|
+    Search::RadiusSuggestionsBuilder::RADIUS_OPTIONS.each do |distance|
```
david-mears-dfe, line 70: I just realised we need to also include legacy radius options (15 and 20) for job alerts that use these.

alexbowen (approved):
this looks ok from frontend point of view, its pretty much removal of code though!

david-mears-dfe (approved):

PR #3716, opened 2021-07-08
Remove Polygons from Locations Polygons

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2802

## Changes in this PR:

- As we no longer use the original polygons retrieved from ONS, there is no need to store them, so they have been removed.
- Removed `buffer_radius`


app/components/jobseekers/search_results/heading_component.rb
```diff
@@ -4,25 +4,26 @@ def initialize(vacancies_search:)
     @keyword = @vacancies_search.keyword
     @location = @vacancies_search.location_search.location
     @polygon_boundaries = @vacancies_search.location_search.polygon_boundaries
+    @radius = @vacancies_search.search_criteria[:buffer_radius] || @vacancies_search.search_criteria[:radius]
```
david-mears-dfe, line 7: I thought you were getting rid of buffer_radius?

app/services/import_polygons.rb
```diff
@@ -45,16 +45,15 @@ def call
...
       location_polygon = LocationPolygon.find_or_create_by(name: location_name)
 
       # Update location_type separately to reduce chance of creating duplicate LocationPolygons for the same location
       # when we change a location's location-type mapping.
       location_polygon.update(location_type: LOCATIONS_MAPPED_TO_HUMAN_FRIENDLY_TYPES[location_name])
 
-      # Skip buffers API call if the points and BUFFER_RADII have not changed since last time we used them to calculate the buffers.
-      location_polygon.update(polygons: polygons_hash, buffers: get_buffers(polygons)) if polygons_hash != location_polygon.polygons || location_polygon.buffers.keys.map(&:to_i) != BUFFER_RADII
+      # Skip buffers API call if the points and the buffer radii have not changed since last time we used them to calculate the buffers.
+      location_polygon.update(buffers: get_buffers(polygons)) if location_polygon.buffers&.keys&.map(&:to_i) != Search::RadiusSuggestionsBuilder::RADIUS_OPTIONS
```
david-mears-dfe, line 56: I think this should be `BUFFER_RADII`

db/schema.rb
```diff
@@ -10,16 +10,17 @@
...
 
-ActiveRecord::Schema.define(version: 2021_07_02_140805) do
+ActiveRecord::Schema.define(version: 2021_07_07_104042) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "fuzzystrmatch"
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
+  enable_extension "postgis"
```
david-mears-dfe, line 20: Please remove the unrelated changes to schema.rb.

spec/services/import_polygons_spec.rb
```diff
@@ -167,106 +157,30 @@
...
+              "distances=#{convert_miles_to_metres(distance)}&f=json&geodesic=false&geometries=%7B%22geometryType"\
+              "%22%3D%3E%22esriGeometryPolygon%22%2C+%22geometries%22%3D%3E%5B%7B%22rings%22%3D%3E%5B%5B%5B52%2C+"\
+              "0%5D%2C+%5B53%2C+1%5D%5D%5D%7D%5D%7D&inSR=4326&outSR=4326&unionResults=true&unit=",
             ).and_return(buffer_response)
           end
           subject.call
         end
 
-        it "imports buffers" do
-          expect(imported_polygon.buffers).to eq(new_buffers)
```
david-mears-dfe, line 178: Don't we want to keep this test?

spec/services/import_polygons_spec.rb
```diff
@@ -92,71 +105,48 @@
...
       before do
         create(:location_polygon,
                name: location_name,
                location_type: api_location_type,
-               polygons: polygons,
-               buffers: original_buffers)
+               buffers: original_polygons)
+      end
+
+      context "when Search::RadiusSuggestionsBuilder::RADIUS_OPTIONS is the same as the last time the task was run" do
```
david-mears-dfe, line 127: You mean `BUFFER_RADII`?

spec/services/import_polygons_spec.rb
```diff
@@ -92,71 +105,48 @@
...
+        it "skips the buffers API call and returns the current buffers" do
+          expect(subject).to_not receive(:get_buffers)
+          expect(imported_polygon.buffers).to eq(original_polygons)
+          subject.call
+        end
       end
 
-      context "when the buffer radius options are different from the last time the task was run" do
-        let(:new_buffers) do
+      context "when Search::RadiusSuggestionsBuilder::RADIUS_OPTIONS are different from the last time the task was run" do
```
david-mears-dfe, line 135: You mean `BUFFER_RADII`?

app/components/jobseekers/search_results/heading_component.rb
josephhull676: Well spotted!

josephhull676: I think this got through in the confusion of splitting PRs and rebasing etc etc

app/services/import_polygons.rb
```diff
@@ -45,16 +45,15 @@ def call
...
       location_polygon = LocationPolygon.find_or_create_by(name: location_name)
 
       # Update location_type separately to reduce chance of creating duplicate LocationPolygons for the same location
       # when we change a location's location-type mapping.
       location_polygon.update(location_type: LOCATIONS_MAPPED_TO_HUMAN_FRIENDLY_TYPES[location_name])
 
-      # Skip buffers API call if the points and BUFFER_RADII have not changed since last time we used them to calculate the buffers.
-      location_polygon.update(polygons: polygons_hash, buffers: get_buffers(polygons)) if polygons_hash != location_polygon.polygons || location_polygon.buffers.keys.map(&:to_i) != BUFFER_RADII
+      # Skip buffers API call if the points and the BUFFER_RADII have not changed since last time we used them to calculate the buffers.
+      location_polygon.update(buffers: get_buffers(polygons)) if location_polygon.buffers&.keys&.map(&:to_i) != BUFFER_RADII
```
david-mears-dfe, line 56: I wonder should we still check that at least one of the buffered sets of polygons is different? At the moment this just checks that the radius options are different. We checked if the polygon itself had changed to make sure we picked up any bugfixes, e.g. if the polygon is in Wales when it should be in Liverpool.

spec/services/import_polygons_spec.rb
```diff
@@ -167,106 +157,30 @@
...
-
-      context "when the buffer radius options are the same as the last time the task was run" do
-        it "skips the buffers API call and returns the current buffers" do
-          expect(subject).to_not receive(:get_buffers)
-          expect(imported_polygon.buffers).to eq(original_buffers)
-          subject.call
-        end
-      end
-
-      context "when the points in the response are the same as the last time the task was run" do
```
david-mears-dfe, line 190: Let's resurrect a version of this test.

app/services/import_polygons.rb
```diff
@@ -114,4 +103,22 @@ def buffer_api_endpoint(coords, distance)
   def take_every_nth_coord(coords, number)
     coords.each_with_index.filter_map { |item, index| item if (index % number).zero? }
   end
+
+  def poylgons_have_changed?(location_polygon, polygons)
```
david-mears-dfe, line 107: "poylgons"

app/services/import_polygons.rb
```diff
@@ -114,4 +103,22 @@ def buffer_api_endpoint(coords, distance)
...
+    return true if location_polygon&.buffers.nil? || location_polygon.buffers&.fetch(BUFFER_RADII.first.to_s, nil).nil?
+
+    get_buffered_polygons(polygons, BUFFER_RADII.first) != location_polygon.buffers[BUFFER_RADII.first.to_s]
+  end
+
+  def get_buffered_polygons(polygons, distance)
```
david-mears-dfe, line 113: You're now calling this method twice per location for the first radius option (once for `poylgons_have_changed?` and once in `buffered_polygons`), which will make the task take longer (I assume it's getting the buffered polygons that takes so much time).

However, I can't immediately think of a good way to avoid this. I can only think of
- _assign_ the output of `get_buffered_polygons` for the first radius option to a variable
- check it against the buffer in the database
- when you update the location polygon, update the first radius option using the variable you already assigned, and then make calls for the remaining radius options.

This seems a bit convoluted.

As the task only runs once per week and isn't causing issues at the moment, I'm happy to leave this duplication alone rather than make the code more complicated and hard to follow.

spec/services/import_polygons_spec.rb
```diff
@@ -82,191 +93,178 @@
...
+          allow(HTTParty).to receive(:get).with(
+            "https://tasks.arcgisonline.com/arcgis/rest/services/Geometry/GeometryServer/buffer?bufferSR=3857&"\
+            "distances=#{convert_miles_to_metres(ImportPolygons::BUFFER_RADII.first)}&f=json&geodesic=false&geometries=%7B%22geometryType%22%3D%3E%22esriGeometryPolygon%"\
+            "22%2C+%22geometries%22%3D%3E%5B%7B%22rings%22%3D%3E%5B%5B%5B51.406361958644%2C+-2.3780576677997%5D%"\
+            "2C+%5B51.4063596372237%2C+-2.3787764623145%5D%5D%5D%7D%5D%7D&inSR=4326&outSR=4326&unionResults=true&"\
+            "unit=",
+          ).and_return(unchanged_buffers_response)
+        end
+
+        it "skips the API call that returns all of the buffered polygons and returns the current buffers" do
```
david-mears-dfe, line 152: Let's use plural `API calls that return` to make it clear to future refactorers that we are saving multiple API calls with this check, so it's actually worth doing.

spec/services/import_polygons_spec.rb
```diff
@@ -82,191 +93,178 @@
...
                name: location_name,
                location_type: api_location_type,
-               polygons: polygons,
-               buffers: original_buffers)
+               buffers: original_polygons)
       end
 
-      context "when the buffer radius options are different from the last time the task was run" do
-        let(:new_buffers) do
+      context "when the first buffer polygon retreived is the same as the last time the task was run" do
```
david-mears-dfe, line 141: `retrieved`

david-mears-dfe (approved):
See comments. A few minor spellings and a rambly comment.

spec/services/search/vacancy_search_spec.rb
```diff
@@ -171,6 +171,7 @@
         before do
           allow(vacancies).to receive(:empty?).and_return(false)
           mock_algolia_search(vacancies, 1, keyword, arguments_to_algolia)
+          freeze_time
```
david-mears-dfe, line 174: I don't understand what the freeze_time is for - what depends on time?

josephhull676: The timestamp in the filters

david-mears-dfe (approved):

PR #3717, opened 2021-07-08
[TEVA-2720] Validate job applications before submitting them

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2720

## Changes in this PR:

- **Refactor/fix up the code that serves the equivalent purpose for vacancy publishers**

Then:

- **Add 'action required' error message to job applications when invalid**

If we change the validations on job application forms after a job application is
started but before it is submitted, a job application can be submitted with invalid
values (e.g. missing a field). We should alert the user to this so that we can avoid
having submitted applications in unexpected states (this has caused a bug recently
owing to us adding a field 'age' to the equal opportunities form - this resulted in
an error being thrown which prevented applications which completed that form before
the change from being able to be submitted).

This solution is modelled on the 'action required' solution which serves a parallel
purpose for the vacancy build process on the hiring staff side. It is not a perfect
solution as it does not validate references, employment histories, qualifications, or
qualification results.

`id` attributes are added to values in the review step so that the 'action required' error
summary can link to the relevant part of the page.

We will not display the review form errors and action required errors at the same time
to avoid confusion/visual overload. The review form errors will take precedence.

### After

Screenshot 2021-07-08 at 17 11 45

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,7 +14,7 @@ class ApplicationController < ActionController::Base
 
   after_action :trigger_page_visited_event, unless: :request_is_healthcheck?
 
-  helper_method :cookies_preference_set?, :referred_from_jobs_path?, :utm_parameters, :current_variant?
+  helper_method :cookies_preference_set?, :referred_from_jobs_path?, :user_type, :utm_parameters, :current_variant?
 
   include Publishers::AuthenticationConcerns
   include DeviseFlashConcerns
@@ -29,6 +29,14 @@ class ApplicationController < ActionController::Base
       format.html { render "errors/not_found", status: :not_found }
       format.json { render json: { error: "Resource not found" }, status: :not_found }
       format.all { render status: :not_found, body: nil }
+    end
+  end
+
+  def user_type
+    if current_jobseeker.present?
+      :jobseeker
+    elsif current_publisher.present?
+      :publisher
     end
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,9 +35,9 @@ class ApplicationController < ActionController::Base
   private
 
   def user_type
-    if current_jobseeker.present?
+    if jobseeker_signed_in?
       :jobseeker
-    elsif current_publisher.present?
+    elsif publisher_signed_in?
       :publisher
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,6 +32,8 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  private
+
   def user_type
     if current_jobseeker.present?
       :jobseeker
@@ -39,8 +41,6 @@ class ApplicationController < ActionController::Base
       :publisher
     end
   end
-
-  private
 
   def cookies_preference_set?
     cookies["consented-to-cookies"].present?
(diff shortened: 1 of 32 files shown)


app/controllers/application_controller.rb
```diff
@@ -34,6 +34,14 @@ def not_found
 
   private
 
+  def user_type
+    if current_jobseeker.present?
```
csutter, line 38: You should be able to do `jobseeker_signed_in?` (and ditto for publisher)

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -80,6 +84,22 @@ def withdraw
...
+  def completed_steps_valid?
+    # Check that all completed steps are valid, in case we have changed the validations since the step was completed.
+    # NB: Only validates top-level step forms. Does not validate individual qualifications, employments, or references.
+    job_application.completed_steps.all? do |step|
+      step_valid?("jobseekers/job_application/#{step}_form".camelize.constantize)
+    end
+  end
+
+  def step_valid?(step_form)
+    form = step_form.new(job_application.slice(*send("#{step_form.to_s.underscore.split('/').last.split('_form').first}_fields")))
```
csutter, line 96: Maybe if `completed_steps_valid?` passes in `step` to this method, and then this method constructs the class, this would be simpler?

I know we've got this multi-step form validation pattern elsewhere, but I'd love for us to have a stab at revisiting it one day (not this PR 😄 )

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -1,11 +1,15 @@
 class Jobseekers::JobApplicationsController < Jobseekers::BaseController
   include QualificationFormConcerns
+  include Jobseekers::Wizardable
 
   before_action :raise_unless_vacancy_enable_job_applications,
                 :redirect_if_job_application_exists, only: %i[new create new_quick_apply quick_apply]
   before_action :redirect_unless_draft_job_application, only: %i[review]
 
   helper_method :job_application, :qualification_form_param_key, :review_form, :vacancy, :withdraw_form
+  helper_method :personal_details_fields, :professional_status_fields, :qualifications_fields, :employment_history_fields,
```
csutter, line 10: Evidently it's too late in the day for me to make sense of the world, but why do these need to be made helpers?

david-mears-dfe: They are `send`ed in app/helpers/job_application_helper.rb#L86

david-mears-dfe: Or maybe not... I'll try removing them and see if tests fail.

david-mears-dfe: Reader, they did not.

app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
```diff
@@ -1,4 +1,4 @@
-= render "shared/error_tag", attributes: %i[contact_email contact_number school_visits how_to_apply application_link], model: @vacancy
+= render "shared/error_tag", applying_for_the_job_fields, model: @vacancy
```
cesidio: This does not seem consistent with how it is used on the other partials in this commit. I would have expected it to be:
`= render "shared/error_tag", attributes: applying_for_the_job_fields, model: @vacancy`
seems there is a missing `attributes:`

cesidio (approved):

cesidio:
Only one suggestions, if you have multiple commits, please make sure each individual commit is meaningful.

PR #3724, opened 2021-07-12
[TEVA-2732] Show job page

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2732

## Screenshots of UI changes:
Screenshot 2021-07-14 at 22 48 56
Screenshot 2021-07-14 at 22 32 35
Screenshot 2021-07-14 at 22 32 23

looking perfectly usable on mobile also, maybe some tweaks i could make there but not a deal breaker i feel

### DIFF
--- a/app/controllers/publishers/vacancies_controller.rb
+++ b/app/controllers/publishers/vacancies_controller.rb
@@ -7,8 +7,6 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   helper_method :applying_for_the_job_fields, :important_dates_fields, :job_details_fields, :job_location_fields, :job_summary_fields, :pay_package_fields, :schools_fields
 
   def show
-    return redirect_to organisation_job_review_path(vacancy.id), notice: t(".notice") unless vacancy.published?
-
     @vacancy = VacancyPresenter.new(vacancy)
   end
 
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -87,6 +87,10 @@ class Vacancy < ApplicationRecord
     published? && expires_at&.future? && (publish_on&.today? || publish_on&.past?)
   end
 
+  def pending?
+    published? && publish_on&.future?
+  end
+
   def application_link=(value)
     # Data may not include a scheme/protocol so we must be careful when creating
     # links that Rails doesn't make them incorrectly relative.
--- a/app/views/publishers/vacancies/job_applications/_banner.html.slim
+++ b/app/views/publishers/vacancies/job_applications/_banner.html.slim
@@ -1,0 +1,43 @@
+= render BannerComponent.new do
+  - if vacancy.expires_at&.past?
+    = govuk_breadcrumbs breadcrumbs: { "#{t("publishers.vacancies_component.expired.tab_heading")}": jobs_with_type_organisation_path(:expired), "#{vacancy.job_title}": "" }
+  - elsif vacancy.pending?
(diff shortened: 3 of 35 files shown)


app/models/vacancy.rb
```diff
@@ -87,6 +87,10 @@ def listed?
     published? && expires_at&.future? && (publish_on&.today? || publish_on&.past?)
   end
 
+  def scheduled?
```
david-mears-dfe, line 90: The scope is called `pending` but it means the same as `scheduled`. We should choose one naming convention to be consistent.

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -39,6 +37,13 @@ def review
     @vacancy = VacancyPresenter.new(vacancy)
   end
 
+  def live
+    return redirect_to organisation_job_review_path(vacancy.id), notice: t(".notice") unless vacancy.published?
+
+    @vacancy = VacancyPresenter.new(vacancy)
+    render "publishers/vacancies/_show"
```
david-mears-dfe, line 44: What differences are there between this view and the _real_ live listing that jobseekers see?

If it's very few, would it make more sense to use the same view for both, plus some conditionals like `unless publisher_signed_in?`? Because if we show publishers a page and call it "the live listing", they might share the link to that page (copying it from the address bar), instead of the link to the _real_ live listing as seen by jobseekers; and a jobseeker who follows that incorrect link will be required to authenticate as a publisher to see the listing.

cesidio: Very good point! In talks with Jesse to define what we want on that

alexbowen:
im looking why but it seems like there is extra padding at bottom of tab menu in screenshot 3?

app/views/publishers/vacancies/job_applications/_banner.html.slim
```diff
@@ -0,0 +1,41 @@
+= render BannerComponent.new do
```
alexbowen, line 1: if you put `= render BannerComponent.new do classes: "govuk-!-padding-bottom-0" do` fixes the padding thing i mentioned below

app/controllers/publishers/vacancies_controller.rb
alexbowen: was this resolved? @cesidio @david-mears-dfe

cesidio: as per this conversation on slack:
https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1626161646232600
the link now points to the public live vacancy

app/views/publishers/vacancies/job_applications/_banner.html.slim
```diff
@@ -0,0 +1,44 @@
...
+        .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:shortlisted)}"
+          .govuk-heading-l = job_applications.shortlisted.count
+          h3.govuk-caption-s = t(".shortlisted", count: job_applications.shortlisted.count)
+
+      .govuk-grid-column-one-quarter
+        .application-count.govuk-tag class="govuk-tag--#{status_tag_colour(:unsuccessful)}"
+          .govuk-heading-l = job_applications.unsuccessful.count
+          h3.govuk-caption-s = t(".rejected", count: job_applications.unsuccessful.count)
+
+.govuk-grid-row
```
cesidio, line 40: Here we need to do the following only when:
- if vacancy.enable_job_applications? && vacancy.published? && !vacancy.pending?

alexbowen: ah ok, i missed, i had to do a certain amount of manual code merging earlier and missed this

alexbowen (approved):

PR #3739, opened 2021-07-14
[TEVA-2841] Increase default search radius to 10

david-mears-dfe:
https://dfedigital.atlassian.net/browse/TEVA-2841

## Changes in this PR:

- We forgot to make the default 10.


alexbowen (approved):
would be nice to rename your wip commit to something else else lgtm

PR #3740, opened 2021-07-14
fix repeat meta descriptions flagged in SEO audit

alexbowen:
no ticket

now that a lot of noise has been removed from SEM dashboard (thanks largely to @csutter updating robots.txt - and we have a lot less vacancies than a month or two ago) few things more apparent like repeated meta data

an example:

Screenshot 2021-07-14 at 21 49 53

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -108,7 +108,7 @@ en:
       heading: Cookies
       subheading: Cookies that measure website use
       title: Cookies
-    page_description: Find out what cookies Teaching Vacancies stores on your computer and what they are for.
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -16,6 +16,7 @@ def meta_description
     view = params[:id].presence || params[:action].presence
     controller_i18n_path = controller_path.tr("/", ".")
 
+    # binding.pry
```
david-mears-dfe, line 19:
```suggestion

```

alexbowen: yep just noticed ;)

csutter (approved):

PR #3742, opened 2021-07-15
Spike new similar jobs location logic

david-mears-dfe:
No ticket.

Currently there is an issue whereby searching for jobs that are similar to a multi-school job will use the postcode of the parent organisation. For MATs, this is suboptimal but acceptable, but for LAs, it is a bug, because we don't have postcodes for LAs (so the search goes ahead without a location).

This is a proposed solution which takes the mean location of the schools the vacancy is at, as the location to search around. This is also probably slightly better for the MAT case, too.

Conversation: https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1626341468288200

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -115,6 +115,15 @@ class Vacancy < ApplicationRecord
     job_applications.after_submission.count >= EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD
   end
 
+  def set_mean_geolocation!
+    # When SimilarJobs searches for jobs similar to a multi-school job, we need to derive a location to search around.
+    # We do not have LAs' postcodes; anyway, LAs can manage schools outside themselves.
+    # Take the mean of the geolocations of the school(s) the vacancy is at.
+    # Centroids are for polygons (ordered points defining a shape), so the mean is more appropriate than a centroid.
+    coords = organisations.map(&:geolocation).compact.map { |loc| [loc.x, loc.y] }.reject { |coord| coord == [0, 0] }
+    self.mean_geolocation = [(coords.sum(&:first) / coords.length), (coords.sum(&:second) / coords.length)] unless coords.count.zero?
+  end
+
   private
 
   def slug_candidates
--- a/app/services/search/criteria_inventor.rb
+++ b/app/services/search/criteria_inventor.rb
@@ -1,19 +1,20 @@
 # Based on a vacancy, devise a plausible set of search criteria for a job alert subscription
-class Search::CriteriaDeviser
+class Search::CriteriaInventor
   attr_reader :criteria
 
   def initialize(vacancy)
     @vacancy = vacancy
     @subjects_from_job_title = get_subjects_from_job_title
-    @criteria = devise_search_criteria
+    set_criteria
   end
 
   private
 
-  def devise_search_criteria
-    {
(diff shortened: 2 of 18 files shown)


app/models/vacancy.rb
```diff
@@ -115,6 +115,15 @@ def publish_equal_opportunities_report?
     job_applications.after_submission.count >= EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD
   end
 
+  def set_mean_geolocation!
+    # When SimilarJobs searches for jobs similar to a multi-school job, we need to derive a location to search around.
+    # We do not have LAs' postcodes; anyway, LAs can manage schools outside themselves.
+    # Take the mean of the geolocations of the school(s) the vacancy is at.
+    # Centroids are for polygons (ordered points defining a shape), so the mean is more appropriate than a centroid.
+    coords = organisations.map(&:geolocation).compact.map { |loc| [loc.x, loc.y] }.reject { |coord| coord == [0, 0] }
+    self.mean_geolocation = [(coords.sum(&:first) / coords.length), (coords.sum(&:second) / coords.length)] unless coords.count.zero?
```
alexbowen, line 124: nice!

app/services/search/criteria_inventor.rb
```diff
@@ -1,19 +1,20 @@
...
 
-  def devise_search_criteria
-    {
-      location: @vacancy.parent_organisation.postcode,
-      radius: (@vacancy.parent_organisation.postcode.present? ? "10" : nil),
+  def set_criteria
+    set_location
+    @criteria ||= {
+      location: @location,
+      radius: (@location.present? ? "10" : nil),
```
csutter, line 17: This could use a constant

app/services/search/criteria_inventor.rb
```diff
@@ -30,6 +31,14 @@ def keyword
     get_keywords_from_job_title.presence unless @vacancy.job_roles.present?
   end
 
+  def set_location
+    @location ||= if @vacancy.organisations.many?
```
csutter, line 35: Will that postcode then show up in the job alert? That might look a bit weird if the user never entered the mean postcode.

spec/rails_helper.rb
```diff
@@ -19,6 +19,7 @@
 
 # Stub Geocoder HTTP requests in specs
 Geocoder::DEFAULT_STUB_COORDINATES = [51.67014192630465, -1.2809649516211556].freeze
+Geocoder::DEFAULT_LOCATION = "TE5 T1NG".freeze
```
csutter, line 22: Might this cause issues down the line not being a real postcode?

csutter (commented):
Liking the approach so far - jotted down a few thoughts to meditate on 🙆

app/services/search/criteria_inventor.rb
david-mears-dfe: Yes. Not sure what we should show up here - I began with using just lat and lng everywhere, but only afterwards did I realise that we'd need something human-readable when a job alert is created from the search!

spec/rails_helper.rb
david-mears-dfe: What kind of issues?

david-mears-dfe:
This was a spike, we don't need to merge it

PR #3762, opened 2021-07-20
[TEVA-2751] Add column encryption with Lockbox: step 1

david-mears-dfe:
- Add ciphertext columns for columns to be encrypted, for 'defence in depth'. These columns are a subset of analytics_pii.yml (excluding '*_id' columns in order not to break joining, and excluding email columns as we may want to join/`where` on those).

There is a single master key environment variable added for this commit, which is used to generate individual keys for each column (so that cracking one column doesn't crack all other columns).

The reason for adding column encryption is to add a layer of defence for data that could be exploited for fraud such as national insurance numbers. Data is encrypted at rest in our database, so column encryption defends against the case where someone gets into the database but not into the app itself.

- Add task for migrating the data from the old to the new columns.

Note that Rails will ship with the ability to encrypt columns in the next major version (7). That could be years away, so this interim solution may last a while (if indeed we do migrate to the Rails-native method in the future). This issue describes how we can migrate away from Lockbox: https://github.com/ankane/lockbox/issues/121

## Side effects of this approach

New secret environment key (a different one has been added to each environment: dev, prod, qa, staging). This means that production database dumps will only be decryptable if you set your local key to the production key.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2751

## Testing

I tested the migrations and task on dev.

Note: This was before the most recent changes, so might be worth checking again once I'm back.


david-mears-dfe:
> Don't think we need to encrypt publisher oid though?

The reason oid is on there is because it's in analytics_pii.yml. If I had to guess/remember, I would say we anonymise that data because oid points to a record in DSI. If DSI, or any of its client services which store OIDs, had a data leak, then OID would help a hacker to interpret any data they stole from us.

david-mears-dfe:
> We'd need to look into what to do with the sanitised DB dumps

Hmm, I hadn't thought about this - staging needs to be able to decrypt data from prod dumps. So should we use the same key across all environments? It feels wrong, but it's strictly safer than the status quo which is having no column encryption at all.

app/models/application_record.rb
```diff
@@ -10,7 +10,7 @@ class ApplicationRecord < ActiveRecord::Base
   private
 
   def event_data
-    { table_name: self.class.table_name }.merge(anonymised_attributes)
+    { "table_name" => self.class.table_name }.merge(anonymised_attributes)
```
cesidio, line 13: Why this key should change to a symbol?

david-mears-dfe: It's merged with attributes, which are all string keys. It's less confusing when all the keys are the same type.

app/models/job_application.rb
```diff
@@ -32,6 +32,12 @@ class JobApplication < ApplicationRecord
   # If you want to add a status, be sure to add a `status_at` column to the `job_applications` table
   enum status: { draft: 0, submitted: 1, reviewed: 2, shortlisted: 3, unsuccessful: 4, withdrawn: 5 }, _default: 0
 
+  encrypts :first_name, :last_name, :previous_names, :street_address, :city, :postcode, :teacher_reference_number,
```
david-mears-dfe, line 35: forgot phone_number

db/migrate/20210721135549_add_lockbox_ciphertext_columns_and_blind_index_columns.rb
```diff
@@ -0,0 +1,52 @@
...
+    add_column :employments, :main_duties_ciphertext, :text
+    add_column :feedbacks, :email_ciphertext, :text
+    add_column :feedbacks, :email_bidx, :string
+    add_index :feedbacks, :email_bidx
+    add_column :job_applications, :first_name_ciphertext, :text
+    add_column :job_applications, :last_name_ciphertext, :text
+    add_column :job_applications, :previous_names_ciphertext, :text
+    add_column :job_applications, :street_address_ciphertext, :text
+    add_column :job_applications, :city_ciphertext, :text
+    add_column :job_applications, :postcode_ciphertext, :text
```
david-mears-dfe, line 18: forgot phone_number

spec/services/copy_vacancy_spec.rb
```diff
@@ -20,8 +20,7 @@
 
         described_class.new(vacancy).call
 
-        expect(Vacancy.find(vacancy.id).attributes == vacancy.attributes)
-          .to eq(true)
+        expect(vacancy.attributes == Vacancy.find(vacancy.id).attributes).to eq(true)
```
csutter, line 23: I know you didn't change this in this PR, but might as well change it to 

```ruby
expect(vacancy.attributes).to eq(Vacancy.find(vacancy.id).attributes)
```

spec/rails_helper.rb
```diff
@@ -3,6 +3,8 @@
 
 require "spec_helper"
 ENV["RAILS_ENV"] ||= "test"
+# Lockbox master key is specific to test environment
+ENV["LOCKBOX_MASTER_KEY"] = "5b1f54d9c713b028b34a34b2bf95b4e22150fcd636069efc67cf7bff64ddfb04"
```
csutter, line 7: Overriding env variables gets a bit faffy - I reckon this would be better off in `config/environments/test.rb` as

```ruby
Lockbox.master_key = 'blahdiblah'
```

csutter (approved):
👍

david-mears-dfe:
Closing for a week until I'm back to save resources

PR #3763, opened 2021-07-20
[TEVA-2860] Dac report initial A11y fixes

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2860

these are a bunch of changes based on suggestions from the recent DAC accessibilty report. i tried to make the commits fairly succint and i have referenced against the audit document in preparation for creating subsequent tickets for the backlog

will squash the commits before merging but leaving them there in case they aid reviewing this PR

### DIFF
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -16,7 +16,7 @@ module VacanciesHelper
 
   def page_title_prefix(vacancy, form_object, page_heading)
     if vacancy.published?
-      "#{form_object.errors.present? ? 'Error: ' : ''}Edit the #{page_heading}"
+      "#{form_object.errors.present? ? 'Error: ' : ''}Edit the #{page_heading} for #{vacancy.job_title}"
     else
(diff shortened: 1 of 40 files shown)


app/helpers/vacancies_helper.rb
```diff
@@ -16,7 +16,7 @@ def uncapitalize_words(location_name)
 
   def page_title_prefix(vacancy, form_object, page_heading)
     if vacancy.published?
-      "#{form_object.errors.present? ? 'Error: ' : ''}Edit the #{page_heading}"
+      "#{form_object.errors.present? ? 'Error: ' : ''}Edit the #{page_heading} for #{vacancy.job_title}"
```
cesidio, line 19: This would be a bit more readable I think if it was written like this:

`"#{'Error: ' if form_object.errors.present?}Edit the #{page_heading} for #{vacancy.job_title}"`

cesidio (approved):

app/helpers/vacancies_helper.rb
alexbowen: i think @cesidio  there are a bunch of changes (well pages missing the 'error' page title state) that needs to be addressed that i have on my list of tickets to create today, so ill leave this as is and mention this in ticket (i didnt think to improve code readability here just added what i added)

PR #3770, opened 2021-07-21
[TEVA-2820] Remove prompt to create account from job alert email

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2820

## Changes in this PR:

- As jobseeker is no longer called in alert.text.erb, @jobseeker is not set and therefore not available when the email event is created. This caused some tests to fail and meant that the jobseeker id was not sent in the email event data. I apparently fixed this with a callback I added to the alert mailer. I don't know if this was the correct way to go about making the tests pass though...

### DIFF
--- a/app/mailers/jobseekers/alert_mailer.rb
+++ b/app/mailers/jobseekers/alert_mailer.rb
@@ -1,4 +1,6 @@
 class Jobseekers::AlertMailer < Jobseekers::BaseMailer
+  after_action :jobseeker
+
   self.delivery_job = AlertMailerJob
   helper DatesHelper
   helper VacanciesHelper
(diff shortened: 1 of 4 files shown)


david-mears-dfe:
> `@jobseeker` is not set and therefore not available when the email event is created

If I understand this right, we need `@jobseeker` to be set so it can be recorded in the event?

josephhull676:
> > `@jobseeker` is not set and therefore not available when the email event is created
> 
> If I understand this right, we need `@jobseeker` to be set so it can be recorded in the event?

Yes, I believe so.

cesidio (approved):

david-mears-dfe (approved):

PR #3777, opened 2021-07-22
[TEVA-2885] Adapt behavior of navbar component

david-mears-dfe:
We wanted the 'Manage jobs' link to be styled as if active when using
the jobs dashboard; as this dashboard uses multiple urls, we need to
check if the current url includes a certain string that's common to the
urls.

Screenshot 2021-07-22 at 17 27 40

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2885


alexbowen (approved):
nice one

PR #3803, opened 2021-08-02
[TEVA-2751] Add column encryption with Lockbox: step A

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2751

## Changes in this PR:

- Add ciphertext columns for columns to be encrypted. These columns are a subset of analytics_pii.yml (excluding '*_id' columns in order not to break joining, and excluding subscription.email as we may want to join on that column and we do not associate jobseeker records with subscription records (meaning that the email on the subscription can't be linked to a jobseeker by a hacker).

We'll not encrypt columns we want to search on, index by, or validate uniqueness of, because that would require a 'blind index', which would add too much complexity.

There is a single master key environment variable added for this commit, which is used to generate individual keys for each column (so that cracking one column doesn't crack all other columns).

The reason for adding column encryption is to add a layer of defence for data that could be exploited for fraud such as national insurance numbers. Data is encrypted at rest in our database, so column encryption defends against the case where someone gets into the database but not into the app itself.

- Change columns (that we will drop in a later step) to allow null values, so that a later step is able to work. See ankane/lockbox#133

- Add task for migrating the data from the old to the new columns.

Note that Rails will ship with the ability to encrypt columns in the next major version (7). That could be years away, so this interim solution may last a while (if indeed we do migrate to the Rails-native method in the future). This issue describes how we can migrate away from Lockbox: ankane/lockbox#121

- Prevent _ciphertext attributes from being exported to BigQuery. We'll export just the decrypted attribute (which is anonymised), rather than, for example, exporting both 'email' and 'email_ciphertext'.


app/models/application_record.rb
```diff
@@ -7,14 +7,18 @@ class ApplicationRecord < ActiveRecord::Base
...
 
+  def attributes_except_ciphertext
+    attributes.reject { |k, _v| k.include?("_ciphertext") }
+  end
+
   private
 
   def event_data
-    { table_name: self.class.table_name }.merge(anonymised_attributes)
+    { "table_name" => self.class.table_name }.merge(anonymised_attributes)
```
david-mears-dfe, line 17: > Why this key should change from a symbol to a string?



It's merged with attributes, which are all string keys. It's less confusing when all the keys are the same type.

csutter (approved):

PR #3805, opened 2021-08-02
[TEVA-2751] Add column encryption with Lockbox: step C

david-mears-dfe:
- Drop unencrypted columns. These attributes are still available in Object.attributes (e.g. `jobseeker.attributes`)

- Add lockbox 'master key' for development to .env.local
 
- Update sanitise.sql
Given that different environments use different column
encryption keys for Lockbox, we can't hard-code values other
than null or empty string for ciphertext columns when
writing a script that could be used in any environment.

### DIFF
--- a/db/scripts/sanitise.sql
+++ b/db/scripts/sanitise.sql
@@ -2,21 +2,10 @@ TRUNCATE TABLE alert_runs;
 TRUNCATE TABLE emergency_login_keys;
 TRUNCATE TABLE sessions;
 
--- Columns encrypted with lockbox should be set to values that can be decrypted by the target environment.
--- These values are encrypted with the staging lockbox master key.
--- Change them if you are using this script for a different environment, as the key will be different.
--- There are multiple values in order to pass the different validations on each attribute.
-SET lorem_string_ciphertext = 'London'
-SET lorem_phone_number_ciphertext
-SET lorem_email_ciphertext
-SET lorem_teacher_reference_number_ciphertext
-SET lorem_national_insurance_number_ciphertext
-SET empty_string_ciphertext
-
 UPDATE employments
-       SET organisation='Example Organisation',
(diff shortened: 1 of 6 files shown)


db/scripts/sanitise.sql
```diff
@@ -53,26 +42,26 @@ UPDATE equal_opportunities_reports
...
-           phone_number='01234567890',  -- encrypt
-           teacher_reference_number='1234567',  -- encrypt
-           national_insurance_number='QQ 12 34 56 C', -- encrypt
-           personal_statement='Lorem ipsum dolor sit amet', -- encrypt
-           support_needed_details='', -- encrypt
-           close_relationships_details='', -- encrypt
-           further_instructions='', -- encrypt
-           rejection_reasons='', -- encrypt
-           gaps_in_employment_details='', -- encrypt
+       SET first_name_ciphertext='',
```
csutter, line 50: Is there something I'm missing - shouldn't these be `NULL`?

david-mears-dfe: I'll check as it's been a while since I wrote this

csutter (approved):
See comment about `NULL`, 👍 -ing it either way pending your investigation.

db/scripts/sanitise.sql
david-mears-dfe: They work either way. I don't see a reason to prefer NULL to `''`.

david-mears-dfe: @csutter I'll wait to hear your reply tomorrow and leave this merge until then.

csutter: I would say `NULL` is more obvious in what it represents - absence of value rather than the value of "empty string". That said, given what these are (ciphertext columns) it should be obvious either way, so happy to keep it at `''`.

PR #3807, opened 2021-08-03
Add expanded versions of acronyms

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2917

## Comments

- The only use of the acronym QTS, as far as I could see, were on the professional status step of the job application form. The first instance of this acronym includes the 'expanded version', so no need to change them. There are some error messages that can be displayed on that page with the acronym in though, would the first of these be considered the first instance of the acronym?

### DIFF
--- a/config/locales/forms.yml
+++ b/config/locales/forms.yml
@@ -542,7 +542,7 @@ en:
         job_roles_options:
           teacher: Teacher
           leadership: Leadership
-          sen_specialist: SEN specialist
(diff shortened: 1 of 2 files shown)


alexbowen:
this wasnt a failiure in our accessibility compliance rather a usuability nice to have. so i think our effort should reflect that and not worry about form errors right now. i think that could be quite complex to fix

alexbowen (approved):

PR #3808, opened 2021-08-03
Convert bullet point symbols in job listing details to li tags

david-mears-dfe:
No ticket.

This is to solve the problem where (particularly) job adverts contain '•' but do not contain corresponding newlines, resulting in inline bullets.

## Before (from https://teaching-vacancies.service.gov.uk/jobs/estates-and-facilities-manager)

Screenshot 2021-08-03 at 14 12 14

## After

Screenshot 2021-08-03 at 14 11 12
Screenshot 2021-08-03 at 14 11 17

## Join the conversation

https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1627990485014100

### DIFF
--- a/app/presenters/vacancy_presenter.rb
+++ b/app/presenters/vacancy_presenter.rb
@@ -20,15 +20,15 @@ class VacancyPresenter < BasePresenter
   end
 
   def job_advert
-    simple_format(model.job_advert)
+    simple_format(fix_bullet_points(model.job_advert))
   end
 
   def about_school
-    simple_format(model.about_school)
+    simple_format(fix_bullet_points(model.about_school))
   end
 
   def school_visits
-    simple_format(model.school_visits) if model.school_visits.present?
+    simple_format(fix_bullet_points(model.school_visits)) if model.school_visits.present?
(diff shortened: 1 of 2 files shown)


app/presenters/vacancy_presenter.rb
```diff
@@ -79,4 +79,30 @@ def contract_type_with_duration
     duration = model.fixed_term? ? "(#{model.contract_type_duration})" : nil
     [type, duration].compact.join(" ")
   end
+
+  private
+
+  def fix_bullet_points(text)
```
alexbowen, line 85: should this look for asterisks also?

david-mears-dfe: Do people use asterisks in this way?

david-mears-dfe: Example! 
https://teaching-vacancies.service.gov.uk/jobs/teacher-of-biology-0bfc9e23-4386-414f-9cbd-939ec4fb629c

alexbowen (approved):

PR #3813, opened 2021-08-04
[TEVA- 2960] replace with current autocomplete with alphagov accessible version

alexbowen:
replaces existing autocomplete with accessible alphagov version

have tested on ie11 safari, firefox, chrome, edge, iphone 8, android

styling has changed slightly to the default for the alphagov module, @jesseyuen has verified this is fine

Screenshot 2021-08-05 at 12 50 32

https://dfedigital.atlassian.net/browse/TEVA-2960


app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -1,37 +1,46 @@
-import 'es6-promise/auto';
-import view from './autocomplete.view';
+import accessibleAutocomplete from 'accessible-autocomplete';
+import 'accessible-autocomplete/dist/accessible-autocomplete.min.css';
+import './autocomplete.scss';
 
-export const renderAutocomplete = (widgetParams) => {
-  view.create(widgetParams.container, widgetParams.input, widgetParams.key);
+const AUTOCOMPLETE_THRESHOLD = 3;
```
david-mears-dfe, line 5: Not immediately obvious what this number represents. Can we rename it to something a bit clearer?

app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -1,37 +1,46 @@
...
-
-export const removeUnwantedCharacters = (searchTerm) => searchTerm.replace('.', ' ');
+const highlightRefinement = (text, refinement) => {
+  const index = text.toLowerCase().indexOf(refinement.toLowerCase());
 
-export const showOptions = (getOptions, container, input, key) => {
-  const searchTerm = removeUnwantedCharacters(input.value);
-  return getOptions(searchTerm).then((options) => view.show(options, container, input, key));
-};
+  if (index >= 0) {
```
david-mears-dfe, line 10: Arguably you don't need this condition, right? If index is 0 we would get `text.substring(0, 0)` and that would (correctly) highlight nothing.

app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -1,37 +1,46 @@
...
+    /* eslint-disable */
```
david-mears-dfe, line 11: Would be good to know what rule is being disabled here - can you disable a specific rule in this way?

alexbowen: its line length. i couldnt get a multiline template to display without there being gaps between the highlighted part for some reason. so was displaying **Dur** ham

app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -1,37 +1,46 @@
...
+      formInput.parentNode.removeChild(formInput);
+
+      accessibleAutocomplete({
+        element: document.querySelector('#accessible-autocomplete'),
+        id: elementId,
+        name: 'location',
+        defaultValue: currentInputValue,
+        source: (query, populateResults) => {
+          currentInputValue = query;
+          return source({ query, populateResults });
```
david-mears-dfe, line 34: is `populateResults` something defined by accessible autocomplete?

alexbowen: yes it an argument passed into this callback

app/views/shared/search/_radius.html.slim
```diff
@@ -3,7 +3,7 @@
   :last,
   :first,
   label: { text: t("jobs.search.within_radius"), size: "s" },
-  class: "govuk-!-width-#{wide ? 'full' : 'one-third'}",
+  class: "govuk-!-width-#{wide ? 'full' : 'one-half'}",
```
david-mears-dfe, line 6: It looks like you've removed the need for us to have a `wide` variable for two out of three search input partials (the other two: keyword and location). So it would be nice to stop setting the variable where it's not needed:

https://github.com/DFE-Digital/teaching-vacancies/blob/be355535078eff4f3254e925beb63efc59d209b9/app/views/home/_search.html.slim#L6
https://github.com/DFE-Digital/teaching-vacancies/blob/be355535078eff4f3254e925beb63efc59d209b9/app/views/home/_search.html.slim#L7
https://github.com/DFE-Digital/teaching-vacancies/blob/be355535078eff4f3254e925beb63efc59d209b9/app/views/vacancies/_search.html.slim#L50
https://github.com/DFE-Digital/teaching-vacancies/blob/be355535078eff4f3254e925beb63efc59d209b9/app/views/vacancies/_search.html.slim#L51

app/frontend/src/components/autocomplete/autocomplete.js
alexbowen: although you could call it whatever you want, this is name in their docs, so ill leave it as is in case someone ever cross references the two

david-mears-dfe:
The main thing preventing me from approving is a question to Jesse on the Slack thread about how accessible it is to have the selection and background share the same blue colour.

alexbowen:
i have 'fixed' the font to force it to GDS @david-mears-dfe

app/views/shared/search/_radius.html.slim
alexbowen: good point yeah can remove these now 👍

app/frontend/src/components/autocomplete/autocomplete.scss
```diff
@@ -1,273 +1,29 @@
...
+    background-color: #fff;
+    font-family: 'GDS Transport', arial, sans-serif;
+    height: 2.5rem;
   }
 
-  .no-js .autocomplete__link:focus {
-    color: #0b0c0c;
+  .autocomplete__option--focused,
+  .autocomplete__option:hover {
+    background-color: #c2dff4;
```
alexbowen, line 22: this is custom color hence no GDS variable

david-mears-dfe (approved):
LG™️

PR #3820, opened 2021-08-05
[TEVA-2981] Fix bug caused by nil expires_at

david-mears-dfe:
This bug occurred when a hiring staff/publisher filled out 'applying for the job' and
enabled applications through TV, without first having set an expires_at time.

- Added regression test
- Stopped showing number of applications of each type when vacancy is in draft state
- Corrected a translation - we had been labelling expires_at with 'Date to be listed',
which is not what expires_at means.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2981

### Before

Screenshot 2021-08-05 at 15 39 26

### After (for a draft job and then for a published job)

Screenshot 2021-08-05 at 15 39 42
Screenshot 2021-08-05 at 15 40 13

### DIFF
--- a/app/views/publishers/vacancies/job_applications/_banner.html.slim
+++ b/app/views/publishers/vacancies/job_applications/_banner.html.slim
@@ -16,31 +16,31 @@
 
   - if vacancy.enable_job_applications?
     .govuk-grid-row.job-applications-summary
-      .govuk-grid-column-one-quarter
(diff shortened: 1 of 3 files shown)


alexbowen (approved):

david-mears-dfe:
https://teaching-vacancies-review-pr-3820.london.cloudapps.digital/

PR #3829, opened 2021-08-09
[TEVA-2991] Include 'how to apply' in the review step

david-mears-dfe:
We omitted to include this information in the review step after creating a vacancy.

Also, for local authorities, there is no option to enable job applications, so we should not display the 'Apply through Teaching Vacancies: No' section (as they have not answered this question/cannot change this).

We should show the application link for local authorities, too; so the check that 'enable_job_applications' was false is updated to check that it's nil or false (since local authority vacancies have this value as nil).

https://dfedigital.atlassian.net/browse/TEVA-2991

## Before (for local authority)

Screenshot 2021-08-09 at 15 42 34

## After (for local authority)

Screenshot 2021-08-09 at 15 47 16

### DIFF
--- a/app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
+++ b/app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
@@ -7,22 +7,28 @@
 
   - review.body do
     = govuk_summary_list do |component|
-      - component.slot(:row,
-        key: t("jobs.enable_job_applications"),
-        value: @vacancy.enable_job_applications? ? "Yes" : "No",
-        html_attributes: { id: "enable_job_applications" })
+      - unless current_organisation.group_type == "local_authority"
+        - component.slot(:row,
+          key: t("jobs.enable_job_applications"),
+          value: @vacancy.enable_job_applications? ? "Yes" : "No",
+          html_attributes: { id: "enable_job_applications" })
 
       - if @vacancy.enable_job_applications?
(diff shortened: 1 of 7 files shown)


app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
```diff
@@ -7,23 +7,29 @@
 
   - review.body do
     = govuk_summary_list do |component|
-      - component.slot(:row,
-        key: t("jobs.enable_job_applications"),
-        value: @vacancy.enable_job_applications? ? "Yes" : "No",
-        html_attributes: { id: "enable_job_applications" })
+      - unless current_organisation.group_type == "local_authority"
```
cesidio, line 10: you can shorten this to:

`- unless current_organisation.local_authority?

spec/support/vacancy_helpers.rb
```diff
@@ -112,6 +112,19 @@ def fill_in_copy_vacancy_form_fields(vacancy)
   end
 
   def verify_all_vacancy_details(vacancy)
+    vacancy.reload
+    vacancy = VacancyPresenter.new(vacancy) unless vacancy.is_a?(VacancyPresenter)
+
+    unless vacancy.parent_organisation.school?
+      if vacancy.job_location == "at_multiple_schools"
```
cesidio, line 119: you can shorten to:

`if vacancy.at_multiple_schools?`

cesidio (approved):

PR #3849, opened 2021-08-12
Drop redundant NQT suitable column

david-mears-2:
This column has been repeatedly [dropped](https://github.com/DFE-Digital/teaching-vacancies/pull/1603) and [readded](https://github.com/DFE-Digital/teaching-vacancies/pull/1858) depending on the flavour of the day of what filters we offer jobseekers on the search page (i.e. whether we consider NQT suitability part of job roles or something separate) and whether we ask the NQT question separately from the job roles question when vacancies are created.

We are going to add another job role soon which will, like NQT suitable, be a separate jobseeker-facing filter on the search page.

It's best if this type of data lives in the job_roles array column, rather than being duplicated into a boolean column as well. Our search ignores the boolean column and uses the job_roles array. The array data model is the most flexible; it need not map perfectly to however the search filters or questions happen to work this month, because that will inevitably vary in the future.

### DIFF
--- a/spec/support/vacancy_helpers.rb
+++ b/spec/support/vacancy_helpers.rb
@@ -24,7 +24,7 @@ module VacancyHelpers
             visible: false
     end
 
-    vacancy.job_roles&.each do |job_role|
+    vacancy.job_roles.excluding("nqt_suitable")&.each do |job_role|
       check I18n.t("helpers.label.publishers_job_listing_job_details_form.job_roles_options.#{job_role}"),
             name: "publishers_job_listing_job_details_form[job_roles][]",
             visible: false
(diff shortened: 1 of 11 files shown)


spec/support/vacancy_helpers.rb
```diff
@@ -24,7 +24,7 @@ def fill_in_job_details_form_fields(vacancy)
             visible: false
     end
 
-    vacancy.job_roles&.each do |job_role|
+    (vacancy.job_roles - ["nqt_suitable"])&.each do |job_role|
```
cesidio, line 27: how about

`vacancy.job_roles.excluding("nqt_suitable")&.each do |job_role|`

cesidio: probably you can drop the safe navigation too

cesidio (approved):
It all looks good to me. The only thing that we may want to address is this redundancy:

Screenshot 2021-08-12 at 14 29 05

PR #3852, opened 2021-08-12
Fix NQT suitable bugs (option 2)

david-mears-dfe:
The NQT suitable 'no' option was selected by default on creating a job listing
after we dropped the boolean column for NQT suitable.

This solution (option 2 of 2) fixes it by introducing a 'nqt_not_suitable' job role which records the user's yes/no answer the the NQT suitable question. This lets us distinguish between yeses, nos, and 'not yet answered's. It also requires a migration task for already-published vacancies, or they will see an 'action required' info box if they edit the vacancy.


csutter:
Closed in favour of #3851.

PR #3861, opened 2021-08-17
[TEVA-2959] add gds focus style to filter remove buttons

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2959

Screenshot 2021-08-17 at 09 20 54

### DIFF
--- a/app/components/filters_component/filters_component.scss
+++ b/app/components/filters_component/filters_component.scss
@@ -236,6 +236,10 @@
           width: 0;
         }
 
+        &:focus {
+          outline: $govuk-focus-width solid $govuk-focus-colour;
+        }
+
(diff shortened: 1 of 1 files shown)


app/components/filters_component/filters_component.scss
```diff
@@ -236,6 +236,10 @@
           width: 0;
         }
 
+        &:focus {
+          outline: 3px solid $govuk-focus-colour;
```
csutter, line 240: Perhaps `$govuk-focus-width`?

alexbowen: didnt realise there was such a thing

csutter (approved):

PR #3863, opened 2021-08-17
[TEVA-3024] improve accessibilty markup for footer links

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-3024

- makes the screen reader readout a bit more sensible
- puts the brackets into pseudo content as i feel would get very repetitive for screenreader with the amount of link t0 tab through

Screenshot 2021-08-17 at 10 01 06


csutter (approved):
This'll need a test change as I think those are still looking for the brackets - other than that, 👍

PR #3880, opened 2021-08-19
refactor of filters js

alexbowen:
no ticket

there was an apparent problem flagged on rollbar yesterday that PR issued and merged yesterday seems to have resolved but i wasnt happy with the JS when i was going over it at that point. i didnt feel it was very clear and there was repetition that i wanted to tidy up

- tried to make the init function clearer what it was initialising
- any functions that the init function calls are from within forEach blocks on doc.getElByClassname and so dont need to repeat that around the script
- general renaming and consolidation of some a11y related functions into `do this for mobile/do this for desktop` type functions to try make things clearer

### DIFF
--- a/app/components/filters_component/filters_component.js
+++ b/app/components/filters_component/filters_component.js
@@ -1,6 +1,5 @@
 import '../../frontend/src/lib/polyfill/closest.polyfill';
 import '../../frontend/src/lib/polyfill/from.polyfill';
-import logger from '../../frontend/src/lib/logging';
 import 'classlist-polyfill';
 
 export const CHECKBOX_CLASS_SELECTOR = 'govuk-checkboxes__input';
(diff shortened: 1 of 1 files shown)


app/components/filters_component/filters_component.js
```diff
@@ -19,99 +18,90 @@ window.addEventListener(
 );
 
 export const init = (removeButtonSelector, clearButtonId, closeButtonId, showFilterPanelId) => {
+  const showFilterPanelEl = document.getElementById(showFilterPanelId);
+  const closeFilterPanelEl = document.getElementById(closeButtonId);
+  const clearFiltersEl = document.getElementById(clearButtonId);
+
```
josephhull676, line 24: Perhaps not necessary, but maybe it's a bit easier to read if they're in the same order that they're in when passed into the init function?

josephhull676 (approved):

PR #3886, opened 2021-08-20
[TEVA-2871] Job roles changes in job listing journey

csutter:
### Introduce "job roles" step and primary job role
- Create new "job role" step at the beginning of job creation flow
- Move job roles from "job details" step to new step
- Introduce concept of a primary job role, which is selected and
  persisted into `job_roles` in this step, and turn checkboxes into
  radios as only one can be selected in this step (implemented as
  a virtual attribute on `Vacancy` which munges the `job_roles` array
  until we improve the data schema)
- Move job role into its own section on review page
- Add new "Education support" and "SENDCo" primary job roles
- Remove ability to "save and return" until job details step has been
  completed

### Introduce "job role details" step and additional job roles

- Change existing "SEN specialist" job role to "SEND responsible"
- Add virtual attribute for additional job roles to `Vacancy`
- Create a new "job role details" sub-step of "job roles" step
- Create a form model dealing with additional job roles (and a
  virtual attribute to handle the "SEND responsible" role being
  yes/no radio buttons
- Show "job role details" step as part of "job roles" step unless
  the primary job role is a SENDCo
- Show check boxes for additional job roles if primary job role is
  teacher
- Show yes/no radio buttons for SEND responsibilities if primary
  job role is leadership or education support

### Update job display page and filters to reflect new roles

- Change "job role" search filters to only include primary roles
- Fix translations (and keep old job_roles_options translations for
  now to avoid having to change use everywhere)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2871

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -6,7 +6,15 @@ class Vacancy < ApplicationRecord
 
   friendly_id :slug_candidates, use: %w[slugged history]
 
-  array_enum job_roles: { teacher: 0, leadership: 1, sen_specialist: 2, nqt_suitable: 3 }
+  # Each vacancy must have *exactly* one primary job role. It may have zero or multiple
+  # additional job roles. Legacy vacancies *may* have more than one primary job role as
+  # we used to allow multiple.
+  # TODO: This is a compromise to keep changes to the data model minimal for now. Once
+  # the legacy vacancies are gone, we should refactor the data model.
+  PRIMARY_JOB_ROLES =  { teacher: 0, leadership: 1, education_support: 4, sendco: 5 }.freeze
+  ADDITIONAL_JOB_ROLES = { sen_specialist: 2, nqt_suitable: 3 }.freeze
+  array_enum job_roles: PRIMARY_JOB_ROLES.merge(ADDITIONAL_JOB_ROLES)
+
   array_enum working_patterns: { full_time: 0, part_time: 100, job_share: 101 }
   # Legacy vacancies can have these working_pattern options too: { compressed_hours: 102, staggered_hours: 103 }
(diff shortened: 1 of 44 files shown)


app/models/vacancy.rb
```diff
@@ -6,7 +6,15 @@ class Vacancy < ApplicationRecord
...
   friendly_id :slug_candidates, use: %w[slugged history]
 
-  array_enum job_roles: { teacher: 0, leadership: 1, sen_specialist: 2, nqt_suitable: 3 }
+  # Each vacancy must have *exactly* one primary job role. It may have zero or multiple
+  # additional job roles. Legacy vacancies *may* have more than one primary job role as
+  # we used to allow multiple.
+  # TODO: This is a compromise to keep changes to the data model minimal for now. Once
+  # the legacy vacancies are gone, we should refactor the data model.
+  PRIMARY_JOB_ROLES =  { teacher: 0, leadership: 1, education_support: 4, sendco: 5 }.freeze
+  ADDITIONAL_JOB_ROLES = { sen_specialist: 2, nqt_suitable: 3 }.freeze
```
josephhull676: These are the legacy job roles right, perhaps LEGACY_JOB_ROLES makes more sense as a name then??

csutter: They're not legacy, they are still applicable - the difference is that every job now _must_ have exactly one of `PRIMARY_JOB_ROLES`, and it _may_ have none, one, or both of the `ADDITIONAL_JOB_ROLES` on top 💡

csutter: `sen_specialist` gets renamed to the new thing in a later commit

alexbowen (approved):

PR #3887, opened 2021-08-20
[TEVA-3007] fix missing notification count in main nav

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-3007

with hidden text so screen reader reads `notifications 2000 unread`

Screenshot 2021-08-20 at 14 27 42

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -47,9 +47,10 @@ html.govuk-template.app-html-class lang="en"
               - if current_organisation.school_group?
                 - navigation.item link_text: t("nav.school_group_index_link", organisation_type: organisation_type_basic(current_organisation)), align: :left, path: organisation_schools_path
               - navigation.item link_text: t("nav.jobseekers_index_link"), align: :left, path: root_path
-              - navigation.item link_text: t("nav.notifications_index_link"), aria: { label: "#{current_publisher.notifications.unread.count} unread notifications" }, align: :left, path: publishers_notifications_path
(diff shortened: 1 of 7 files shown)


app/views/layouts/application.html.slim
```diff
@@ -47,9 +47,10 @@ html.govuk-template.app-html-class lang="en"
               - if current_organisation.school_group?
                 - navigation.item link_text: t("nav.school_group_index_link", organisation_type: organisation_type_basic(current_organisation)), align: :left, path: organisation_schools_path
               - navigation.item link_text: t("nav.jobseekers_index_link"), align: :left, path: root_path
-              - navigation.item link_text: t("nav.notifications_index_link"), aria: { label: "#{current_publisher.notifications.unread.count} unread notifications" }, align: :left, path: publishers_notifications_path
-              - if current_publisher.notifications.unread.any?
-                span.notification-badge = current_publisher.notifications.unread.count
+              - navigation.item link_text: t("nav.notifications_index_link_html", count: current_publisher.notifications.unread.any? ? current_publisher.notifications.unread.count : 0),
```
csutter, line 50: Does `current_publisher.notifications.unread.count` not return 0 anyway if there are none, meaning we can do away with the conditional?

csutter (approved):

app/views/layouts/application.html.slim
alexbowen: if it doesnt it should i think

PR #3896, opened 2021-08-23
[TEVA-2873] add changes to apply form for education support role

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2873

### DIFF
--- a/app/views/publishers/vacancies/build/applying_for_the_job.html.slim
+++ b/app/views/publishers/vacancies/build/applying_for_the_job.html.slim
@@ -13,7 +13,18 @@
 
         h2.govuk-heading-l = t("publishers.vacancies.steps.applying_for_the_job")
 
-        - if vacancy.listed? || current_organisation.local_authority?
+        - if vacancy.primary_job_role == "education_support"
+          details class="govuk-details" data-module="govuk-details"
+            summary.govuk-details__summary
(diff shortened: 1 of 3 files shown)


app/views/publishers/vacancies/build/applying_for_the_job.html.slim
```diff
@@ -13,7 +13,18 @@
 
         h2.govuk-heading-l = t("publishers.vacancies.steps.applying_for_the_job")
 
-        - if vacancy.listed? || current_organisation.local_authority?
+        - if vacancy.job_roles.include?("education_support")
```
csutter, line 16: If you rebase, you should be able to do `vacancy.primary_job_role == "education_support"` instead

spec/support/vacancy_helpers.rb
```diff
@@ -74,7 +74,7 @@ def upload_document(form_id, input_name, filepath)
   end
 
   def fill_in_applying_for_the_job_form_fields(vacancy, local_authority_vacancy: false)
-    if !local_authority_vacancy && vacancy.enable_job_applications?
+    if !local_authority_vacancy && vacancy.enable_job_applications? && !vacancy.job_roles.include?("education_support")
```
csutter, line 77: Ditto

csutter (approved):
See comments but 👍

PR #3898, opened 2021-08-24
Implement filter for SEND responsibilities

csutter:
Note that this does not yet include the work identified as part of
implementing this change to make the "primary" job role filters
work independently of the NQT/SEND filters (rather than being one
big OR across all of them).

- Add SEND responsibilities filter to search filters
- Add SEND responsibilities filter to job alert creation filters
- Change wording of filters

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2872

### DIFF
--- a/app/views/vacancies/_search.html.slim
+++ b/app/views/vacancies/_search.html.slim
@@ -13,12 +13,21 @@ ruby:
     {
       title: t("jobs.filters.nqt_suitable"),
       key: "nqt_suitable",
-      hidden_text: t("jobs.filters.nqt_suitable_hidden_prefix"),
       attribute: :job_roles,
       selected: @form.job_roles.include?("nqt_suitable") ? @form.job_roles : [],
       options: @form.nqt_suitable_options,
       value_method: :first,
-      text_method: :last,
+      text_method: :second,
+      selected_method: :last,
+    },
+    {
+      title: t("jobs.filters.send_responsible"),
+      key: "send_responsible",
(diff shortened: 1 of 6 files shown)


app/views/vacancies/_search.html.slim
```diff
@@ -13,12 +13,21 @@ ruby:
     {
       title: t("jobs.filters.nqt_suitable"),
       key: "nqt_suitable",
-      hidden_text: t("jobs.filters.nqt_suitable_hidden_prefix"),
```
alexbowen, line 16: is @jesse ok with this? i seem to remember i added it after discussion with himn so the labels and text in the remove buttons wasnt so verbose and hence visually a bit ugly

csutter: @alexbowen Excellent spot! The new label text includes the phrasing previously in the hidden prefix, so it's always part of the label:
```

alexbowen: if sticking with the visually hidden prefix, can be bit more generic than `nqt_suitable_hidden_prefix` for sure, but the ui started looking a bit odd with the super long labels but needed to be descriptive for screen readres

csutter: I do think it's a bit long, but the new design definitely has the whole "Only show results" shebang visible to all users, with nothing only for screenreaders

alexbowen (approved):

PR #3902, opened 2021-08-24
Add working patterns step

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2876

## Changes in this PR:

- Added working patterns step with working patterns and working patterns details fields
- Added 'Term time' working pattern
- Added error message for working patterns field
- Added step to show pages, review page and edit pages
- Added 'Term time' filter

### Working patterns step

Screenshot 2021-09-03 at 12 20 37

### Working patterns step errors

Screenshot 2021-09-03 at 12 20 12

### Review page

Screenshot 2021-09-03 at 12 22 27

### Manage listing / edit page

Screenshot 2021-09-03 at 12 25 21

### Jobseeker view

Screenshot 2021-09-03 at 12 25 52

### Term time fliter

Screenshot 2021-09-03 at 12 31 06

### DIFF
--- a/app/form_models/publishers/job_listing/working_patterns_form.rb
+++ b/app/form_models/publishers/job_listing/working_patterns_form.rb
@@ -1,0 +1,5 @@
+class Publishers::JobListing::WorkingPatternsForm < Publishers::JobListing::VacancyForm
+  attr_accessor :working_patterns, :working_patterns_details
+
+  validates :working_patterns, presence: true, inclusion: { in: Vacancy.working_patterns.keys }
+end
--- a/app/views/publishers/vacancies/build/working_patterns.html.slim
+++ b/app/views/publishers/vacancies/build/working_patterns.html.slim
@@ -1,0 +1,24 @@
+- content_for :page_title_prefix, page_title_prefix(vacancy, form, t("publishers.vacancies.steps.working_patterns"))
+
+.govuk-main-wrapper
+  .govuk-grid-row
+    .govuk-grid-column-full
+      = render Publishers::VacancyFormPageHeadingComponent.new(vacancy, current_step_number, total_steps(steps_config))
+      = govuk_back_link text: session[:current_step] == :review ? t("buttons.cancel_and_return") : t("buttons.back_to_previous_step"), href: back_path
+
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      = form_for form, url: wizard_path, method: :patch do |f|
+        = f.govuk_error_summary
+
+        h2.govuk-heading-l = t("publishers.vacancies.steps.working_patterns")
+
+        = f.govuk_collection_check_boxes :working_patterns, Vacancy.working_patterns.keys, :to_s, :to_s
+
+        = f.govuk_text_area :working_patterns_details, label: { size: "s", id: "working-pattern-details-label" }
+
+        = render "publishers/vacancies/vacancy_form_partials/submit", f: f
+
+    - unless vacancy.published?
+      .govuk-grid-column-one-third
+        = render "steps"
--- a/config/locales/activerecord.yml
+++ b/config/locales/activerecord.yml
@@ -136,8 +136,10 @@ en:
       too_short: Job title must be at least %{count} characters
       too_long: Job title must not be more than %{count} characters
       invalid_characters: Job title must not contain any HTML tags
+  working_patterns_errors: &working_patterns_errors
     working_patterns:
       blank: Select a working pattern
+      inclusion: Select a working pattern
   pay_package_errors: &pay_package_errors
     salary:
       blank: Enter a salary
@@ -247,6 +249,9 @@ en:
         publishers/job_listing/job_details_form:
           attributes:
             <<: *job_details_errors
+        publishers/job_listing/working_patterns_form:
+          attributes:
+            <<: *working_patterns_errors
         publishers/job_listing/pay_package_form:
           attributes:
             <<: *pay_package_errors
@@ -500,6 +505,7 @@ en:
             <<: *job_location_errors
             <<: *schools_errors
             <<: *job_details_errors
+            <<: *working_patterns_errors
             <<: *pay_package_errors
             <<: *important_dates_errors
             <<: *applying_for_the_job_errors
--- a/config/locales/forms.yml
+++ b/config/locales/forms.yml
@@ -227,9 +227,6 @@ en:
           For secondary school roles include subject and, if relevant,
           level of seniority (‘Subject leader for science’, for example).
         subjects: What subject(s) will the teacher focus on?
-        working_patterns: >-
-          Select all working patterns you will consider for the role.
-          Flexible working options may attract more candidates.
       publishers_job_listing_job_role_form:
         primary_job_role_options:
           leadership: Includes Head of Year/Subject, Deputy/Assistant Head, Headteacher, and trust-level roles
@@ -254,6 +251,9 @@ en:
       publishers_job_listing_schools_form:
         add_school: Add a school to your account
         edit_schools: "Can't see the school you are looking for?"
+      publishers_job_listing_working_patterns_form:
+        working_patterns: Select all working patterns you will consider for the role. Flexible working options may attract more candidates
+        working_patterns_details: For example, '20 hours a week' or 'Tuesday and Wednesday mornings'
 
     label:
       general_feedback_form:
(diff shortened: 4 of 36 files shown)


app/form_models/publishers/job_listing/working_patterns_form.rb
```diff
@@ -0,0 +1,5 @@
+class Publishers::JobListing::WorkingPatternsForm < Publishers::JobListing::VacancyForm
+  attr_accessor :working_patterns, :working_patterns_details
+
+  validates :working_patterns, presence: true
```
david-mears-dfe, line 4: Could validate inclusion in the set of possible working patterns, too. We tend to validate inclusion for checkbox/radio button fields, I would guess 'to be on the safe side' (a user could put whatever parameters they liked into the post url, and then in a job listing we'd be rendering user-input data that we don't expect to come from a user, which could be dangerous)

app/views/publishers/vacancies/build/working_patterns.html.slim
```diff
@@ -0,0 +1,24 @@
+- content_for :page_title_prefix, page_title_prefix(vacancy, form, t("publishers.vacancies.steps.pay_package"))
```
david-mears-dfe, line 1: Wrong title mate

david-mears-dfe: It's not pay_package

config/locales/activerecord.yml
```diff
@@ -247,6 +248,9 @@ en:
         publishers/job_listing/job_details_form:
           attributes:
             <<: *job_details_errors
+        publishers/job_listing/working_patterns_form:
+          attributes:
+            <<: *working_patterns_errors
```
david-mears-dfe, line 253: Needs to be added around line 500 too. That's why this is assigned to a variable using the `&`: so it can be reused without duplication.

config/locales/forms.yml
```diff
@@ -582,6 +578,13 @@ en:
         actual_salary_html: Actual salary (optional<span class='govuk-visually-hidden'> field</span>)</span>
         benefits_html: Additional allowances (optional<span class='govuk-visually-hidden'> field</span>)</span>
         salary: Annual or full-time equivalent (FTE) salary
+      publishers_job_listing_working_patterns_form:
+        working_patterns_details_html: How many hours or days does the role cover? (optional<span class='govuk-visually-hidden'> field</span>)</span>
```
david-mears-dfe, line 582: It's simpler to use `, form_group: { classes: "optional-field" }`, which is a class we added so that we did not have to include html in the translations.

db/schema.rb
```diff
@@ -466,6 +466,7 @@
     t.boolean "enable_job_applications"
     t.string "completed_steps", default: [], null: false, array: true
     t.string "actual_salary"
+    t.text "working_patterns_details"
```
david-mears-dfe, line 469: This is more of a general question than a critique of what you've decided here, but is there any reason that we should ever use `string` instead of `text`? There is some discussion [here](https://stackoverflow.com/questions/3354330/difference-between-string-and-text-in-rails). There is no performance penalty in postgres for using text instead of string, but we might want to change database one day.

spec/form_models/publishers/job_listing/job_details_form_spec.rb
```diff
@@ -7,8 +7,6 @@
   it { is_expected.to allow_value("Job &amp; another job").for(:job_title) }
   it { is_expected.not_to allow_value("Title with <p>tags</p>").for(:job_title).with_message(I18n.t("job_details_errors.job_title.invalid_characters")) }
 
-  it { is_expected.to validate_presence_of(:working_patterns) }
-
   # it { is_expected.to validate_inclusion_of(:contract_type).in_array(Vacancy.contract_types.keys) }
```
david-mears-dfe, line 10: I wonder why this test is commented out. When I run it, it passes. Could you uncomment it please? (I know it wasn't you.)

spec/presenters/subscription_presenter_spec.rb
```diff
@@ -35,7 +35,7 @@
       let(:search_criteria) { { working_patterns: %w[part_time] } }
 
       it "formats and returns the working pattern" do
-        expect(presenter.filtered_search_criteria["working_patterns"]).to eq("Part-time")
+        expect(presenter.filtered_search_criteria["working_patterns"]).to eq("Part time")
```
david-mears-dfe, line 38: Are we deliberately unhyphenating them? If so there are more cases of hyphenating this lying around in the translations.

spec/presenters/vacancy_presenter_spec.rb
```diff
@@ -112,12 +112,24 @@
...
 
-    it "returns a working patterns string if working_patterns is set" do
-      vacancy = VacancyPresenter.new(create(:vacancy, working_patterns: %w[full_time part_time]))
-      vacancy.organisation_vacancies.create(organisation: create(:school, name: "Smith High School"))
+    context "when only working_patterns is set" do
+      it "returns a string only containing the working pattern" do
+        vacancy = VacancyPresenter.new(create(:vacancy, working_patterns: %w[full_time part_time], working_patterns_details: nil))
+        vacancy.organisation_vacancies.create(organisation: create(:school, name: "Smith High School"))
+
+        expect(vacancy.show_working_patterns).to eq(I18n.t("jobs.working_patterns_info_many", patterns: "full-time, part-time"))
```
david-mears-dfe, line 120: I don't think you need to do this unless you feel like it, but it annoys me that we are not using I18n's intelligent counting thing where you can say
```
working_patterns_info:
  one: "Blah blah"
  many: "Blah blah but different"
```
and let it choose which one to use based on the `count` parameter.

config/locales/forms.yml
csutter: I thought we moved away from that? 😕  Paging @alexbowen

db/schema.rb
csutter: I (personally) prefer using both `string` and `text` to imply that one takes a bit of text, and the other a lot - but that's mostly because I came of age in ancient times when MySQL was all the rage and DB space was at a premium 👴 

I think using `text` across the board is probably fine.

spec/form_models/publishers/job_listing/job_details_form_spec.rb
csutter: @josephhull676 I feel like you asked that just the other week and I didn't have an answer for you 🤣 

Agree we should just uncomment it.

app/form_models/publishers/job_listing/working_patterns_form.rb
josephhull676: I believe I used inclusion at first but had problems with. I'll give it another go.

app/views/publishers/vacancies/build/working_patterns.html.slim
josephhull676: Whoopsy!

config/locales/forms.yml
josephhull676: Moved away from which one? The approach I've gone with or David's? 😅

db/schema.rb
josephhull676: I went with text as I saw other text_area fields had their corresponding attributes set to this type. From looking at the post you linked to I can see string maps to the type VARCHAR(255) in MySQL, with a limit of 255 characters I assume? text maps to a type with a limit of 4294967296, as far as I can see from that post?

I read that string, when mapped to a datatype in postgres (character varying(n)) has a much bigger limit, is that correct?

If that's the case and I'm understanding things correctly, if we're concerned about problems arising from changing database, surely using text makes sense (if the same limit exists to the string's mapped types in other databases)?

josephhull676: Please let me know if I haven't made any sense there 🤣

spec/form_models/publishers/job_listing/job_details_form_spec.rb
josephhull676: Sweet! 

And yes you did @csutter 😂

spec/presenters/subscription_presenter_spec.rb
josephhull676: I unhyphenated them because that's how they appear in the prototype. I'll ask Jesse!


Screenshot 2021-09-03 at 17 18 03

csutter (approved):

PR #3926, opened 2021-09-03
show all roles with vacancies in footer links

alexbowen:
no ticket

change criteria for showing fotter role links to show all roles that have a count.

### DIFF
--- a/app/services/vacancy_facets.rb
+++ b/app/services/vacancy_facets.rb
@@ -2,7 +2,7 @@ class VacancyFacets
   CACHE_DURATION = 24.hours
 
   def job_roles
-    cached(:job_roles) { sort_and_limit(job_role_facet.select { |role| role.in?(Vacancy.primary_job_role_options) }, 4) }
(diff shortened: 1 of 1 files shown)


app/services/vacancy_facets.rb
```diff
@@ -2,7 +2,7 @@ class VacancyFacets
   CACHE_DURATION = 24.hours
 
   def job_roles
-    cached(:job_roles) { sort_and_limit(job_role_facet.select { |role| role.in?(Vacancy.primary_job_role_options) }, 4) }
+    cached(:job_roles) { job_role_facet.select { |role, count| role.in?(Vacancy.primary_job_role_options) && count.positive? } }
```
david-mears-dfe, line 5: We can still `sort_and_limit` though, in order to have the most popular listed first. Also, `sort_and_limit` checks whether count is positive. We just need to set `number_of_results` to the number of primary_job_role_options.

cesidio (approved):

david-mears-dfe (approved):

PR #3936, opened 2021-09-06
Remove job creation "save and return" functionality

csutter:
This replaces the existing "save and return" functionality when
creating a job listing as a publisher with a "cancel and return" link,
removing the need to persist possibly invalid data and simplifying the
way our forms work, while making the experience less confusing for our
users.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3082

## Screenshots of UI changes:

image


app/views/publishers/vacancies/build/job_location.html.slim
```diff
@@ -15,7 +15,10 @@
 
         = f.govuk_collection_radio_buttons :job_location, t("helpers.options.publishers_job_listing_job_location_form.job_location.#{current_organisation.group_type}"), :first, :last
 
-        = f.govuk_submit t("buttons.continue")
+        .govuk-button-group
```
csutter, line 18: This is a little bit of duplication on two step views for now (this one and `job_role`, with the plan being to refactor once we know whether or not all submit buttons will just say "Continue" from now on)

cesidio (approved):
🔥

alexbowen (approved):

PR #3938, opened 2021-09-07
[TEVA-3069] Bug related to changing published vacancy's job role

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3069

- When a vacancy is edited, we validate_all_steps. One of these validations, on the applying_for_job_form, checks that the value of enable_job_applications is one of [true, false, "true", "false"]. 

- When a job role has been selected that doesn't allow applications, the enable_job_applications field is not rendered, so enable_job_applications is nil. In this case the validation for enable_job_applications is not run as vacancy.allow_enabling_job_applications? = false

- When a hiring staff user changes the job role to one that does accept applications, vacancy.allow_enabling_job_applications? = true, so the validation is run. However, as the value of enable_job_applications is still nil, the validation does not pass, causing an error to be displayed.

## Changes in this PR

- Added a before_validation callback to set enable_job_applications to false if the current user is from an LA, the vacancy is listed and enable_job_applications is nil.

## Error message displayed previously

Screenshot 2021-09-07 at 11 00 46

### DIFF
--- a/app/form_models/publishers/job_listing/applying_for_the_job_form.rb
+++ b/app/form_models/publishers/job_listing/applying_for_the_job_form.rb
@@ -1,7 +1,7 @@
 class Publishers::JobListing::ApplyingForTheJobForm < Publishers::JobListing::VacancyForm
   attr_accessor :application_link, :enable_job_applications, :contact_email, :contact_number, :personal_statement_guidance, :school_visits, :how_to_apply, :current_organisation
 
-  before_validation :override_enable_job_applications_for_local_authority!
+  before_validation :override_enable_job_applications!
 
   validates :enable_job_applications, inclusion: { in: [true, false, "true", "false"] }, if: proc { vacancy.allow_enabling_job_applications? }
   validates :how_to_apply, presence: true, unless: proc { enable_job_applications.in?(["true", true]) }
@@ -14,14 +14,16 @@ class Publishers::JobListing::ApplyingForTheJobForm < Publishers::JobListing::Va
 
   private
 
-  # If a publisher is signed in as an LA, we do not allow them to set the job applications feature.
-  # This forces the field to be set so the validations pass when submitting the form (despite the
-  # field not being there).
-  # TODO: Remove me (and conditional in view) when we start allowing applications feature for LAs
-  def override_enable_job_applications_for_local_authority!
-    return if params[:current_organisation].nil?
-    return unless params[:current_organisation].local_authority? && enable_job_applications.blank?
+  def override_enable_job_applications!
+    # If a publisher is signed in as an LA, we do not allow them to set the job applications feature.
+    # This forces the field to be set so the validations pass when submitting the form (despite the
+    # field not being there).
+    # TODO: Remove params[:current_organisation].local_authority? (and conditional in view) when we start allowing applications feature for LAs
 
-    self.enable_job_applications = false
+    # If a Publisher publishes a vacancy for a job role that does not allow enabling job applications
+    # but then changes the job role to one that does, enable_job_applications is nil, meaning the validation
+    # for this field does not pass. We want the validations for the enable_job_applications field to pass
+    # to prevent an error from being displayed on the review page in this situation when validate_all_steps is run.
+    self.enable_job_applications = false if (params[:current_organisation].local_authority? || vacancy.listed?) && enable_job_applications.blank?
   end
 end
--- a/spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
+++ b/spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
@@ -1,8 +1,9 @@
 require "rails_helper"
 
 RSpec.describe Publishers::JobListing::ApplyingForTheJobForm, type: :model do
-  subject { described_class.new(vacancy: vacancy) }
+  subject { described_class.new(current_organisation: organisation, vacancy: vacancy) }
 
+  let(:organisation) { build_stubbed(:trust) }
   let(:vacancy) { build_stubbed(:vacancy) }
 
   it { is_expected.to allow_value("https://www.this-is-a-test-url.tvs").for(:application_link) }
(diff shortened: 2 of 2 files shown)


app/form_models/publishers/job_listing/applying_for_the_job_form.rb
```diff
@@ -14,6 +14,11 @@ class Publishers::JobListing::ApplyingForTheJobForm < Publishers::JobListing::Va
 
   private
 
+  def override_enable_job_applications
+    override_enable_job_applications_for_local_authority!
+    override_enable_job_applications_when_vacancy_published! unless params[:current_organisation].local_authority?
```
david-mears-dfe, line 19: Do we need this `unless`?

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
```diff
@@ -24,4 +29,12 @@ def override_enable_job_applications_for_local_authority!
 
     self.enable_job_applications = false
   end
+
+  # If a Publisher publishes a vacancy for a job role that does not allow enabling job applications
+  # but then changes the job role to one that does, we don't want them to be able to allow applications
+  # but we do want the validations for the enable_job_applications field to pass to prevent an error
+  # from being displayed on the review page.
```
david-mears-dfe, line 36: Is it specifically about job roles? I thought it was just that
1) we don't want `enable_job_applications` to be able to change after a job is published
but
2) we do want the step to be valid

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
josephhull676: Well I just figured it saved executing (is that the correct term?)  `override_enable_job_applications_when_vacancy_published!` if the user is signed in as an LA, as enable_job_applications would've already been overridden

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
josephhull676: The step only becomes invalid when the job role is changed to a job role that allows enabling job applications as the validation is only run if `vacancy.allow_enabling_job_applications?`

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
```diff
@@ -24,4 +29,12 @@ def override_enable_job_applications_for_local_authority!
...
+  def override_enable_job_applications_when_vacancy_published!
+    self.enable_job_applications = false if vacancy.published? && self.enable_job_applications.nil?
```
csutter, line 38:
```ruby
self.enable_job_applications ||= false if vacancy.published?
```

josephhull676: I've made a change to this. It seems this works if `self.enable_job_applications` is nil, which it will be when the previous job role didn't allow applications (as the field isn't rendered). When validate_all_steps is run, the value of `enable_job_applications` is successfully overridden.

However, if the user then goes to the applying for the job step and changes something, the validation fails. This is because the value submitted for `enable_job_applications` is now a blank string (as this field is now a hidden field?). 

To fix this, instead of the above, it's now:

```
  def override_enable_job_applications_when_vacancy_listed!
    return if params[:current_organisation].nil?

    self.enable_job_applications = false if enable_job_applications.blank?
  end
```

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
david-mears-dfe: Isn't validation run whenever a hiring staff tries to edit a listing?

josephhull676: Yeah it is. But it's only invalid when they change job roles from one which doesn't accept applications to one that does.

If they select "Teaching assistant" for example, the field for enable_job_applications is not rendered because this returns false:
```
def allow_enabling_job_applications?
  %w[teacher leadership sendco].include?(main_job_role)
end
```
This then means that `enable_job_applications` is nil. The validation is not run because the condition (which uses the above method - see app/models/form_models/publishers/job_listing/applying_for_the_job.html.slim) evaluates to false.

If they then edit the vacancy and change the job role, when all of the steps are validated, this validation is now run. This is because `allow_enabling_job_applications?` now evaluates to true. The validation then fails as nil is not in `[true, false, "true", "false"]`

david-mears-dfe: Sorry, which validation is not run?

david-mears-dfe: Oh, I found it:

```
   validates :enable_job_applications, inclusion: { in: [true, false, "true", "false"] }, if: proc { vacancy.allow_enabling_job_applications? }
```

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
```diff
@@ -18,15 +19,14 @@
...
+    subject { described_class.new(current_organisation: organisation, enable_job_applications: "false", vacancy: vacancy) }
 
     it { is_expected.to validate_presence_of(:how_to_apply) }
   end
 
   describe "enable job applications override" do
-    subject { described_class.new(current_organisation: organisation, vacancy: vacancy) }
-
     context "when the current organisation given is a local authority" do
+      subject { described_class.new(current_organisation: organisation, vacancy: vacancy) }
```
david-mears-dfe, line 29: I don't think this line adds anything above what we already get from line 4.

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
```diff
@@ -38,14 +38,36 @@
     end
 
     context "when the current organisation given is not a local authority" do
-      let(:organisation) { build_stubbed(:trust) }
-      let(:vacancy) { build_stubbed(:vacancy, job_roles: %w[teacher]) }
+      subject { described_class.new(current_organisation: organisation, vacancy: vacancy) }
```
david-mears-dfe, line 41: Same here.

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
```diff
@@ -38,14 +38,36 @@
...
+        it "overrides enable_job_applications" do
+          expect { subject.valid? }.to change { subject.enable_job_applications }.from(nil).to(false)
+        end
+
+        it "does not validate enable_job_applications and how_to_apply" do
+          subject.valid?
 
-        expect(subject.enable_job_applications).to be_nil
-        expect(subject.errors).to include(:enable_job_applications)
+          expect(subject.errors).to_not include(:enable_job_applications)
```
david-mears-dfe, line 69: and how_to_apply

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
```diff
@@ -38,14 +38,36 @@
...
 
-      it "does not override enable_job_applications" do
-        subject.valid?
+      context "when the vacancy is in draft" do
+        let(:vacancy) { build_stubbed(:vacancy, :draft, job_roles: %w[teacher]) }
+
+        it "validates enable_job_applications and how_to_apply" do
+          subject.valid?
+
+          expect(subject.errors).to include(:enable_job_applications)
```
david-mears-dfe, line 49: and how_to_apply

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
david-mears-dfe: So either change the test description to exclude how_to_apply, or test for it

david-mears-dfe: It would be good to test that is was valid, too? `expect(subject).to be_valid`

app/form_models/publishers/job_listing/applying_for_the_job_form.rb
josephhull676: That's the one!

spec/form_models/publishers/job_listing/applying_for_the_job_form_spec.rb
josephhull676: It's not valid though as it does not provide values for the other fields. Should I provide values for the other fields, or should I just test for the absence of the error related to `enable_job_applications` ?

david-mears-dfe (approved):

PR #3939, opened 2021-09-07
[TEVA-3094] Do not use params[:commit] in controllers

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-3094

### DIFF
--- a/app/controllers/publishers/vacancies/job_applications_controller.rb
+++ b/app/controllers/publishers/vacancies/job_applications_controller.rb
@@ -54,12 +54,9 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
   end
 
   def status
-    case params[:commit]
-    when t("buttons.shortlist")
-      "shortlisted"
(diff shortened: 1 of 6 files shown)


app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -54,12 +54,9 @@ def job_application
   end
 
   def status
-    case params[:commit]
-    when t("buttons.shortlist")
-      "shortlisted"
-    when t("buttons.confirm_rejection")
-      "unsuccessful"
-    end
+    return "shortlisted" if params[:publishers_job_application_update_status_form].key?("further_instructions")
```
david-mears-dfe, line 57: This doesn't feel that abstract - is it possible to test the action_name or the path instead?

cesidio: the action_name is the same `update_status`

josephhull676 (approved):

PR #3943, opened 2021-09-08
Use step process for vacancy flow header

csutter:
- Change `VacancyFormPageHeadingComponent` to use step process instead
  of manually passed in numbers, and change usage everywhere
- Expose step process to step views
- Remove `Publishers::Vacancies::BaseController#current_step_number`
  and `#steps_adjust`
- Remove `VacanciesHelper#total_steps`
- Make `StepProcess` raise a more informative error if the current step
  that is passed in does not exist in the set of steps
- Take session into account when determining steps (needed for
  `job_location` step)

## Jira ticket URL

### DIFF
--- a/app/controllers/publishers/vacancies/base_controller.rb
+++ b/app/controllers/publishers/vacancies/base_controller.rb
@@ -3,14 +3,7 @@ require "indexing"
 class Publishers::Vacancies::BaseController < Publishers::BaseController
   include Publishers::Wizardable
 
-  helper_method :current_step_number, :step_current, :steps_adjust, :steps_config, :vacancy
-
-  def steps_adjust
-    # Only adjust *after* job role step as the extra step for school groups comes after that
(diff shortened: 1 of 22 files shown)


cesidio (approved):
much nicer!

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -25,8 +18,17 @@ def step_current
...
   end
 
-  def current_step_number
-    steps_config[step_current][:number] - steps_adjust
+  def step_process
+    # TODO: We currently have to do this kinda thing in a lot of places thanks to `wicked`
+    #       Find a better way!
+    current_step = defined?(step) ? step : :review
+
+    ::Publishers::Vacancies::VacancyStepProcess.new(
```
josephhull676, line 26: Why the extra :: before Publishers??

josephhull676 (approved):

PR #3950, opened 2021-09-09
Update vacancy factory with latest steps

david-mears-dfe:
We have added some steps since we wrote this. Hard to access the steps_config from within this file so hard-coding.


csutter (approved):

josephhull676 (approved):
Kinda weird for schools to be in the completed steps if the jobs at one school? I know that was there already but I only just noticed!

PR #3952, opened 2021-09-09
[TEVA-3025] Update the URL of the NQT job alert subscription form

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3025

## Changes in this PR:

- Added route for new ECT job alert URL
- Redirect from old URL
- Rename keys in params to use ECT instead of NQT 

## Comments: 

- Haven't changed the value in the job_roles array as the job role is still called nqt_suitable. I think this may change as part of the work on: https://dfedigital.atlassian.net/browse/TEVA-3075

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -204,28 +204,6 @@ en:
     sign_in: List a teaching job
     sign_out: Sign out
 
-  nqt_job_alerts:
-    confirm:
-      heading: What happens next
-      splash:
-        heading: Your job alert has been set up
-        message: We sent a confirmation email to %{email}
-      keywords: "Keywords:"
-      location: "Location:"
-      intro: "We'll email you any teaching jobs suitable for NQTs that match your search criteria:"
-      unsubscribe: You can unsubscribe by following the link at the bottom of the email.
-    heading: Sign up for NQT job alerts
-    intro:
-      content: >-
-        Setting up an alert can help make your search for a teaching job easier. You’ll receive an email every morning
-        with any new job that’s been listed that matches your search criteria.
-      teaching_vacancies: >-
-        Teaching Vacancies is a free national job-listing service from the Department for Education. It helps schools to
-        save money on recruitment and is here to help you to take the first step in your career.
-      unsubscribe: >-
-        You can unsubscribe at any time by following the instructions in the job alert
-    page_title: Sign up for NQT job alerts
-
   number:
     currency:
       format:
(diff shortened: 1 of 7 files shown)


david-mears-dfe:
You could write a request spec to test the redirect from old URL

config/locales/en.yml
```diff
@@ -201,28 +201,6 @@ en:
     sign_in: List a teaching job
     sign_out: Sign out
 
-  nqt_job_alerts:
```
csutter, line 204: I'm sure I'm missing something but where did this go?

josephhull676: I don't think it's used anymore. I think initially it was used for the confirmation page after an nqt job alert was created? I checked the confirmation page for normal job alerts and nqt job alerts after deleting it and there wasn't any content missing!

spec/requests/subscriptions_spec.rb
```diff
@@ -23,11 +23,24 @@
     end
 
     context "when hit via the nqt job alerts url" do
```
david-mears-dfe, line 25: ECT

david-mears-dfe: So what is the difference between this context and the next one?

spec/requests/subscriptions_spec.rb
```diff
@@ -23,11 +23,24 @@
...
-      let(:origin) { "/sign-up-for-NQT-job-alerts" }
-      let(:params) { { "nqt_job_alert" => true, "origin" => "/sign-up-for-NQT-job-alerts", "search_criteria" => { "job_roles" => ["nqt_suitable"] } } }
+      let(:origin) { "/sign-up-for-ECT-job-alerts" }
+      let(:params) { { "ect_job_alert" => true, "origin" => "/sign-up-for-ECT-job-alerts", "search_criteria" => { "job_roles" => ["nqt_suitable"] } } }
```
david-mears-dfe, line 27: You don't use this `params`.

david-mears-dfe: or `origin`

josephhull676 (commented):
.

spec/requests/subscriptions_spec.rb
josephhull676: Ah yes!

david-mears-dfe (approved):

PR #3954, opened 2021-09-09
[TEVA-2918] add skip links to dashboard results areas

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2918

### DIFF
--- a/app/views/publishers/notifications/index.html.slim
+++ b/app/views/publishers/notifications/index.html.slim
@@ -1,22 +1,27 @@
 - content_for :page_title_prefix, t(".title", count: notifications.unread.total_count)
 
+- if notifications.any?
+  - content_for :skip_links do
+    a.govuk-skip-link href="#notifications-results" = t(".skip_link")
+
 .govuk-main-wrapper
   .govuk-grid-row
-    .govuk-grid-column-two-thirds
+    .govuk-grid-column-full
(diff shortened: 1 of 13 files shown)


app/views/shared/_skip_links.html.slim
```diff
@@ -1,5 +1,5 @@
 - unless cookies_preference_set?
   a.govuk-skip-link href="#cookie-consent" Skip to cookie consent
 a.govuk-skip-link href="#main-content" Skip to main content
```
csutter, line 3: Is it worth having this as an `else` for the `if content_for?(:skip_links)` instead of always showing it?

app/views/publishers/notifications/index.html.slim
```diff
@@ -1,22 +1,27 @@
 - content_for :page_title_prefix, t(".title", count: notifications.unread.total_count)
 
+- if notifications.any?
+  - content_for :skip_links do
+    a.govuk-skip-link href="#notifications-results" Skip to notifications
```
csutter, line 5: I feel like we might want to put these in as translations, but I'm not hell bent on it.

csutter (approved):

app/views/publishers/notifications/index.html.slim
alexbowen: i think they should

PR #3957, opened 2021-09-09
Remove use of `steps_config` in vacancy validation

csutter:
- Add `VacancyStepProcess#validatable_steps` to define steps that
  should be validated when validating all steps for a vacancy
- Use this instead of `steps_config`
- Simplify `all_steps_valid?` to not skip steps that are no longer
  included

## Jira ticket URL

### DIFF
--- a/app/services/publishers/vacancies/vacancy_step_process.rb
+++ b/app/services/publishers/vacancies/vacancy_step_process.rb
@@ -20,6 +20,10 @@ class Publishers::Vacancies::VacancyStepProcess < StepProcess
     })
   end
 
+  def validatable_steps
+    steps - %i[documents review]
+  end
+
   private
 
   def job_role_steps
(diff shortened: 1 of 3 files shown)


cesidio (approved):

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -20,6 +20,10 @@ def initialize(current_step, vacancy:, organisation:, session:)
     })
   end
 
+  def validatable_steps
+    steps - %i[job_location schools documents review]
```
david-mears-dfe, line 24: Why minus job_location and schools? I think these steps should be validated (if they are among `steps`).

csutter: They were here before, I took them out, still works, good shout! 🏆

PR #3959, opened 2021-09-09
[TEVA-2980] Add postcode (from mean geolocation) for generating similar jobs for multiple school vacancies

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2980

## Changes in this PR:

When SimilarJobs searches for jobs similar to a multi-school job, we need to derive a location to search around.

Until now, we could not generate a location to search around for LAs because we don’t have their postcodes. And anyway, even if we did have its location, it wouldn’t necessarily be helpful since they can manage schools outside of themselves, and, like trusts, they are big enough that their head office might be quite far from the school(s) whose job listing the user is looking at.

Take the mean of the geolocations of the school(s) the vacancy is at, and use that as the location to search around. The corresponding job alert we suggest the user creates derives a location from this mean geolocation by using a new method on Geocoding that queries an API for the best postcode given a certain latitude and longitude. Used nominatim API as the backup if Google is over query limit, because it was the default for the geocoder gem.

Centroids are for polygons (ordered points defining a shape), so the mean is more appropriate than a centroid.

Mean geolocation of a vacancy is updated whenever:
- The easting/northing of a school it associates to is updated (these are what we derive
its geolocation from)
- An organisation-vacancy association is destroyed, created, or updated.

http_headers: Configured Geocoder to include our contact details as this
is a condition of using the Nominatim 'OpenStreetsMap' data source.

Corrected some seed vacancy geolocations that should be negative instead
of positive.

 Replace 'distinct' with 'group' in vacancy scope. This is because SQL's DISTINCT operator cannot be called on entities (such as vacancies) which have a point attribute, because postgres does not have an equality operator for points. Instead, we will use SQL's GROUP BY operator.

## Screenshots of it in action:

Screenshot 2021-09-09 at 17 05 06

Click the 'get job alert' button from the above, and see:

Screenshot 2021-09-09 at 17 07 51


app/controllers/publishers/vacancies_controller.rb
```diff
@@ -59,8 +59,8 @@ def summary
 
   private
 
-  def devise_job_alert_search_criteria
-    @devised_job_alert_search_criteria = Search::CriteriaDeviser.new(vacancy).criteria
+  def invent_job_alert_search_criteria
+    @invented_job_alert_search_criteria = Search::CriteriaInventor.new(vacancy).criteria
```
josephhull676, line 63: This method sets `@invented_job_alert_search_criteria`, right? Why is this being called before the show action in a callback when it's set in the show action anyway?

lib/geocoding.rb
```diff
@@ -15,13 +15,37 @@ def coordinates
...
       Rails.logger.error("Google Geocoding API responded with OVER_QUERY_LIMIT")
       Geocoder.coordinates(location, lookup: :uk_ordnance_survey_names)
-    end || no_match
+    end || no_coordinates_match
+  end
+
+  def postcode_from_coordinates
+    return Geocoder::DEFAULT_LOCATION if Rails.application.config.geocoder_lookup == :test
+
+    Rails.cache.fetch([:geocoding, location], expires_in: CACHE_DURATION, skip_nil: true) do
```
csutter, line 24: This needs to be a different key from the one on line 13 otherwise it'll overwrite the other cache

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -159,6 +159,7 @@ def and_perform_a_search
       check I18n.t("helpers.label.publishers_job_listing_job_details_form.job_roles_options.teacher")
       check I18n.t("jobs.filters.nqt_suitable_only")
       check I18n.t("jobs.filters.send_responsible_only")
+      # check I18n.t("jobs.education_phase_options.primary")
```
csutter, line 162: Is this commented out on purpose?

csutter (commented):
Couple of comments, and you might need to rebase ⚾

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -26,7 +26,7 @@ def all_steps_valid?
     steps_to_skip = if current_organisation.is_a?(School)
                       []
                     else
-                      vacancy.job_location == "at_central_office" ? [:job_location] : %i[job_location schools]
+                      vacancy.job_location == "central_office" ? [:job_location] : %i[job_location schools]
```
cesidio, line 29: you can do
`vacancy.central_office? ? [:job_location] : %i[job_location schools]`

app/models/school.rb
```diff
@@ -48,16 +48,14 @@ def religious_character
...
-        easting: easting.to_i,
-        northing: northing.to_i,
-        type: :gb,
-      ).to_wgs84
-
-      geolocation = [wgs84[:latitude], wgs84[:longitude]]
-    end
-
-    self.geolocation = geolocation
+    self.geolocation = if easting && northing && (easting_changed? || northing_changed?)
```
cesidio, line 51: a guard clause would be more readable here:

https://rubystyle.guide/#no-nested-conditionals

spec/system/jobseekers_can_view_a_vacancy_spec.rb
```diff
@@ -1,22 +1,34 @@
...
-    let(:vacancy) { create(:vacancy, :published) }
+    let(:vacancy_trait) { :published }
 
     scenario "jobseekers can view the vacancy" do
       verify_vacancy_show_page_details(vacancy)
     end
 
     context "when the publish_on date is in the future" do
-      let(:vacancy) { create(:vacancy, :future_publish) }
+      let(:vacancy_trait) { :future_publish }
```
josephhull676, line 31: Should all of this cleaning up stuff be put in a separate PR?

spec/system/jobseekers_can_view_a_vacancy_spec.rb
```diff
@@ -1,22 +1,34 @@
...
 
 RSpec.describe "Viewing a single published vacancy" do
-  let(:school) { create(:school) }
+  let(:school_group) { create(:local_authority) }
+  let(:school1) { create(:school) }
+  let(:school2) { create(:school) }
+  let(:vacancy) do
+    create(:vacancy, vacancy_trait, :at_multiple_schools, :with_supporting_documents, :no_tv_applications,
+           organisation_vacancies_attributes: [
+             { organisation: school1 }, { organisation: school2 }
```
josephhull676, line 10: Why on multiple lines?

spec/system/jobseekers_can_view_a_vacancy_spec.rb
```diff
@@ -1,22 +1,34 @@
...
+           ],
+           working_patterns: %w[full_time part_time],
+           subjects: %w[Physics])
+  end
 
   before do
-    vacancy.organisation_vacancies.create(organisation: school)
+    SchoolGroupMembership.find_or_create_by(school_id: school1.id, school_group_id: school_group.id)
```
josephhull676, line 17: Why find_or_create_by ?

lib/geocoding.rb
david-mears-dfe: V.g. point

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
david-mears-dfe: No

app/models/vacancy.rb
```diff
@@ -174,6 +174,19 @@ def publish_equal_opportunities_report?
     job_applications.after_submission.count >= EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD
   end
 
+  def set_mean_geolocation!
+    # When SimilarJobs searches for jobs similar to a multi-school job, we need to derive a location to search around.
+    # We do not have LAs' postcodes; anyway, LAs can manage schools outside themselves.
```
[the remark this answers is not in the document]
david-mears-dfe: I've got rid of the line as it's not very interesting.

app/services/search/criteria_inventor.rb
```diff
@@ -1,25 +1,31 @@
+require "geocoding"
+
 # Based on a vacancy, devise a plausible set of search criteria for a job alert subscription
-class Search::CriteriaDeviser
+class Search::CriteriaInventor
   attr_reader :criteria
 
+  DEFAULT_RADIUS_IN_MILES = 25
```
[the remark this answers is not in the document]
david-mears-dfe: The ticket says 20, but the closest radius option to that is 25. I should check that this is OK.

app/services/search/location_builder.rb
```diff
@@ -18,7 +18,7 @@ def initialize(location, radius)
     elsif search_with_polygons?
       initialize_polygon_boundaries
     elsif @location.present?
-      @location_filter = build_location_filter(@location, @radius)
+      @location_filter = build_location_filter
```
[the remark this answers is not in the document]
david-mears-dfe: Because there is no need to pass arguments.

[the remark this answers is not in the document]
david-mears-dfe: Instead of passing needless arguments we can use the attr_readers.

spec/models/school_spec.rb
```diff
@@ -49,41 +52,48 @@
...
-
-          expect(school.geolocation).to eq(nil)
+          expect(school.geolocation).to be_nil
         end
       end
     end
 
     context "when there is a previous geolocation" do
       let(:school) { create(:school, easting: 100, northing: 200) }
+      let!(:vacancy) { create(:vacancy, organisation_vacancies_attributes: [{ organisation: school }]) }
```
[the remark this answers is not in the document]
david-mears-dfe: Whoops, these are left over and unnecessary

spec/system/jobseekers_can_view_a_vacancy_spec.rb
david-mears-dfe: No reason

spec/system/jobseekers_can_view_a_vacancy_spec.rb
david-mears-dfe: Arguably... I'm touching this file because I had to update it with this PR anyway.

david-mears-dfe: It's more than cleaning up. I was updating this file to test that the mean geolocation is used when we create a job alert out of a similar jobs section on a job listing.

spec/system/jobseekers_can_view_a_vacancy_spec.rb
david-mears-dfe: I forget the precise reason, but I think for some tests the membership already existed, so, if it does, we shouldn't create it again.

lib/geocoding.rb
```diff
@@ -15,13 +15,37 @@ def coordinates
...
+      result.data["address_components"].find { |line| "postal_code".in?(line["types"]) }["short_name"]
+    rescue Geocoder::OverQueryLimitError
+      Rails.logger.error("Google Geocoding API responded with OVER_QUERY_LIMIT")
+      result = Geocoder.search(location, lookup: :nominatim).first.data
+      if result["error"].present?
+        Rails.logger.error("Geocoding Nominatim API responded with error: #{result['error']}")
+        return no_postcode_match
+      end
+      result["address"]["postcode"]
+    end
```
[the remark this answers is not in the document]
david-mears-dfe: Sure.

spec/lib/organisation_import/import_school_data_spec.rb
```diff
@@ -102,6 +102,19 @@
...
     end
 
+    context "updating an existing school's location" do
+      let(:wgs84_latitude) { 51.51396894535262 }
+      let(:wgs84_longitude) { -0.07751626505544208 }
+      let!(:school) { create(:school, urn: "100000", geolocation: [1, 2]) }
+      let!(:vacancy) { create(:vacancy, organisation_vacancies_attributes: [{ organisation: school }]) }
+
+      it "changes the mean_geolocation on the school's vacancies" do
+        expect { subject.run! }.to change { school.reload.easting }.from("1").to("533498")
```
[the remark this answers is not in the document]
david-mears-dfe: I'm checking that the school's easting attribute was changed to what GIAS says (example_schools_data.csv). But it's unnecessary so I'll get rid of it.

spec/models/organisation_vacancy_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe OrganisationVacancy do
+  context "when a vacancy is associated to an organisation" do
+    let(:vacancy) { create(:vacancy, mean_geolocation: [1, 2]) }
+    let(:school) { create(:school, geolocation: [3, 4]) }
+    let!(:organisation_vacancy1) { OrganisationVacancy.create(organisation: school, vacancy: vacancy) }
+
+    it "recalculates the vacancy's mean geolocation" do
+      expect(vacancy.mean_geolocation).to eq(school.geolocation)
+    end
```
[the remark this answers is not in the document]
david-mears-dfe: Exactly. Here I am not trying to test the details of _how_ the mean geolocation is calculated, I just want to know that it is triggered to recalculate it. So I am using the simplest type of mean geolocation (average location of 1 location).

app/services/search/criteria_inventor.rb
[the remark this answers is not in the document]
david-mears-dfe: Ben says 25

app/models/school.rb
david-mears-dfe: I tried a guard clause but actually it changes the behaviour since we want to set geolocation to nil if the condition is false.

spec/system/jobseekers_can_view_a_vacancy_spec.rb
josephhull676: Ah you were removing `"when creating a job alert"` and in the process you cleaned up?

spec/system/jobseekers_can_view_a_vacancy_spec.rb
josephhull676: Ah that makes sense

spec/system/jobseekers_can_view_a_vacancy_spec.rb
david-mears-dfe: OK, now the above comment is no longer true, and I'm going to revert this file

csutter (approved):

PR #3961, opened 2021-09-10
[TEVA-2748] Update govuk-components gem to version 2.1

cesidio:
## Jira ticket URL
https://dfedigital.atlassian.net/browse/TEVA-2748


app/components/timeline_component.rb
```diff
@@ -1,6 +1,10 @@
 class TimelineComponent < GovukComponent::Base
   include ViewComponent::SlotableV2
```
alexbowen, line 2: do we still need these includes with this update?

app/components/banner_component.rb
```diff
@@ -1,4 +1,8 @@
 class BannerComponent < GovukComponent::Base
+  def initialize(classes: [], html_attributes: {})
+    super(classes: classes, html_attributes: html_attributes)
+  end
```
david-mears-dfe, line 4: What's the purpose of this method? It looks the same as super?

cesidio: have a look at how other GOVUK Components have been implemented, like this one:

https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/back_link_component.rb

classes and html_attributes must always be passed in in the base class, but are optional in the subclasses

csutter (approved):

PR #3964, opened 2021-09-10
Refactor `_field` methods and use of `send`

csutter:
`Wizardable` has methods for the fields (`Vacancy` model attributes)
for each individual form which are then invoked using `send`  - these
should just be owned by each form model itself.

## Jira ticket URL


cesidio (approved):
much nicer

PR #3965, opened 2021-09-10
[TEVA-3075] New URL for 'Suitable for early career teachers' search results

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3075

## Changes in this PR:

- Changed nqt_suitable job role to ect_suitable. Now when request hits /teaching-jobs-for-ect_suitable a search is made for all ECT suitable roles.
- Changed all other references from NQTs to ECT


app/presenters/subscription_presenter.rb
```diff
@@ -63,7 +63,7 @@ def render_phases_filter(value)
     { education_phases: value.map { |role| I18n.t("jobs.education_phase_options.#{role}") }.join(", ") }
   end
 
-  def render_nqt_filter(value)
-    { '': "Suitable for NQTs" } if value.eql?("true")
+  def render_ect_filter(value)
+    { '': "Suitable for ECTs" } if value.eql?("true")
```
csutter, line 67: In places where we display this to the user, should this be spelled out fully?

csutter (approved):
Looks good!

Do we need to do an Algolia reindex? 🤔

PR #3967, opened 2021-09-10
[TEVA-3063] Update heading on vacancy show page

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3063

## Changes in this PR:

ronseal


csutter (approved):
Normally I wouldn't change the translation key when updating translations, but it does look like a good cleanup (I hope one day we can whittle our use of synonyms of "job" down to one or two 😂 )

PR #3976, opened 2021-09-13
Improve email generation in factories

csutter:
- Ensure all email addresses we generate in factories are on the
  `example.com` domain
- Rename publishers factory to what it actually contains


josephhull676 (approved):
Why do they have to be on the example.com domain?

PR #3979, opened 2021-09-13
[TEVA-3050] Update back links in job listing journey

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3050

### DIFF
--- a/app/components/publishers/vacancy_form_page_heading_component.rb
+++ b/app/components/publishers/vacancy_form_page_heading_component.rb
@@ -12,7 +12,7 @@ class Publishers::VacancyFormPageHeadingComponent < ViewComponent::Base
 
   private
 
-  attr_reader :vacancy, :copy
+  attr_reader :vacancy, :copy, :step_process
 
   def page_title
     return t("jobs.edit_job_title", job_title: vacancy.job_title) if vacancy.published?
@@ -22,5 +22,15 @@ class Publishers::VacancyFormPageHeadingComponent < ViewComponent::Base
 
   def organisation_from_job_location
     vacancy.at_multiple_schools? ? "multiple schools" : vacancy.parent_organisation_name
+  end
+
+  def back_path
+    if step_process.previous_step_or_review == :review
+      return organisation_job_path(vacancy.id) if vacancy.completed_steps.size != step_process.steps.size
+
+      organisation_job_review_path(vacancy.id)
+    else
+      organisation_job_build_path(vacancy.id, step_process.previous_step)
+    end
   end
 end
--- a/app/components/publishers/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
+++ b/app/components/publishers/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
@@ -1,5 +1,10 @@
-h1.govuk-heading-m
-  = heading
-  - unless vacancy.published?
-    span.govuk-caption-m
-      = t("jobs.current_step", step: @step_process.current_step_group_number, total: @step_process.total_step_groups)
+- if step_process.previous_step_or_review.present? && step_process.current_step != :review
+  = govuk_back_link text: t("buttons.back"), href: back_path, classes: "govuk-!-margin-bottom-7"
+
+.govuk-grid-row
+  .govuk-grid-column-full
+    h1.govuk-heading-m
+      = heading
+      - unless vacancy.published?
+        span.govuk-caption-m
+          = t("jobs.current_step", step: step_process.current_step_group_number, total: step_process.total_step_groups)
--- a/app/services/publishers/vacancies/vacancy_step_process.rb
+++ b/app/services/publishers/vacancies/vacancy_step_process.rb
@@ -24,6 +24,12 @@ class Publishers::Vacancies::VacancyStepProcess < StepProcess
     steps - %i[documents review]
   end
 
+  def previous_step_or_review
+    return steps.last if session[:current_step] == :review && first_of_group?
+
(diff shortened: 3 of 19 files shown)


csutter:
Ah sorry I didn't quite set the right context on this 😅  I reckon this whole thing here is identical across the steps and should be able to go into a partial:

```slim

.govuk-main-wrapper
  = govuk_back_link text: t("buttons.back"), href: step == :documents ? organisation_job_build_path(vacancy.id, :important_dates) : back_path
  .govuk-grid-row
    .govuk-grid-column-full
      = render Publishers::VacancyFormPageHeadingComponent.new(vacancy, step_process)
```

Otherwise I think it'd be a bit prematurely optimised if it's _just_ for the back link.

app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -1,7 +1,7 @@
 require "google/apis/drive_v3"
 
 class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseController
-  helper_method :form
+  helper_method :form, :step
```
csutter, line 4: You can use `step_process.current_step` in the views I think, no need to expose this.

app/views/publishers/vacancies/build/_back_link.html.slim
```diff
@@ -0,0 +1 @@
+= govuk_back_link text: t("buttons.back"), href: step == :documents ? organisation_job_build_path(vacancy.id, :important_dates) : back_path
```
csutter, line 1: (see controller comment)

app/components/publishers/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
```diff
@@ -1,5 +1,9 @@
-h1.govuk-heading-m
-  = heading
-  - unless vacancy.published?
-    span.govuk-caption-m
-      = t("jobs.current_step", step: @step_process.current_step_group_number, total: @step_process.total_step_groups)
+- unless @back_path.nil?
```
csutter, line 1: Might read nicer as a one-liner?

app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -39,6 +39,12 @@ def destroy
 
   private
 
+  def back_path
+    return edit_organisation_job_path(vacancy.id) if session[:current_step] == :review
+
+    organisation_job_build_path(vacancy.id, :important_dates)
```
csutter, line 45: Let's use `step_process` here too instead of hardcoding the previous step

csutter (commented):
Looks lovely now! Two more things to address then we're good to go 👍

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -24,6 +24,14 @@ def validatable_steps
     steps - %i[documents review]
   end
 
+  def previous_visited_step
+    return steps.last if vacancy.completed_steps.include?(steps.last.to_s) && first_of_group?
+
+    return nil if current_step == steps.first
```
csutter, line 30: Previous step should already return `nil` for this case so I don't think this line is necessary

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -24,6 +24,14 @@ def validatable_steps
     steps - %i[documents review]
   end
 
+  def previous_visited_step
+    return steps.last if vacancy.completed_steps.include?(steps.last.to_s) && first_of_group?
```
csutter, line 28: I don't think this is right, it does a different thing compared to what the `back_path` does right now.

`review` doesn't get added to `completed_steps` until all steps are valid: https://github.com/DFE-Digital/teaching-vacancies/blob/master/app/controllers/publishers/vacancies_controller.rb#L29

So you might come from the review page because you're editing an incomplete draft, but with this logic the "Back" link would take you to the previous step rather than the review page.

app/components/publishers/vacancy_form_page_heading_component.rb
```diff
@@ -23,4 +23,14 @@ def page_title
   def organisation_from_job_location
     vacancy.at_multiple_schools? ? "multiple schools" : vacancy.parent_organisation_name
   end
+
+  def back_path
+    if step_process.previous_step_or_review == :review
+      return organisation_job_path(vacancy.id) if vacancy.completed_steps.size != step_process.steps.size
```
csutter, line 29: Is this the show path? 🧐

csutter (approved):

app/components/publishers/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
```diff
@@ -1,5 +1,10 @@
-h1.govuk-heading-m
-  = heading
-  - unless vacancy.published?
-    span.govuk-caption-m
-      = t("jobs.current_step", step: @step_process.current_step_group_number, total: @step_process.total_step_groups)
+- if step_process.previous_step_or_review.present? && step_process.current_step != :review
+  = govuk_back_link text: t("buttons.back"), href: back_path, classes: "govuk-!-margin-bottom-7"
+
+.govuk-grid-row
+    .govuk-grid-column-full
```
cesidio, line 5: is it indented 4 space?

josephhull676: So it is!

PR #3984, opened 2021-09-15
[TEVA 3059] improve a11y of search subjects input

alexbowen:
- add aria live area so user is ihnformed when subjects updates with input change
- adds aria tags so user is informed when they tab to option it is option x of total

https://dfedigital.atlassian.net/browse/TEVA-3059

### DIFF
--- a/app/components/searchable_collection_component/searchable_collection_component.html.slim
+++ b/app/components/searchable_collection_component/searchable_collection_component.html.slim
@@ -2,7 +2,8 @@
   div class="#{border_class} #{scrollable_class}"
     - if searchable
       .searchable-collection-component__search
-        input.govuk-input.searchable-collection-component__search-input.icon.icon--left.icon--search.js-action placeholder="Search" aria-label="#{label_text}"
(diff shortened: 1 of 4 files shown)


app/components/searchable_collection_component/searchable_collection_component.html.slim
```diff
@@ -2,7 +2,8 @@
   div class="#{border_class} #{scrollable_class}"
     - if searchable
       .searchable-collection-component__search
-        input.govuk-input.searchable-collection-component__search-input.icon.icon--left.icon--search.js-action placeholder="Search" aria-label="#{label_text}"
+        input.govuk-input.searchable-collection-component__search-input.icon.icon--left.icon--search.js-action placeholder="Search" aria-label="#{label_text}" aria-expanded="true" aria-describedby="publishers-job-listing-job-details-form-subjects-hint" aria-owns="subjects__listbox" role="combobox"
```
cesidio, line 5: very minor, `"Search"` should come from I18n files

cesidio (approved):

PR #3987, opened 2021-09-15
[TEVA-3167] Stop semantic logger from draining to log file

felixtheflex:
## Jira ticket URL

## Changes in this PR:

set semantic logger config to drain to file as `false`


csutter (approved):
Ooooh 💡

PR #3993, opened 2021-09-16
[TEVA-3078] Update bullet formatting rules

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3078

## Changes in this PR:

- Update bullet formatting rules to fix below-pictured problems.

Commit message:

Job adverts nowadays contain a lot of html elements instead of the newline characters they formerly used.

Do not mess with the formatting when the unit we're operating on
(named 'paragraph') contains one bullet point, since the formatting
is probably already correct.

### Before

Screenshot 2021-09-16 at 13 47 34
Screenshot 2021-09-16 at 13 51 00
Screenshot 2021-09-16 at 14 07 52
Screenshot 2021-09-16 at 13 47 45

### After

Screenshot 2021-09-16 at 13 44 12
Screenshot 2021-09-16 at 14 05 07
Screenshot 2021-09-16 at 14 07 20
Screenshot 2021-09-16 at 13 44 30

### DIFF
--- a/app/presenters/vacancy_presenter.rb
+++ b/app/presenters/vacancy_presenter.rb
@@ -6,6 +6,8 @@ class VacancyPresenter < BasePresenter
   delegate :working_patterns, to: :model, prefix: true
   delegate :job_roles, to: :model, prefix: true
 
+  HTML_STRIP_REGEX = %r{(&nbsp;|<div>|</div>|<!--block-->)+}
+
   def share_url(utm_source: nil, utm_medium: nil, utm_campaign: nil, utm_content: nil)
     params = {}
     if utm_source.present?
@@ -36,7 +38,7 @@ class VacancyPresenter < BasePresenter
   end
 
   def benefits
-    simple_format(model.benefits) if model.benefits.present?
+    simple_format(fix_bullet_points(model.benefits)) if model.benefits.present?
   end
 
   def expired?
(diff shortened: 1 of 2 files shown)


app/presenters/vacancy_presenter.rb
```diff
@@ -115,31 +115,32 @@ def greyed_additional_job_role(role)
...
-          index == first_bulleted_line_idx ? "<ul>#{item}" : item
-        }.join.concat("</ul>")
-      end
-    }.join("\n")
+    text.split("<br>").map { |para|
+      next para if para.count(bullet) <= 1 # If paragraph only has one bullet point it is probably correctly formatted
+
+      first_bulleted_line_idx = strip(para).first == bullet ? 0 : 1
+      para.split(bullet).reject { |line| strip(line).blank? }.map.with_index { |line, index|
+        item = "<li>#{line.gsub(%r{(&nbsp;|<div>|</div>|<!--block-->)+}, '')}</li>"
```
csutter, line 130: I feel that regex can probably be extracted out into a constant - it's used both here and in `#strip`

csutter (approved):
This logic is really messy by its nature, but this actually makes it a little bit nicer so... 👍 

Tackling this "properly" might be something for the long-term tech debt pile.

PR #3995, opened 2021-09-16
Improve organisation import scripts

csutter:
- Make `run!` methods more explicit
- Make `save_csv_file` less verbose and remove specs that test
  private method
- Extract repeated `GIAS_BASE_URL`
- Remove open-uri and superfluous use of #each on `CSV.foreach`


david-mears-dfe (approved):
The life-changing magic of tidying up, by Marie Kondo

cesidio (approved):

PR #3997, opened 2021-09-16
Use PostGIS for school geolocation

csutter:
This sets up PostGIS and as a first actual change includes adding a `geopoint` column to the `schools` table which uses a PostGIS geography point, populated at the same time as the `geolocation` column (when easting and northing are set on a vacancy).

The `geolocation` column will stay around for now until:
* The import task has had a chance to successfully run in production, **and then**
* All current uses of the `geolocation` column have been replaced by `geopoint` and we've quadruple-checked everything works

## ⚠ Pain

This will require some local environment changes. You need to have Postgis installed to run the app locally going forward, on Mac with Homebrew it _should_ be as easy as:

```
brew install postgis
```


.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:11.6-alpine
+        image: postgis/postgis:12-3.1-alpine
```
cesidio, line 25: production has 12.8, not that it's that relevant

cesidio (approved):
exciting changes!

.github/workflows/test.yml
csutter: I think the `3.1` here is the PostGIS version, only the `12` refers to Postgres itself

PR #4005, opened 2021-09-17
Remove edit action from vacancies controller

josephhull676:
## Changes in this PR:

- Whilst pairing with Cesi on the ticket related to the back links, we noticed that the only time you hit the edit action is when you make a change on a published vacancy and click 'update listing'.
- The only difference between the two actions was that we `validate_all_steps` in the edit action. This has been moved to the show action.
- The only differences between the templates were that the errors and the various buttons (to copy listing, extend deadline etc) were not rendered on the edit page. To us, it make sense to have these buttons when a publisher makes a change to a published vacancy.
- Now, when a publisher goes to a published vacancy's show page, changes something and clicks 'update listing' they are taken to the show action / template.


app/controllers/publishers/vacancies_controller.rb
```diff
@@ -3,6 +3,9 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   before_action :invent_job_alert_search_criteria, only: %i[show preview]
 
   def show
+    redirect_to organisation_job_review_path(vacancy.id) unless vacancy.published?
+
+    validate_all_steps
```
david-mears-dfe, line 8: So if I have completed half the steps and go for a lunch break and come back and log in again, and go to this draft job, I am greeted with a long list of all the fields I have not yet filled in. Right? This seems overly aggressive.

josephhull676: We have since removed this redirect, but we essentially do the same thing. We link to the show page from the dashboard and just display the errors on here. 

We ran this by Jesse and he seemed happy with it :)

david-mears-dfe: As long as Jesse is on board that's cool

cesidio (approved):

PR #4010, opened 2021-09-20
[TEVA-3188] Unset users from production paas space every day

saliceti:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3188

## Changes in this PR:
Users should not keep permanent access to production for security
reasons. Only the service account can stay.


saliceti:
This was tested manually successfully

csutter (approved):

PR #4011, opened 2021-09-21
Update vacancy.set_postcode_from_mean_geolocation to use new geopoints

david-mears-dfe:
We can use postgis' native functions to perform the calculation of th
midpoint of some points.

ST_Centroid is here being used on *points*; if the points were passed to
it as a line or polygon the centroid would be calculated differently and
incorrectly.

ST_Centroid only takes geometries as its input so we have to cast the
geographies as geometries and then back again.


app/models/vacancy.rb
```diff
@@ -187,16 +187,16 @@ def set_postcode_from_mean_geolocation!
...
-                                          end
+    if at_one_school?
+      self.postcode_from_mean_geolocation = organisation.postcode
+    else
+      schools = organisations.schools.where.not(geopoint: nil)
+      centroid = schools.pluck(Arel.sql("ST_Centroid(ST_Collect(geopoint::geometry))::geography")).first
+
+      return unless centroid
+
+      self.postcode_from_mean_geolocation = Geocoding.new([centroid.x, centroid.y]).postcode_from_coordinates || postcode_from_mean_geolocation
```
csutter, line 198: Do we want to not override any existing value for this field? If so you can write this more concisely as:

```ruby
self.postcode_from_mean_geolocation = Geocoding.new([centroid.x, centroid.y]).postcode_from_coordinates
```

csutter (approved):
Looks good (see comment).

We might want to hold off merging for a day though until I've sorted the missing geopoints for some orgs!

app/models/vacancy.rb
david-mears-dfe: This is to avoid overwriting the attribute with nil when the Geocoder returns a nil response.

david-mears-dfe: I'm not sure how your suggestion works?

csutter: Apologies, that was pre-coffee! I meant:

```ruby
self.postcode_from_mean_geolocation ||= Geocoding.new([centroid.x, centroid.y]).postcode_from_coordinates
```

(with the `||=` operator)

csutter: Sorry that was pre-coffee and I typed it wrong, and also I see what you mean now.

I think I'd prefer a slightly longer but more explicit way of writing this bit, so it's more obvious what the intention is, e.g.:

```ruby
postcode = Geocoding.new([centroid.x, centroid.y]).postcode_from_coordinates
return unless postcode

self.postcode_from_mean_geolocation = postcode
```

david-mears-dfe:
I don't feel comfortable merging this right now as it's the end of my last working day this week. See you Monday.

PR #4012, opened 2021-09-21
Improve location columns setting for `School`s

csutter:
- Use a callback instead of overriding School easting/northing setters
- Set attributes in import without using #send
- Improve School model specs
  - Remove test that just checks a column exists
  - Improve order of tests
  - Fix tests to include new callback behaviour


app/models/school.rb
```diff
@@ -4,6 +4,8 @@ class School < Organisation
   has_many :school_group_memberships
   has_many :school_groups, through: :school_group_memberships
 
+  before_save :set_geolocation_from_easting_and_northing
```
david-mears-dfe, line 7: what about 
```suggestion
  before_save :set_geolocation_from_easting_and_northing, if: { easting_changed? || northing_changed? }
```

csutter: Could do - although it's not an expensive operation and not having an `if` there means we can change the logic that sets the columns without having to manually force `set_geolocation_from_easting_and_northing` on existing vacancies.

What do you think? Maybe we can add it later when the geo logic is a bit more solidified?

cesidio (approved):

app/models/school.rb
david-mears-dfe: Sure!

david-mears-dfe (approved):

PR #4031, opened 2021-09-23
Rewrite organisation import to be more efficient

csutter:
- Add `activerecord-import` gem for more efficient bulk data import
- Add `Gias::Data` as a GIAS data downloader and parser
- Add `Gias::ImportSchoolsAndLocalAuthorities` to import school and
  LA data
- Add `Gias::ImportTrustData` to import MAT data
- Move logic to delete stale `SchoolGroupMembership`s into the model
  itself
- Add `LogBenchmark` for simple benchmarking
- Change import task to just call import job to avoid duplication
- Remove legacy organisation import scripts

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2999


app/models/school_group.rb
```diff
@@ -1,4 +1,6 @@
 class SchoolGroup < Organisation
   has_many :school_group_memberships
   has_many :schools, through: :school_group_memberships
+
+  scope :trusts, -> { where.not(uid: nil) }
```
cesidio, line 5: this scope is available from the super class

csutter: 🤦

cesidio (approved):
Love this refactor

app/models/school_group_membership.rb
```diff
@@ -1,4 +1,21 @@
...
+    update_all(do_not_delete: false)
+  end
+
+  def self.delete_records_marked_for_deletion
+    to_delete = marked_for_deletion.map { |m| m.school_group.name }
+                                   .group_by { |x| x }
+                                   .transform_values(&:length)
+    Rollbar.log(:info, "Memberships to delete, by SchoolGroup: #{to_delete}")
+
+    raise "Skipped deleting suspiciously high number of `SchoolGroupMembership`s (#{marked_for_deletion.count})" if marked_for_deletion.count > 150
```
cesidio, line 17: Maybe we want to use a constant that explains that number 150

PR #4032, opened 2021-09-23
Tighten rubocop ABC size metric

david-mears-dfe:
Refactor by tightening rubocop metrics

    A few things are rewritten to decrease ABC size.

    Other metrics are reduced to as low as they can go without forcing me to
    refactor more things, in order to keep the code relatively simple going
    forwards.

    We do not need to index the slug as 'permalink': this is a hangover
    from the initial migration from ElasticSearch to Algolia, when there
    was a possibility that we would use Algolia javascript in the front end
    to update search results without reloading the page. This is not going
    to happen.

    https://github.com/DFE-Digital/teaching-vacancies/commit/1ebc5b3fcfbda8580d5cfb803829616af4a8b256
    is the commit when we renamed 'slug' to 'permalink' for the sake of readability
    on the Algolia dashboard, but we don't need to index this attribute, as I say.

### DIFF
--- a/app/services/import_polygons.rb
+++ b/app/services/import_polygons.rb
@@ -24,28 +24,7 @@ class ImportPolygons
       next unless location_in_scope?
 
       geometry_rings = region_response.dig("geometry", "rings")
-
-      # When the ONS API returns multiple 'rings' for a location, some of these are small islands or similar. Some of
-      # these islands are inhabited and have a school on, and others do not. (E.g. Mersea Island in Essex, Hayling Island
-      # in Hampshire, Sheppey Island in Kent, St Mary's in the Isles of Scilly.)
-      #
-      # Store the 'rings' in a hash because Postgres doesn't support multidimensional arrays of varying extents.
-      #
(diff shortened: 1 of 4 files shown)


app/services/import_polygons.rb
```diff
@@ -65,6 +46,28 @@ def location_in_scope?
...
+  def transform_geometry_rings_to_polygons(geometry_rings)
+    # When the ONS API returns multiple 'rings' for a location, some of these are small islands or similar. Some of
+    # these islands are inhabited and have a school on, and others do not. (E.g. Mersea Island in Essex, Hayling Island
+    # in Hampshire, Sheppey Island in Kent, St Mary's in the Isles of Scilly.)
+    #
+    # Store the 'rings' in a hash because Postgres doesn't support multidimensional arrays of varying extents.
+    #
+    # If algolia searches by polygon are slow, (some of) these boundaries could be downsampled significantly.
+
+    polygons = []
```
csutter, line 58: While you're here (although hopefully we'll get to kill all this next sprint): Both the outer and inner loop should really be using `map` instead of creating empty arrays and pushing into them!

csutter (approved):

PR #4035, opened 2021-09-24
[TEVA-3147] replace url underscores with hyphens for SEO

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-3147

new sitemap https://teaching-vacancies-review-pr-4035.london.cloudapps.digital/sitemap

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -1,10 +1,11 @@
 class VacanciesController < ApplicationController
   helper_method :job_application
 
-  before_action :set_landing_page_description, :set_map_display, only: %i[index]
+  before_action :set_landing_page_description, :set_map_display, :set_params_from_pretty_landing_page_params, only: %i[index]
 
   def index
-    params[:location] = params[:location_facet] if params[:location_facet]
+    set_params_from_pretty_landing_page_params
+
     @vacancies_search = Search::VacancySearch.new(
       search_form.to_hash,
       sort_by: search_form.jobs_sort,
@@ -32,6 +33,12 @@ class VacanciesController < ApplicationController
   end
 
   private
+
+  def set_params_from_pretty_landing_page_params
+    params[:location] = params[:location_facet].titleize if params[:location_facet]
+    params[:job_roles] = params[:job_role].parameterize.underscore if params[:job_role]
+    params[:subject]&.titleize
+  end
 
   def algolia_search_params
     strip_empty_checkboxes(%i[job_roles phases working_patterns])
--- a/app/views/home/_vacancy_facets.html.slim
+++ b/app/views/home/_vacancy_facets.html.slim
@@ -6,7 +6,7 @@
         ul.govuk-list.two-column-list aria-label="#{t(".cities")} list"
           - @vacancy_facets.cities.each do |city, count|
             li
-              = govuk_link_to location_path(city) do
+              = govuk_link_to location_path(city.parameterize) do
                 = "#{city} "
                 = t(".link_text_html", count: count)
 
@@ -15,7 +15,7 @@
         ul.govuk-list.two-column-list aria-label="#{t(".counties")} list"
           - @vacancy_facets.counties.each do |county, count|
             li
-              = govuk_link_to location_path(county) do
+              = govuk_link_to location_path(county.parameterize) do
                 = "#{county} "
                 = t(".link_text_html", count: count)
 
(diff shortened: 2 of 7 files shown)


david-mears-dfe:
Are you going to add some redirects for the old urls?

alexbowen:
> Are you going to add some redirects for the old urls?

indeed, this isnt finished yet in a few respects. not working today but redirects defo on my list to sort tommorow

config/routes.rb
```diff
@@ -185,15 +191,15 @@
   # If parameters are used that are the same as those in the search form, pagination with kaminari will break
   match "teaching-jobs-in-:location_facet",
         to: "vacancies#index", as: :location, via: :get,
-        constraints: ->(request) { LocationPolygon.include?(request.params[:location_facet]) }
+        constraints: ->(request) { LocationPolygon.include?(request.params[:location_facet].tr("-", " ")) }
 
   match "teaching-jobs-for-:job_role",
         to: "vacancies#index", as: :job_role, via: :get,
-        constraints: ->(request) { Vacancy.job_roles.key?(request.params[:job_role]) },
+        constraints: ->(request) { Vacancy.job_roles.key?(request.params[:job_role].tr("-", "_")) },
```
josephhull676, line 198: In the review app when I click on:

Screenshot 2021-09-28 at 13 27 13
 
The URL is still teaching-jobs-for-ect_suitable.

app/controllers/vacancies_controller.rb
```diff
@@ -4,6 +4,9 @@ class VacanciesController < ApplicationController
   before_action :set_landing_page_description, :set_map_display, only: %i[index]
 
   def index
+    if params[:job_role]
```
david-mears-dfe, line 7: Could you make this into a before_action applying only to `index`?

config/routes.rb
```diff
@@ -169,8 +171,14 @@
...
 
-  # Legacy publisher sign in path (users may still have this bookmarked)
-  get "/identifications/new", to: redirect("/publishers/sign_in")
+  # Legacy authentication paths (users may still have this bookmarked)
+  get "/identifications/new", to: redirect("/publishers/sign-in")
+  get "/publishers/sign_in", to: redirect("/publishers/sign-in")
+  get "/publishers/sign_out", to: redirect("/publishers/sign-out")
+  get "/publishers/account_requests/new", to: redirect("/publishers/account-requests/new")
+  get "/jobseekers/sign_in", to: redirect("/jobseekers/sign-in")
+  get "/teaching-jobs-for-:job_role", to: redirect { |path_params| "/teaching-jobs-for-#{path_params[:job_role].tr('_', '-')}" },
```
david-mears-dfe, line 180: It looks like this is under the heading `# Legacy authentication paths (users may still have this bookmarked)` - could you edit that comment to not refer to authentication?

app/controllers/vacancies_controller.rb
```diff
@@ -4,7 +4,10 @@ class VacanciesController < ApplicationController
   before_action :set_landing_page_description, :set_map_display, only: %i[index]
 
   def index
-    params[:location] = params[:location_facet] if params[:location_facet]
+    # Set search parameters from pretty landing page params
+    params[:location] = params[:location_facet].titleize if params[:location_facet]
```
david-mears-dfe, line 8: Do we need to titleize params[:location] as well as params[:location_facet]?

david-mears-dfe: (I don't know what the titleizing is for)

csutter: `params[:location]` would be user input, so no. The `titleize`ing is to convert the pretty param (e.g. teaching-jobs-in-**london-town** to "London Town")

app/views/home/_vacancy_facets.html.slim
```diff
@@ -15,7 +15,7 @@
         ul.govuk-list.two-column-list aria-label="#{t(".counties")} list"
           - @vacancy_facets.counties.each do |county, count|
             li
-              = govuk_link_to location_path(county) do
+              = govuk_link_to location_path(county).parameterize do
```
csutter, line 18: It would be better to just call `parameterize` on `county`, not the whole path (here and in the other places in this file), because that's the only bit that needs it.

app/controllers/vacancies_controller.rb
```diff
@@ -4,7 +4,9 @@ class VacanciesController < ApplicationController
   before_action :set_landing_page_description, :set_map_display, only: %i[index]
 
   def index
-    params[:location] = params[:location_facet] if params[:location_facet]
+    # Set search parameters from pretty landing page params
+    search_params
```
csutter, line 8: As this has side effects, can we call this method something more explicit? `set_params_from_pretty_landing_page_params` perhaps, then we don't even need the comment.

csutter (approved):

app/controllers/vacancies_controller.rb
csutter: And then it could be a before_action:

```ruby
before_action :set_params_from_pretty_landing_page_params, only: :index
```

PR #4036, opened 2021-09-24
[TEVA-3189] Remove row.html_attributes and fix anchors

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3189

## Changes in this PR:
This PR fixes the links in the error summary by adding divs corresponding to each anchor to the top of the relevant section.
It also removes the row.html_attributes calls when the summary list rows are generated (these calls do not generate any html attributes).

### DIFF
--- a/app/views/jobseekers/job_applications/review/_declarations.html.slim
+++ b/app/views/jobseekers/job_applications/review/_declarations.html.slim
@@ -1,4 +1,5 @@
-#right_to_work_in_uk
+= render_divs_for_fields(Jobseekers::JobApplication::DeclarationsForm)
+
 = render ReviewComponent.new id: "jobseekers-job-application-review-form-declarations-field-error" do |review|
   - review.heading title: t("jobseekers.job_applications.build.declarations.heading"),
                    text: job_application_review_edit_section_text(job_application, :declarations),
--- a/app/views/jobseekers/job_applications/review/_declarations.html.slim
+++ b/app/views/jobseekers/job_applications/review/_declarations.html.slim
@@ -13,4 +13,3 @@
       - summary_list.row do |row|
         - row.key text: t("helpers.legend.jobseekers_job_application_declarations_form.right_to_work_in_uk")
         - row.value text: job_application.right_to_work_in_uk.humanize
-        - row.html_attributes { { id: "right_to_work_in_uk" } }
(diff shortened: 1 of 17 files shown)


app/views/jobseekers/job_applications/review/_equal_opportunities.html.slim
```diff
@@ -9,12 +9,10 @@
       - summary_list.row do |row|
         - row.key text: t("helpers.legend.jobseekers_job_application_equal_opportunities_form.disability")
         - row.value text: job_application.disability.humanize
-        - row.html_attributes { { id: "disability" } }
```
david-mears-dfe, line 12: I think removing this will mean that, when there is an error related to the `disability` field in the error summary / action required box, the link will not go anywhere. So can you add the divs in this partial as you did in the others? The same applies to a couple of other sections.

robyoung91: Ah right, I think I understand what you mean. I had been looking at the **"There is a problem"** error summary for the job application review page and saw that it didn't seem to rely on the row ids to link to the errors (those row.html_attributes weren't even successfully adding any ids to the rows) so thought we didn't need them.

I've just seen [your PR](https://github.com/DFE-Digital/teaching-vacancies/pull/3717) and realised we can get an **"Action required"** summary here too, just like the job listing review page which does indeed need ids!

cesidio (approved):

app/views/jobseekers/job_applications/review/_declarations.html.slim
```diff
@@ -1,3 +1,4 @@
+#right_to_work_in_uk
```
david-mears-dfe, line 1: We also need `#close_relationships`

david-mears-dfe (commented):
Actually, some summary list rows correspond to two fields. E.g. the gender row in `app/views/jobseekers/job_applications/review/_equal_opportunities.html.slim` has one id for the gender field, and one id for the gender description field (the free text field).

david-mears-dfe (approved):

app/views/publishers/vacancies/edit_vacancy_sections/_edit_job_location.html.slim
```diff
@@ -1,5 +1,7 @@
 = render "publishers/vacancies/error_tag", attributes: (Publishers::JobListing::JobLocationForm.fields + Publishers::JobListing::SchoolsForm.fields)
 
+= render_divs_for_fields(Publishers::JobListing::JobLocationForm)
```
david-mears-dfe, line 3: Just realised this one needs the schools form fields too (see line 1)

PR #4043, opened 2021-09-27
[TEVA-3047] WIP: Education phase step

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3047

## Changes in this PR:

- Coding in the open in case of preliminary feedback


app/controllers/concerns/publishers/wizardable.rb
```diff
@@ -61,19 +68,17 @@ def job_details_params(params)
...
-    params.require(:publishers_job_listing_working_patterns_form)
-          .permit(:working_patterns_details, working_patterns: [])
+    params.require(:publishers_job_listing_working_patterns_form).permit(:working_patterns_details, working_patterns: [])
           .merge(completed_steps: completed_steps)
   end
 
   def pay_package_params(params)
-    params.require(:publishers_job_listing_pay_package_form)
-          .permit(:actual_salary, :salary, :benefits)
+    params.require(:publishers_job_listing_pay_package_form).permit(:actual_salary, :salary, :benefits)
```
cesidio, line 81: I preferred this to be on 3 lines instead of 2, but it's just personal preference

app/views/publishers/vacancies/build/job_details.html.slim
```diff
@@ -10,11 +10,19 @@
 
         h2.govuk-heading-l = t("publishers.vacancies.steps.job_details")
 
+        = govuk_inset_text do
+          - if vacancy.one_phase?
+            = t("publishers.vacancies.build.job_details.phase_inset_text.one_phase", phase: vacancy.education_phases.first.humanize)
+          - else
```
david-mears-dfe, line 16: Note to self: should explicitly check `allow_phase_to_be_set?` here.

config/analytics.yml
```diff
@@ -318,3 +318,4 @@ shared:
     - candidate_hired_from
     - completed_steps
     - working_patterns_details
+    - phase
```
david-mears-dfe, line 321: Note to self: add 'key_stages'

app/models/concerns/vacancy_phaseable.rb
```diff
@@ -0,0 +1,25 @@
+module VacancyPhaseable
```
cesidio, line 1: The name suggest you want to have it in namespace Vacancy::Phaseable. Or you can just call it EducationPhaseable

app/controllers/publishers/vacancies/build_controller.rb
```diff
@@ -2,8 +2,8 @@ class Publishers::Vacancies::BuildController < Publishers::Vacancies::BaseContro
   include Wicked::Wizard
   include OrganisationsHelper
 
-  steps :job_role, :job_role_details, :job_location, :schools, :job_details, :working_patterns, :pay_package,
-        :important_dates, :documents, :applying_for_the_job, :job_summary
+  steps :job_role, :job_role_details, :job_location, :schools, :education_phases, :job_details, :working_patterns,
+        :pay_package, :important_dates, :documents, :applying_for_the_job, :job_summary
```
david-mears-dfe, line 6: Cesi says

> We fetched your branch and were testing the process of creating a vacancy and solved a bug for which it was asking the education phase when it was not supposed to. you have to add:
```ruby
when :education_phases
  skip_step if vacancy.one_phase?
```
on line 20 in app/controllers/publishers/vacancies/build_controller.rb

david-mears-dfe: Maybe this `show` action should use vacancy_step_process instead of rolling its own logic?

app/models/vacancy.rb
```diff
@@ -216,4 +207,10 @@ def enable_job_applications_cannot_be_changed_once_listed
 
     errors.add(:enable_job_applications, :cannot_be_changed_once_listed)
   end
+
+  def drop_attributes_if_no_longer_applicable
```
david-mears-dfe, line 211: I'd like to draw reviewers attention to this method. When we remove a step, we should remove the attributes that are no longer permitted, in order for the data to be self-consistent (e.g. so we don't display something on the review step or the listing that is not applicable.)

david-mears-dfe: OK, this is now going to be changed to use the new Resettable module that @robyoung91 is creating in https://github.com/DFE-Digital/teaching-vacancies/pull/4070

app/models/vacancy.rb
```diff
@@ -5,6 +5,7 @@ class Vacancy < ApplicationRecord
   extend ArrayEnum
 
   include Indexable
+  include Vacancy::Phaseable
```
david-mears-dfe, line 8: Might not need `Vacancy::`

PR #4045, opened 2021-09-27
Lint code

david-mears-dfe:
Tighten rubocop metrics. Exclude algolia from block length metric as it is an extreme outlier (70 or so lines), and because we are planning to remove this code soon.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -83,17 +83,17 @@ Naming/AccessorMethodName:
 # We should consider slowly moving closer to the default settings.
 
 Metrics/AbcSize:
-  Max: 32 # Default 17
+  Max: 31 # Default 17
   Exclude:
     - "config/**/*"
-    - "lib/**/*"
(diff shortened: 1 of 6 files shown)


app/models/concerns/indexable.rb
```diff
@@ -140,3 +141,4 @@ def _geoloc
                  .map { |organisation| { lat: organisation.geolocation.x.to_f, lng: organisation.geolocation.y.to_f } }
   end
 end
+# rubocop:enable Metrics/BlockLength
```
csutter, line 144: You don't need this if it applies to the whole file. Even better, you can also put the `rubocop:disable` comment right next to the beginning of the block it applies to, i.e.:

```ruby
module Indexable # rubocop:disable Metrics/BlockLength
  # ...
end
```

cesidio: or maybe exclude this file from this rule in rubocop.yml file

PR #4054, opened 2021-09-29
[TEVA-3198] Update hiring staff sign-in journey

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3198

## Screenshots of UI changes:

Screenshot 2021-10-05 at 09 32 42


app/mailers/admins/account_mailer.rb
```diff
@@ -1,9 +0,0 @@
-class Admins::AccountMailer < Mail::Notify::Mailer
-  def account_creation_request(record, _opts = {})
-    @template = NOTIFY_ADMIN_ACCOUNT_CREATION_REQUEST_TEMPLATE
```
david-mears-dfe, line 3: Please delete this constant from initializers/notification.rb too, and, once merged, you could delete the template from Notify itself.

app/models/account_request.rb
```diff
@@ -1,2 +0,0 @@
-class AccountRequest < ApplicationRecord
```
david-mears-dfe, line 1: Do we want to delete this table at some future date? Or leave the records around? @dfe-teaching-vacancies

josephhull676: Hmm we didn't do anything with it besides passing it into a mailer. I know the table was included in analytics.yml, but I'm not sure what's done with this file to be honest.

csutter: I think killing it is fine - don't know why this existed in the first place given all that happens is it being sent somewhere 👍

cesidio:
https://github.com/DFE-Digital/teaching-vacancies/commit/153257400f9b05268b960272fb936db8409c037e

this is what has been added for account requests, you want to double check with it. I think you have removed everything except some I18n keys in `config/locales/mailers.yml`

db/migrate/20210929121317_drop_account_requests_table.rb
```diff
@@ -0,0 +1,12 @@
+class DropAccountRequestsTable < ActiveRecord::Migration[6.1]
+  def change
+    drop_table "account_requests", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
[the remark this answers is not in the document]
david-mears-dfe: Don't want to speak for Joe but one reason to include the other stuff is in order to make this migration able to be rolled back. We probably don't want to rollback this migration, but if we wanted to do many at once, this one would interrupt you with a 'can't be rolled back' error unless you include the info on how to recreate the table.

bencmitchell:
Can we change the redirect on this "Log in" link to the DFE sign in page the current process links to (directly to where you can enter your email and password, as opposed to this interstitial page). 
image

image

bencmitchell:
Button should read "Request a DFE Sign-in account 
image

cesidio (approved):

PR #4059, opened 2021-09-30
[TEVA-3043] Add status tags to job listing edit sections

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3043

## Changes in this PR:

- Are the tests a bit overkill / unnecessary? 
- Wasn't sure where to put the translations for the tags. What do you think?

## Screenshots of UI changes:

Screenshot 2021-09-30 at 11 53 12

### DIFF
--- a/app/helpers/job_applications_helper.rb
+++ b/app/helpers/job_applications_helper.rb
@@ -80,18 +80,6 @@ module JobApplicationsHelper
     t("buttons.complete_section")
   end
 
-  def job_application_review_section_tag(job_application, step, form_class)
-    tag_attributes = if form_class.fields.any? { |field| field.in?(job_application.errors.messages.keys) }
-                       { text: t("shared.status_tags.action_required"), colour: "orange" }
-                     elsif step.to_s.in?(job_application.completed_steps)
-                       { text: t("shared.status_tags.complete") }
-                     else
-                       { text: t("shared.status_tags.not_started"), colour: "red" }
-                     end
-
-    govuk_tag(**tag_attributes)
-  end
-
   def job_application_build_submit_button_text
     if redirect_to_review?
       t("buttons.save")
--- a/app/helpers/job_applications_helper.rb
+++ b/app/helpers/job_applications_helper.rb
@@ -80,14 +80,13 @@ module JobApplicationsHelper
     t("buttons.complete_section")
   end
 
-  def job_application_review_section_tag(job_application, step)
-    form_class = "jobseekers/job_application/#{step}_form".camelize.constantize
+  def job_application_review_section_tag(job_application, step, form_class)
     tag_attributes = if form_class.fields.any? { |field| field.in?(job_application.errors.messages.keys) }
-                       { text: t("messages.jobs.action_required.label"), colour: "orange" }
+                       { text: t("shared.status_tags.action_required"), colour: "orange" }
                      elsif step.to_s.in?(job_application.completed_steps)
-                       { text: "complete" }
+                       { text: t("shared.status_tags.complete") }
                      else
-                       { text: "not started", colour: "red" }
+                       { text: t("shared.status_tags.not_started"), colour: "red" }
                      end
 
     govuk_tag(**tag_attributes)
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -80,24 +80,6 @@ module VacanciesHelper
     t("school_groups.job_location_heading.at_multiple_schools", organisation_type: organisation_type_basic(vacancy.parent_organisation))
   end
 
-  def vacancy_review_section_tag(vacancy, steps)
-    form_classes = steps.map { |step| "publishers/job_listing/#{step}_form".camelize.constantize }
-
-    return govuk_tag(text: t("shared.status_tags.not_started"), colour: "grey") if steps.none? { |step| vacancy_step_completed?(vacancy, step) }
-
-    return govuk_tag(text: t("shared.status_tags.action_required"), colour: "red") if vacancy_forms_contain_errors?(vacancy, form_classes)
-
-    govuk_tag(text: t("shared.status_tags.complete"))
-  end
-
-  def vacancy_forms_contain_errors?(vacancy, form_classes)
-    form_classes.each do |form_class|
-      return true if form_class.fields.any? { |field| field.in?(vacancy.errors.messages.keys) }
-    end
-
-    false
-  end
-
   def vacancy_school_visits_hint(vacancy)
     organisation = organisation_type_basic(vacancy.parent_organisation).tr(" ", "_")
     t("helpers.hint.publishers_job_listing_applying_for_the_job_form.#{organisation}_visits")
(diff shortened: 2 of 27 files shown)


csutter (approved):

app/helpers/job_applications_helper.rb
```diff
@@ -83,11 +83,11 @@ def job_application_review_edit_section_text(job_application, step)
   def job_application_review_section_tag(job_application, step)
     form_class = "jobseekers/job_application/#{step}_form".camelize.constantize
```
david-mears-dfe, line 84: Would it be simpler to pass in the form class directly (in both the helper methods) rather than reconstruct it this way?

app/views/publishers/vacancies/edit_vacancy_sections/_edit_job_role.html.slim
```diff
@@ -1,19 +1,18 @@
...
-                   href: organisation_job_build_path(@vacancy.id, :job_role)
+                   href: organisation_job_build_path(vacancy.id, :job_role)
+    = vacancy_review_section_tag(vacancy, :job_role)
 
   - review.body do
     = govuk_summary_list do |summary_list|
       - summary_list.row do |row|
         - row.key text: t("jobs.main_job_role")
-        - row.value text: @vacancy.show_main_job_role || t("jobs.not_defined")
+        - row.value text: vacancy.show_main_job_role || t("jobs.not_defined")
```
david-mears-dfe, line 13: This kind of correction applies in all the _edit_whatever.html.slim files, it seems! Worth doing a separate PR for all those?

spec/system/publishers_can_edit_a_draft_vacancy_as_a_school_spec.rb
```diff
@@ -3,26 +3,53 @@
...
-      expect(page).to have_content(I18n.t("pay_package_errors.salary.blank"))
-      expect(page).to_not have_content(I18n.t("buttons.submit_job_listing"))
+      scenario "correct status tag is displayed" do
+        within(".review-component#pay_package") do
+          expect(page).to have_content(I18n.t("shared.status_tags.action_required"))
+        end
+      end
+    end
+
+    context "when a step has not been completed" do
```
david-mears-dfe, line 32: Can you make this description more precise? The ticket says the status tag is 'not started' *when no question has been answered* in that section.

david-mears-dfe:
> Are the tests a bit overkill / unnecessary?

They would run more quickly if you made them into unit tests in a helper spec, rather than entire system specs.

app/helpers/job_applications_helper.rb
josephhull676: Makes sense

app/helpers/job_applications_helper.rb
```diff
@@ -80,14 +80,13 @@ def job_application_review_edit_section_text(job_application, step)
     t("buttons.complete_section")
   end
 
-  def job_application_review_section_tag(job_application, step)
-    form_class = "jobseekers/job_application/#{step}_form".camelize.constantize
+  def job_application_review_section_tag(job_application, step, form_class)
```
david-mears-dfe, line 83: No need for `step`

app/helpers/vacancies_helper.rb
```diff
@@ -80,6 +80,18 @@ def vacancy_job_location_heading(vacancy)
     t("school_groups.job_location_heading.at_multiple_schools", organisation_type: organisation_type_basic(vacancy.parent_organisation))
   end
 
+  def vacancy_review_section_tag(vacancy, step, form_class)
```
david-mears-dfe, line 83: No need for `step`

app/views/jobseekers/job_applications/review/_ask_for_support.html.slim
```diff
@@ -4,7 +4,7 @@
   - review.heading title: t("jobseekers.job_applications.build.ask_for_support.heading"),
                    text: job_application_review_edit_section_text(job_application, :ask_for_support),
                    href: jobseekers_job_application_build_path(job_application, :ask_for_support) do
-      = job_application_review_section_tag(job_application, :ask_for_support)
+      = job_application_review_section_tag(job_application, :ask_for_support, Jobseekers::JobApplication::AskForSupportForm)
```
david-mears-dfe, line 7: So in all these you can leave out the step

app/views/publishers/vacancies/edit_vacancy_sections/_edit_job_location.html.slim
```diff
@@ -1,12 +1,11 @@
-= render "publishers/vacancies/error_tag", attributes: (Publishers::JobListing::JobLocationForm.fields + Publishers::JobListing::SchoolsForm.fields)
-
 = render_divs_for_fields(Publishers::JobListing::JobLocationForm)
 = render_divs_for_fields(Publishers::JobListing::SchoolsForm)
 
 = render ReviewComponent.new id: "job_location" do |review|
   - review.heading title: t("publishers.vacancies.steps.job_location"),
                    text: t("buttons.change"),
                    href: organisation_job_build_path(@vacancy.id, :job_location)
+    = vacancy_review_section_tag(vacancy, :job_location, Publishers::JobListing::JobLocationForm)
```
david-mears-dfe, line 8: See the deleted line in this file. We need to somehow incorporate the SchoolsForm or the schools step, as well as the job location step. Perhaps the helpers should take an array of form fields, in the same way as the deleted line in this file, since they don't directly use the form class itself.

spec/system/publishers_can_edit_a_draft_vacancy_as_a_school_spec.rb
```diff
@@ -44,13 +45,23 @@
...
+        %w[job_role job_details pay_package working_patterns important_dates documents applying_for_the_job job_summary]
+      end
 
+      scenario "can submit a completed draft from the manage job listing page" do
         click_on I18n.t("buttons.submit_job_listing")
 
         expect(page.current_path).to eq(organisation_job_summary_path(complete_vacancy.id))
       end
+
+      scenario "correct status tag is displayed" do
```
david-mears-dfe, line 58: I don't think this quite fits in the describe "submitting a completed draft" because that doesn't happen in this scenario, not sure what to rename it to though

app/views/publishers/vacancies/edit_vacancy_sections/_edit_job_location.html.slim
josephhull676: I've made some changes to account for this. Take a look.

spec/system/publishers_can_edit_a_draft_vacancy_as_a_school_spec.rb
josephhull676: Made some changes. Take a look!

david-mears-dfe (approved):
In an ideal world I would want to get rid of the duplicated helpers so you don't have to make changes to both methods. E.g. when you made the vacancy helper able to account for multiple steps at once, you didn't get that for free with the other helper. But I will approve so that you can decide for yourself if you want to address this.

PR #4070, opened 2021-10-04
[TEVA-3213] Add callback to reset dependent fields

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3213

## Changes in this PR:

This PR adds a module, Resettable, which is included by the Vacancy model and implements a callback which ensures that the Actual Salary field is reset and hidden if a job listing is changed to Full-time working pattern from any working pattern which uses Actual Salary.

The callback also ensures that if a job listing with a Fixed Term contract and defined contract_type_duration is changed to a Permanent contract, then the contract_type_duration is reset.

### DIFF
--- a/app/views/publishers/vacancies/edit_vacancy_sections/_edit_pay_package.html.slim
+++ b/app/views/publishers/vacancies/edit_vacancy_sections/_edit_pay_package.html.slim
@@ -13,9 +13,10 @@
         - row.key text: t("jobs.annual_salary")
         - row.value text: vacancy.salary
 
-      - summary_list.row do |row|
-        - row.key text: t("jobs.actual_salary")
-        - row.value text: vacancy.actual_salary.presence || t("jobs.not_defined")
+      - unless vacancy.model_working_patterns == ["full_time"]
+        - summary_list.row do |row|
+          - row.key text: t("jobs.actual_salary")
+          - row.value text: vacancy.actual_salary.presence || t("jobs.not_defined")
 
       - summary_list.row do |row|
         - row.key text: t("jobs.benefits")
--- a/app/models/concerns/vacancy/resettable.rb
+++ b/app/models/concerns/vacancy/resettable.rb
@@ -1,0 +1,24 @@
+module Vacancy::Resettable
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :reset_dependent_fields
+  end
+
+  def reset_dependent_fields
+    reset_actual_salary
+    reset_contract_type_duration
+  end
+
+  def reset_actual_salary
+    return unless working_patterns_changed? && working_patterns == ["full_time"]
+
+    self.actual_salary = ""
+  end
+
+  def reset_contract_type_duration
+    return unless contract_type_changed? && contract_type == "permanent"
+
+    self.contract_type_duration = ""
+  end
+end
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -5,6 +5,7 @@ class Vacancy < ApplicationRecord
   extend ArrayEnum
 
   include Indexable
+  include Resettable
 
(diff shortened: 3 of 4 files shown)


.rubocop.yml
```diff
@@ -96,7 +96,7 @@ Metrics/BlockLength:
     - "app/models/concerns/indexable.rb"
 
 Metrics/ClassLength:
-  Max: 156 # Default 100
+  Max: 167 # Default 100
```
david-mears-dfe, line 99: I want to resist us raising these limits. They should gradually trend downwards over time. Can we extract the resetting callbacks to a separate module?

david-mears-dfe (commented):
I solved a parallel problem in my open PR:

https://github.com/DFE-Digital/teaching-vacancies/blob/3b5ba2ef2fa1b80049691b70ebc4396fbd9d333f/app/models/vacancy.rb#L211

Which style do we prefer - putting all the resets into a single method or multiple methods?

app/views/publishers/vacancies/edit_vacancy_sections/_edit_pay_package.html.slim
```diff
@@ -13,9 +13,10 @@
         - row.key text: t("jobs.annual_salary")
         - row.value text: vacancy.salary
 
-      - summary_list.row do |row|
-        - row.key text: t("jobs.actual_salary")
-        - row.value text: vacancy.actual_salary.presence || t("jobs.not_defined")
+      - unless vacancy.__getobj__.working_patterns == ["full_time"]
```
josephhull676, line 16: I'm not too sure what `__getobj__` is doing there, but `vacancy.working_patterns` will do I think!

.rubocop.yml
robyoung91: Good idea - I have moved these resetting callbacks into the Resettable module

app/views/publishers/vacancies/edit_vacancy_sections/_edit_pay_package.html.slim
robyoung91: The __getobj__ allows us to access working_patterns directly from the Vacancy model rather than the VacancyPresenter which maps them to a different format (which could potentially change and cause us issues in the future?)

david-mears-dfe: This line means we're find to call `working_patterns` on the presenter, I think.

```ruby
delegate :working_patterns, to: :model, prefix: true
```

app/models/concerns/resettable.rb
```diff
@@ -0,0 +1,24 @@
+module Resettable
```
david-mears-dfe, line 1: Could you namespace this under concerns/vacancy/resettable? The module will be called Vacancy::Resettable. I ask because I'm creating the vacancy directory in my open PR.

app/views/publishers/vacancies/edit_vacancy_sections/_edit_pay_package.html.slim
david-mears-dfe: Oh, I just saw that the presenter has a working_patterns method as well as that delegator. Confused now!

david-mears-dfe: Why does it have two presenter methods for working_patterns AND a delegator :/

robyoung91: Cesi tells me we can use the prefix specified in the delegator i.e. `model_working_patterns` to access the delegator which looks nicer than using `__getobj__`

david-mears-dfe (approved):

PR #4077, opened 2021-10-05
Enable entity events for organisation import

csutter:
Since we have refactored the organisation import logic, the import has
started working again, but we've lost the automatic entity events
fired from `ApplicationRecord` (due to the bulk import bypassing AR).

This adds a hash of GIAS data to organisation records, which is
refreshed on a nightly basis after the import scheduled job runs, and
triggers an update (and hence an event) if the GIAS data has changed.

It also adds `gias_data` for LAs (name and code) which wasn't there
previously.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3244


cesidio (approved):
Very elegant. Love it

PR #4078, opened 2021-10-05
Use govuk_footer component

cesidio:


csutter (approved):

alexbowen (approved):
nice, good to put header/footer into partials and make the app layout more succinct

alexbowen:
dont we need the ogl licence bit though?

PR #4089, opened 2021-10-06
[TEVA-3168] Change Action required summary banner title

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3168

## Changes in this PR:

This PR amends the banner title of the "Action required" summary to "Action required" rather than "Warning". It also removes the "Action required" heading.

### Before

image

### After

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -18,6 +18,7 @@ en:
     important: Important
     success: Success
     warning: Warning
+    action_required: Action required
 
   breadcrumbs:
     active_jobs: Active jobs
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -18,6 +18,7 @@ en:
     important: Important
     success: Success
     warning: Warning
+    action_required: Action required
```
robyoung91, line 21: For reviewers: I've added this new header under banners rather then reusing the existing [jobs.action_required.heading](https://github.com/DFE-Digital/teaching-vacancies/blob/master/config/locales/notifications.yml#L8) one, my logic being that it's a banner title. Is this the best thing to do?

Furthermore, should I delete `jobs.action_required.heading`, as it isn't being used anywhere now?

csutter: Sounds good! Yeah, feel free to kill it.

cesidio: Yes you can remove it now that it's not used anymore

cesidio (approved):

PR #4093, opened 2021-10-06
Add application form preview file

david-mears-dfe:
Necessary for upcoming tickets.


robyoung91 (commented):
did someone mention in the technical review meeting that we might want to put it in `public/pdf/` or something similar? (would that work?)

PR #4129, opened 2021-10-12
[TEVA-2679] Reduce chance of encoding errors with open/close quotes

david-mears-2:
Using the html & character directly, rather than appending something using css pseudo
classes, is more conventional and removes an encoding step.

I was unable to duplicate the reported bug where these quotes had been seen in a mangled
form: â€™

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2679


david-mears-dfe:
https://teaching-vacancies-review-pr-4129.london.cloudapps.digital/

PR #4138, opened 2021-10-13
[TEVA-3270] remove phase banner border bottom from home page

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-3270


jesseyuen:
I'd say we are close to removing this banner so we will need to apply the same fix here.

Screenshot 2021-10-13 at 08 34 33

PR #4193, opened 2021-10-25
[TEVA-2883] change important date to radio selectors with reveal content

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-2883

Screenshot 2021-10-25 at 11 05 40

Also removes some JS behaviour for clearing fields as not needed with conditional reveal

### DIFF
--- a/app/views/publishers/vacancies/build/important_dates.html.slim
+++ b/app/views/publishers/vacancies/build/important_dates.html.slim
@@ -32,11 +32,10 @@
           ->(option) { option },
           ->(option) { t("helpers.options.publishers_job_listing_extend_deadline_form.expiry_time.#{option}") }
 
-        .clear-form
-          = f.govuk_date_field :starts_on
-
-          .clear-form__checkbox
-            = f.govuk_check_box :starts_asap, "true", 0, multiple: false, link_errors: false
+        = f.govuk_radio_buttons_fieldset :starts_asap, legend: { text: t("helpers.legend.publishers_job_listing_important_dates_form.starts_on") }, classes: "clear-form" do
+          = f.govuk_radio_button :starts_asap, "false", link_errors: true, label: { text: t("helpers.legend.publishers_job_listing_important_dates_form.start_date_specific") } do
+            = f.govuk_date_field :starts_on, legend: { text: "Enter date" }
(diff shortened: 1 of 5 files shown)


app/views/publishers/vacancies/build/important_dates.html.slim
```diff
@@ -32,11 +32,11 @@
...
 
-        .clear-form
-          = f.govuk_date_field :starts_on
+        = f.govuk_radio_buttons_fieldset :starts_asap, legend: { text: t("helpers.legend.publishers_job_listing_important_dates_form.starts_on") } do
+          = f.govuk_radio_button :starts_asap, 0, link_errors: true, label: { text: t("helpers.legend.publishers_job_listing_important_dates_form.start_date_specific") } do
+            = f.govuk_date_field :starts_on, legend: { text: "Enter date" }
 
-          .clear-form__checkbox
-            = f.govuk_check_box :starts_asap, "true", 0, multiple: false, link_errors: false
+          = f.govuk_radio_button :starts_asap, 1, label: { text: t("helpers.legend.publishers_job_listing_important_dates_form.start_asap") }
```
david-mears-dfe, line 39: (recapitulating Slack message)

I can’t get it to save the update if you select 'start as soon as possible'.

When I save ‘as soon as poss’ I get this in the console:
```ruby
v.reload.starts_on
=> Tue, 25 Oct 2022
```

These lines in ImportantDatesForm are no longer doing their jobs, because starts_asap is coming through as “1” rather than “true”:
```ruby
starts_on: (starts_on unless starts_asap == "true"),
# ...
errors.add(:starts_on, :date_and_asap) if starts_on.present? && starts_asap == "true"
```

I suggest we rewrite this form to be coded more defensively: instead of
```ruby
starts_asap == "true"
```
use
```ruby
["true", "1"].include?(starts_asap)
```

david-mears-dfe: The validation `starts_on_and_starts_asap_not_present` might need a rethink as to whether it's still necessary now that the questions are rolled into one thing.

david-mears-dfe (changes requested):
See comments!

david-mears-dfe (approved):

PR #4212, opened 2021-10-26
[TEVA-3319] Add Search form A:B test

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3319

## Changes in this PR

This PR implements an A:B test for the search results filters on the mobile version of our service by creating a panel component which contains the code and styling for a panel overlay. If the search overlay is not used and removed following AB testing, the panel component is intended to allow overlay functionality to be decoupled (e.g. from the filters component) in a later refactor.

### Before

image

### After

image


alexbowen (approved):

david-mears-dfe (commented):
I'm wondering why you decided to write PanelComponent as a component, rather than directly writing the content in the slim file, since the component's only used once?

robyoung91:
> I'm wondering why you decided to write PanelComponent as a component, rather than directly writing the content in the slim file, since the component's only used once?

Our thinking behind this was that the FilterComponent requires (and implements) similar overlay functionality and that we could refactor this at a later time to use the PanelComponent code, making it less tightly coupled. We can also use PanelComponent for any other similar overlays we may want to implement in the future.

david-mears-dfe (commented):
Screenshot 2021-11-01 at 13 11 05

Looks like one button has a border at the top and the other doesn't?

PR #4221, opened 2021-10-27
[TEVA-3352] Vary the default search radius depending on location type

david-mears-dfe:
Do so by splitting out some new radius logic from Search::LocationBuilder into a Search::RadiusBuilder class, which is consumed by LocationBuilder, SubscriptionForm, and SearchForm.

The logic is as follows:

If a user has selected the (new) '0' option ('This area only') for the radius of a search/job alert, we change this to a default value for point-based location searches. Otherwise, we keep the radius entered.

Other changes:

- Set the default radius to the default for polygon searches, which is now 0.

- Fix bug in subscription presenter

The render_location_filter was ignoring radiuses for polygon-based location searches, e.g. said 'In Oxford' when the search was for Oxford plus 100 miles radius.

-  Update and refactor search heading component

We can drop the logic related to distinguishing between polygon and location searches, and also simplify by creating each section of the heading separately before stringing them together (rather than writing out every possible permutation in the translations).

I have written the tests for the search heading component to test the strings directly rather than to test against a translation, primarily because the translations are more brittle than the strings themselves (if the tests had already been strings I would have had to change very few of them in this commit) and secondarily because it is safer to test the strings because translations, when not found, return a string rather than an error.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3352

### DIFF
--- a/app/services/search/vacancy_search.rb
+++ b/app/services/search/vacancy_search.rb
@@ -67,8 +67,8 @@ class Search::VacancySearch
                              Search::Strategies::PgSearch.new(
                                keyword,
                                filters: search_criteria,
-                               location: search_criteria[:location],
-                               radius: search_criteria[:radius],
+                               location: location_search.location_filter[:location],
+                               radius: location_search.radius,
                                page: page,
                                per_page: per_page,
                                sort_by: sort_by,
--- a/spec/components/jobseekers/search_results/heading_component_spec.rb
+++ b/spec/components/jobseekers/search_results/heading_component_spec.rb
@@ -14,14 +14,14 @@ RSpec.describe Jobseekers::SearchResults::HeadingComponent, type: :component do
     allow(vacancies_search).to receive(:keyword).and_return(keyword)
     allow(vacancies_search).to receive_message_chain(:location_search, :location).and_return(location)
     allow(vacancies_search).to receive_message_chain(:total_count).and_return(count)
-    allow(vacancies_search).to receive_message_chain(:search_criteria, :[]).and_return(radius)
+    allow(vacancies_search).to receive_message_chain(:location_search, :radius).and_return(radius)
   end
 
   context "when landing_page is a job role" do
     let(:landing_page) { "teaching-assistant" }
 
     it "renders correct heading" do
-      expect(subject.text).to eq(I18n.t("jobs.search_result_heading.landing_page_html", jobs_count: count, landing_page: landing_page.titleize.downcase, count: count))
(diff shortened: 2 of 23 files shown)


app/services/search/vacancy_search.rb
```diff
@@ -67,8 +67,8 @@ def search_strategy
                              Search::Strategies::PgSearch.new(
                                keyword,
                                filters: search_criteria,
-                               location: search_criteria[:location],
-                               radius: search_criteria[:radius],
+                               location: location_search.location_filter[:location],
+                               radius: location_search.radius,
```
david-mears-dfe, line 71: @csutter `location_search` has some logic in it (from `Search::LocationBuilder`) that alters the radius attribute. Should we
(a) pass in this attribute to `PgSearch`, as above, or
(b) replicate this logic within `VacancyLocationQuery`?

csutter: I reckon a) does the job for now.

My general approach for the `PgSearch` stuff has been to reuse as much as we can from the existing logic, and then when we go all in on the new search, slowly unpicking everything into new and possibly nicer code!

spec/components/jobseekers/search_results/heading_component_spec.rb
```diff
@@ -1,93 +1,64 @@
...
-    let(:location) { nil }
-    let(:polygon_boundaries) { nil }
-
-    it "renders correct heading" do
-      expect(rendered_component).to include(
-        I18n.t("jobs.search_result_heading.keyword_html", jobs_count: count, keyword: keyword, count: count),
-      )
+  context "when searching with a keyword and a location with a radius (polygon or point)" do
+    it "renders correct heading text" do
+      expect(subject.text).to eq("#{count} jobs match ‘#{keyword}’ within #{radius} miles of ‘#{location}’")
```
cesidio, line 38: much nicer to read 👍

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -14,7 +14,7 @@
         allow_any_instance_of(ApplicationController).to receive(:verify_recaptcha).and_return(false)
       end
 
-      it "redirects to invalid_recaptcha path" do
+      scenario "redirects to invalid_recaptcha path" do
```
cesidio, line 17: I find `it` more readable than `scenario`

cesidio (approved):

PR #4232, opened 2021-11-02
[TEVA-3359] Update home page CTAs for hiring staff

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3359

## Changes in this PR:

The ticket wants the 'create a job listing' link to go the first step of the create a job listing process. This page requires a POST request as it creates a vacancy. This means we have to use a button in a form to create a post request, and style it like a link.

An alternative idea would be to create a get action that redirects to a POST request.

Open to further alternative ideas!

### Signed in as a jobseeker

Screenshot 2021-11-02 at 12 04 38

### Signed out of everything

Screenshot 2021-11-02 at 12 04 55

### Signed in as a publisher

Screenshot 2021-11-02 at 12 05 21

### DIFF
--- a/app/frontend/src/styles/base/_utilities.scss
+++ b/app/frontend/src/styles/base/_utilities.scss
@@ -1,3 +1,12 @@
+input.button-as-link {
+  background: none;
+  border: 0;
+  cursor: pointer;
+  margin-bottom: 0;
+  outline: inherit;
+  padding: 0;
+}
+
 .wordwrap {
   word-wrap: break-word;
 }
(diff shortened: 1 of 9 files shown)


app/frontend/src/styles/base/_utilities.scss
```diff
@@ -1,3 +1,12 @@
+input.button_as_link {
```
csutter, line 1: Dashes please!

csutter (approved):

PR #4234, opened 2021-11-02
Fix new features page loop bug

josephhull676:
Whilst working on another ticket, I noticed that when visiting the new features page after logging in, I got stuck in a loop:

https://user-images.githubusercontent.com/30624173/139918223-3e8e7518-869a-4af0-8194-931cc311ae6a.mov

Before redirecting to the new features page, we check session[:visited_new_features_page]. If it is nil then we redirect to the new features page. 

To prevent the loop, session[:visited_new_features_page] was set to true when the new features controller's show action was hit. It appears someone removed this at some point, causing this bug.

### DIFF
--- a/app/controllers/publishers/new_features_controller.rb
+++ b/app/controllers/publishers/new_features_controller.rb
@@ -4,6 +4,7 @@ class Publishers::NewFeaturesController < Publishers::BaseController
   def show
     @new_features_form = Publishers::NewFeaturesForm.new
     current_publisher.update(viewed_new_features_page_at: Time.current)
+    session[:visited_new_features_page] = true
   end
 
   def update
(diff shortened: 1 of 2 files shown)


david-mears-dfe:
> To prevent the loop, session[:visited_new_features_page] was set to true when the new features controller's show action was hit. It appears someone removed this at some point, causing this bug.

In the original PR, there was no setting of the session variable to true in the `show` action of the controller:

https://github.com/DFE-Digital/teaching-vacancies/pull/4112/files#diff-7ecfd768a3a467dcf2035db82a90e336e272b61deaa3ee17ec05798d599b1f90

So how did it ever work??

david-mears-dfe (changes requested):
I think the session[:thing] = true should be in the `update` action, according to that original PR

david-mears-dfe (approved):

PR #4249, opened 2021-11-05
Remove legacy scripts in `bin/`

csutter:
These are unused and hail from the days of a previous Docker-based setup
or are Rails defaults we don't use.


david-mears-dfe:
Goodbye, little scripts. 👋

david-mears-dfe (approved):

PR #4258, opened 2021-11-08
Update vacancy overviews with top and inline errors on publish attempt

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3353

## Changes in this PR:

```
Publishers can edit a draft vacancy
  when a school group
    shows the status of each stage
    can edit job location
    indicates that you're reviewing a draft
    when incomplete and previewed
      provides inline validation errors which link to the relevant form parts
      provides top-of-page validation errors which link to the relevant form parts
    when completed and previewed
      shows the preview
    when completed and submitted
      publishes the draft
    when incomplete and submitted for publication
      provides inline validation errors which link to the relevant form parts
      provides top-of-page validation errors which link to the relevant form parts
  when a single school
    indicates that you're reviewing a draft
    shows the status of each stage
    when incomplete and previewed
      provides inline validation errors which link to the relevant form parts
      provides top-of-page validation errors which link to the relevant form parts
    when completed and previewed
      shows the preview
    when incomplete and submitted for publication
      provides top-of-page validation errors which link to the relevant form parts
      provides inline validation errors which link to the relevant form parts
    when completed and submitted
      publishes the draft

17 examples, 0 failures
```

## Screenshots of UI changes:

Screenshot 2021-11-08 at 15 56 23


david-mears-dfe (commented):
I see you've used `local_assigns[:variable]` in various places. Why do you use that rather than referencing `variable` directly?

elliotcm:
> I see you've used `local_assigns[:variable]` in various places. Why do you use that rather than referencing `variable` directly?

@david-mears-dfe They're optional values, so it's cleaner than having to check with `defined?` (which I hate doing anyway).  Passing in nil and having the target complain if it needs it is IMO better than having "method not defined" errors.

app/components/validatable_summary_list_component.rb
```diff
@@ -0,0 +1,21 @@
+class ValidatableSummaryListComponent < GovukComponent::SummaryListComponent
+  registered_slots.delete(:rows)
```
david-mears-dfe, line 2: Checking I understand what this does: it removes the rows defined in the superclass so we can write our own?

elliotcm: That's correct, yes.  Unfortunately there's no "good" way to do that.

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -20,8 +20,18 @@ def initialize(current_step, vacancy:, organisation:, session:)
     })
   end
 
-  def validatable_steps
-    steps - %i[documents review]
+  def validatable_steps(top_level: false)
```
david-mears-dfe, line 23: I'm skeptical of including a parameter that is by default false and is not set to true anywhere but the tests. What is its purpose?

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -32,6 +42,24 @@ def previous_step_or_review
 
   private
 
+  def validate_step(step_name)
```
david-mears-dfe, line 45: I have a slight question mark about whether validation should really be the job of the VacancyStepProcess, whose main responsibility has been to house the logic that tracks which step should be rendered next. I'm open to it though.

app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
```diff
@@ -3,39 +3,22 @@
...
-        - summary_list.row do |row|
-          - row.key text: t("jobs.personal_statement_guidance")
-          - row.value text: vacancy.personal_statement_guidance.presence || t("jobs.not_defined")
-
+        - list.row :personal_statement_guidance, optional: true
       - else
-        - summary_list.row do |row|
-          - row.key text: t("jobs.application_link")
-          - row.value text: vacancy.application_link.present? ? open_in_new_tab_link_to(vacancy.application_link, vacancy.application_link, "aria-label": t("jobs.aria_labels.apply_link")) : t("jobs.not_defined")
+        - list.row :application_link, if_present: open_in_new_tab_link_to(vacancy.application_link, vacancy.application_link, "aria-label": t("jobs.aria_labels.apply_link"))
```
david-mears-dfe, line 17: Trying to understand what `if_present` is up to.

The RowComponent will do its default logic for displaying the attribute, unless there is an `if_present` present, in which case it will display it according to that (as long as the attribute is present). Right?

app/views/publishers/vacancies/edit_vacancy_sections/_edit_job_details.html.slim
```diff
@@ -4,30 +4,19 @@
...
     = review_section_tag(vacancy, %i[education_phases job_details], [Publishers::JobListing::EducationPhasesForm, Publishers::JobListing::JobDetailsForm])
 
   - review.body do
-    = govuk_summary_list do |summary_list|
+    = validatable_summary_list(vacancy, show_errors: local_assigns[:show_errors], error_path: organisation_job_build_path(vacancy.id, :job_details, back_to: local_assigns[:back_to])) do |list|
       - if vacancy.allow_phase_to_be_set?
-        - summary_list.row do |row|
-          - row.key text: t("jobs.filters.phases")
-          - row.value text: vacancy.phase&.humanize.presence || t("jobs.not_defined")
+        - list.row :phase, text: vacancy.phase&.humanize, label: t("jobs.filters.phases"), optional: true
```
david-mears-dfe, line 13: I appreciate this optional: parameter, it's a nice touch.

spec/controllers/publishers/vacancies/base_controller_spec.rb
```diff
@@ -9,7 +9,7 @@
 
       it "does perform the task" do
         expect(UpdateGoogleIndexQueueJob).to receive(:perform_later)
-        controller.update_google_index(vacancy)
+        controller.send(:update_google_index, vacancy)
```
david-mears-dfe, line 12: Not sure why!

spec/helpers/status_tag_helper_spec.rb
```diff
@@ -11,7 +11,7 @@
       context "when there is an error on the step's form object" do
         let(:steps) { [:personal_details] }
 
-        before { allow(job_application).to receive_message_chain(:errors, :messages).and_return({ city: "Invalid city" }) }
+        before { job_application.errors.add(:city) }
```
david-mears-dfe, line 14: 👍

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -3,13 +3,11 @@
...
+  private
 
-  def step_process
-    # TODO: We currently have to do this kinda thing in a lot of places thanks to `wicked`
-    #       Find a better way!
-    current_step = defined?(step) ? step : :review
+  helper_method :current_step, :step_process, :vacancy
 
+  def step_process
     ::Publishers::Vacancies::VacancyStepProcess.new(
```
david-mears-dfe, line 11: Do we need this `::`?

app/services/publishers/vacancies/vacancy_step_process.rb
elliotcm: I don't think I originally intended it to be that way.  The step grouping is a bit of a bodge from the look of it, so doing it from outside the class would require reaching inside a bit to know what to remove from the set, but I agree it probably shouldn't be there if it's not used elsewhere, though I do have a preference for leaving useful functionality in place, even if it's not used.

app/services/publishers/vacancies/vacancy_step_process.rb
elliotcm: I'm not married to it being the job of the step process (though I'm fine with it), so long as it's not scattered around all the controllers and such.

app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
elliotcm: Yeah there's some logic in some of the templates where "if the attribute is set, display this other thing".  It's not very elegant and I'm open to other approached.

spec/controllers/publishers/vacancies/base_controller_spec.rb
elliotcm: I made the methods in the base controller private, because there should be no public methods in a controller that you don't want to be actions.

Having made them private, we need to `send` in the tests in order to access them.  They were AFAICT only public to allow this testing.

A better way would be to properly subclass the base controller in the tests, but this felt good enough for now.

app/controllers/publishers/vacancies/base_controller.rb
elliotcm: Probably not, but I assumed they'd put it there for a reason.

app/services/publishers/vacancies/vacancy_step_process.rb
david-mears-dfe: Reasonable! Leaving conversation up in case others want to take sides.

app/services/publishers/vacancies/vacancy_step_process.rb
david-mears-dfe: Is 'put the validation bits into a module/class and include the module' a good or a bad pattern (here, and in general)? It has the advantage of separating the concerns (_somewhat_).

From the look of it, `all_steps_valid?` and `validate_all_steps?` know nothing about vacancies, so they could be taken out, and/or live in parent class `StepProcess`.

app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
david-mears-dfe: A low-hanging fruit improvement that would make it easier for me to read and understand in the view (without having to go read the component), is to call it `if_attribute_present`.

spec/controllers/publishers/vacancies/base_controller_spec.rb
david-mears-dfe: Makes sense!

app/controllers/publishers/vacancies/base_controller.rb
david-mears-dfe: I'll try PRing it out and see if anyone complains.

david-mears-dfe:
Worth making sure this branch is covered by a test? https://sonarcloud.io/code?id=DFE-Digital_teaching-vacancies&pullRequest=4258&selected=DFE-Digital_teaching-vacancies%3Aapp%2Fcontrollers%2Fpublishers%2Fvacancies%2Fbase_controller.rb&line=30

app/views/publishers/vacancies/edit_vacancy_sections/_edit_applying_for_the_job.html.slim
elliotcm: How about `value_if_attribute_present`?

elliotcm:
> Worth making sure this branch is covered by a test? https://sonarcloud.io/code?id=DFE-Digital_teaching-vacancies&pullRequest=4258&selected=DFE-Digital_teaching-vacancies%3Aapp%2Fcontrollers%2Fpublishers%2Fvacancies%2Fbase_controller.rb&line=30

Yes.  I'm a little surprised that it isn't.

app/services/publishers/vacancies/vacancy_step_process.rb
elliotcm: Putting things in modules for inclusion is good for composing functionality in different combinations, but since we're not doing that here I don't think it achieves much other than hide the code away.

I'm seeing the step process as a model that wraps the various form models, but perhaps that can be split out?

app/services/publishers/vacancies/vacancy_step_process.rb
elliotcm: I've removed this.

app/services/publishers/vacancies/vacancy_step_process.rb
elliotcm: I've created a new model to handle this.

david-mears-dfe: I like it! Thanks!

cesidio (approved):

david-mears-dfe:
I see lots of system specs but not unit tests for the new components/models. Would it duplicate things to add unit tests for those?

elliotcm:
@david-mears-dfe 
> I see lots of system specs but not unit tests for the new components/models. Would it duplicate things to add unit tests for those? In fact maybe I think we should have unit tests even if the models are already covered by system specs, because the unit tests will help us find bugs faster.

I agree.  I wouldn't normally take this approach, it's just that the higher level tests were in place before I saw the need to create new classes.

I would like to get this out, though, and adding unit tests will take a while.  Would you be OK with me coming back with another PR to add those unit tests?

david-mears-dfe:
> I would like to get this out, though, and adding unit tests will take a while. Would you be OK with me coming back with another PR to add those unit tests afterwards?

Good idea!

david-mears-dfe (approved):

elliotcm:
@david-mears-dfe FYI I've added the unit tests to this PR.

jesseyuen (approved):

PR #4270, opened 2021-11-10
[TEVA-3204] set compression on Cloudfront except Prod

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3240

### DIFF
--- a/terraform/app/modules/cloudfront/main.tf
+++ b/terraform/app/modules/cloudfront/main.tf
@@ -51,18 +51,27 @@ resource "aws_cloudfront_distribution" "default" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
-  dynamic "ordered_cache_behavior" {
-    for_each = local.cloudfront_cached_paths
+  ordered_cache_behavior {
 
-    content {
-      allowed_methods        = ["GET", "HEAD"]
-      cached_methods         = ["GET", "HEAD"]
-      path_pattern           = ordered_cache_behavior.value
-      target_origin_id       = "${var.service_name}-${var.environment}-default-origin"
-      viewer_protocol_policy = "redirect-to-https"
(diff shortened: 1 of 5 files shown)


terraform/workspace-variables/qa.tfvars
```diff
@@ -13,6 +13,8 @@ distribution_list = {
   }
 }
 
+enable_cloudfront_compress = "true"
```
saliceti, line 16: We could make `true` default and remove it from here
Make sure you use `true` as boolean and not `"true"` as string, it's cleaner

terraform/app/modules/cloudfront/main.tf
```diff
@@ -51,19 +51,29 @@ resource "aws_cloudfront_distribution" "default" {
...
 
+  ordered_cache_behavior {
+
+      allowed_methods        = ["GET", "HEAD"]
+      cached_methods         = ["GET", "HEAD"]
+      path_pattern           = "/attachments/*"
+      target_origin_id       = "${var.service_name}-${var.environment}-default-origin"
+      viewer_protocol_policy = "redirect-to-https"
+      cache_policy_id        = data.aws_cloudfront_cache_policy.managed-caching-optimized.id
+      compress               =  false
```
saliceti, line 73: The default value is false: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#compress
So it can be omitted here

terraform/workspace-variables/qa.tfvars
```diff
@@ -13,6 +13,8 @@ distribution_list = {
   }
 }
 
+enable_cloudfront_compress = true
```
saliceti, line 16: Please remove this, true is default now

saliceti (approved):

PR #4274, opened 2021-11-10
Implement query parser for Postgres search

csutter:
We use a managed Postgres database through GOV.UK PaaS and it does not
support custom synonym dictionaries for full-text search (because we do
not get filesystem access on the database server which Postgres requires
for loading dictionaries). This means we can't just add a synonym
dictionary/thesaurus to Postgres, use `websearch_to_tsquery`, and call
it a day - instead we need to parse our own queries and take synonyms
into account when doing so.

This implements a basic query parser with synonym support, using the
[`parslet` gem](http://kschiess.github.io/parslet) to parse user input
into an abstract "syntax" (well, ish) tree that can then be transformed
into an Arel AST which ActiveRecord can use to query.

- Add basic synonym dictionaries from Algolia as YAML files and load
  them into Rails configuration
- Add `Search::Postgres::QueryParser` to parse user search queries into
  a tree
- Add `Search::Postgres::QueryTransformer` to translate that tree into
  an Arel AST


david-mears-dfe:
💥

app/services/search/postgres/query_parser.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  rule(:synonym_term) { synonym_atoms.reduce(:|).as(:synonym_term) >> term_terminator }
+  rule(:oneway_synonym_term) { oneway_synonym_atoms.reduce(:|).as(:oneway_synonym_term) >> term_terminator }
+  rule(:plain_term) { match('\S').repeat(1).as(:plain_term) >> term_terminator }
+
+  rule(:term) { synonym_term | oneway_synonym_term | plain_term }
+
+  rule(:query) { (space | term).repeat(0).as(:query) }
+
+  root(:query)
```
david-mears-dfe, line 24: I prefer the code to be in the order you read it. This is Bob Martin's "step-down rule": as you read down, functions should go down a level of abstraction.

app/services/search/postgres/query_parser.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def synonym_atoms
```
david-mears-dfe, line 26: Is 'atoms' a name from Parslet? Alternative ideas: instances, tokens, strings

config/initializers/search.rb
```diff
@@ -0,0 +1,26 @@
+Rails.application.configure do
+  config_dir = Rails.root.join("config/search")
+
+  synonyms = YAML.load_file(config_dir.join("synonyms.yml"))
+  oneway_synonyms = YAML.load_file(config_dir.join("oneway_synonyms.yml"))
```
david-mears-dfe, line 5: I've struggled to come up with a different name for `terms_with_synonyms` and `terms_with_oneway_synonyms`. Maybe we could rename `synonyms` and `oneway_synonyms` instead to `thesaurus` and `one_way_thesaurus`, since the difference is that these contain information about relationships and mappings, whereas the `terms_with*` are lists of words that can be looked up in the thesauruses.

csutter: Ditto. I think "thesaurus" is nice, but on the other hand "synonym" is a bit more approachable.

I will sleep on it!

csutter: Found some slightly better names – we can always improve them further later 😄

alexbowen (approved):

PR #4276, opened 2021-11-11
[TEVA-3361] AB test alternative search results

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3361

## Changes in this PR:

- This PR adds an AB test variant which moves the search fields from the left column to the right column on the results page _for desktop users only_

### Before
image

### After
image

### DIFF
--- a/app/components/panel_component/panel.scss
+++ b/app/components/panel_component/panel.scss
@@ -11,6 +11,12 @@
         border-bottom: 0;
       }
 
+      &--mobile {
+        @include govuk-media-query($from: desktop) {
+          display: none;
+        }
+      }
+
       &:last-of-type {
         border-bottom: 0;
       }
--- a/app/frontend/src/styles/application/jobseekers/search-header.scss
+++ b/app/frontend/src/styles/application/jobseekers/search-header.scss
@@ -6,6 +6,25 @@
     flex-direction: row;
   }
 
+  .govuk-form-group,
+  .search-results__header__submit {
+    flex-basis: 25%;
+  }
+
+  .autocomplete,
+  .govuk-form-group {
+    margin-right: govuk-spacing(2);
+  }
+
+  &--border {
+    border-bottom: 1px solid $govuk-border-colour;
+    margin-bottom: govuk-spacing(6);
+
+    .govuk-heading-m {
+      display: none;
+    }
+  }
+  
   &-stats {
     @include govuk-media-query($from: tablet) {
       order: 2;
@@ -29,5 +48,11 @@
     .govuk-label {
       display: inline;
     }
+  }
+}
+
+.grid-column-right .search-results__header {
+  @include govuk-media-query($until: desktop) {
+    display: none;
   }
 }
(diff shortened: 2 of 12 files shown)


app/components/panel_component/panel.scss
```diff
@@ -28,6 +28,16 @@
       }
     }
   }
+
+  .search-controls__alternative {
```
david-mears-dfe, line 32: I think we should use double hyphen here. We're following this convention: https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

app/views/subscriptions/_fields.html.slim
```diff
@@ -1,5 +1,5 @@
 = render "shared/search/keyword", f: f, wide: false
-= render "shared/search/location", f: f, wide: false
+= render "shared/search/location", f: f, hint: true, wide: false
```
david-mears-dfe, line 2: Looks like `wide` doesn't do anything so we can get rid.

config/ab_tests.yml
```diff
@@ -50,3 +53,6 @@ test:
   2021_10_mobile_search_panel_test:
     present: 1
     overlay: 0
+  2021_11_desktop_search_results_page_test:
+    present: 1
+    alternative: 0
```
david-mears-dfe, line 58: Doesn't this mean 'always use the present version and never the alternative'?

robyoung91: Yes - I believe we set it like that under `test:` so we don't get the variants complicating things when we are running tests?

app/components/panel_component/panel.scss
alexbowen: yes correct @david-mears-dfe - alternative is a modifier strictly speaking

Dockerfile
```diff
@@ -1,11 +1,11 @@
-ARG PROD_PACKAGES="libxml2=2.9.12-r1 libxslt=1.1.34-r1 libpq=13.4-r0 tzdata=2021e-r0 shared-mime-info=2.1-r0 util-linux=2.37.2-r0 busybox=1.33.1-r6"
```
alexbowen, line 1: this file change should not be in this PR, some a bit screwy happened withe your rebasing

app/frontend/src/styles/application/jobseekers/search-header.scss
```diff
@@ -31,3 +50,9 @@
     }
   }
 }
+
+div.grid-column-right .search-results__header {
```
alexbowen, line 54: can we remove the `div.` from this selector, i do not think it is neccessary

config/ab_tests.yml
alexbowen, line 58: i would rather these names were a bit more descriptive `alternative` doesnt mean much to me, whereas `right` or `right_column` for example gives a bit more clue to whats going on

alexbowen:
doesnt appear to be working on ipad safari, probably needs a media query tweak - sorry
Screenshot 2021-11-16 at 21 39 54

alexbowen:
have tested in samsung android, iphone safari and ipad safari on browserstack - clean bill of health

cesidio (approved):

alexbowen:
this is good with me but for someone else can approve as i worked on it

PR #4288, opened 2021-11-15
[TEVA-3467] Track subscription origin better

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3467

## Changes in this PR:

- Add missing origin param to header links

- Track whether form is autopopulated in job_alert_subscription_created events

    This is to make Performance Analysis' life easier, by simplying the logic in PA that
    works out whether we expect the form to be autopopulated, into a progammatic check
    whether the form is in fact autopopulated, which is streamed as part of the event data
    for the job_alert_subscription_created event.

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -4,6 +4,7 @@ class SubscriptionsController < ApplicationController
   def new
     @point_coordinates = params[:coordinates_present] == "true"
     @ect_job_alert = params[:ect_job_alert]
+    session[:subscription_autopopulated] = params[:search_criteria].present?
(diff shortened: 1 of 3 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -89,12 +90,13 @@ def track_origin_and_trigger_event
   def trigger_subscription_event(type, subscription)
     request_event.trigger(
       type,
```
david-mears-dfe, line 92: This is an alphabetization plus the adding of `autopopulated`

cesidio (approved):

PR #4289, opened 2021-11-15
Update job to remove jobs from google index

josephhull676:
## Changes 

- We now batch the requests sent to the Indexing API to batches of 500. This is because we were hitting our daily limit of 1000 requests per day. 

- We set a flag on the vacancies that have been removed so that we only send one request per job.

### DIFF
--- a/app/jobs/remove_expired_vacancies_from_google_index_job.rb
+++ b/app/jobs/remove_expired_vacancies_from_google_index_job.rb
@@ -3,8 +3,9 @@ class RemoveExpiredVacanciesFromGoogleIndexJob < ApplicationJob
 
   def perform
     Rails.logger.info("Removing expired jobs from Google index")
-    Vacancy.expired.find_each do |vacancy|
-      RemoveGoogleIndexQueueJob.perform_later(Rails.application.routes.url_helpers.job_url(vacancy))
+    Vacancy.expired.where(google_index_removed: false).limit(500).each do |vacancy|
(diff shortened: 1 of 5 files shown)


app/jobs/remove_expired_vacancies_from_google_index_job.rb
```diff
@@ -3,8 +3,9 @@ class RemoveExpiredVacanciesFromGoogleIndexJob < ApplicationJob
 
   def perform
     Rails.logger.info("Removing expired jobs from Google index")
-    Vacancy.expired.find_each do |vacancy|
-      RemoveGoogleIndexQueueJob.perform_later(Rails.application.routes.url_helpers.job_url(vacancy))
+    Vacancy.expired.where(google_index_removed: false).limit(500).each do |vacancy|
+      RemoveGoogleIndexQueueJob.perform_now(Rails.application.routes.url_helpers.job_url(vacancy))
```
david-mears-dfe, line 7: Don't we prefer `perform_later`?

josephhull676: So this was based on a suggestion from Cesi. If we go with perform_later, these jobs are added to the queue. These jobs are then run concurrently. Cesi was concerned there may be a chance that Google doesn't like receiving concurrent requests, so we went with perform_now. There's only 500, so it shouldn't take too long.

cesidio (approved):

PR #4291, opened 2021-11-16
Remove csutter's double colon operator

david-mears-dfe:
I wanted to understand what this was for. I thought proposing to remove it would focus the mind. It has been asked before, to no reply: https://github.com/DFE-Digital/teaching-vacancies/pull/3943#discussion_r705204499


alexbowen (approved):
doesnt look right to me

csutter:
Oh yeah this can go.

The leading `::` _would_ be necessary if we had nested the controller definition in a module like so:

```ruby
module Publishers
  module Vacancies
    class BaseController
      # ...
```

... but because we use the "short" form of module nesting here it's superfluous (but not actively harmful).

csutter (approved):

PR #4292, opened 2021-11-16
Stop sending origin param for job alert subscription events

david-mears-dfe:
Performance Analysis will now rely solely on the new `autopopulated` parameter for these events, until such time as we deem it necessary to track the origin (this is unlikely as page_view events contain the request referer).

## Slack conversation

https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1637078984243600?thread_ts=1636970213.215900&cid=CG9A8H1HP

## Next

- [ ] Let Steven know when the PR is merged.


david-mears-dfe:
Closing because we think there is an existing bug (or some other reason) meaning that request events' referers come through as null in most cases.

david-mears-dfe:
Cesi and Steven got to the bottom of those blank referers

We have millions of bots sending us pageview requests without referers

But we just filter them out for analytics

So it's not a problem

cesidio (approved):

PR #4299, opened 2021-11-17
Update the email previews

david-mears-dfe:
We can preview the emails in the browser. Some were missing. These were added to aid content design to have overview of all content.

### DIFF
--- a/spec/mailers/previews/publishers/job_application_preview.rb
+++ b/spec/mailers/previews/publishers/job_application_preview.rb
@@ -1,0 +1,6 @@
+# Documentation: app/mailers/previewing_emails.md
+class Publishers::JobApplicationPreview < ActionMailer::Preview
+  def applications_received
+    # TODO
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/mailers/previews/publishers/job_application_preview.rb
```diff
@@ -0,0 +1,11 @@
+# Documentation: app/mailers/previewing_emails.md
+class Publishers::JobApplicationPreview < ActionMailer::Preview
+  def applications_received
+    publisher = Publisher.all.find { |p| p.vacancies.any? }
+    vacancy = publisher.vacancies.last
+
+    FactoryBot.create(:job_application, vacancy: vacancy, submitted_at: Date.yesterday, status: "submitted")
+
+    Publishers::JobApplicationMailer.applications_received(publisher: Publisher.first)
```
cesidio, line 9: I don't quite understand this. We first choose a publisher that has vacancies, then we create a job application for one of those vacancies and on the last line we use `Publisher.first`. Mail preview can assume that the seed has run and there are already submitted job applications, can't we use one of those?

david-mears-dfe: Ah yes, `Publisher.first` should be `publisher`. But I can rewrite this to use the seed data.

david-mears-dfe: I've just replaced the method with # TODO because it's not a priority right now. I have already generated the previews that were needed by Content Design.

cesidio (approved):

PR #4300, opened 2021-11-17
[TEVA-3441] convert autocomplete to stimulus component

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-3441

## Changes in this PR:

- make autocomplete a stimulus controller
- make controller responsible for adding markup as is JS only
- utilise the autocomplete `displayMenu` config for positioning

### DIFF
--- a/app/frontend/src/application.js
+++ b/app/frontend/src/application.js
@@ -2,6 +2,7 @@ import { Application } from '@hotwired/stimulus';
 
 import PanelController from '../../components/panel_component/panel';
 import UtilsController from './components/utils';
+import AutocompleteController from './components/autocomplete/autocomplete';
 
 const application = Application.start();
 
@@ -11,3 +12,4 @@ window.Stimulus = application;
 
 application.register('panel', PanelController);
 application.register('utils', UtilsController);
+application.register('autocomplete', AutocompleteController);
--- a/app/frontend/src/components/autocomplete/autocomplete.js
+++ b/app/frontend/src/components/autocomplete/autocomplete.js
@@ -1,3 +1,4 @@
+import { Controller } from '@hotwired/stimulus';
 import accessibleAutocomplete from 'accessible-autocomplete';
 import 'accessible-autocomplete/dist/accessible-autocomplete.min.css';
 import './autocomplete.scss';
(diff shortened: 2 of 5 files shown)


app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -6,40 +7,47 @@ import api from '../../lib/api';
...
   /* eslint-enable */
 };
 
-const autocomplete = () => {
-  Array.from(document.querySelectorAll('.autocomplete')).forEach((el) => {
-    const formInput = el.querySelector('input');
-    const dataSource = api[el.dataset.source];
+export default class extends Controller {
+  connect() {
+    const formInput = this.element.querySelector('input');
```
david-mears-dfe, line 20: Should we use a Stimulus' target API for this?

app/frontend/src/components/autocomplete/autocomplete.js
```diff
@@ -6,40 +7,47 @@ import api from '../../lib/api';
...
+
+    this.addContainers();
 
     if (formInput && dataSource) {
       let currentInputValue = formInput.value;
       formInput.parentNode.removeChild(formInput);
 
       accessibleAutocomplete({
-        element: el.querySelector('.accessible-autocomplete'),
+        element: this.element.querySelector('.accessible-autocomplete'),
```
david-mears-dfe, line 31: and for this?

app/frontend/src/components/autocomplete/autocomplete.js
alexbowen: i had a quandry about this tbh but i settled on no. thats because in this case its a stimulus controller being applied to a bit of html (input) that shouldnt/doesnt have knowledge of autocomplete being applied to it so it shouldnt have to explicitly define the input is a target.

app/frontend/src/components/autocomplete/autocomplete.js
alexbowen: similar, had a quandry, but i think it should be internal to the autocomplete js. its an odd one this as stimulus doesnt bring much to the party in this case. all we are doing is encapsulating govuk-accessible-autocomplete so i dont thing need to over engineer it. good spot pointing these out, i did give it a bit of thought but decided against. having said that i did think some more with your comments and have made things a little tighter by removing some uneccessary mark up and styles

alexbowen: having said that i did think some more because of your comment and have made things a little tighter by removing some uneccessary mark up and styles which i have updated this PR with

app/frontend/src/application.js
```diff
@@ -2,6 +2,7 @@ import { Application } from '@hotwired/stimulus';
 
 import PanelController from '../../components/panel_component/panel';
 import UtilsController from './components/utils';
+import AutoCompleteController from './components/autocomplete/autocomplete';
```
cesidio, line 5: shouldn't this be `AutocompleteController` ?

cesidio (approved):

app/frontend/src/application.js
alexbowen: whilst it doesnt actually matter can call it what we want, in keeping with the filename/folder name you are right

app/frontend/src/components/autocomplete/autocomplete.js
david-mears-dfe: How come the input shouldn't have knowledge of autocomplete being applied to it?

PR #4303, opened 2021-11-18
Update nodejs version in .tool-versions

cesidio:


.tool-versions
```diff
@@ -1,3 +1,3 @@
 ruby 3.0.2
-nodejs 12.18.3
+nodejs 12.22.0
 yarn 1.22.10
```
elliotcm, line 3: Slightly newer versions are available:
```suggestion
nodejs 12.22.7
yarn 1.22.17
```

elliotcm (changes requested):

PR #4311, opened 2021-11-22
Only try to look up local hostname in dev container

elliotcm:
This breaks outside the dev container.


cesidio (approved):

elliotcm:
We tried to check this for a dev containers user but apparently even `master` isn't working, so @csutter can chip in when he's back if this causes an issue.

PR #4318, opened 2021-11-23
[TEVA-3446] Sign in user after account confirmation even when already confirmed

david-mears-dfe:
This is one possible approach to the problem we suspect is caused by BingPreview clicking on one-time-use links in confirmation emails. As things stand, for some users the link will not work (will show as 'expired') when the account is already confirmed, because of BingPreview (or similar bots).

The Devise controller’s ‘show’ action has a yield resource if block_given? which I am hooking into here, in order to (in effect) remove any check that the account has already been confirmed. If the account’s already confirmed, the user has the same experience as when they are confirming for the first time: they are signed in.

## Alternative ideas

Feel free to suggest alternative ideas.

We could instead block the page from requests with BingPreview in the user agent, but then we'd have to keep a denylist of bots, which could never be exhaustive; and it might play havoc with the bots' intended behaviour.

We could also vary the content on the 'confirmation invalid' page depending if the jobseeker is confirmed yet or not, but that would have the downsides of interrupting the users and of requiring new content to be written for this very rare scenario.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3446

### DIFF
--- a/app/controllers/jobseekers/confirmations_controller.rb
+++ b/app/controllers/jobseekers/confirmations_controller.rb
@@ -1,6 +1,12 @@
 class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
   after_action :remove_devise_flash!, only: %i[create]
 
+  def show
+    super do |jobseeker|
+      jobseeker.errors.delete(:email, :already_confirmed) if jobseeker.errors.added?(:email, :already_confirmed)
+    end
+  end
+
   protected
 
   def after_confirmation_path_for(resource_name, resource)
--- a/spec/system/jobseekers_can_sign_up_to_an_account_spec.rb
+++ b/spec/system/jobseekers_can_sign_up_to_an_account_spec.rb
@@ -51,5 +51,15 @@ RSpec.describe "Jobseekers can sign up to an account" do
         end
       end
     end
+
+    context "when the confirmation link is followed after the account has already been confirmed" do
+      before { jobseeker.confirm }
+
+      it "still signs them in" do
+        visit first_link_from_last_mail
+        expect(current_path).to eq(jobseekers_saved_jobs_path)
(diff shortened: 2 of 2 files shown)


cesidio (approved):
❤️  very elegant

elliotcm:
I think this is fine.  I would however suggest that if we're having problems with bots causing side-effects on a GET, we should be using a POST instead.  With this approach we're treating bot confirmations as legit, when someone has just opened (or maybe even received) the email.

IMO we should add a form to the page with a button for them to click, or click it for them with JS if there's a bot-safe way to do it.

david-mears-dfe:
> I would however suggest that if we're having problems with bots causing side-effects on a GET, we should be using a POST instead.

You make a good point.

Having side-effects (model updates) as a consequence of a GET request is in this case part of Devise:

https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/app/controllers/devise/confirmations_controller.rb#L21

So this would be a heavy departure from the Devise defaults (which are dubiously RESTful).

We could add another GET endpoint to the controller, with a button that triggers the `show` action.

The POST request to the controller is already 'taken'; it sends the initial confirmation email (with a link to the GET endpoint), so any button we add could be a PUT request.

elliotcm:
@david-mears-dfe 
> The POST request to the controller is already 'taken'; it sends the initial confirmation email (with a link to the GET endpoint), so any button we add could be a PUT request.

Are we prevented from adding another action to the controller?

david-mears-dfe:
This is being closed in favour of the approach discussed above by elliot.

david-mears-dfe:
Another reason for not doing the above is that then the confirmation email would become a permanent password-equivalent that anyone could use at any time to log into that account... I think?

PR #4319, opened 2021-11-23
[TEVA-3305] Updates to Jobseeker Flows

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3305

## Changes in this PR:

- Ensures that unless a user signs in from an "Apply for this job" or "Save this job for later" link, they are directed to either the "My applications" dashboard (if they have any applications) or the "Saved jobs" dashboard (if they have no applications. 
- Ensures that the `:jobseeker_root_path` session variable is destroyed upon signing out.
- Ensures that the `:jobseeker_root_path` session variable does not persist if a user navigates away from the sign in page after being sent there after clicking an "Apply for this job" or "Save this job for later" link - meaning if they sign in from elsewhere, they will be directed to the correct page on their dashboard.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,10 +12,11 @@ class ApplicationController < ActionController::Base
   before_action :store_jobseeker_redirect_to!, if: -> { redirect_to_param.present? }
   before_action :trigger_click_event, if: -> { click_event_param.present? }
   before_action { EventContext.request_event = request_event }
+  before_action :destroy_jobseeker_return_to!
 
   after_action :trigger_page_visited_event, unless: :request_is_healthcheck?
 
-  helper_method :cookies_preference_set?, :referred_from_jobs_path?, :user_type, :utm_parameters, :current_variant?
+  helper_method :cookies_preference_set?, :current_variant?, :jobseeker_root_path, :referred_from_jobs_path?, :user_type, :utm_parameters
   helper GOVUKDesignSystemFormBuilder::BuilderHelper
 
   include Publishers::AuthenticationConcerns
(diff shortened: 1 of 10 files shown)


app/controllers/jobseekers/base_controller.rb
```diff
@@ -9,6 +9,8 @@ def storable_location?
   end
 
   def store_jobseeker_location!
+    return if request.referrer && URI(request.referrer).path == "/jobseekers/sign-in"
```
david-mears-dfe, line 12: I think it's more often spelt `referer`, due to a historic misspelling when the protocol was agreed...

Bear in mind not all browsers will set the referer header, so this wont' work for everyone.

app/controllers/jobseekers/sessions_controller.rb
```diff
@@ -11,8 +12,19 @@ class Jobseekers::SessionsController < Devise::SessionsController
...
 
+  def destroy
+    session.delete(:jobseeker_return_to)
+    super
+  end
+
   private
 
+  def destroy_jobseeker_return_to
+    return if request.referrer && (URI(request.referrer).path.include? "/job_application/new")
```
david-mears-dfe, line 23: Bear in mind not all browsers will set the referer header, so this wont' work for everyone.

app/controllers/application_controller.rb
```diff
@@ -47,6 +48,14 @@ def user_type
     end
   end
 
+  def jobseeker_root_path
+    if current_jobseeker.job_applications.any?
+      jobseekers_job_applications_path
+    else
+      jobseekers_saved_jobs_path
+    end
+  end
```
elliotcm, line 57: A small issue here is that the entry you've removed from `routes.rb` does a little more than just this.  You'd want to:

1. also define `jobseeker_root_url`
2. pass any args through to the more specific methods in case people are passing stuff in

elliotcm (changes requested):
Looks good generally, one issue we'll need to correct.

app/controllers/application_controller.rb
elliotcm:
```suggestion
  def jobseeker_root_path(*args)
    if current_jobseeker.job_applications.any?
      jobseekers_job_applications_path(*args)
    else
      super(*args)
    end
  end

  def jobseeker_root_url(*args)
    if current_jobseeker.job_applications.any?
      jobseekers_job_applications_url(*args)
    else
      super(*args)
    end
  end
```

elliotcm (approved):

PR #4356, opened 2021-12-01
Fix smoke test

cesidio:


david-mears-dfe (approved):

david-mears-dfe:
This fixes the immediate problem but it would be nice to use capybara when we have time!

PR #4359, opened 2021-12-01
Fix flaky spec

cesidio:


elliotcm (approved):
Highly suspicious, but, sure.

csutter (approved):

PR #4377, opened 2021-12-03
[TEVA-3554] Remove status tags on published jobs

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3554

### Before
Screenshot 2021-12-03 at 14 22 00

### After
Screenshot 2021-12-03 at 14 21 10


csutter (approved):
I think one day I would prefer refactoring the `ReviewComponent` to encapsulate whether or not we show the tags – but for now this is okay.

PR #4378, opened 2021-12-03
TEVA-3509 Add the new "copy existing or create new" vacancy creation flow

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3509

## Screenshots of UI changes:

See review app and comments below.


elliotcm:
Screenshot 2021-12-07 at 15 11 19
Screenshot 2021-12-07 at 15 11 24
Screenshot 2021-12-07 at 15 11 29
Screenshot 2021-12-07 at 15 11 34

spec/factories/support/random_helpers.rb
```diff
@@ -0,0 +1,23 @@
...
+    if Rails.env.test?
+      range.min
+    else
+      rand(range)
+    end
+  end
+
+  def factory_sample(things, num = 1)
+    if Rails.env.test?
+      num == 1 ? things.first : things[0...num]
```
csutter, line 12:
```ruby
things.take(num)
```

app/components/vacancy_selector_component/vacancy_selector.js
```diff
@@ -0,0 +1,125 @@
...
+      },
+    });
+  }
+
+  static inputValueTemplate(choice) {
+    if (choice) {
+      return choice['job-title'].innerText;
+    }
+
+    return undefined;
```
csutter, line 95: This should return `undefined` by default anyway, or does ESlint want it to be explicit?

app/models/organisation.rb
```diff
@@ -46,6 +46,10 @@ def local_authority?
     local_authority_code.present?
   end
 
+  def trust_or_la?
+    trust? || local_authority?
```
csutter, line 50: This is identical to `school_group?`

csutter (approved):
Couple of comments but looks good 👍

spec/factories/support/random_helpers.rb
elliotcm: Will still need to check for `1` since `sample` returns an item for 1 and an array for anything else, but I can use it for the second part.

app/components/vacancy_selector_component/vacancy_selector.js
elliotcm: Yeah the linter complains if there's no explicit return.

app/models/organisation.rb
elliotcm: I have no idea what the intent behind all these is, but in terms of the code/DB they're different.

`trust?` and `local_authority?` are based on the presence of strings whereas `school_group?` checks for a subclass (which is unrelated to those two values).

elliotcm:
> @elliotcm looks good apart from one issue where I see "date: not set" I created a job listing and, from the confirmation screen, I selected "Create another job listing" . When I went to Select a job to copy the closing date of my job was not shown, despite it being defined. image This is also shown if I go in through the Create a job listing action. image
> 
> It is however shown when I go into the listing itself.

@bencmitchell Thanks, I'll take a look.

bencmitchell:
Looks sorted 👍 
image

PR #4379, opened 2021-12-03
[TEVA-3577] validate makekfile execution arguments

felixtheflex:
## Jira ticket URL

## Changes in this PR:

As is, condition element (bool) of the `if statments` in the makefile arguments execuate as `true` if the content of the variable is not empty; i.e. we only consider if the variable is not empty and not the value of the variable. E.g. $(CONFIRM_PRODUCTION) i.e. CONFIRM_PRODUCTION=no. 

If someone types `make production terraform-deploy CONFIRM_PRODUCTION=no`, terraform deploy will run successfully despite CONFIRM_PRODUCTION being set No. This could be dangerous, especially for someone new to the project.

This PR is to ensure with other variables such as CONFIRM_DESTROY, only gets executed when they are set to YES as an argument.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,8 @@ staging: ## staging
 		$(eval backend_config=-backend-config="key=$(env)/app.tfstate")
 
 .PHONY: production
-production: ## production # Requires `CONFIRM_PRODUCTION` to be present (any value)
-		$(if $(CONFIRM_PRODUCTION), , $(error Can only run with CONFIRM_PRODUCTION))
+production: ## production # Requires `CONFIRM_PRODUCTION=YES` to be present
+		@if [[ "$(CONFIRM_PRODUCTION)" != "YES" ]]; then echo "Please enter "CONFIRM_PRODUCTION=YES" to run workflow"; exit 1; fi
 		$(eval env=production)
 		$(eval space=teaching-vacancies-production)
 		$(eval var_file=$(env))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -56,7 +56,7 @@ staging: ## staging
 
 .PHONY: production
 production: ## production # Requires `CONFIRM_PRODUCTION` to be present (any value)
```
david-mears-dfe, line 58: This comment should be updated, correct?

david-mears-dfe (commented):
I like it a lot! As I don't know the language involved here I'll let someone else give it a stamp of approval.

Makefile
```diff
@@ -55,8 +55,9 @@ staging: ## staging
 		$(eval backend_config=-backend-config="key=$(env)/app.tfstate")
 
 .PHONY: production
-production: ## production # Requires `CONFIRM_PRODUCTION` to be present (any value)
+production: ## production # Requires `CONFIRM_PRODUCTION` to be present
 		$(if $(CONFIRM_PRODUCTION), , $(error Can only run with CONFIRM_PRODUCTION))
+		$(shell if ! [[ $(CONFIRM_PRODUCTION) = "YES" ]]; then echo "Please enter "CONFIRM_PRODUCTION=YES" to run workflow"; fi)
```
saliceti, line 60: I'm getting an error:
```
% make production CONFIRM_PRODUCTION=whatever                                
Please enter CONFIRM_PRODUCTION=YES to run workflow
make: Please: No such file or directory
make: *** [production] Error 1
```

saliceti (approved):

PR #4383, opened 2021-12-06
[TEVA-3417] Convert map.js to use Stimulus and OpenStreetMap

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3417

### Before
Screenshot 2021-12-06 at 11 09 41

### After
Screenshot 2021-12-06 at 17 28 45


app/components/map_component/map.js
```diff
@@ -0,0 +1,49 @@
...
+          this[item.type](item.data);
+        });
+      });
+    });
+  }
+
+  create(point, zoom) {
+    this.map = L.map('map').setView(point, zoom);
+
+    L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(this.map);
```
csutter, line 31: We need to add OSM attribution to the map to comply with their licence:

```js
L.tileLayer(
  'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
  {
    attribution: 'Map data &copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
  }
).addTo(this.map);
```

csutter (changes requested):
Looks good - see comment about the licence attribution.

I reckon we can go live with OSM, and then maybe switch to Mapbox if we think that's useful.

david-mears-dfe:
This version of Essex (which has multiple polygons because it has islands) worries me. Is this a flaw in the debugging tool or in the location search itself? If the debugging tool is making me think there is a bug when there isn't, we could consider getting rid of the debugging tool rather than trying to invest time in debugging it. Or I can put it on the Christmas firebreak list...

Screenshot 2021-12-06 at 21 33 36

csutter (approved):

alexbowen:
tested and working on IE11

alexbowen (approved):

cesidio:
> This version of Essex (which has multiple polygons because it has islands) worries me. Is this a flaw in the debugging tool or in the location search itself? If the debugging tool is making me think there is a bug when there isn't, we could consider getting rid of the debugging tool rather than trying to invest time in debugging it. Or I can put it on the Christmas firebreak list...
> 
> Screenshot 2021-12-06 at 21 33 36

This has now been fixed, can you please test it again as soon as the new review app will build? Thanks for reporting it David! :)

PR #4406, opened 2021-12-09
Stop using Algolia for all searches

csutter:
- Use `PgSearch` strategy exclusively in `VacancySearch`
- Remove search AB test
- Simplify `VacancySearch` specs
- Fix `Subscription` spec to not rely on Algolia

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -37,7 +37,7 @@ class Subscription < ApplicationRecord
 
   def vacancies_for_range(date_from, date_to)
     criteria = search_criteria.symbolize_keys.merge(from_date: date_from, to_date: date_to)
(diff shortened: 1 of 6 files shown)


app/models/subscription.rb
```diff
@@ -37,7 +37,7 @@ def unsubscribe
 
   def vacancies_for_range(date_from, date_to)
     criteria = search_criteria.symbolize_keys.merge(from_date: date_from, to_date: date_to)
-    Search::VacancySearch.new(criteria, per_page: MAXIMUM_RESULTS_PER_RUN, fuzzy: false).vacancies
+    Search::VacancySearch.new(criteria, per_page: MAXIMUM_RESULTS_PER_RUN).vacancies
```
elliotcm, line 40: I think we can remove this param from the initializer too.

elliotcm (changes requested):

elliotcm (approved):

PR #4410, opened 2021-12-09
[TEVA-3447] Implement job alert banner location AB test

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3447

## Changes in this PR:

This PR originally implemented an AB test which changes the location "Job alert" banner on the job alerts email and changes the copy. We have decided that this test would not be able to obtain statistically significant data in a reasonable time, so this PR now implements the proposed content changes directly.

As upcoming tickets will implement AB tests in mailers, the code which allows for AB tests to be added to mailers has been retained.

### Before

image

### After

image

### DIFF
--- a/app/mailers/jobseekers/base_mailer.rb
+++ b/app/mailers/jobseekers/base_mailer.rb
@@ -2,7 +2,11 @@ class Jobseekers::BaseMailer < ApplicationMailer
   private
 
   def email_event
-    @email_event ||= EmailEvent.new(template, to, uid, jobseeker: @jobseeker)
+    @email_event ||= EmailEvent.new(template, to, uid, jobseeker: @jobseeker, ab_tests: ab_tests)
+  end
+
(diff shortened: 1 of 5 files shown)


app/helpers/notify_views_helper.rb
```diff
@@ -22,6 +22,11 @@ def edit_subscription_link(subscription)
     notify_link(url, t(".edit_link_text"))
   end
 
+  def edit_subscription_link_job_alert_banner_location_ab_test(subscription)
```
david-mears-dfe, line 25: From the name of this method I would expect it to return an ab test, not a link.

robyoung91: how about `job_alert_banner_location_ab_test_edit_subscription_link` ?

cesidio (approved):

PR #4411, opened 2021-12-09
Remove Algolia indexing and configuration

csutter:
This removes indexing of records to Algolia, as well as configuration
and the Algolia gem.


cesidio (approved):
🔥

PR #4414, opened 2021-12-10
[TEVA-3572] Disaster-recovery-lost-db-instance

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3572

## Changes in this PR:

Simulate complete loss of database instance:

create new `makefile` recipe to recover recreate and restore backup from s3
Document disaster recovery process

### DIFF
--- a/documentation/disaster-recovery.md
+++ b/documentation/disaster-recovery.md
@@ -1,20 +1,12 @@
 # Disaster recovery
 
-## Data loss
 
-In the case of a data loss, we need to recover the data as soon as possible in order to resume normal service. Declare a major incident
-and document the incident timelines as you go along.
+This documentation covers two scenarios
 
-The application's database is a postgres instance, which resides on PaaS. This provides a point-in-time backup with
-the resolution of 1 second, available between 5min and 7days ago - this is not applicable to postgres `tiny` plans. We can use
-terraform to create a new database using a point-in-time backup of the affacted database instance.
+- [Data loss](#data-loss)
+- [Loss of database instance](#loss-of-database-instance)
 
-### Make note of database failure time
-
-Make note of the time the database failure occurred, and then use this to calculate when the integrity of the data in the database was still viable. For instance,
-if data loss or corruption happened at 1200hrs, use this to work out what snapshot time is best for the product (consult with the PM if you are unsure what would be best from the product perspective). This would determine the value of `SNAPSHOT_TIME` env.
-
-___Important___: You should convert the time to UTC before actually using it. When you record the time, note what timezone you are using. Especially during BST (British Summer Time).
+In case of any above database disaster, please do the following:
 
 ### Freeze pipeline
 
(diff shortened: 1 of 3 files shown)


documentation/disaster-recovery.md
```diff
@@ -1,20 +1,12 @@
 # Disaster recovery
 
-## Data loss
 
-In the case of a data loss, we need to recover the data as soon as possible in order to resume normal service. Declare a major incident
-and document the incident timelines as you go along.
+This documentation covers two disaster scenario
```
saliceti, line 4: scenarios

documentation/disaster-recovery.md
```diff
@@ -86,3 +93,29 @@ cf service teaching-vacancies-postgres-production
...
+
+- Recreate the lost postgress database instance
+- Restore data from daily backup
+
+### Recreate the lost postgress database instance
+
+In order to recreate the lost postgress database instance, use the following make recipes `terraform-app-plan` and `terraform-app-apply`:
+
+```
+make <env> terraform-app-plan passcode=MyPasscode tag=master-08406f04dd9eadb7df6fcda5213be880d7df37ed-20201022090714 ## to see the proposed changes.
```
saliceti, line 109: We could leave the tag out so it uses the default
BTW this must be changed here: https://github.com/DFE-Digital/teaching-vacancies/blob/main/Makefile#L110

documentation/disaster-recovery.md
```diff
@@ -86,3 +93,29 @@ cf service teaching-vacancies-postgres-production
...
+## Loss of database instance
+
+In case the database instance is lost, the objectives are:
+
+- Recreate the lost postgress database instance
+- Restore data from daily backup
+
+### Recreate the lost postgress database instance
+
+In order to recreate the lost postgress database instance, use the following make recipes `terraform-app-plan` and `terraform-app-apply`:
```
saliceti, line 106: We may want to make sure the service is really gone and not in a broken state. We may have to run `cf delete-service` or even remove from terraform state

documentation/disaster-recovery.md
```diff
@@ -86,3 +93,29 @@ cf service teaching-vacancies-postgres-production
...
+```
+
+```
+make <env> terraform-app-apply passcode=MyPasscode tag=master-08406f04dd9eadb7df6fcda5213be880d7df37ed-20201022090714 ## apply proposed changes i.e. create new database instance.
+```
+This will create a new postgres database instance as described in the terraform configuration file.
+
+### Restore Data From Daily Backup
+
+Once the lost database instance has been recreated, the last daily backup will need to be restored. To achieve this, use the following makefile recipe `restore-daily-backup`.
```
saliceti, line 119: Can we add an example as above

documentation/disaster-recovery.md
felixtheflex: wouldn't the terraform plan confirm that anyway?

documentation/disaster-recovery.md
felixtheflex: since it is production, don't we want to use exactly same tag?

saliceti: OK then please show how to retrieve the tag

saliceti (approved):

documentation/disaster-recovery.md
saliceti: This may cause unexpected issues. Can we add something like:
> In case the database service is deleted or in an inconsistent state we must recreate it and repopulate it. First make sure it is full gone by running `cf services` and `cf delete-service` if necessary.
> Then recreate the lost postgress database instance using the following make recipes...

PR #4421, opened 2021-12-11
[TEVA-3484] Introduce SortComponent

david-mears-dfe:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3484

## Changes in this PR:

- The ticket wants similar logic in many pages (and there are a few pages it didn't mention that I think want it too, such as in the jobseeker dashboard (saved jobs and job alerts pages)), namely: if there are only two sorting options, use links instead of a drop-down.

- This seemed like a good time to wrap that logic (and styling!) in a component.

- This component chooses whether to use a drop down or a pair of links based on the number
of sorting options, hence its name 'accessible' sort component.

- In order for it to be polymorphic, the interface of 'sort' objects has been regularized
so that all sort options will have a 'sort_by' and a 'display_name' attribute (rather than
'column', which would exclude virtual attributes and other kinds of sorting such as by
relevance to a keyword).

- Also fixes an existing related bug where publishers could not sort
job applications by last_name, since that is not a database column, but only a virtual
attribute, so it cannot be used with 'order' (which relies on SQL).


app/services/publishers/job_application_sort.rb
```diff
@@ -1,13 +1,15 @@
...
 
-  def options
-    [
-      SortOption.new("submitted_at", "desc", I18n.t("publishers.vacancies.job_applications.index.sort_by.date_received")),
-      SortOption.new("last_name", "asc", I18n.t("publishers.vacancies.job_applications.index.sort_by.applicant_last_name")),
-    ]
+  private
+
+  def last_name_option
+    @last_name_option ||= SortOption.new("last_name", I18n.t("publishers.vacancies.job_applications.index.sort_by.applicant_last_name"))
```
csutter, line 9: I reckon this isn't worth memoizing - it's not all that expensive to compute (ditto below).

app/services/search/vacancy_sort.rb
```diff
@@ -0,0 +1,31 @@
...
+  private
+
+  attr_reader :keyword
+
+  def default_sort_option
+    options.include?(relevance_option) ? relevance_option : publish_on_desc_option
+  end
+
+  def relevance_option
+    @relevance_option ||= SortOption.new("relevance", I18n.t("jobs.sort_by.most_relevant"))
```
csutter, line 25: Ditto

spec/components/accessible_sort_component_spec.rb
```diff
@@ -0,0 +1,80 @@
+require "rails_helper"
+
+RSpec.describe AccessibleSortComponent, type: :component do
+  let(:path) { Rails.application.routes.url_helpers.method(:page_path) }
+  let(:url_params) { { id: "terms-and-conditions" } }
+  let(:sort) do
+    sort_class = Class.new(RecordSort)
```
csutter, line 7: I might be missing something, but could this be:

```ruby
Class.new(RecordSort) do
  def self.options
    Array.new(number_of_sorting_options) do |i|
      SortOption.new("sort_by_#{i}", "sorting algorithm number #{i + 1}")
    end
  end
end.new
```

csutter (commented):
Like it – got a couple of comments but looks great for the most part.

I don't like the name though, we wouldn't ever have an `InaccessibleSortingComponent`, right? 😕 How about just `SortComponent`?

spec/components/accessible_sort_component_spec.rb
david-mears-dfe: `number_of_sorting_options` is not in the class's scope.

david-mears-dfe:
Demo video:

https://drive.google.com/file/d/1TXNl4Uobzuz9qQJTMiWyt5WPW9YF6nt5/view?usp=sharing

I show the sorting component on the search page, and then on the hiring staff and jobseeker dashboards.

I forgot to show that, when you get results from a search with no keyword, you don’t get any sorting options, because ‘relevance’ doesn’t make sense in that scenario, and so there’s only one sorting option left.

When I show the jobseeker dashboard you can see I’ve made it not give any sorting options until there are at least two things to sort (saved jobs in this case)

app/services/search/vacancy_sort.rb
david-mears-dfe: I believe in this one relevance has to be memoized in order for `default_sort_option` to work as planned.

spec/components/accessible_sort_component_spec.rb
csutter: Hmm, I would have thought it is given that it's dynamically defining the class inside a block. I'll try and give it a whirl if only to update my own mental model 😄

csutter: Right – it's the `def` that is the issue, not the `Class.new`, because the `def` doesn't carry a closure.

Interestingly, the RSpec `let`s are still not accessible when using `define_method` either, but something like this works:

```ruby
  let(:sort) do
    num = number_of_sorting_options

    Class.new(RecordSort) do
      define_method(:options) do
        Array.new(num) do |i|
          SortOption.new("sort_by_#{i}", "sorting algorithm number #{i + 1}")
        end
      end
    end.new
  end
```

csutter: Might take another look after lunch!

csutter: How about this:

```ruby
let(:sort_klass) do
  Class.new(RecordSort) do
    attr_reader :options

    def initialize(options)
      @options = options
    end
  end
end
let(:sort) do
  sort_klass.new(
    (1..number_of_sorting_options).map do |i|
      SortOption.new("sort_by_#{i}", "sorting algorithm number #{i}")
    end,
  )
end
```

A bit more explicit and less meta 😃

app/presenters/subscription_presenter.rb
```diff
@@ -33,7 +33,7 @@ def available_filter_hash
 
   def search_criteria_field(field, value)
     return if field.eql?("radius")
-    return if field.eql?("jobs_sort")
+    return if field.eql?("sort_by")
```
csutter, line 36: I know you didn't change the method, but I wonder why this is `eql?` 🧐

app/components/sort_component.rb
```diff
@@ -0,0 +1,42 @@
+# When there are only two sorting options, these should be presented as links, not
+# options in a drop-down. This component manages that display logic.
+class SortComponent < GovukComponent::Base
+  attr_reader :url_params, :path, :sort
+
+  # @param [Method] path The helper method which generates the destination path for the links or forms
```
csutter, line 6: Another nitpick: This argument takes a `String` (we just happen to pass in a method that returns a string)

csutter (approved):
Couple of tiny nitpicks left, but 👍

app/components/sort_component.rb
david-mears-dfe: Actually, I'm passing a method itself (e.g. `method(:page_path)`) because I need to be able to pass the `url_params` to that method to do with as it sees fit (merged with the sort_by param).

PR #4426, opened 2021-12-13
[TEVA-3516] Split applying for the job step

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3516

## Screenshots of UI changes:
Screenshot from 2021-12-13 18-36-38

Screenshot from 2021-12-13 18-37-00

Screenshot from 2021-12-13 18-37-19

### DIFF
--- a/app/controllers/concerns/publishers/wizardable.rb
+++ b/app/controllers/concerns/publishers/wizardable.rb
@@ -96,8 +96,16 @@ module Publishers::Wizardable
   end
 
   def applying_for_the_job_params(params)
-    params.require(:publishers_job_listing_applying_for_the_job_form)
-          .permit(:application_link, :enable_job_applications, :contact_email, :contact_number, :personal_statement_guidance, :school_visits, :how_to_apply)
+    if params[:publishers_job_listing_applying_for_the_job_form]
+      params.require(:publishers_job_listing_applying_for_the_job_form).permit(:enable_job_applications)
+    else
+      {}
+    end.merge(completed_steps: completed_steps, current_organisation: current_organisation)
+  end
+
+  def applying_for_the_job_details_params(params)
+    params.require(:publishers_job_listing_applying_for_the_job_details_form)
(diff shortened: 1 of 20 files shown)


app/controllers/concerns/publishers/wizardable.rb
```diff
@@ -96,8 +96,16 @@ def important_dates_params(params)
   end
 
   def applying_for_the_job_params(params)
-    params.require(:publishers_job_listing_applying_for_the_job_form)
-          .permit(:application_link, :enable_job_applications, :contact_email, :contact_number, :personal_statement_guidance, :school_visits, :how_to_apply)
+    if params[:publishers_job_listing_applying_for_the_job_form]
```
david-mears-dfe, line 99: When is `params[:publishers_job_listing_applying_for_the_job_form]` not present?

app/services/publishers/vacancies/vacancy_step_process.rb
```diff
@@ -54,4 +54,12 @@ def job_details_steps
       %i[job_details]
     end
   end
+
+  def applying_for_the_job_steps
```
david-mears-dfe, line 58: Is it possible to reuse this logic in `override_enable_job_applications` so that we don't have to remember to update it in both places?

david-mears-dfe (approved):

app/controllers/concerns/publishers/wizardable.rb
cesidio: it's not present when you submit the form without selecting any radio button on it. Because it's the only field now on that page, and we know that the govuk_form_builder does not send any forms if all the values are empty, we need that.

app/services/publishers/vacancies/vacancy_step_process.rb
cesidio: the form model does not know about the vacancy_step_process service

PR #4427, opened 2021-12-14
Add Kernel#fail_safe for graceful degradation

csutter:
- Add `FailSafe` lib file to add `Kernel#fail_safe`
- Refactor existing uses of rescuing-and-sending-to-Rollbar to use new
  method
- Refactor `FailSafe` view component module to use `Kernel#fail_safe`
- Fix several specs that were only passing because unrelated code was
  silently swallowing errors
- Stub out BigQuery in test environment


cesidio (approved):
nice

PR #4428, opened 2021-12-14
Fix bug affecting seeds

josephhull676:
## Changes in this PR:

```
Created database 'tvs_test'
rails aborted!
TypeError: no implicit conversion of String into Array
/Users/josephhull/Dev/teaching-vacancies/app/models/vacancy.rb:165:in `+'
/Users/josephhull/Dev/teaching-vacancies/app/models/vacancy.rb:165:in `additional_job_roles='
```

This bug was introduced after changes to `factory_sample` in `spec/factories/support/random_helpers.rb`.

In vacancies factory we have the following:

```
when "teacher"
        factory_rand_sample(Vacancy.additional_job_role_options, 0..2)
```

If `factory_rand` in `spec/factories/support/random_helpers.rb` returns 1, a string is returned by `factory_rand_sample`. In the `additional_job_roles=` we then attempt to add this string to an array, causing the error.

### DIFF
--- a/spec/factories/support/random_helpers.rb
+++ b/spec/factories/support/random_helpers.rb
@@ -1,4 +1,6 @@
 module RandomHelpers
+  module_function
+
   def factory_rand(range)
     if Rails.env.test?
       range.min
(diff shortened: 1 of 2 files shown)


elliotcm:
Good catch.  IMO this should be fixed by changing the behaviour of the helper method again.  The odd behaviour of the native `sample` method is what's throwing this.

cesidio (approved):

PR #4473, opened 2022-01-06
[TEVA-3434] Re-order create a job listing steps

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3434

## Changes in this PR:

This PR moves the "Supporting documents" step of the Create a job listing journey to after the "Applying for the job details" steps.

### Before

image

### After

image

### DIFF
--- a/spec/system/publishers_can_publish_a_vacancy_as_a_school_spec.rb
+++ b/spec/system/publishers_can_publish_a_vacancy_as_a_school_spec.rb
@@ -83,9 +83,6 @@ RSpec.describe "Creating a vacancy" do
 
       fill_in_important_dates_fields(vacancy)
       click_on I18n.t("buttons.continue")
-      expect(current_path).to eq(organisation_job_documents_path(created_vacancy.id))
-
-      click_on I18n.t("buttons.continue")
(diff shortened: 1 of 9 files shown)


spec/system/publishers_can_publish_a_vacancy_as_a_school_spec.rb
```diff
@@ -141,14 +141,14 @@
...
-      click_on I18n.t("buttons.continue")
-
       fill_in_applying_for_the_job_form_fields(vacancy)
       click_on I18n.t("buttons.continue")
 
       fill_in_applying_for_the_job_details_form_fields(vacancy)
       click_on I18n.t("buttons.continue")
 
+      click_on I18n.t("buttons.continue")
+
```
josephhull676, line 151: Can't figure out why we click continue twice here?

josephhull676 (approved):

spec/system/publishers_can_publish_a_vacancy_as_a_school_spec.rb
robyoung91: We fill in the job details form fields, then press continue on that page and at that point we are at the supporting documents step - we then click continue again (without uploading anything) to go to the next step

PR #4480, opened 2022-01-07
TEVA-3446 Add an interstitial page with a button for confirming email address

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3446

## Screenshots of UI changes:

Screenshot 2022-01-07 at 15 38 55

### DIFF
--- a/spec/system/jobseekers_can_register_spec.rb
+++ b/spec/system/jobseekers_can_register_spec.rb
@@ -15,6 +15,7 @@ RSpec.describe "Jobseekers can register" do
 
     click_on I18n.t("jobseekers.registrations.check_your_email.resend_link"), visible: false
     visit first_link_from_last_mail
+    click_on "Confirm"
 
     expect(current_path).to eq jobseekers_saved_jobs_path
   end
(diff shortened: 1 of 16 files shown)


spec/system/jobseekers_can_register_spec.rb
```diff
@@ -15,6 +15,7 @@
 
     click_on I18n.t("jobseekers.registrations.check_your_email.resend_link"), visible: false
     visit first_link_from_last_mail
+    click_on "Confirm"
```
josephhull676, line 18: Would you `confirm_email_address` here too?

josephhull676 (approved):

spec/system/jobseekers_can_register_spec.rb
elliotcm: While they do the same thing, I wanted to be more explicit about the two steps here as this is what's actually under test, whereas in the other cases it's a step on the way to testing something else.

spec/system/jobseekers_can_sign_up_to_an_account_spec.rb
```diff
@@ -37,41 +37,13 @@
       context "when the confirmation period has expired" do
         before { travel_to 3.hours.from_now }
 
-        it "does not confirm email and redirects to resend confirmation page" do
```
josephhull676, line 40: Why are you deleting these tests?

elliotcm: The logging was put in place by David when it was under investigation.  I removed that code so these events are no longer fired.

PR #4481, opened 2022-01-07
Fix delete documents

cesidio:


csutter:
Two thoughts:
* Could we just pass in `name`, `size`, and `content_type` as parameters to the function instead of `document`? Saves us the `if`
* Should we wrap this in a `fail_safe` block to make sure `send_event` never causes an exception in prod? (as it just sends events)

josephhull676 (approved):

PR #4486, opened 2022-01-10
Remove use of slim files for email view templates

josephhull676:
## Jira ticket URL

Nae ticket


app/views/publishers/authentication_fallback_mailer/sign_in_fallback.text.erb
```diff
@@ -0,0 +1,7 @@
+# <%= t("app.title") %>
+
+<%= t("publishers.temp_login.email.heading") %>
```
csutter, line 3: Missing the `##` second level heading here

josephhull676: So before changing anything, the email looked like this:

Screenshot 2022-01-10 at 12 01 01

I was playing around with this and it seemed both 

```
# <%= t("app.title") %>

## <%= t("publishers.temp_login.email.heading") %>

<%= t("publishers.temp_login.email.body") %>

<%= choose_organisation_link(@login_token) %>

```
and

```
# <%= t("app.title") %>

<%= t("publishers.temp_login.email.heading") %>

<%= t("publishers.temp_login.email.body") %>

<%= choose_organisation_link(@login_token) %>

```

look the same. Both seem to create p elements?

josephhull676: With second level heading 

Screenshot 2022-01-10 at 12 07 30

Without:

Screenshot 2022-01-10 at 12 08 13

csutter: Huh, I reckon it might be because Notify doesn't actually support second-level headings. In which case, carry on!

PR #4488, opened 2022-01-10
Update Ruby to 3.1

csutter:
- (Temporarily) add `net-*` gems to Gemfile (which have been split out
  from stdlib into separate gems in Ruby 3.1, but aren't explicit
  dependencies of Rails yet)


app/views/jobseekers/job_applications/index.html.slim
```diff
@@ -16,25 +16,25 @@
...
-            = render "job_application", job_application: job_application if job_application.vacancy.expires_at.future?
+            = render("job_application", job_application:) if job_application.vacancy.expires_at.future?
 
           - current_jobseeker.job_applications.includes(:vacancy).order(submitted_at: :desc).submitted.each do |job_application|
-            = render "job_application", job_application: job_application
+            = render "job_application", job_application:
 
           - current_jobseeker.job_applications.includes(:vacancy).order(submitted_at: :desc).reviewed.each do |job_application|
-            = render "job_application", job_application: job_application
+            = render "job_application", job_application:
```
elliotcm, line 25:
```suggestion
            = render("job_application", job_application:)
```

app/views/jobseekers/job_applications/index.html.slim
```diff
@@ -16,25 +16,25 @@
...
 
           - current_jobseeker.job_applications.includes(:vacancy).order(submitted_at: :desc).shortlisted.each do |job_application|
-            = render "job_application", job_application: job_application
+            = render "job_application", job_application:
 
           - current_jobseeker.job_applications.includes(:vacancy).order(submitted_at: :desc).unsuccessful.each do |job_application|
-            = render "job_application", job_application: job_application
+            = render "job_application", job_application:
```
elliotcm, line 31:
```suggestion
            = render("job_application", job_application:)
```

app/views/jobseekers/job_applications/index.html.slim
```diff
@@ -16,25 +16,25 @@
...
 
           - current_jobseeker.job_applications.includes(:vacancy).order(submitted_at: :desc).withdrawn.each do |job_application|
-            = render "job_application", job_application: job_application
+            = render "job_application", job_application:
```
elliotcm, line 34:
```suggestion
            = render("job_application", job_application:)
```
These can potentially all be replaced by `= render(job_application)`.

app/views/jobseekers/registrations/edit.html.slim
```diff
@@ -9,8 +9,8 @@
         = f.govuk_error_summary
 
         - if password_update?
-          = render "edit_password", f: f
+          = render "edit_password", f:
         - else
-          = render "edit_email", f: f
+          = render "edit_email", f:
```
elliotcm, line 14:
```suggestion
          = render("edit_email", f:)
```

elliotcm (changes requested):
I'm going to pause this review at 100 files in, should we make it so that we never end a line with `:`?  Typically in views where we'd e.g. end with `f: f` it now just ends `f:` and looks like a Python block.

csutter:
Given the outstanding questions about hash literals in certain contexts (@elliotcm raises some excellent points), and the fact that Brakeman doesn't support 3.1 syntax yet anyway, I'll close this PR and open a new one without the changes and with the Rubocop rule disabled.

Let's revisit this in the future though.

PR #4502, opened 2022-01-12
Update application to Rails 7

csutter:
- Update all dependencies in advance of Rails update
- Update Rails to 7.0.1
- Run `rails app:update` and integrate meaningful config changes
- Generate and run ActiveStorage migrations
- Update Spring to 4.0 (and temporarily disable `spring-watcher-listen`
  due to incompatibility)
- Disable `config.cache_classes` in test environment (should no longer
  be enabled when using Spring)
- Set test values for ActiveRecord encryption in test environment
- Fix deprecated use of `to_s` on `Date`/`Time`/`TimeWithZone`
- Fix use of renamed `ActionMailer::DeliveryJob`


csutter:
Closing as `rspec-rails` is not ready yet and we're hitting a few edge cases with niche Rails 7 deprecations (and might not be for a while).

Will create a new PR sorting out the bulk of the deprecations so we're ready to hit the ground running when everything falls into place.

PR #4506, opened 2022-01-13
[TEVA-3487] Reposition flash messages

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3487

### Before
Screenshot from 2022-01-13 12-50-54

### After
Screenshot from 2022-01-13 12-49-23


app/views/publishers/vacancies/end_listing/show.html.slim
```diff
@@ -3,8 +3,9 @@
 .govuk-main-wrapper
   .govuk-grid-row
     .govuk-grid-column-two-thirds
-      .govuk-caption-l = vacancy.job_title
```
josephhull676, line 6: Did you mean to remove this?

josephhull676 (changes requested):

PR #4518, opened 2022-01-14
Fix DidYouMean deprecation warning

elliotcm:
https://github.com/rubygems/rubygems/issues/5234#issuecomment-1002544751


cesidio (approved):

csutter:
This'll need bumping of 2.3.5 in the `Dockerfile` also, and possibly the devcontainer too.

elliotcm:
> This'll need bumping to 2.3.5 in the `Dockerfile` also, and possibly the devcontainer too.

@csutter I've updated the dockerfile.  The dev container doesn't seem to specify it.

PR #4534, opened 2022-01-17
[TEVA-3656] Update end listing form to align with expired vacancy feedback

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3656

## Questions

- I considered removing the EndListingForm and using the ExpiredFeedbackForm instead, but decided to stick with the EndListingForm. I did this because it seemed to make more sense (to me anyway) to render a form with the same name as the controller.  What are your thoughts?


robyoung91 (commented):
@josephhull676 this is looking good to me - just one thing - if the user submits the form without selecting either radio button, they receive a duplicated error message in the error summary (see below). Is this maybe something we want to change?

image

robyoung91 (approved):

PR #4537, opened 2022-01-18
[TEVA-3596] Implement template pages

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3596

## Screenshots of UI changes:
image


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,24 @@
+class PostsController < ApplicationController
+  def show
+    @content = Kramdown::Document.new(parsed.content).to_html
+    @title = parsed.front_matter["title"]
+  end
+
+  private
+
+  def parsed
+    FrontMatterParser::Parser.new(:md).call(file_content)
```
elliotcm, line 10: Should we cache some or all of these methods?

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def file_content
+    File.read(Rails.root.join("app", "views", "content", section, file_name))
+  end
+
+  def section
+    request.path.split("/").second
```
elliotcm, line 18: We could make this part of the route if you like.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  private
+
+  def parsed
+    FrontMatterParser::Parser.new(:md).call(file_content)
+  end
+
+  def file_content
+    File.read(Rails.root.join("app", "views", "content", section, file_name))
```
elliotcm, line 14: Does this currently 500 if the file is missing?  If so can we 404 it?

app/views/content/get-help-hiring/find-the-best-teachers.md
```diff
@@ -0,0 +1,19 @@
+---
+order: 1
+title: Find the best teachers
+---
+## Include the school perks
+List all the perks you can have working at your school
+
+## Another subtitle
+Bla bla bla
```
elliotcm, line 9: I'm a little wary of this going live, even if it's not linked from anywhere yet.  Should we rename the files to something random so at least we're not associating temp content with possibly real future URLs?

elliotcm (changes requested):
I know this is still a draft PR, but I had a spare moment to review this.

app/views/content/get-help-hiring/find-the-best-teachers.md
josephhull676: Hey mate, we're going to remove this file before the branch is merged. This is only there so we can see the page in the review app and locally.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,15 @@
+class PostsController < ApplicationController
+  before_action :redirect_if_document_not_found, only: :show
+
+  helper_method :document
+
+  private
+
+  def redirect_if_document_not_found
+    redirect_to not_found_path unless document
```
csutter, line 9: This should use `not_found` from `ApplicationController` rather than redirecting:

```ruby
def show
  not_found unless document
end
```

app/services/markdown_document.rb
```diff
@@ -0,0 +1,51 @@
+class MarkdownDocument
+  def initialize(section, filename)
+    @section = section
+    @file_name = filename
+  end
+
+  def parse
+    return unless File.file?(file_path)
```
csutter, line 8: I'm not sure about all the methods on this class that implicitly require `#parse` to have been called or else they'll blow up.

Should the parsing step just happen in `initialize` maybe?

config/routes.rb
```diff
@@ -115,6 +115,10 @@
 
   get "/pages/*id" => "pages#show", as: :page, format: false
 
+  get "/:section/:file_name" => "posts#show",
+      constraints: ->(request) { request.params[:section].in?(%w[jobseeker-guides get-help-hiring]) },
```
csutter, line 119: I think you can just do this:

```ruby
constraints: { section: %w[jobseeker-guides get-help-hiring] }
```

csutter: It should also probably have a constraint on `file_name` (which might be better as a `post_name` as we don't really care that it's a file)?

e.g.:
```ruby
{ post_name: /[\w-]+/ }
```

app/services/markdown_document.rb
csutter: Also `@parsed` seems like it should be called `@front_matter` as that's what it specifically is

app/views/posts/show.html.slim
```diff
@@ -0,0 +1,17 @@
+- content_for :page_title_prefix, document.title
+- content_for :page_description, document.meta_description
+
+= govuk_breadcrumbs breadcrumbs: { t("breadcrumbs.home") => root_path,
+                                   params[:section].titleize.capitalize => "#",
+                                   document.title => "" }
+
+.govuk-main-wrapper.kramdown
```
csutter, line 8: I'd probably call this something a bit more generic than `kramdown`?

csutter (commented):
I really love where you've gotten to with this 🎉 

Couple of minorish comments but I think this is almost ready!

csutter (approved):

PR #4541, opened 2022-01-19
Use and configure Sentry for stack traces

felixtheflex:
## Jira ticket URL

## Changes in this PR:

## Next steps:


csutter:
Commenting here so we don't forget – we need to:

* switch JS error reporting to Sentry too, and
* port over any custom Rollbar configuration we have

PR #4542, opened 2022-01-20
Implement frontend event tracking

csutter:
We want to be able to track some user interactions on the frontend in a
more coherent way than we currently do (a `click_event` param on some
requests, and redirecting some external links through the backend). This
adds a lightweight frontend event tracking framework to achieve this.

- Add an "events" API endpoint to receive events from the frontend
- Add a `FrontendEvent` subtype of `RequestEvent` that overrides it to
  have the correct event data for frontend events
- Add `triggerEvent` function to allow sending an event from Javascript
- Add `trackedLink` Stimulus component to encapsulate links that should
  be tracked
- Add `LinksHelper#tracked_link_to`/`#tracked_open_in_new_tab_link_to`
  to mark up links with the Stimulus controller and event data
- Start tracking link clicks for links to organisation websites on
  job listings

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-2627

### DIFF
--- a/app/frontend/src/components/trackedLink/trackedLink.js
+++ b/app/frontend/src/components/trackedLink/trackedLink.js
@@ -1,0 +1,49 @@
+import { Controller } from '@hotwired/stimulus';
+import { triggerEvent } from '../../lib/events';
+
+const EVENT_TYPE = 'tracked_link_clicked';
+
+// Stimulus controller to send link click data to the app's events endpoint
+export default class extends Controller {
+  static targets = ['link'];
+
+  // Tracks a user click on the link connected to this controller
+  track(e) {
+    // Whether or not we should briefly inhibit the default behaviour of the click (i.e. the browser
+    // following the link) so we have a chance to send event data.
+    // True if and only if the user has used the primary mouse button (or tapped the link on a
+    // touchscreen device) AND the link is not intended to open in a new tab via a target attribute.
+    // Otherwise on e.g. a middle click to open in new tab, or a right click (which for
+    // simplicity's sake we also count as a click as the most likely reason is the user is opening
+    // in a new tab or copying the URL) we are happy for the default behaviour to go ahead normally
+    // (because the page isn't going away and there is ample opportunity for the event to complete).
+    const shouldWaitAndRedirect = e.button === 0 && !this.linkTarget.target === '_blank';
+
+    if (shouldWaitAndRedirect) {
+      // Don't let the browser follow the link immediately (before our event request has had a
+      // chance to be initiated)
+      e.preventDefault();
+
+      // Set a (hopefully) barely noticeable timeout for the redirect to happen anyway regardless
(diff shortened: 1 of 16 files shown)


app/frontend/src/components/trackedLink/trackedLink.js
```diff
@@ -0,0 +1,49 @@
...
+        href: this.linkTarget.href,
+        mouse_button: e.button,
+      },
+    ).then(() => {
+      if (shouldWaitAndRedirect) this.#redirect();
+    });
+  }
+
+  // Redirect to the target of the link (used if default behaviour has been inhibited)
+  #redirect() {
```
alexbowen, line 46: why the # prepending the method name?

app/frontend/src/lib/events.js
```diff
@@ -0,0 +1,21 @@
...
+import { railsCsrfToken } from './utils';
+
+const EVENTS_ENDPOINT = '/api/events';
+
+// Triggers an event and makes a request to the backend endpoint with the event data.
+export const triggerEvent = (type, data) => {
+  const headers = { 'X-CSRF-Token': railsCsrfToken() };
+
+  return new Promise((resolve) => {
+    axios.post(EVENTS_ENDPOINT, { type, data }, { headers })
```
alexbowen, line 12: i think this should return the axios promise to the outer one?

app/frontend/src/lib/utils.js
```diff
@@ -61,3 +61,5 @@ export const storageAvailable = (type, logMessage = false) => {
           && (storage && storage.length !== 0);
   }
 };
+
+export const railsCsrfToken = () => document.getElementsByName('csrf-token')[0].content;
```
alexbowen, line 65: does this need to export a function? `export const [railsCsrfToken] = blah;` also if there is no el with name csrf-token, it will fail trying to find `.content` of undefined

app/frontend/src/lib/events.js
csutter: The idea is that this promise always resolves, regardless of whether or not the Axis promise does. If Axios fails we still want to resolve not reject (but log the error to Rollbar/Sentry/you name it).

app/frontend/src/lib/utils.js
csutter: I think it needs to be a function to actually have access to the document - or am I missing something?

Have rewritten it to fail less badly when the token isn't there (although it _should_ always be 🤞 )

app/frontend/src/components/trackedLink/trackedLink.js
alexbowen: i would remove in case of any wierdness, it is a visual indicator right?

csutter: Yup fair enough, have removed 👍

alexbowen (approved):

cesidio (approved):

PR #4549, opened 2022-01-24
[TEVA-3597] Implement landing pages

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3597

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -119,6 +119,10 @@ Rails.application.routes.draw do
       constraints: { section: /(jobseeker-guides|get-help-hiring)/, post_name: /[\w-]+/ },
       as: :post
 
+  get "/:section" => "posts#index",
+      constraints: { section: /(jobseeker-guides|get-help-hiring)/ },
+      as: :posts
+
   get "/list-school-job" => "pages#show", defaults: { id: "list-school-job" }
 
   get "/cookies-preferences", to: "cookies_preferences#new", as: "cookies_preferences"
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -119,6 +119,10 @@
       constraints: { section: /(jobseeker-guides|get-help-hiring)/, post_name: /[\w-]+/ },
       as: :post
 
+  get "/:section" => "posts#index",
+      constraints: { section: /(jobseeker-guides|get-help-hiring)/ },
```
elliotcm, line 123: This will also match e.g. `/spooky-jobseeker-guides-for-ghosts`.

elliotcm (changes requested):
Just a lil regex issue to fix.

config/routes.rb
robyoung91: Ah right, well spotted - do we need `^jobseeker-guides$|^get-help-hiring$` ?

josephhull676 (approved):

config/routes.rb
elliotcm: @robyoung91 Turns out I was wrong, they already anchor them (and in fact the anchors will break it):
https://guides.rubyonrails.org/routing.html#segment-constraints

You can change it to `get "/:section" => "posts#index", section: /(jobseeker-guides|get-help-hiring)/` to be more compact if you want but no need if you don't want to.

elliotcm (approved):

PR #4566, opened 2022-01-28
TEVA-3595 Working patterns update

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3595

### Before

Screenshot 2022-01-28 at 13 00 25

### After

Screenshot 2022-01-28 at 13 23 16

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -20,7 +20,7 @@ class Vacancy < ApplicationRecord
   # When removing a job_role or working_pattern, remember to update *subscriptions* that have the old values.
   array_enum job_roles: MAIN_JOB_ROLES.merge(ADDITIONAL_JOB_ROLES)
   array_enum key_stages: { early_years: 0, ks1: 1, ks2: 2 }
-  array_enum working_patterns: { full_time: 0, part_time: 100, job_share: 101, term_time: 102 }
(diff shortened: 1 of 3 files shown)


app/models/vacancy.rb
```diff
@@ -20,7 +20,7 @@ class Vacancy < ApplicationRecord
   # When removing a job_role or working_pattern, remember to update *subscriptions* that have the old values.
   array_enum job_roles: MAIN_JOB_ROLES.merge(ADDITIONAL_JOB_ROLES)
   array_enum key_stages: { early_years: 0, ks1: 1, ks2: 2 }
-  array_enum working_patterns: { full_time: 0, part_time: 100, job_share: 101, term_time: 102 }
+  array_enum working_patterns: { full_time: 0, part_time: 100, flexible: 104, job_share: 101, term_time: 102 }
```
alexbowen, line 23: if i were super pedantic these should be in some order, numeric probably, but ill approve pr anyway :)

alexbowen (approved):

app/models/vacancy.rb
alexbowen: scrub that i guess this is the order in the UI

elliotcm: Yeah they requested the specific order, and inserting a new one while keeping numeric order would mean a data migration.

cesidio (approved):

PR #4589, opened 2022-02-01
Remove `initially_indexed` field from Vacancy

csutter:
This was used way back when as part of the Algolia migration and is no
longer needed.

### DIFF
--- a/db/migrate/20220201101037_remove_initially_indexed_from_vacancies.rb
+++ b/db/migrate/20220201101037_remove_initially_indexed_from_vacancies.rb
@@ -1,0 +1,6 @@
+class RemoveInitiallyIndexedFromVacancies < ActiveRecord::Migration[6.1]
+  def change
+    remove_index :vacancies, :initially_indexed
+    remove_column :vacancies, :initially_indexed, :boolean
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20220201101037_remove_initially_indexed_from_vacancies.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveInitiallyIndexedFromVacancies < ActiveRecord::Migration[6.1]
+  def change
+    remove_index :vacancies, :initially_indexed
```
elliotcm, line 3: This happens automatically.

> Indexes and table constraints involving the column will be automatically dropped as well

https://www.postgresql.org/docs/12/sql-altertable.html

elliotcm (approved):
The `remove_index` is redundant, but shouldn't cause any trouble.

db/migrate/20220201101037_remove_initially_indexed_from_vacancies.rb
csutter: Ah cool – I think it's still worth having in there though in order for the migration to be cleanly reversible (not really a concern in this case as I can't think of a reason we would want to)!

elliotcm: Good call, though in that case you'd also need to provide the type, null, defaults etc so it can be recreated.  You can pass `index: true` along with that, but at that point it's a style distinction rather than anything actually important :)

PR #4610, opened 2022-02-07
TEVA-3521 Rework login code (i.e. devise)

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3521


cesidio (approved):

elliotcm:
Product signoff by @bencmitchell 
https://ukgovernmentdfe.slack.com/archives/C0323RT5QRL/p1644934809743489?thread_ts=1644928680.318519&cid=C0323RT5QRL

app/controllers/home_controller.rb
```diff
@@ -1,12 +1,58 @@
 class HomeController < ApplicationController
+  include ReturnPathTracking::Helpers
+
   def index
     @form = Jobseekers::SearchForm.new
     @vacancy_facets = VacancyFacets.new
   end
 
+  def jobseeker_forced_login
```
csutter, line 9: I feel like these don't really feel at home (ha) in the `HomeController` 😕 

Given this adds an `AuthenticationController`, is there anything that would stop these actions living there (and that being used for recall)? Feels like that's a more natural place for them to live. Or a `WardenRecallController` if not.

csutter: Also: Most of the other changes are self-explanatory, and I feel like I get what's going on here based on our conversations yesterday, but this bit looks like the most likely to provoke head-scratching if we look at it again a few months down the line, so could probably use a brief comment outlining the approach!

csutter (approved):
Looks really great! One/two final comments but otherwise good to go 🙏

app/controllers/home_controller.rb
elliotcm: @csutter I've moved these into a "warden" controller and added an explanatory comment.

I don't think the authentication controller would be appropriate for this, as that would add these actions to every devise controller.  `AuthenticationController` should remain abstract.

csutter (approved):

PR #4613, opened 2022-02-08
TEVA-3852 Deduplicate job title when indexing for search

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3852

## Next steps:

Reindex in prod after deploy?

### DIFF
--- a/app/models/concerns/database_indexable.rb
+++ b/app/models/concerns/database_indexable.rb
@@ -22,7 +22,7 @@ module DatabaseIndexable
     # `job_title` and `subject` are used for ranking (and weighted with 'A' here, the other
     # searchable fields get the lowest possible 'D' weight)
     Search::Postgres::TsvectorGenerator.new(
-      a: [job_title, subjects],
+      a: [unique_words(job_title), subjects],
       d: [
         readable_phases,
         VacancyPresenter.new(self).show_job_roles,
@@ -41,5 +41,13 @@ module DatabaseIndexable
         organisations.map(&:town).reject(&:blank?).uniq,
       ],
     ).tsvector
+  end
+
+  private
+
+  def unique_words(text)
+    return if text.nil?
+
+    text.downcase.scan(/[\w-]+/).uniq
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/concerns/database_indexable.rb
```diff
@@ -42,4 +42,10 @@ def generate_searchable_content # rubocop:disable Metrics/AbcSize,Metrics/Method
       ],
     ).tsvector
   end
+
+  private
+
+  def deduplicate(text)
+    text.squeeze(" ").gsub(/[^\w\s]/, "").downcase.split.uniq.join(" ")
```
csutter, line 49: You don't have to join at the end - `TsvectorGenerator` will join arrays internally! Maybe this can then be called `unique_words`?

elliotcm: 👍

elliotcm: I also forgot `#scan` exists.

app/models/concerns/database_indexable.rb
```diff
@@ -42,4 +42,12 @@ def generate_searchable_content # rubocop:disable Metrics/AbcSize,Metrics/Method
       ],
     ).tsvector
   end
+
+  private
+
+  def unique_words(text)
+    return if text.nil?
+
+    text.downcase.scan(/\w+/).uniq
```
csutter, line 51: Niche thing: I think `\w` doesn't include dashes - not sure whether or not this is a big deal. Perhaps `[\w-]+`?

csutter (approved):

app/models/concerns/database_indexable.rb
stevenleggdfe: @csutter @elliotcm looking at what people search for they don't search for things using dashes. But we do have the occasional use of a dash e.g. in a job title like "Part-time class teacher" etc. So it might matter for that?

elliotcm: @csutter Updated.

elliotcm:
Product signoff from @stevenleggdfe https://ukgovernmentdfe.slack.com/archives/CQ3UX00P6/p1644486857124089?thread_ts=1644485022.986749&cid=CQ3UX00P6

PR #4615, opened 2022-02-08
Componentise landing page links on homepage

csutter:
This is step one of many towards moving to a more sensible way of
handling landing pages and the links to them. This will make it easier
to implement the upcoming changes to these links too hopefully.

- Create `LandingPageLinkComponent` and `LandingPageLinkGroupComponent`
  to represent the landing page links on the homepage
- Move count outside of links and rejig a11y of links


app/frontend/src/styles/application.scss
```diff
@@ -144,18 +144,6 @@ dd {
   }
 }
 
-.home_index {
```
alexbowen, line 147: 🔥

app/components/landing_page_link_group_component/landing_page_link_group_component.html.slim
```diff
@@ -0,0 +1,5 @@
+div class="#{classes.join(" ")}"
```
alexbowen, line 1: i think you can use this (to be consistent with other components) `= tag.div(class: classes, **html_attributes) do` and then you can add the shared example tests

josephhull676 (approved):

alexbowen (approved):

PR #4625, opened 2022-02-09
use re-usable action

felixtheflex:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3771

## Changes in this PR:

We used third party action such as benc-uk/workflow-dispatch@v1.1, to trigger `workflow dispatch` event on some workflows
since then, GitHub Action has become matured and has introduced features such as re-usable actions with inputs. This means
we can do away with `benc-uk/workflow-dispatch@v1.1` and redesign the workflow such that `re-usable actions` now acts like
a `method` with inputs from the calling workflow; and with the desired logic to execute.

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -1,27 +1,27 @@
-name: Deploy App to Environment
+name: deploy workflow
+description: deploy's application
 
-on:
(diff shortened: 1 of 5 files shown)


sfawcett123 (approved):
LGTM

PR #4628, opened 2022-02-09
[TEVA-3838] Fix name method bug

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3838

## Changes in this PR:

This PR fixes the bug causing "local authority" to be added to the end of a local authority name and then persisted in the database.

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -22,7 +22,7 @@ class Organisation < ApplicationRecord
   end
 
   def name
-    @name ||= read_attribute(:name)&.concat(local_authority? ? " local authority" : "")
+    local_authority? ? "#{read_attribute(:name)} local authority" : read_attribute(:name)
   end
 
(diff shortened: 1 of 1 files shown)


alexbowen:
deoes this not mean there is something needs tweaking in a test too?

robyoung91:
> deoes this not mean there is something needs tweaking in a test too? as in its testing that the string includes the la name when it should be an exact match - i havent actually looked at the test tho!

@alexbowen do you mean the test for the name method on the model spec? In which case, I think the current test seems to cover it (it tests equality to "Camden local authority" for organisation name Camden).

The change here means we just aren't assigning the new concatenated name to an instance variable (this was causing a problem when creating a vacancy, as the organisation name was being updated in the db with the value of the instance variable).

alexbowen (approved):

PR #4636, opened 2022-02-10
[TEVA-3857] Remove redundant govuk-main-wrapper div

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3857

In this PR we also render breadcrumbs and back links above flash messages, have a look at `app/views/layouts/application.html.slim`


elliotcm (approved):
I've largely skimmed it on the assumption that everything still looks and behaves correctly.

Do we need to move that `back_path` test somewhere else?  The component still renders it even if it's rendering outside of itself (which perhaps is a thing we should avoid in components).

PR #4637, opened 2022-02-11
Allow filters content to be passed in slots

alexbowen:
no ticket

Refactor filters view component to allow filter groups to passed in as slots so components are not hardcoded in the filters view, making it more flexible and hence no internal logic is needed.

we decided to refactor this in preparation for filters AB test but in a separate PR. Also regardless of AB test outcome, it makes filters more flexible for future changes

there were also a bunch of legacy arguments being passed to filters that were no longer used and so have removed them

### DIFF
--- a/app/components/filters_component.rb
+++ b/app/components/filters_component.rb
@@ -1,34 +1,57 @@
 class FiltersComponent < GovukComponent::Base
-  attr_accessor :filters, :form, :items, :options
+  attr_accessor :filters, :form, :options
 
   def self.variants
     %w[default]
   end
 
-  def initialize(filters:, form:, items:, options:, classes: [], html_attributes: {})
+  def initialize(form:, options:, filters: {}, classes: [], html_attributes: {})
     super(classes: classes, html_attributes: html_attributes)
 
     @filters = filters
     @form = form
-    @items = items
     @options = options
   end
 
-  def render?
-    filters.present?
+  renders_many :groups, lambda { |key:, component:|
+    tag.div(key, class: "filters-component__groups__group", data: { group: key }) do
+      tag.div(component)
+    end
+  }
+
+  renders_one :remove_buttons, "RemoveButtons"
+
+  class RemoveButtons < ApplicationComponent
+    def initialize(classes: [], html_attributes: {})
+      super(classes: classes, html_attributes: html_attributes)
+    end
+
+    renders_many :groups, lambda { |selected:, key:, options:, value_method:, selected_method:, legend: false|
+      if selected&.any?
+        safe_join([
+          tag.h4(legend, class: "govuk-heading-s"),
+          tag.ul(class: "filters-component__remove-tags") do
+            options.collect do |option|
+              next unless selected.include?(option.public_send(value_method))
+
+              concat(tag.li do
+                tag.button(class: "filters-component__remove-tags__tag icon icon--left icon--cross", data: { group: key, key: option.public_send(value_method) }) do
+                  safe_join([tag.span(t("shared.filter_group.remove_filter_hidden"), class: "govuk-visually-hidden"),
+                             option.public_send(selected_method)])
+                end
+              end)
+            end
+          end,
+        ])
+      end
+    }
   end
 
   def display_remove_buttons?
-    filters[:total_count].positive? && options[:remove_buttons]
-  end
-
-  def mobile_modifier(selector)
-    options[:mobile_variant] ? "#{selector}--mobile" : selector
+    options[:remove_buttons] && filters[:total_count]&.positive?
   end
 
   def default_classes
-    ["filters-component"].tap do |applied_classes|
-      applied_classes.push(mobile_modifier("filters-component")) if options[:mobile_variant]
-    end
+    %w[filters-component]
   end
 end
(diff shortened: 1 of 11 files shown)


app/components/filters_component.rb
```diff
@@ -1,34 +1,57 @@
...
+      tag.div(component)
+    end
+  }
 
-  def display_remove_buttons?
-    filters[:total_count].positive? && options[:remove_buttons]
+  renders_one :remove_buttons, "RemoveButtons"
+
+  class RemoveButtons < GovukComponent::Base
+    def initialize(classes: [], html_attributes: {})
```
elliotcm, line 25: If you like, this can inherit from `ApplicationComponent` which will give you this same initializer.

app/components/filters_component.rb
```diff
@@ -1,34 +1,57 @@
...
+      super(classes: classes, html_attributes: html_attributes)
+    end
+
+    renders_many :groups, lambda { |selected:, key:, options:, value_method:, selected_method:, legend: false|
+      if selected&.any?
+        safe_join([
+          tag.h4(legend, class: "govuk-heading-s"),
+          tag.ul(class: "filters-component__remove-tags") do
+            options.collect do |option|
+              next unless selected&.include?(option.send(value_method))
```
elliotcm, line 35: You know `selected` exists at this point.

elliotcm: Can we also make this a `public_send`?

app/components/filters_component.rb
```diff
@@ -1,34 +1,57 @@
...
+
+              concat(tag.li do
+                tag.button(class: "filters-component__remove-tags__tag icon icon--left icon--cross", data: { group: key, key: option.send(value_method) }) do
```
elliotcm, line 38: Can we also make this a `public_send`?

app/components/filters_component.rb
```diff
@@ -1,34 +1,57 @@
...
+                  safe_join([tag.span(t("shared.filter_group.remove_filter_hidden"), class: "govuk-visually-hidden"),
+                             option.send(selected_method)])
```
elliotcm, line 40: Can we also make this a `public_send`?

app/components/publishers/vacancies_component/vacancies_component.html.slim
```diff
@@ -11,27 +11,16 @@
...
-                                                small: true }],
-                                      options: { remove_buttons: true,
-                                                mobile_variant: true,
-                                                autosubmit_form: true,
-                                                publisher_preference: (@publisher_preference if @organisation.local_authority?) },
-                  classes: "govuk-!-padding-left-3 govuk-!-padding-right-3 govuk-!-padding-top-0")
+        = filters(form: f,
+                  filters: { total_count: @publisher_preference.organisations.count },
+                  options: { remove_buttons: true, publisher_preference: (@publisher_preference if @organisation.local_authority?) },
+                  html_attributes: { tabindex: "-1" }) do |c|
```
elliotcm, line 19: I know the documentation uses `c` for "component", but could we use something else that better indicates which component this variable represents?

app/views/subscriptions/_fields.html.slim
```diff
@@ -3,69 +3,18 @@
...
-    },
-  ]
+.divider-bottom class="govuk-!-margin-top-6 govuk-!-margin-bottom-4"
 
 .plain-styling
   = filters(form: f,
-    filters: { total_count: 0, title: t("subscriptions.edit.filters") },
-    items: items,
-    options: { remove_buttons: false, mobile_variant: false, close_all: false, autosubmit_form: false })
+    options: { remove_buttons: false }) do |c|
```
elliotcm, line 10: Same here regarding `c`.

app/views/vacancies/_filters.html.slim
```diff
@@ -1,64 +1,57 @@
...
     },
   ]
 
 = filters(form: f,
   filters: { total_count: @form.total_filters },
-  items: items,
-  options: { heading_text: "Filter results", remove_buttons: true, mobile_variant: true, close_all: true, autosubmit_form: true },
-  html_attributes: { tabindex: "-1", id: id })
+  options: { heading_text: "Filter results", remove_buttons: true },
+  html_attributes: { tabindex: "-1", id: id }) do |c|
```
elliotcm, line 48: Same here regarding `c`.

elliotcm (changes requested):
A few small change requests here.  I would probably also extract the rendering of the `RemoveButtons` groups into a proper template, but not a blocker to merging.

app/components/filters_component.rb
alexbowen: regarding the selected thing - yes i realise it exists as there is an outer check and then this line checks wether the current group of filters has one of its items currently selected i.e wether to display a remove button for it. am i misunderstanding your comment @elliotcm ?

elliotcm: > and the comment above about extracting this - yes i was thinking it would be nice to do something like that as this groups lamba function does look particuarly nice but i was unsure whats the best way as it contains a bit of logic and the public_send stuff. maybe we can spend 20mins refactoring in mob on monday if nothing better to do. also a little concious of unblocking @josephhull676 abtest at the same time if we dont and its not a deal breaker...

Yeah it's not a big deal.

elliotcm: > regarding the selected thing - yes i realise it exists as there is an outer check and then this line checks wether the current group of filters has one of its items currently selected i.e wether to display a remove button for it. am i misunderstanding your comment @elliotcm ?

What I mean here is that the `&.` is unnecessary because you know `selected` isn't nil.

PR #4666, opened 2022-02-16
[TEVA-3846] visual testing jobseeker authentication

alexbowen:
WIP


saliceti:
As discussed, there is a concern that storing images in the git repository may not be efficient as it will grow over time since all versions are kept, each git clone will get slower and the pipeline will slow down.
Can we consider an S3 bucket instead?

PR #4667, opened 2022-02-16
[TEVA-3776] Fix date job starts pre-selected radio bug

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3776

## Changes in this PR:

- Previously the default value for `starts_asap` was false, causing the radio that submitted the value "false" to be pre-selected.
- Whilst working on this I noticed that the same radios were not present in `app/views/publishers/vacancies/copy/new.html.slim` or `app/views/publishers/vacancies/extend_deadline/show.html.slim,` so I've changed that to be consistent with `app/views/publishers/vacancies/build/important_dates.html.slim`. I assume at some point we decided to change to radios but didn't update the other pages.


elliotcm (approved):
Looks great.

PR #4669, opened 2022-02-17
[TEVA-3811] Update manage your alerts page

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3811

## Changes in this PR:

This PR affects the jobseekers job alert (subscriptions) index page:
- Remove sort options
- Move "manage alert" and "unsubscribe" links to the left and change "manage alert" to a secondary button
- Implement radio buttons to amend subscription frequency, autosubmitted with javascript or with a submit button if no javascript running in browser

Additionally, this PR changes the job alert subscription email, moving the "edit" link.

### Before
image

### After
image

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -132,6 +132,7 @@ class SubscriptionsController < ApplicationController
       redirect_to jobseekers_subscriptions_path, success: t(".success")
     else
       @jobseeker = Jobseeker.find_by(email: subscription.email)
+      store_return_location(jobseekers_subscriptions_path)
       render :confirm
     end
   end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -40,23 +40,20 @@ class SubscriptionsController < ApplicationController
 
   def update
     subscription = Subscription.find_and_verify_by_token(token)
-    @form = Jobseekers::SubscriptionForm.new(subscription_params)
-    @subscription = SubscriptionPresenter.new(subscription)
 
-    if @form.valid?
-      subscription.update(@form.job_alert_params)
-      Jobseekers::SubscriptionMailer.update(subscription.id).deliver_later
-      trigger_subscription_event(:job_alert_subscription_updated, subscription)
-
-      if jobseeker_signed_in?
-        redirect_to jobseekers_subscriptions_path, success: t(".success")
-      else
-        @jobseeker = Jobseeker.find_by(email: subscription.email)
-        store_return_location(jobseekers_subscriptions_path)
-        render :confirm
-      end
+    if updating_frequency?
+      subscription.update(frequency: params.dig(:subscription, :frequency))
+      notify_and_redirect subscription
     else
-      render :edit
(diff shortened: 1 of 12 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -122,4 +119,21 @@ def token
   def vacancy_id
     params.permit(:vacancy_id)[:vacancy_id]
   end
+
+  def updating_frequency?
+    params[:subscription].present?
```
robyoung91, line 124: Do I need to implement some kind of strong params pattern here?

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -101,4 +101,67 @@
       end
     end
   end
+
+  describe "#update" do
```
robyoung91, line 105: is this test necessary, given the system spec? have I tested correctly?

app/controllers/subscriptions_controller.rb
elliotcm: No, you only need that when you're passing the params into something, so that people can't craft params that change fields you don't want changing.  In this case you're just inspecting a param so it's fine.

spec/controllers/subscriptions_controller_spec.rb
elliotcm: Are you testing anything more here than you are in the other test? If so, perhaps, if not, no.

app/frontend/src/styles/application/jobseekers/jobseeker.scss
```diff
@@ -1,6 +1,8 @@
 .jobseeker {
   .create-alert__button {
-    text-align: right;
+    .govuk-button {
+      margin-bottom: govuk-spacing(1);
```
alexbowen, line 4: instead of this could you not just put an inline govuk style on the element in template? `class="govuk-!-margin-bottom-1"` i think thats better for a style like this that is very specific to one instance of something on one page

app/frontend/src/styles/application/jobseekers/subscription.scss
```diff
@@ -0,0 +1,34 @@
+.js-enabled {
+  .subscription {
```
alexbowen, line 2: could you call this `subscription-card` or something a bit more descriptive? subscription is a little generic to me and could mean a few things

app/views/jobseekers/subscriptions/index.html.slim
```diff
@@ -6,15 +6,9 @@
...
-  .govuk-grid-row
-    .govuk-grid-column-two-thirds
-      - if subscriptions.many?
-        = render SortComponent.new path: method(:jobseekers_subscriptions_path), sort: sort
-      p.empty
-
-    .govuk-grid-column-one-third.create-alert__button
-      = govuk_button_link_to t(".button_create"), new_subscription_path
+.govuk-grid-row
+  .govuk-grid-column-one-third.create-alert__button
```
alexbowen, line 10: see above

alexbowen (approved):
looks from from my pov

PR #4693, opened 2022-02-21
TEVA-3868 supportal

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3868

## Screenshots of UI changes:

Screenshot 2022-02-21 at 16 23 40

## Next steps:

Before deploying this, ensure that 
1. a new AWS secret env var has been added called `DFE_SIGN_IN_PUBLISHER_ROLE_ID` with the same value as `DFE_SIGN_IN_SERVICE_ACCESS_ROLE_ID`
2. a new AWS secret env var has been added called `DFE_SIGN_IN_SUPPORT_USER_ROLE_ID` with the new DSI role ID

Once this PR is fully deployed, we can unblock this card:
https://dfedigital.atlassian.net/browse/TEVA-3914

- [x] Confirm with Steven Legg the new event names and event data before merging


app/views/support_users/dashboard/dashboard.html.slim
```diff
@@ -0,0 +1,3 @@
+- content_for :page_title_prefix, "Hello support user"
+
+<p>Hello support user</p>
```
cesidio, line 3: this is a slim file

cesidio: ah but I guess this is just temporary. cool

elliotcm: Fair point, I don't remember which of us added that.  Probably me.  Not sure I can be bothered to change it since it's being replaced anyway and still works as intended.

cesidio (approved):

.env.example
```diff
@@ -2,13 +2,14 @@ AUTHENTICATION_FALLBACK=
 BIG_QUERY_API_JSON_KEY=
 BIG_QUERY_DATASET=
 DATABASE_URL=
+DFE_SIGN_IN_HIRING_STAFF_ROLE_ID=
```
csutter, line 5: Should this be `PUBLISHER` rather than hiring staff to match what we call them in most places in the app?

Gemfile
```diff
@@ -117,7 +117,7 @@ group :test do
   gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "simplecov"
-  gem "vcr"
+  gem "vcr", github: "vcr/vcr", ref: "7ac8292c289ca98dcb4254b59e1ee29e2205d8b3"
```
csutter, line 120: Can you add a comment (e.g. `# fixes a syntax error bug between VCR and Webmock`) to explain why it's pinned?

app/services/dsi_client.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    DSIClient::Response.new(response).tap do |r|
+      raise DSIClient::RequestInvalid, response.code if r.invalid?
+      raise DSIClient::RequestFailed, response.code if r.failure?
+    end
+  end
+
+  def uri_for_path(path)
+    URI(ENV["DFE_SIGN_IN_URL"]).tap do |uri|
```
csutter, line 36: This and possibly some of the other `ENV.[]` calls in this class might benefit from `ENV.fetch` just for a slightly earlier and more meaningful error in case the env var isn't set

csutter (approved):
💪

Gemfile
elliotcm: This is in the git blame, but sure.

.env.example
elliotcm: OK

csutter (approved):

alexwiskar:
@elliotcm I can access this now and it all looks good

PR #4696, opened 2022-02-22
Remove no_vacancies_component

cesidio:
# Before
Screenshot 2022-02-22 at 10 45 26

# After
Screenshot 2022-02-22 at 10 46 57


csutter (approved):
😱

PR #4703, opened 2022-02-23
[TEVA-3847] convert filter component js to stimulus controller

alexbowen:
- refactor filters JS to be a stimulus controller
- refactor to inject submit button rather than full formbuilder form as submit button is only thing the component conditionally shows if no JS on page

because of work i was experimenting with to utilise preview components for testing, the PR also removes old design system code but keeps the filters preview which i want to utilise (along with other components) in visual testing

# preview of component

Screenshot 2022-02-23 at 11 30 07


app/components/filters_component/filters_component.html.slim
```diff
@@ -1,4 +1,4 @@
-= tag.div(class: classes, **html_attributes) do
+= tag.div(class: classes, **html_attributes.merge({ data: { controller: "filters" } })) do
```
cesidio, line 1: `.merge({ data: { controller: "filters" } })` maybe should go in the ruby file

PR #4705, opened 2022-02-23
refactor/reduce searchable collection args

alexbowen:
this i felt needed doing before i do any stimulus JS on this component

a stab at simplfying and removing unused arguments

also injects collection components to remove internal logic that is nothing to do with searchable collection. Also i felt injecting a form into this should not be neccessary for what this does.


cesidio:
The refactoring makes the size of checkboxes different
Screenshot 2022-02-23 at 16 57 27
Screenshot 2022-02-23 at 16 57 35

alexbowen:
@cesidio  should be easy fix, i obviously missed off an argument porting it over

app/components/searchable_collection_component/searchable_collection_component.html.slim
```diff
@@ -1,30 +1,8 @@
-= tag.div(class: classes, **html_attributes) do
+= tag.div(class: classes, **html_attributes.merge({ data: { controller: "searchable-collection" } })) do
```
cesidio, line 1: it's better to do `.merge({ data: { controller: "searchable-collection" } })` in the ruby file than in the view

cesidio (approved):

PR #4715, opened 2022-02-24
Remove unused styles and partials

cesidio:


app/views/shared/search/_current_location.html.slim
```diff
@@ -1,3 +1,3 @@
-.location-finder.govuk-body class="govuk-!-margin-top-1 govuk-!-margin-bottom-4"
```
alexbowen, line 1: i dont think you want to remove the class

alexbowen (approved):
assume all is well, no review app. some of that css removed looks ancient!

PR #4730, opened 2022-02-28
Make rack-mini-profiler opt-in

elliotcm:
## `bin/rails server`
Screenshot 2022-02-28 at 15 37 58

## `RACK_MINI_PROFILER=true bin/rails server`
Screenshot 2022-02-28 at 15 37 30


alexbowen (approved):
sorts my issue and feels like it should be opt in, maybe another rails dev approval though?!

PR #4733, opened 2022-02-28
[TEVA-3913] Cleanup views and styles

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3913


app/views/publishers/vacancies/_actions.html.slim
```diff
@@ -1,9 +1,9 @@
 .vacancy-actions
   - if vacancy.published?
-    = govuk_button_link_to t("buttons.copy_listing"), new_organisation_job_copy_path(vacancy.id), class: "govuk-button--secondary"
+    = govuk_button_link_to t("buttons.copy_listing"), new_organisation_job_copy_path(vacancy.id), class: "govuk-button--secondary govuk-!-margin-right-2"
```
alexbowen, line 3: can you use the `govuk-button-group` thing so all the spacing is correct and you dont need the  repeated `govuk-!-margin-right-2` ?

alexbowen: i think `.govuk-button-group` could replace `.vacancy-actions`

app/frontend/src/styles/application/publishers/publisher.scss
```diff
@@ -1,21 +1,3 @@
...
-      @include govuk-media-query($from: desktop) {
-        margin-right: govuk-spacing(2);
-      }
-    }
-  }
-
-  .job-application-actions {
-    display: flex;
-  }
+.job-application-actions {
```
alexbowen, line 1: wondering if this could dissappear also and where its used replaced with `govuk-button-group` also?

alexbowen (approved):
whilst i havent physically checked pages, all looks sensible

PR #4736, opened 2022-03-01
use button group for listing application buttons

alexbowen:
[see thread](https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1646127635925769)

before

Screenshot 2022-03-01 at 11 25 36

after

Screenshot 2022-03-01 at 11 25 27


cesidio:
This work has already been done in the cleanup PR that I just merged.

PR #4756, opened 2022-03-04
[TEVA-3796] Link tracking

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3796


cesidio (approved):

app/helpers/links_helper.rb
```diff
@@ -1,23 +1,146 @@
...
+      **kwargs,
+    )
+  end
+
+  def map_link(text, url, vacancy_id: nil, **kwargs)
+    tracked_link_to(
+      text,
+      url,
+      link_type: :school_website_from_map,
+      link_subject: anon(vacancy_id),
```
stevenleggdfe, line 102: This is probably just my ignorance but just to check - should this read:
`link_subject: anon(vacancy.id),`
instead of
`link_subject: anon(vacancy_id),`
?

app/helpers/links_helper.rb
```diff
@@ -1,23 +1,146 @@
...
+  def anon(value)
+    StringAnonymiser.new(value).to_s
+  end
+
+  def school_website_link(organisation, vacancy: nil, **kwargs)
+    tracked_open_in_new_tab_link_to(
+      t("schools.website_link_text", organisation_name: organisation.name),
+      organisation.url,
+      link_type: :school_website,
+      link_subject: anon(vacancy&.id),
```
stevenleggdfe, line 58: What does the & after vacancy do here? Probably my ignorance of Ruby but just to check!

alexbowen (approved):

app/helpers/links_helper.rb
elliotcm: Because the vacancy is optional, we have to account for situations where it isn't present.  `vacancy&.id` means "call `id` on `vacancy` if `vacancy` is present, otherwise return nil`.

elliotcm: No, because in this case we're passing in `vacancy_id` rather than the whole vacancy, due to where it's being used.  I think due to some changes we could now pass the whole vacancy in but it would be the same effect.

elliotcm: Because the vacancy is optional, we have to account for situations where it isn't present.  `vacancy&.id` means "call `id` on `vacancy` if `vacancy` is present, otherwise return nil`.

app/helpers/links_helper.rb
elliotcm: No, because in this case we're passing in `vacancy_id` rather than the whole vacancy, due to where it's being used.  I think due to some changes we could now pass the whole vacancy in but it would be the same effect.

elliotcm:
Product signoff from @stevenleggdfe:
https://ukgovernmentdfe.slack.com/archives/CQ3UX00P6/p1647441445574429?thread_ts=1647429914.208379&cid=CQ3UX00P6

PR #4760, opened 2022-03-04
Fix N+1 query in `VacancySearch`

csutter:
In almost all situations where we search for a set of vacancies, we are
interested in the organisation data these vacancies belong to. Currently
this suffers from an N+1 query situation because we do not include the
organisations as part of the generated scope.

### Search results page: Before (65 SQL queries 🤯)

image

### Search results page: After (15 SQL queries 😅)

image


cesidio (approved):
little change, big impact

PR #4778, opened 2022-03-10
Refactor vacancy_presenter

cesidio:


app/presenters/vacancy_presenter.rb
```diff
@@ -50,14 +45,14 @@ def publish_today?
     model.publish_on == Date.current
   end
 
-  def working_patterns
-    model_working_patterns.map { |working_pattern|
+  def show_working_patterns
```
elliotcm, line 48: Could we change this to something like `humanized_working_patterns`?  I keep expecting this to be a boolean.

elliotcm (changes requested):

app/presenters/vacancy_presenter.rb
cesidio: `humanized_working_patterns` is more explicit, but I would change all these methods to be something like `readable_method`, what do you think about it?

elliotcm (approved):

PR #4807, opened 2022-03-16
[TEVA-3944] Add action to strip whitespace from attributes

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3944

## Changes in this PR:

This PR adds a `before_save` action to the `ApplicationRecord` model to ensure that leading or trailing whitespace (such as that added by user input) is stripped from model parameters before saving to the database.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,6 +1,8 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
+  before_save :strip_attributes
+
   after_create { EventContext.trigger_event(:entity_created, event_data) }
   after_update { EventContext.trigger_event(:entity_updated, event_data) unless saved_change_to_attribute?(:last_activity_at) }
   after_destroy { EventContext.trigger_event(:entity_destroyed, event_data) }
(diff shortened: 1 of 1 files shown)


app/models/application_record.rb
```diff
@@ -1,6 +1,8 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
+  before_save :strip_attributes
```
robyoung91, line 4: Should this be called in the ApplicationRecord model or in the relevant classes which inherit from it (e.g. `Vacancy`) ?

josephhull676: I'm not sure, but I guess you could do a little test to check? 

We have other callbacks defined in this file (I think that's what they're called). 

`after_create { EventContext.trigger_event(:entity_created, event_data) }` for eg.

Perhaps you could put a breakpoint in the `trigger_event` method and then create an entity (a vacancy for eg) and see if you hit the breakpoint. If you do, I guess you can just put the before_save in this file.

josephhull676 (approved):

app/models/application_record.rb
robyoung91: It does work when placed in here (have tested), but was wondering if it is costly to be calling it on all models (rather than just the ones that have user input and might have whitespace)

PR #4808, opened 2022-03-16
[TEVA-3932] Change feature nudge copy and reset dismiss flag

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3932

### Before
Screenshot 2022-03-16 at 14 56 27

### After
Screenshot 2022-03-16 at 15 24 58

### DIFF
--- a/app/controllers/publishers/new_features_controller.rb
+++ b/app/controllers/publishers/new_features_controller.rb
@@ -1,7 +1,7 @@
 class Publishers::NewFeaturesController < Publishers::BaseController
   skip_before_action :check_terms_and_conditions, only: %i[show update reminder]
 
-  NEW_FEATURES_PAGE_UPDATED_AT = DateTime.new(2022, 1, 7).freeze # This constant lives here so that we remember to update it.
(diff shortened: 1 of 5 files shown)


db/migrate/20220316152547_reset_feature_dismiss.rb
```diff
@@ -0,0 +1,5 @@
+class ResetFeatureDismiss < ActiveRecord::Migration[6.1]
+  def change
+    Publisher.update_all(dismissed_new_features_page_at: nil)
```
csutter, line 3: This has been something that's been bothering me for a while, but it feels like instead of resetting this we should have a thing somewhere that has the date the features page has last been changed 😕

csutter: (and then check against that)

elliotcm: Yeah probably.  Maybe one for the supportal?

elliotcm:
Product signoff by @alexwiskar:
https://ukgovernmentdfe.slack.com/archives/CG9A8H1HP/p1647507099407189?thread_ts=1647448825.357529&cid=CG9A8H1HP

csutter (approved):

PR #4811, opened 2022-03-16
[TEVA-3691] Add activity logs for vacancies

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3691

## Screenshots

Screenshot 2022-03-16 at 16 37 20


app/controllers/publishers/vacancies/activity_log_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Publishers::Vacancies::ActivityLogController < Publishers::Vacancies::BaseController
+  helper_method :activity_log
+
+  def show; end
+
+  private
+
+  def activity_log
+    vacancy.versions.map { |version| { created_at: version.created_at, publisher: version.actor, changeset: version.changeset } }.reverse
```
csutter, line 9: Still think this isn't necessary 🤔 We should probably just have something like this in the show:

```ruby
def show
  @versions = vacancy.versions.order(created_at: :desc)
end
```

And then iterate over `@versions` in the view.

app/helpers/vacancies_helper.rb
```diff
@@ -91,4 +91,10 @@ def vacancy_breadcrumbs(vacancy)
       "#{vacancy.job_title}": "",
     }
   end
+
+  def vacancy_activity_log_item(attribute, new_value)
+    new_value.map! { |value| Vacancy.array_enums[attribute.to_sym].key(value).humanize } if attribute.to_sym.in?(Vacancy.array_enums)
+
+    t("publishers.activity_log.show_value_changed", attribute: attribute.humanize, new_value: (new_value.is_a?(Array) ? new_value.to_sentence : new_value.humanize))
```
csutter, line 98: Could this use field translations instead of humanising?

app/models/vacancy.rb
```diff
@@ -17,6 +17,12 @@ class Vacancy < ApplicationRecord
   MAIN_JOB_ROLES = { teacher: 0, leadership: 1, teaching_assistant: 6, education_support: 4, sendco: 5 }.freeze
   ADDITIONAL_JOB_ROLES = { send_responsible: 2, ect_suitable: 3 }.freeze
 
+  ATTRIBUTES_TO_TRACK = %i[
```
csutter, line 20: Maybe `ATTRIBUTES_TO_TRACK_IN_ACTIVITY_LOG`?

app/models/vacancy.rb
```diff
@@ -67,13 +73,22 @@ class Vacancy < ApplicationRecord
...
 
   before_save :on_expired_vacancy_feedback_submitted_update_stats_updated_at
   before_save :refresh_geolocation, if: -> { job_location_changed? }
 
   EQUAL_OPPORTUNITIES_PUBLICATION_THRESHOLD = 5
   EXPIRY_TIME_OPTIONS = %w[7:00 8:00 9:00 10:00 11:00 12:00 13:00 14:00 15:00 16:00 17:00 23:59].freeze
 
+  def self.array_enums
+    {
+      job_roles: { "teacher" => 0, "leadership" => 1, "teaching_assistant" => 6, "education_support" => 4, "sendco" => 5, "send_responsible" => 2, "ect_suitable" => 3 },
```
csutter, line 86: It's good that this is explicit now, but I think you can simplify it a bit and not type out the explicit values:

```ruby
def self.array_enums
  {
    job_roles: job_roles,
    key_stages: key_stages,
    working_patterns: working_patterns,
  }
end
```

db/schema.rb
```diff
@@ -433,6 +433,16 @@
     t.datetime "unsubscribed_at"
   end
 
+  create_table "support_users", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
csutter, line 436: Don't think you meant to check this in?

csutter (changes requested):
Couple of comments but looking great otherwise 👍

app/controllers/application_controller.rb
```diff
@@ -108,4 +109,8 @@ def current_organisation
     @current_organisation ||= Organisation.find_by(id: session[:publisher_organisation_id])
   end
   helper_method :current_organisation
+
+  def user_for_paper_trail
+    current_publisher
```
csutter, line 114: Possibly `current_publisher || current_support_user` to future-proof this? Given that soon they will be able to edit stuff too!

app/views/publishers/vacancies/activity_log/show.html.slim
```diff
@@ -0,0 +1,17 @@
...
+  .govuk-grid-column-two-thirds
+    - unless @versions.present?
+      p.govuk-body-m = t("publishers.activity_log.no_changes")
+    - @versions.each do |version|
+      - version.changeset.each_pair do |attribute, diff|
+        p.govuk-body class="govuk-!-font-weight-bold"
+          = vacancy_activity_log_item(attribute, diff.last, organisation_type_basic(vacancy.organisation))
+
+      p.govuk-body-s class="govuk-!-margin-top-1"
+        | By #{version.actor.given_name} #{version.actor.family_name} on #{version.created_at}
```
csutter, line 16: I wonder if we should have a `papertrail_display_name` method on `Publisher` and `SupportUser` that we can use here instead of having given name and family name here in the view 🤔

For example:
```ruby
class Publisher
  def papertrail_display_name
    "#{given_name} #{family_name} (#{organisations.first.name})"
  end
end
```

and

```ruby
class SupportUser
  def papertrail_display_name
    "#{given_name} #{family_name} (DfE Teaching Vacancies team)"
  end
end
```

config/locales/publishers.yml
```diff
@@ -12,6 +12,38 @@ en:
...
+      salary: Salary changed to %{new_value}
+      school_visits: "%{organisation_type} visits changed to %{new_value}"
+      starts_on: Expected start date changed to %{new_value}
+      subjects:
+        one: Subject changed to %{new_value}
+        other: Subjects changed to %{new_value}
+      working_patterns:
+        one: Working pattern changed to %{new_value}
+        other: Working patterns changed to %{new_value}
+
```
csutter, line 42: Big space is big

db/migrate/20220309085809_create_versions.rb
```diff
@@ -0,0 +1,31 @@
+# This migration creates the `versions` table, the only schema PT requires.
+# All other migrations PT provides are optional.
+class CreateVersions < ActiveRecord::Migration[6.1]
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, { null: false }
```
csutter, line 6: There are stray `{}` here which lead to a corrupted schema (see `schema.rb`)

db/schema.rb
```diff
@@ -495,6 +495,18 @@
     t.index ["searchable_content"], name: "index_vacancies_on_searchable_content", using: :gin
   end
 
+  create_table "versions", force: :cascade do |t|
+    t.string "item_type"
+    t.string "{:null=>false}"
```
csutter, line 500: (see migration)

csutter (changes requested):
A couple of things but we're pretty much there! 💪

app/models/vacancy.rb
```diff
@@ -67,13 +73,26 @@ class Vacancy < ApplicationRecord
   validate :enable_job_applications_cannot_be_changed_once_listed
 
   has_noticed_notifications
+  has_paper_trail on: [:update],
+                  only: ATTRIBUTES_TO_TRACK_IN_ACTIVITY_LOG,
+                  if: proc { |vacancy| vacancy.listed? }
```
csutter, line 78: Do you maybe want `published?`?

I think `listed?` will only apply once it's gone live, so wouldn't cover changes made _after_ publishing but _before_ the date it's published for.

Not saying that `published?` is better, just worth thinking about/checking!

csutter (approved):

app/models/vacancy.rb
josephhull676: Don't scheduled jobs also have the status `published`? The ticket only requires the activity log tab for active and closed jobs, so changes to scheduled jobs (before they're live but after they're 'published') won't be displayed anyway.

My goal here was to only show changes that have been made after the job goes live.

PR #4819, opened 2022-03-17
[TEVA-4002] Button click tracking

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4002

### DIFF
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -1,6 +1,7 @@
 class Api::EventsController < Api::ApplicationController
   # Prevent arbitrary events/data from being triggered by the frontend
   EVENT_ALLOWLIST = {
+    copied_to_clipboard: %i[description subject],
     tracked_link_clicked: %i[link_type link_subject text href mouse_button],
   }.freeze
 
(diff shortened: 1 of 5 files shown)


csutter (approved):
Looks good – one more thing, do you think it would be worth to have a `subject` for the `copied_to_clipboard` event too? So we could e.g. track the anonymised ID for the vacancy whose URL has been copied.

elliotcm:
> Looks good – one more thing, do you think it would be worth to have a `subject` for the `copied_to_clipboard` event too? So we could e.g. track the anonymised ID for the vacancy whose URL has been copied.

@csutter That sounds like a @stevenleggdfe question.

stevenleggdfe:
> > Looks good – one more thing, do you think it would be worth to have a `subject` for the `copied_to_clipboard` event too? So we could e.g. track the anonymised ID for the vacancy whose URL has been copied.
> 
> @csutter That sounds like a @stevenleggdfe question.

Yes please @elliotcm !

elliotcm:
@csutter @stevenleggdfe Done!

csutter (approved):

PR #4826, opened 2022-03-18
Fix working_patterns_for_job_schema

cesidio:
Compliant with 

https://developers.google.com/search/docs/advanced/structured-data/job-posting

### DIFF
--- a/spec/presenters/vacancy_presenter_spec.rb
+++ b/spec/presenters/vacancy_presenter_spec.rb
@@ -5,24 +5,6 @@ RSpec.describe VacancyPresenter do
 
   let(:vacancy) { build(:vacancy) }
 
-  describe "#expired?" do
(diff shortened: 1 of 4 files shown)


spec/presenters/vacancy_presenter_spec.rb
```diff
@@ -164,18 +146,18 @@
 
   describe "#working_patterns_for_job_schema" do
     context "when working_patterns is unset" do
-      let(:vacancy) { build_stubbed(:vacancy, :without_working_patterns) }
+      let(:vacancy) { build_stubbed(:vacancy, :without_working_patterns, fixed_term_contract_duration: "") }
```
josephhull676, line 149: Would it be better to just add `fixed_term_contract_duration: ""` to the trait?

josephhull676 (approved):

PR #4835, opened 2022-03-22
Remove CardComponent

cesidio:
Our `CardComponent` is just a duplicate of the `govuk_summary_list` component

### Before
Screenshot 2022-03-22 at 16 53 02

### After
Screenshot 2022-03-22 at 16 53 25


josephhull676 (approved):
Depending on whatever the designers think of the styling of course!

PR #4844, opened 2022-03-24
[TEVA-3998] Add alerts after redirect to sign in page

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-3998

## Comments:

- Not sure about my approach and wasn't too sure about where to put the translations.


elliotcm (approved):
Seems good enough.  It's not super elegant but nothing in that part of the app is (because the warden/devise stuff is a bit opaque).

PR #4851, opened 2022-03-28
Remove Sentry error when user fails to log in

elliotcm:
We don't really use this and it's very noisy.


cesidio (approved):

csutter (changes requested):
Can we have this as a log instead of a Sentry message? There is _something_ there to be investigated so still worth having a paper trail.

elliotcm:
> Can we have this as a log instead of a Sentry message? There is _something_ there to be investigated so still worth having a paper trail.

@csutter Does this not also happen just when someone gets their password wrong?

elliotcm:
Closed in favour of investigation:
https://dfedigital.atlassian.net/browse/TEVA-4048

PR #4860, opened 2022-03-30
[TEVA-4050] Add a rake task for fetching recruited candidates for a given cycle year

elliotcm:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4050


elliotcm:
Closing this as it turns out you can't actually get applications that are older than the current cycle via the API.  @duncanjbrown will be sending Yvonne a CSV directly.

I'll leave this branch around in case this comes up again in future.

cc @csutter

PR #4867, opened 2022-03-30
add full screen plugin for leaflet maps

alexbowen:
no ticket

LeafletJs is a quality library and there is a plethora of functionality and plugins that we could take advantage of. obviously dont want to go overboard but i thought this (as a simple plugin introduction) was quite neat and i feel is of benefit for a mobile/small screen user to use. use the control to toggle default/full screen view and works smoothly cross browser.

these screen shots probably not the best example, would be useful more on busy maps with many markers/polygons

Screenshot 2022-03-30 at 23 56 44

Screenshot 2022-03-30 at 23 57 22

Screenshot 2022-03-30 at 23 57 51

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6063,6 +6063,11 @@ last-call-webpack-plugin@^3.0.0:
     lodash "^4.17.5"
     webpack-sources "^1.1.0"
 
+leaflet.fullscreen@^2.2.0:
+  version "2.2.0"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1234,7 +1234,7 @@
     mkdirp "^1.0.4"
     rimraf "^3.0.2"
 
-"@rails/webpacker@^5.4.3":
+"@rails/webpacker@^5.4.2":
```
cesidio, line 1237: why this change?

alexbowen: well spotted, i shouldnt have. even tho dont think it matters, ill remove it from this pr

cesidio (approved):

PR #4905, opened 2022-04-11
[TEVA-4083] Add helper to prevent GTM use in non-production environments

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4083


app/helpers/application_helper.rb
```diff
@@ -37,4 +37,8 @@ def footer_links
       "Accessibility" => page_path("accessibility"),
     }
   end
+
+  def include_google_tag_manager?
+    cookies["consented-to-cookies"] == "yes" && Rails.application.config.app_role.production?
```
cesidio, line 42: you can use `Rails.configuration.app_role.production?` that is a bit shorter

cesidio (approved):

PR #4911, opened 2022-04-12
Remove google analytics profile id from non prod envs

josephhull676:
## Changes in this PR:

- GOOGLE_ANALYTICS_PROFILE_ID not needed in non-production environments
- GOOGLE_TAG_MANAGER_CONTAINER_ID not a secret, so have added it to the production ENV variables

# Questions

- I could hard code the GOOGLE_TAG_MANAGER_CONTAINER_ID instead. It's only used in two places. Should I?


josephhull676:
> 

Doesn't look like it. I'll remove it from production env var as well then?

csutter (approved):

PR #4933, opened 2022-04-20
Fix supportal page width and select styling

robyoung91:
## Changes in this PR:

This PR fixes the page width on the supportal feedback pages, ensuring it is full width on each page, including the header. The category select boxes are also changed to use govuk styling.

### Before
image

### After
image


alexbowen (approved):

cesidio:
do we still need `app-width-container--full` defined if it's not used anymore?

cesidio (approved):

PR #4946, opened 2022-04-21
Change supportal feedback tables

robyoung91:
## Changes in this PR:

This PR adds different colours for the "Who" category tags in the general feedback page. The repetition of the "Who" category tag in the "Source" column is removed. 

Additionally, the `unauthenticated_user_type` method is now able to determine if a user is a jobseeker or hiring staff based on their answer to the `visit_purpose` question; a new `"unknown"` value is added for users who do not specify this answer.

A "Contact email" column is added to the table, displaying the user's email address _if they have opted-in to being contacted._

The satisfaction rating tables are moved into sub-tabs.

### Before
image

image

### After
image

image

### DIFF
--- a/app/components/supportal_table_component.rb
+++ b/app/components/supportal_table_component.rb
@@ -8,6 +8,8 @@ class SupportalTableComponent < GovukComponent::Base
     text
   ].freeze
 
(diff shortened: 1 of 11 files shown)


alexbowen (approved):
wondered if there is a system test for this needs updating/adding? might be overkill, not the end of the world either way tbh

josephhull676 (approved):

PR #4947, opened 2022-04-21
refactor definition list usage/styles

alexbowen:


app/frontend/src/styles/layouts/vacancies/index.scss
```diff
@@ -17,6 +13,18 @@
         border-top: 1px solid $govuk-border-colour;
         padding-top: $govuk-gutter-half;
       }
+
+      dl {
```
cesidio, line 17: in the marker popup we need to render this `dl` outside the `li` element

cesidio (approved):

PR #4954, opened 2022-04-26
[TEVA-4090] Use searchable collection to select a job to copy

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4090

## Changes in this PR:

- Select a job to copy form now uses the `searchable_collection` component
- It seems that a user would receive an error when not selecting an option when using the `vacancy_selector` component. I have now added a validation for this form to prevent that from happening


spec/system/copy_or_new_vacancy_flow_spec.rb
```diff
@@ -4,94 +4,98 @@
...
-    expect(page).to have_content("What do you want to do?")
-
-    choose "Start with a blank template"
-    click_on "Continue"
-    expect(page).to have_content("Step 1 of 9")
+  before { login_publisher(publisher: publisher, organisation: school) }
 
-    click_on "Cancel and return to manage jobs"
-    click_on "Create a job listing"
+  context "if there are published vacancies" do
```
cesidio, line 11: context should always start with either `with` or `when`

app/views/publishers/vacancies/select_a_job_for_copying.html.slim
```diff
@@ -2,10 +2,21 @@
 
 .govuk-grid-row
   .govuk-grid-column-full
-    = form_tag redirect_to_copy_job_organisation_jobs_path do
-      .govuk-form-group
-        = vacancy_selector(vacancies.published.order(:job_title), organisation: current_organisation, label_text: t(".heading"))
+    fieldset.govuk-fieldset
```
cesidio, line 5: can't we use `govuk_fieldset` helper method?
https://govuk-form-builder.netlify.app/building-blocks/fieldsets/

app/views/publishers/vacancies/select_a_job_for_copying.html.slim
```diff
@@ -2,10 +2,21 @@
 
 .govuk-grid-row
   .govuk-grid-column-full
-    = form_tag redirect_to_copy_job_organisation_jobs_path do
-      .govuk-form-group
-        = vacancy_selector(vacancies.published.order(:job_title), organisation: current_organisation, label_text: t(".heading"))
```
cesidio, line 7: if the vacancy_selector component is used only here, why don't we get rid of it?

app/views/publishers/vacancies/select_a_job_for_copying.html.slim
```diff
@@ -2,10 +2,18 @@
...
+    = form_for @form, url: redirect_to_copy_job_organisation_jobs_path, method: :get do |f|
+      = f.govuk_fieldset legend: { text: t(".heading"), size: "l" }
```
cesidio, line 6: there is a missing `do` at the end of this line

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -11,10 +11,18 @@ def show
 
   def create_or_copy; end
 
-  def select_a_job_for_copying; end
+  def select_a_job_for_copying
+    @form = Publishers::JobListing::SelectAJobForCopyingForm.new
+  end
 
   def redirect_to_copy_job
```
cesidio, line 18: I am not 100% sure about the name of this method, but cannot think a better one for now

alexbowen (approved):
+1 for removing this javascript 'component' - i was putting off having to refactor it. deleting is the best refactor!

PR #4967, opened 2022-04-28
Import United Learning vacancies

csutter:
Adds the ability to import vacancies from external data sources such as
XML feeds or APIs, and adds an initial feed import for a trial launch.

This will probably be extended and refactored extensively in the future
once we have additional data sources that we integrate with and can
abstract/extract common code.

- Add a concept of external sources to import data from
- Add an initial data source for United Learning based on an XML feed
- Add some display tweaks for imported vacancies
- Add a job for running the import and add it to the schedule
- De-link imported vacancies in publisher dashboard component (and add
  notice that they are imported and cannot be edited)
- Scope vacancy lookup in publisher dashboard to only internal vacancies
  (so that they cannot edit imported vacancies)

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4115

#### Hiring staff dashboard showing imported vacancies
image

### DIFF
--- a/app/controllers/publishers/vacancies/base_controller.rb
+++ b/app/controllers/publishers/vacancies/base_controller.rb
@@ -21,7 +21,8 @@ class Publishers::Vacancies::BaseController < Publishers::BaseController
   end
 
   def vacancy
-    @vacancy ||= vacancies.find(params[:job_id].presence || params[:id])
+    # Scope to internal vacancies to disallow editing of external ones
+    @vacancy ||= vacancies.internal.find(params[:job_id].presence || params[:id])
   end
 
   def form_sequence
--- a/app/jobs/import_from_vacancy_sources_job.rb
+++ b/app/jobs/import_from_vacancy_sources_job.rb
@@ -1,0 +1,49 @@
+class ImportFromVacancySourcesJob < ApplicationJob
+  SOURCES = [UnitedLearningVacancySource].freeze
+
+  queue_as :default
+
+  def perform
+    return if DisableExpensiveJobs.enabled?
+
+    SOURCES.each do |source_klass|
+      import_start_time = Time.zone.now
+
+      source_klass.new.each do |vacancy|
+        PaperTrail.request(whodunnit: "Import from external source") do
+          if vacancy.save
+            Rails.logger.info("Imported vacancy #{vacancy.id} from feed #{source_klass.source_name}")
+          else
+            report_validation_errors(source_klass, vacancy)
+            Rails.logger.error("Failed to save imported vacancy: #{vacancy.errors.inspect}")
+          end
+        end
+      end
+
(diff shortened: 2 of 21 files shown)


app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -21,7 +21,8 @@ def vacancies
   end
 
   def vacancy
-    @vacancy ||= vacancies.find(params[:job_id].presence || params[:id])
+    # Scope to internal vacancies to disallow editing of external ones
```
ddippolito, line 24: 🙋🏻  

Can I ask why we want to prevent editing external ones and who usually does it?

app/jobs/import_from_vacancy_sources_job.rb
```diff
@@ -0,0 +1,19 @@
+class ImportFromVacancySourcesJob < ApplicationJob
+  SOURCES = [UnitedLearningVacancySource].freeze
+
+  queue_as :default
+
+  def perform
+    SOURCES.each do |source_klass|
+      source_klass.new.each do |vacancy|
+        PaperTrail.request(whodunnit: "Import from external source") do
```
ddippolito, line 9: Do you want to genuinely leave `whodunnit` or did you forget to change? 😅

app/vacancy_sources/united_learning_vacancy_source.rb
```diff
@@ -0,0 +1,99 @@
...
+#
+# Allows enumerating over the feed's contents and yields intialized `Vacancy` objects that can be
+# manipulated and persisted by calling code (e.g. an import job).
+#
+# Notes:
+#   - Ruby's `RSS` class doesn't recognise the demo feed as an Atom feed for some reason, so this
+#     uses slightly uglier vanilla XML parsing
+class UnitedLearningVacancySource
+  FEED_URL = ENV.fetch("VACANCY_SOURCE_UNITED_LEARNING_FEED_URL").freeze
+  UNITED_LEARNING_TRUST_UID = "5143".freeze
```
ddippolito, line 12: Should the `5143` live in an ENV variable? 🧠 ?

ddippolito (approved):
I think we could remove some of the comments. But good stuff!!! 🚀🚀🚀

cesidio (approved):

app/controllers/publishers/vacancies/base_controller.rb
csutter: Sure! The main reason in the short term is that we've made a decision to always consider the external system the source of truth, and every time the import runs it will overwrite the existing vacancy. This means that if someone were to edit the vacancy on our service, the next time the import runs their changes would be lost.

app/vacancy_sources/united_learning_vacancy_source.rb
csutter: Not opposed to that - but it's a constant that's unlikely to ever change (it's their GIAS ID).

PR #4978, opened 2022-05-03
Show map results only within polygon

cesidio:


alexbowen (approved):

app/models/organisation.rb
```diff
@@ -12,7 +12,7 @@ class Organisation < ApplicationRecord
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
   scope :local_authorities, -> { school_groups.where.not(local_authority_code: nil) }
-  scope :within_polygon, ->(location_polygon) { where("ST_Intersects(?, geopoint)", location_polygon.area.to_s) }
+  scope :within_polygon, ->(location_polygon) { where("ST_Intersects(?, geopoint)", location_polygon.area.to_s) if location_polygon }
```
csutter, line 15: I'm having a bit of a Monday morning moment but what happens if `location_polygon` is nil and the scope is nil?

PR #5009, opened 2022-05-10
Fix rendering of multypoligon on map

cesidio:


csutter:
I think this is what I was talking to Joe about the other day.

Is there a way we can use [GeoJSON](https://leafletjs.com/examples/geojson/) to deal with getting polygon data onto the map, possibly with [the `rgeo-geojson` gem](https://github.com/rgeo/rgeo-geojson)? Might make it a lot cleaner than having to faff around with arrays of coordinates.

josephhull676 (approved):

PR #5024, opened 2022-05-12
Fix vacancy phases in United Learning source

csutter:
This is coming through unexpectedly as `Multiple phases` instead of
`multiple_phases` (ditto for capitalised other phases). They are looking
into it but in the meantime this makes the phase import able to handle
this.


josephhull676 (approved):
Don't know much about what's going on with this MAT import stuff but it makes sense based on what you described.

PR #5050, opened 2022-05-20
[TEVA-4199] implement new pop ui for map vacancy markers

alexbowen:
create custom sidebar stimulus component for marker details due to leaflet popups not being flexible enough to position in this way and current leaflet plugins to do similar are either too complex for our requirement or just dont work with recent versions of leaflet

- add all keyboard/voiceover a11y
- add all required interaction events
- ensure existing maps on vacancy details pages work

Screenshot 2022-05-20 at 11 44 10

Screenshot 2022-05-20 at 11 43 55

(both the above have school name and phase added to the data displayed, as per requirements)


app/components/map_component/map.js
```diff
@@ -12,34 +18,84 @@ const Map = class {
...
     });
   }
 
-  static createMarkerPopup(marker, { data, open }) {
-    marker.bindPopup('', { className: 'map-component__map__popup' });
-    marker.on('keydown', (e) => e.target.closePopup());
-    marker.on('popupopen', async () => {
-      const markerData = await data(marker);
-      marker.setPopupContent(popupTemplate(markerData));
+  createMarkerSidebar(marker, { target }) {
```
alexbowen, line 65: maybe a bit much going on in this method, im going to split a few things out

app/controllers/vacancies_controller.rb
```diff
@@ -1,5 +1,5 @@
 class VacanciesController < ApplicationController
-  MAX_TOTAL_RESULTS_FOR_MAP = 100
+  MAX_TOTAL_RESULTS_FOR_MAP = 200
```
alexbowen, line 2: all changes in this file will be reverted

spec/requests/api/markers_spec.rb
```diff
@@ -34,10 +34,10 @@
         subject
         expect(json).to include(heading_text: vacancy.job_title)
         expect(json).to include(heading_url: job_path(vacancy))
+        expect(json).to include(name: organisation.name)
         expect(json).not_to include(anonymised_id: StringAnonymiser.new(vacancy.id).to_s)
         expect(json).to include(address: full_address(organisation))
         expect(json).to include(description: nil)
-        expect(json[:details].size).to be 4
```
cesidio, line 40: why have you removed this test?

alexbowen: i was a little unsure what to do here tbh, but the reason it came about was that the popup data requirements needed to change and one field `phases` was often not present and so i filtered all the details field in the controller so no null values were present. happy to talk about it in the morning, deleting the test was a bit of a copout im sure!

app/frontend/src/application.js
```diff
@@ -43,6 +44,7 @@ application.register('form', FormController);
 application.register('location-finder', LocationFinderController);
 application.register('manage-qualifications', ManageQualificationsController);
 application.register('map', MapController);
+application.register('sidebar', MapSidebarController);
```
cesidio, line 47: shouldn't this be
`application.register('map-sidebar', MapSidebarController);`
to be consistent with others?

PR #5051, opened 2022-05-20
Use Markers to speed-up rendering of map

cesidio:


app/helpers/maps_helper.rb
```diff
@@ -1,19 +1,12 @@
...
-    vacancies.map { |vacancy|
-      vacancy.organisations.select { |organisation| organisation.id.in?(organisation_ids) }.map do |organisation|
-        {
-          id: vacancy.id,
-          parent_id: organisation.id,
-          geopoint: RGeo::GeoJSON.encode(organisation.geopoint)&.to_json,
-        }
-      end
-    }.flatten
+    Marker.live.within_polygon(polygon).within_area(coordinates, radius).where(vacancy_id: vacancies.pluck(:id)).map do |marker|
```
csutter, line 3: I reckon this can be in the `VacancySearch` service - so instead of that fetching _all_ matching vacancies in the service, passing them to the view, and then into the helper, only to use their ID to fetch the markers, we could have something like this in `VacancySearch`:

```ruby
def markers
  @markers ||= Marker.live.search_by_location(location, radius).pluck(:vacancy_id, :organisation_id, :geopoint)
end
```

app/controllers/vacancies_controller.rb
```diff
@@ -12,9 +12,7 @@ def index
     )
     @vacancies = @vacancies_search.vacancies
 
-    @show_map = support_user_signed_in? &&
-                @vacancies_search.location &&
-                @vacancies_search.total_count <= MAX_TOTAL_RESULTS_FOR_MAP
+    @show_map = @vacancies_search.location
```
csutter, line 15: I think you've just removed this for testing but don't forget to put it back 🙂

db/migrate/20220520105754_create_markers.rb
```diff
@@ -0,0 +1,15 @@
+class CreateMarkers < ActiveRecord::Migration[6.1]
+  def change
+    create_table :markers, id: :uuid do |t|
+      t.belongs_to :vacancy, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :organisation, null: false, foreign_key: true, type: :uuid
+      t.st_point :geopoint, geographic: true
+      t.date :publish_on, null: false
+      t.datetime :expires_at, null: false
```
csutter, line 8: It might also be worth to add an index on the date/datetime columns because we do a _lot_ of scoping by date in the app. Just noticed we don't have any on the main vacancies table so we might want to do that too (separately) 💡

csutter (commented):
I think this might work nicely – added some comments, especially around where we fetch the markers.

db/migrate/20220520105754_create_markers.rb
csutter: Correction – vacancies do have an index on `expires_at`, just not `publish_on` 🤨

app/services/search/vacancy_search.rb
```diff
@@ -49,14 +49,18 @@ def page_to
...
   def vacancies
     @vacancies ||= scope.page(page).per(per_page)
   end
 
+  def markers
+    @markers ||= Marker.live
+                       .search_by_location(location, radius)
+                       .where(vacancy_id: scope.pluck(:id))
+                       .pluck(:vacancy_id, :organisation_id, :geopoint)
+                       .map { |marker| marker_for_map(marker.first, marker.second, marker.third) }
```
csutter, line 61: `marker_for_map(*marker)` for that extra shortness?

csutter (approved):
Beautiful 👌

PR #5070, opened 2022-05-26
few final refinmements to map frontend

alexbowen:
few minor changes, not making any more unless essential now until map is live
* removed any mention of vacancy/organisation from body of map component JS
* updated skiplinks as i think need to have a map skiplink now
* minor style tweaks due to map repositioning
* made sidebar close happen more specifically on map click so doesnt close then open on each marker click

### DIFF
--- a/app/views/vacancies/index.html.slim
+++ b/app/views/vacancies/index.html.slim
@@ -1,5 +1,6 @@
 - content_for :skip_links do
-  = govuk_skip_link(href: "#search-results", text: t("jobs.skip_link"))
+  = govuk_skip_link(href: "#map-results", text: t("jobs.skip_link_map"))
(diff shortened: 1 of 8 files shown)


app/views/vacancies/index.html.slim
```diff
@@ -16,13 +17,15 @@
     .govuk-grid-column-two-thirds-at-desktop
       = render "vacancies/search/fields_and_button", f: f, form: @form
       = render "vacancies/search/open_filters_button", form: @form
-      #search-results
-        - if @show_map
+      - if @show_map
+        #map-results
```
cesidio, line 21: this should be passed to the html_attributes of the component if it's required by the component

cesidio (approved):

PR #5090, opened 2022-05-31
map js controllers initialisation

alexbowen:
do not let map js controllers init if not showing map. e.g if results over threshold, map JS still runs when doesnt need to

shown in sentry error logs


alexbowen:
actually going to change - should use the built in view component `def render?` method

PR #5091, opened 2022-05-31
Add sandbox environment

ddippolito:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4174

## Changes in this PR:

This is a temporary PR. It comes with some changes to AWS Parameter Store for the new Sandbox Environment and we need to check if we have all the correct Secrets and Environment Variable set ups

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -53,14 +53,16 @@ RSpec.describe ApplicationHelper do
 
   describe "#phase_banner_text" do
     subject { helper.phase_banner_text }
-    let(:environment) { "development" }
+    let(:sandbox) { false }
 
-    before { allow(Rails).to receive(:env).and_return(environment) }
+    before do
+      allow(Rails.configuration.app_role).to receive(:sandbox?).and_return(sandbox)
+    end
 
     it { is_expected.to eq("beta") }
 
     context "sandbox" do
-      let(:environment) { "sandbox" }
+      let(:sandbox) { true }
       it { is_expected.to eq("sandbox") }
     end
   end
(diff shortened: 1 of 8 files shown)


terraform/workspace-variables/sandbox_app_env.yml
```diff
@@ -0,0 +1,19 @@
+---
+APP_ROLE: sandbox
+AUTHENTICATION_FALLBACK: false
+BIG_QUERY_DATASET: sandbox_dataset
+DFE_SIGN_IN_ISSUER: https://oidc.signin.education.gov.uk
+DFE_SIGN_IN_REDIRECT_URL: https://teaching-vacancies.service.gov.uk/auth/dfe/callback
```
csutter, line 6: Needs to be `sandbox.tea...`

terraform/workspace-variables/sandbox_app_env.yml
```diff
@@ -0,0 +1,19 @@
...
+DFE_SIGN_IN_REGISTRATION_URL: https://profile.signin.education.gov.uk/register
+DFE_SIGN_IN_URL: https://api.signin.education.gov.uk
+DISABLE_EXPENSIVE_JOBS: true
```
csutter, line 9: Let's leave this for now but need to change later

spec/helpers/application_helper_spec.rb
```diff
@@ -50,4 +50,20 @@
       expect(helper.recaptcha).to eq(captcha)
     end
   end
+
+  describe "#phase_banner_text" do
+    subject { helper.phase_banner_text }
+    let(:app_role) { "development" }
+
+    before do
+      allow(Rails.configuration).to receive(:app_role).and_return(ActiveSupport::StringInquirer.new(app_role))
```
csutter, line 59: I think this is a bit too much implementation detail. How about:

```ruby
let(:sandbox) { false }
# ...
allow(Rails.configuration.app_role).to receive(:sandbox).and_return(sandbox)
```

PR #5106, opened 2022-06-06
[TEVA-4163] Replace leadership with senior leader

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4163


robyoung91 (approved):

bencmitchell:
Hey @cesidio, looks good but noticed one thing worth looking at.
The job role on the listing does not appear as a link when Senior Leader is selected like it does for other roles like Teacher 
image 
image

PR #5120, opened 2022-06-08
Simplify TF state bucket access policy

csutter:
This isn't realy guarding against any real potential problems.


csutter:
Note: this has already been applied 🚢

saliceti (approved):

PR #5123, opened 2022-06-09
Add breadcrumbs to organisation landing pages

robyoung91:
## Changes in this PR:

This PR adds breadcrumbs to the organisation landing pages. Also, the organisation slug param is no longer disregarded if a user performs a keyword/location search.

### Before

image

### After

image


config/locales/jobs.yml
```diff
@@ -109,7 +109,7 @@ en:
         other: >-
           within <span class='govuk-!-font-weight-bold'>%{count} %{units}</span> of
       location_html: <span class='govuk-!-font-weight-bold text-capitalize'>&lsquo;%{location}&rsquo;</span>
-      organisation_html: at %{organisation}
+      organisation_html: at <span class='govuk-!-font-weight-bold'>%{organisation}</span>
```
alexbowen, line 112: anyway html can easily be kept out of translations yml (and below)? not a deal breaker necessarily (the _html postfix is there for a reason but should be avoided if poss imo)

app/helpers/vacancies_helper.rb
```diff
@@ -131,6 +131,14 @@ def referrer_organisation_slug(referrer)
     organisation_slug
   end
 
+  def organisation_landing_page_breadcrumbs(organisation_slug)
+    landing_page = OrganisationLandingPage[organisation_slug]
+    {
+      "#{t("breadcrumbs.home")}": root_path,
```
cesidio, line 137: to be consistent with next line I'd do:
`t("breadcrumbs.home") => root_path`

cesidio (approved):

PR #5143, opened 2022-06-16
[TEVA-4270] do not use Javascript to clear all filters

alexbowen:
remove js from performing clear all filters and construct a url for link instead so clear filters works when JS off

### DIFF
--- a/app/components/filters_component.rb
+++ b/app/components/filters_component.rb
@@ -5,13 +5,18 @@ class FiltersComponent < ApplicationComponent
     %w[default]
   end
 
-  def initialize(submit_button:, options:, filters: {}, title: nil, classes: [], html_attributes: {})
+  def initialize(submit_button:, options:, filters: {}, clear_filters_link: {}, title: nil, classes: [], html_attributes: {})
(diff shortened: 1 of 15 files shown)


app/components/filters_component/filters_component.scss
```diff
@@ -12,8 +12,6 @@
     position: relative;
 
     &-link-button {
```
cesidio, line 14: I think `&-link-button` can be completely removed as this is a link, not a button link. If you completely remove it, it will look like a link, which is less misleading I think

app/components/filters_component.rb
```diff
@@ -5,15 +5,20 @@ def self.variants
...
 
     @submit_button = submit_button
     @options = options
     @filters = filters
+    @clear_filters_link = clear_filters_link
     @title = title
   end
 
+  def clear_filters_link
+    govuk_link_to @clear_filters_link[:text], @clear_filters_link[:url], method: @clear_filters_link[:method], class: "filters-component__heading-link-button"
```
cesidio, line 19: no need to add 
`class: "filters-component__heading-link-button"`

cesidio (approved):

PR #5150, opened 2022-06-21
[TEVA-4013] implement editor component using conteneditable

alexbowen:
- introduces contenteditable area to job advert field for sanitizing copy and pasted content (sources PDF, google doc, MS word, e-teach, TES, TV)
- all editable content is wrapped in an `editor-content` custom element so can be targeted for styling when its used as content on pages
- toggle format bullet point control
- preview of component available for future debugging and/or addition of new controls, can be viewed at `/components/editor_component/preview`
- tested and working consistently in Chrome, FF, Edge, Safari. IE11 no go unfortunately but default form behaviour works fine as this is essentially an enhancement.

## Vacancy Edit screenshot

Screenshot 2022-07-07 at 13 57 28

## Debug Preview screenshot

Screenshot 2022-07-05 at 03 25 20


app/components/review_component/section/section.html.slim
```diff
@@ -4,4 +4,5 @@ li
 
   = tag.div(class: classes, id: id, **html_attributes) do
     = heading
-    .review-component__section__body = content
+    .review-component__section__body
+      == content
```
alexbowen, line 8: ?

spec/presenters/vacancy_presenter_spec.rb
```diff
@@ -7,109 +7,6 @@
 
   describe "#job_advert" do
     let(:vacancy) { build_stubbed(:vacancy, job_advert: "<script> call();</script>Sanitized content") }
-
```
cesidio, line 10: we either get rid of that code, or instead of testing job_advert, we test any of the other fields that use that code

cesidio: you have to update this test to use job_advert, change it to test any other field like `how_to_apply`

robyoung91 (approved):

app/assets/javascript/components/editor/editor.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+  }
+
+  static contentWrapper = (content) => {
+    const strippedContent = content.replace('<editor-content>', '').replace('</editor-content>', '');
+    return `<editor-content>${strippedContent}</editor-content>`;
+  };
+
+  update() {
+    this.formInput.value = null;
```
adamsilver, line 126: Why set to null first?

app/assets/javascript/components/editor/editor.js
```diff
@@ -0,0 +1,145 @@
...
+    ul: 'insertUnorderedList',
+  };
+
+  static ALLOWED_TAGS = ['p', 'ul', 'li', '#text'];
+
+  connect() {
+    this.enabled = true;
+
+    // this will disable IE11 but make sure enabled for JS tests
+    if ((typeof process !== 'undefined' && process.env.NODE_ENV !== 'test') || typeof window.ClipboardEvent === 'undefined') {
```
adamsilver, line 17: Move the first part of this statement to the tests. This script should not be concerned with the unit tests.

adamsilver, line 17: Instead of hiding the toolbar after detecting that the browser cannot do something which relies on JS, I would progressively enhance by adding the toolbar and listening to events in browsers that _do_ support the features.

That way it works without JS and browsers without the right features (perhaps like IE11)

PR #5162, opened 2022-06-22
Remove explicit PostCSS dependency

csutter:
We don't actually use this for anything (except as an implicit dev
dependency from stylelint)


csutter:
Turns out it's used in some magic fashion by the existing pipeline 🙄

PR #5180, opened 2022-06-28
Enable Cloudfront compression for assets

csutter:
Propshaft-served assets are not compressed in the same way as the
legacy Webpacker ones were. This enables Cloudfront compression for
assets while manually setting compression to false for the legacy
`packs/` folder.


terraform/workspace-variables/production.tfvars.json
```diff
@@ -18,7 +18,7 @@
         }
     },
   "documents_s3_bucket_force_destroy": false,
-  "enable_cloudfront_compress": false,
+  "enable_cloudfront_compress": true,
```
saliceti, line 21: the default value is true so you can remove the variable from tfvars

saliceti (approved):

PR #5198, opened 2022-06-29
Fix basic legacy browser support

csutter:
IE11 support for our service has been hit-and-miss for a while since it
is no longer officially supported and about to be completely eradicated
by Microsoft themselves. This tries to add some basic support back in to
ensure users of some legacy browsers get at least a slightly more
tolerable experience.

- Split out "basic" Javascript (GOV.UK Frontend and Rails UJS) that
  should still work okay-ish on some legacy browsers
- Add a `prebuild` script to compile this basic Javascript with an ES5
  target as part of the build flow


cesidio (approved):

csutter:
Closing in favour of a different solution.

PR #5201, opened 2022-06-30
[TEVA-4288] refactor serachable collection to stimulus controller

alexbowen:


cesidio (approved):

app/assets/javascript/components/searchable_collection_component/searchable_collection.js
```diff
@@ -0,0 +1,69 @@
+import { Controller } from '@hotwired/stimulus';
+
+const SearchableCollectionComponent = class extends Controller {
```
csutter, line 3: This should probably be called `Controller` not `Component` for consistency?

alexbowen: gonna do this all in another PR, there is more inconsistency than just this

PR #5220, opened 2022-07-05
Update GOV.UK Components and fix compatibility

csutter:
Since GOV.UK Components 3.0.6, `GovukComponents::Base` has a different
internal API than the one we previously relied on.

This adds a workaround for the new internal API until we decide whether
to follow the new internal API, or decouple ourselves from it.


ddippolito (approved):

ddippolito:
>  or decouple ourselves from it

When you say this you mean locking version to `3.0.4`?

PR #5223, opened 2022-07-05
Add sample images to content-assets

robyoung91:
## Changes in this PR:

This PR adds `.keep` files to `public/content-assets/get-help-hiring/` and `public/content-assets/jobseeker-guides/` so the content team can see where to add new images.


csutter:
Alternative idea: If you just want the folders to exist (because Git doesn't track empty folders), a pattern for this is to create an empty file named `.keep` in the folder 💡

csutter (approved):
Can you change the commit message to reflect that there's no images anymore? Otherwise 👍

PR #5228, opened 2022-07-06
Move pagination of notifications to `pagy`

csutter:
- Add `pagy` gem and configure according to previous Kaminari settings
  and GOV.UK Design System recommendations
- Stop paginating hiring staff notifications with Kaminari and start
  paginating them with `pagy`
- Add `pagy_stats` helper to show pagination statistics with our "house
  style"


alexbowen:
i just added deply review app from this, would be good to see how/if pagy handles a11y/aria on the controls it creates (im assuming a gov component would be on it with this tho)

alexbowen (approved):

PR #5249, opened 2022-07-13
[TEVA-4223] remove jobseeker account feedback cta

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-4223

the footer link changes depending on signed in status


app/views/layouts/_footer.html.slim
```diff
@@ -12,7 +12,10 @@
           = t("footer.headings.feedback")
         ul.govuk-footer__list
           li.govuk-footer__list-item
-            = link_to t("footer.provide_feedback"), new_feedback_path, class: "govuk-footer__link"
+            - if jobseeker_signed_in?
```
cesidio, line 15: I am not sure about this condition here. From the ticket it seems that this link in the footer should not change

alexbowen: well spotted but i discussed this with ben mitchell and he does need to update acceptance criteria

PR #5250, opened 2022-07-14
Update Alpine image

csutter:


alexbowen:
does the devcontainer need updating also?

cesidio (approved):

PR #5282, opened 2022-07-26
Update Email templates for Notify

ddippolito:
Our usage of templates for Gov Notify is far from ideal. We use a large
number of identical templates with the sole purpose of being able to
track emails sent in the Notify Dashboard.
This tracking is not used anymore as it's been superseeded by our
analytical platform which makes the usage of so many templates useless.

This is why in the will start removing templates and only adopt a
production or a sandbox email templetes.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4238


app/mailers/application_mailer.rb
```diff
@@ -29,4 +29,8 @@ def uid
   def utm_campaign
     email_event_type
   end
+
+  def general_template
```
alexbowen, line 33: very minor but general seems an odd word to use? wouldnt default or standard or generic be better word?

ddippolito: It's still in WIP. I should have specified in the title. There's so much that breaks around. I had no intention to leave it to `general_template` 🤓 

It will probably just be `template` at the end

alexbowen: cool, i didnt see the WIP bit in description tbh!

spec/mailers/publishers/expired_vacancy_feedback_prompt_mailer_spec.rb
```diff
@@ -36,5 +36,17 @@
     it "triggers a `publisher_prompt_for_feedback` email event" do
       expect { mail.deliver_now }.to have_triggered_event(:publisher_prompt_for_feedback).with_data(expected_data)
     end
+
+    context "from Sandbox environment" do
+      let(:notify_template) { NOTIFY_SANDBOX_TEMPLATE }
+
+      before do
```
cesidio, line 43: very minor thing, these could be one-liners:

`before { allow(Rails.env).to receive(:sandbox?).and_return(true) }`

cesidio (approved):

app/mailers/jobseekers/devise_emails.rb
```diff
@@ -15,7 +15,7 @@ def confirmation_instructions(record, token, _opts = {})
     send_email(
       jobseeker: record,
       subject: subject,
-      template: NOTIFY_JOBSEEKER_CONFIRMATION_TEMPLATE,
```
alexbowen, line 18: seeing as these 4 methods are almost identical, could they not be refactored into 1 method with the token being optional arg where its needed?

alexbowen (approved):

app/mailers/jobseekers/devise_emails.rb
ddippolito: I thought about when changing it. 
I ended up not extracting the method, because I didn't think it'd be that much of a benefit as `send_email` is already that method.

I'm going to have to do a last PR cleanup after I deploy this and I'll try extracting it and see if it's better 🤓

PR #5284, opened 2022-07-26
[TEVA-4337] move pagination stats to below pagination controls

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-4337

## After

Screenshot 2022-07-26 at 16 22 15

### DIFF
--- a/app/views/vacancies/index.html.slim
+++ b/app/views/vacancies/index.html.slim
@@ -35,7 +35,7 @@
 
       #search-results
         - if @vacancies.any?
-          = render "vacancies/search/sort_and_stats", form: @form, vacancies_search: @vacancies_search, vacancies: @vacancies
+          = render "vacancies/search/sort", form: @form, vacancies_search: @vacancies_search, vacancies: @vacancies
           = render "vacancies/search/results", vacancies: @vacancies
         - elsif @vacancies_search.organisation_slug
           = render "vacancies/search/no_results_organisation", organisation_name: @vacancies_search.organisation.name
@@ -44,3 +44,5 @@
         - else
           = render "vacancies/search/no_results"
       = govuk_pagination(pagy: @pagy)
+      - if @vacancies.any?
+        = render "vacancies/search/stats", form: @form, vacancies_search: @vacancies_search, vacancies: @vacancies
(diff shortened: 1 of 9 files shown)


app/views/vacancies/index.html.slim
```diff
@@ -44,3 +43,5 @@
         - else
           = render "vacancies/search/no_results"
       = govuk_pagination(pagy: @pagy)
+      - if @vacancies.any?
+        = render "vacancies/search/sort_and_stats", form: @form, vacancies_search: @vacancies_search, vacancies: @vacancies
```
cesidio, line 47: will the options to sort vacancies appear at the bottom of the page?

alexbowen: they shouldnt, i didnt pay enough attention, sort should be top stats bottom, will change

app/views/vacancies/search/_sort.html.slim
```diff
@@ -0,0 +1,4 @@
+.govuk-grid-row
+  .govuk-grid-column-full
```
cesidio, line 2: Didn't we say that we should not use `govuk-grid-column-full` when you have something that takes the full row?

app/views/vacancies/search/_stats.html.slim
```diff
@@ -0,0 +1,4 @@
+.govuk-grid-row
+  .govuk-grid-column-full
```
cesidio, line 2: Same here, `govuk-grid-column-full` seems redudntant

spec/system/jobseekers_can_search_for_jobs_spec.rb
```diff
@@ -10,7 +10,7 @@
 
     it "displays page 1 jobs" do
       within("ul.search-results") { expect(page).to have_css("li", count: 2) }
-      expect(page).to have_css(".search-results-sorting__stats", text: strip_tags(I18n.t("app.pagy_stats_html", from: 1, to: 2, total: 6, type: "results")))
+      expect(page).to have_css(".govuk-grid-column-full", text: strip_tags(I18n.t("app.pagy_stats_html", from: 1, to: 2, total: 6, type: "results")))
```
cesidio, line 13: Checking that something is inside `govuk-grid-column-full` is not useful as it's not a specific class. I'd just check the standard
`expect(page).to have_content strip_tag....`

cesidio (approved):

PR #5285, opened 2022-07-26
[TEVA-4339] Change all times across the service to govuk style

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-4339

9am
10:15pm
12pm (midday)
12am (midnight)


ddippolito (approved):
Pretty!

PR #5289, opened 2022-07-27
[TEVA-4333] change instances of 16-19 -> 16 to 19

alexbowen:
https://dfedigital.atlassian.net/browse/TEVA-4333

this is primarily for a11y reasons as 16-19 does not read well to a screen reader, for e.g on these form elements

Screenshot 2022-07-27 at 10 03 24

but so far what ive done does not change what is sorted in DB, that is still 16-19. this means in places like this it is still 16-19 as it is the value that is used as it does not use a value -> label mapping for summary list row - AGREED THS CAN BE DONE AT FUTURE DATE

Screenshot 2022-07-27 at 10 05 54


ddippolito (approved):
Marvellous 🍬

PR #5297, opened 2022-08-01
[TEVA-4349] Update Job Location Step

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4349

### Before
image
image
image

### After
Screenshot 2022-08-04 at 10 49 35
Screenshot 2022-08-04 at 10 49 57


alexbowen (approved):
looks pretty tididy from my pov. one tiny comment, had to put something!

app/models/vacancy.rb
```diff
@@ -98,13 +96,19 @@ def external?
   end
 
   def organisation
-    @organisation ||= organisations.one? ? organisations.first : organisations.first&.school_groups&.first
+    return organisations.first if organisations.one? || organisations.first&.school_group?
```
ddippolito, line 99: Just some 💅🏻 

I think this would be more readable with a standard `if...else`

```
if organisations.one? || organisations.first&.school_group?
  organisations.first
else
  organisations.first&.school_groups&.first
end
  ```
  
Or you could create a predicate method for `organisations.one? || organisations.first&.school_group?` and use the one liner version.
Thoughts?

File: app/helpers/vacancies_helper.rb [hunk 2]
```diff
@@ -66,28 +66,42 @@ def vacancy_about_school_value(vacancy)
     vacancy_or_organisation_description(vacancy)
   end
 
+  def vacancy_job_location_summary(vacancy)
```
ddippolito, line 69: I think it could be a good idea to create a `JobLocationHelper` class at this point. This helper is 170 lines long and we've got a few helpers exclusively related to job location.

What do you think?

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -22,7 +22,10 @@ def vacancies
 
   def vacancy
     # Scope to internal vacancies to disallow editing of external ones
-    @vacancy ||= vacancies.internal.find(params[:job_id].presence || params[:id])
+
+    # As the vacancy is not associated with an organisation upon creation, calling the vacancies method will return an empty array as an organisation is not associated
+    # with it. To fix this, before the vacancy's status is set (and therefore before an organisation is associated), we find the job from the vacancies where status is nil.
```
ddippolito, line 27: I'm not sure I understand why we need `where(status: nil)` 

Would `Vacancy.find(params[:job_id].presence || params[:id]))` not work?
So:

```
@vacancy ||= (vacancies.internal.find_by(id: params[:job_id].presence || params[:id]) || Vacancy.find(params[:job_id].presence || params[:id]))
```

ddippolito (approved):
Bellissimo! 🍫

app/models/vacancy.rb
cesidio: This is a known pattern:
https://www.rubydoc.info/gems/rubocop/0.22.0/Rubocop/Cop/Style/GuardClause

ddippolito: @cesidio Guard clauses are used to break early and avoid executing the rest of the method when it's unnecessary.

This wouldn't need a guard clause but a normal `if... else` as it would read better.

But happy to go with what you prefer as it's just aesthetics.

bencmitchell:
@josephhull676 The trust name needs to be updated on the Review page to match "Trust head office" like on the selection page.
This might need a discussion with a designer actually, because looking at the bottom screenshot from QA we have a different heading for location when head office has been selected

image
image
image

josephhull676:
> @josephhull676 The trust name needs to be updated on the Review page to match "Trust head office" like on the selection page. This might need a discussion with a designer actually, because looking at the bottom screenshot from QA we have a different heading for location when head office has been selected
> 
> image image image

Have updated this so it reads "Trust central office"

PR #5314, opened 2022-08-04
[TEVA-4356] Amend pay package question

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4356

## Changes in this PR:

This PR makes changes to the "Pay package" question of the vacancy listing process.

- There are now three salary checkboxes; one for each salary "type" that the user would like to add. At least one salary type must be provided by the user, and they can provide all three if they wish. The "Actual salary" checkbox is only shown if a user has selected "Part-time" as a working pattern.

- The "additional allowances" text field (previously saved as `benefits` in the database) is now only shown if the user selects "Yes" on the new radio button question asking if there are any. The `benefits` field has been renamed to `benefits_details` and a new boolean field `benefits` has been added, which stores the user's response to the Yes/No radio button question.

### Before
image

### After
image

## Next steps:

- [ ] Run a job to set the new boolean field `benefits` to the correct value for existing vacancies.
- [ ] Update guidance in `creating-the-perfect-teacher-job-advert.md`

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -195,6 +195,14 @@ class Vacancy < ApplicationRecord
     end
   end
 
+  def salary_types
+    [
+      salary.present? ? "full_time" : nil,
+      actual_salary.present? ? "part_time" : nil,
+      pay_scale.present? ? "pay_scale" : nil,
+    ]
+  end
+
   private
 
   def slug_candidates
--- a/config/locales/activerecord.yml
+++ b/config/locales/activerecord.yml
@@ -133,10 +133,24 @@ en:
       blank: Select a working pattern
       inclusion: Select a working pattern
   pay_package_errors: &pay_package_errors
+    actual_salary:
+      blank: Enter actual salary
+      too_long: Actual salary must not be more than %{count} characters
+      invalid_characters: Actual salary must not contain any HTML tags
+    benefits:
+      inclusion: Select yes if you want to include additional allowances
+    benefits_details:
+      blank: Enter additional allowances
+    pay_scale:
+      blank: Enter pay scale
+      too_long: Pay scale must not be more than %{count} characters
+      invalid_characters: Pay scale must not contain any HTML tags
     salary:
-      blank: Enter a salary
+      blank: Enter full-time salary
       too_long: Salary must not be more than %{count} characters
       invalid_characters: Salary must not contain any HTML tags
+    salary_types:
+      invalid: Select salary details
   important_dates_errors: &important_dates_errors
     expires_at:
       after: The closing date must be after the date the job will be listed
--- a/spec/form_models/publishers/job_listing/pay_package_form_spec.rb
+++ b/spec/form_models/publishers/job_listing/pay_package_form_spec.rb
@@ -1,8 +1,47 @@
 require "rails_helper"
 
+SALARIES = {
+  salary: "full_time",
+  actual_salary: "part_time",
+  pay_scale: "pay_scale",
+}.freeze
+
 RSpec.describe Publishers::JobListing::PayPackageForm, type: :model do
-  it { is_expected.to validate_presence_of(:salary) }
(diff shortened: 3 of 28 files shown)


app/models/vacancy.rb
```diff
@@ -174,6 +174,14 @@ def reset_markers
     end
   end
 
+  def salary_types
+    [
+      salary.present? ? "full_time" : nil,
+      actual_salary.present? ? "part_time" : nil,
+      pay_scale.present? ? "pay_scale" : nil,
+    ]
```
cesidio, line 182: how about we compact it at the end so that it does not return nil values?

config/locales/activerecord.yml
```diff
@@ -124,10 +124,24 @@ en:
       blank: Select a working pattern
       inclusion: Select a working pattern
   pay_package_errors: &pay_package_errors
+    actual_salary:
+      blank: Enter the actual salary
+      too_long: Actual salary must not be more than %{count} characters
+      invalid_characters: Actual salary must not contain any HTML tags
```
josephhull676, line 130: Don't think this translation is needed. There's no validation for it.

config/locales/activerecord.yml
```diff
@@ -124,10 +124,24 @@ en:
...
+    benefits:
+      inclusion: Select if there are additional allowances
+    benefits_details:
+      blank: Enter the details of the additional allowances
+    pay_scale:
+      blank: Enter the pay scale
+      too_long: Pay scale must not be more than %{count} characters
+      invalid_characters: Pay scale must not contain any HTML tags
```
josephhull676, line 138: Same here

spec/support/vacancy_helpers.rb
```diff
@@ -229,6 +239,7 @@ def vacancy_json_ld(vacancy)
       "@type": "JobPosting",
       title: vacancy.job_title,
       jobBenefits: vacancy.benefits,
+      jobBenefitsDetails: vacancy.benefits_details,
```
josephhull676, line 242: I think this would have to remain as `jobBenefits`.

Take a look at https://schema.org/JobPosting

spec/system/publishers_can_edit_a_published_vacancy_as_a_school_spec.rb
```diff
@@ -111,7 +111,8 @@
         fill_in "publishers_job_listing_pay_package_form[salary]", with: ""
         click_on I18n.t("buttons.update_job")
 
-        within_row_for(text: I18n.t("helpers.label.publishers_job_listing_pay_package_form.salary")) do
+        within_row_for(text: I18n.t("helpers.label.publishers_job_listing_pay_package_form.salary"),
+                       element: ".govuk-checkboxes__label") do
           expect(page).to have_content(I18n.t("pay_package_errors.salary.blank"))
```
josephhull676, line 116: I wonder, is this checking within this row, or within the page as a whole if you use `page`?

I wonder if you could use a block parameter instead?

Something like

```
within_row_for(text: I18n.t("helpers.label.publishers_job_listing_pay_package_form.salary"), element: ".govuk-checkboxes__label") do |row|
          expect(row).to have_content(I18n.t("pay_package_errors.salary.blank"))
```

robyoung91: Ah yeah I think this is to ensure we get the error message on the field as well as in the summary up top:
image

spec/form_models/publishers/job_listing/pay_package_form_spec.rb
```diff
@@ -1,8 +1,46 @@
 require "rails_helper"
 
+SALARIES = {
```
cesidio, line 3: You can just have `Publishers::JobListing::PayPackageForm::SALARIES`

josephhull676 (approved):

PR #5336, opened 2022-08-12
[TEVA-4357] Change education phases and key stages

cesidio:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4357


app/models/vacancy.rb
```diff
@@ -133,11 +144,15 @@ def can_receive_job_applications?
...
+    phases.any? { |phase| phase.in? %w[primary middle secondary through] }
+  end
+
+  def key_stages_for_phases
+    phases.map { |phase| PHASES_TO_KEY_STAGES_MAPPINGS[phase.to_sym] }.flatten.uniq.sort
   end
 
   def allow_subjects?
-    readable_phases != ["primary"]
+    phases != ["primary"]
```
robyoung91, line 155: Do we need to include `"nursery"` here? It used to be mapped to `"primary"`

cesidio: Thank you, it should be addressed now

spec/models/vacancy_spec.rb
```diff
@@ -551,100 +490,6 @@
...
-      end
-
-      context "when the phase is secondary" do
-        let(:phase) { "secondary" }
-
-        it_behaves_like "allows the user to set the subjects"
-
-        it_behaves_like "does not allow the user to set the key stage"
-      end
-    end
```
josephhull676, line 579: Are we not going to test for any of this now?  The two methods below still determine which attributes can be set, right? Don't we want to test them?

```
def allow_subjects?
    phases.any? { |phase| phase.in? %w[middle secondary through] }
end
```

```
def allow_key_stages?
    phases.any? { |phase| phase.in? %w[primary middle secondary through] }
end
```

spec/services/publishers/vacancies/vacancy_step_process_spec.rb
```diff
@@ -144,22 +138,6 @@
...
-          end
-
-          context "when the vacancy has a phase already" do
-            let(:vacancy) { create(:vacancy, phase: "secondary", organisations: [school, school2]) }
-
-            it "still includes the `education_phases` step" do
-              expect(subject.steps).to include(:education_phases)
-            end
-          end
-        end
```
josephhull676, line 162: So it seems you've removed all the tests that check to see if education_phase is in the steps. Don't we still need to test for this? It's dependant on `allow_phase_to_be_set?` ,right?

Shouldn't we have two contexts here, one for when `allow_phase_to_be_set?` is true and one for when it's false? 

Then we test `allow_phase_to_be_set?` in the vacancy model spec?

spec/services/publishers/vacancies/vacancy_step_process_spec.rb
```diff
@@ -135,7 +129,7 @@
 
       context "when the vacancy is at multiple schools" do
         let(:school) { create(:school, :secondary) }
-        let(:vacancy) { create(:vacancy, organisations: [school, school2]) }
+        let(:vacancy) { create(:vacancy, phases: %w[primary], organisations: [school, school2]) }
 
         context "when the schools have the same phase" do
           let(:school2) { create(:school, :secondary) }
```
josephhull676, line 135: Seems this test is no longer relevant? 

As far as I understand, we now only show the education phases step if the vacancy has no school phases. In this test, both schools have a phase.

app/models/concerns/resettable.rb
```diff
@@ -33,12 +31,8 @@ def reset_parental_leave_cover_contract_duration
...
   end
 
-  def reset_phase
-    self.phase = nil unless allow_phase_to_be_set?
-  end
-
-  def reset_key_stages
-    self.key_stages = [] unless allow_key_stages?
+  def set_default_key_stage
+    self.key_stages = key_stages_for_phases if key_stages_for_phases.one?
```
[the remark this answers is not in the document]
cesidio: because in that case you can have only one possible key stage and we do not ask the question

PR #5348, opened 2022-08-18
[TEVA-4379] enhance cookie banner with JS xhr request for better UX

alexbowen:


app/controllers/cookies_preferences_controller.rb
```diff
@@ -10,7 +10,10 @@ def create
 
     if @cookies_preferences_form.valid?
       cookies["consented-to-cookies"] = { value: @cookies_preferences_form.cookies_consent, expires: 6.months.from_now }
-      redirect_to session[:previous_url]
```
cesidio, line 13: this could just be
`redirect_to session[:previous_url] unless params[:no_redirect]`

cesidio (approved):

PR #5402, opened 2022-09-12
Fix key stages validation

cesidio:


josephhull676 (approved):
Should we not also add a test to check the right key stages have been submitted for the education phase?

PR #5404, opened 2022-09-12
Add Guidance page for sandbox environment

ddippolito:
Add guidance page and link to the environmental banner 

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4320

### DIFF
--- a/app/views/pages/sandbox.html.slim
+++ b/app/views/pages/sandbox.html.slim
@@ -1,46 +1,42 @@
 - content_for :page_title_prefix do
-  | Sign in to Teaching Vacancies
+  | Sandbox Guidance
 
 - content_for :breadcrumbs do
   = govuk_back_link text: t("buttons.back"), href: root_path
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
-    markdown:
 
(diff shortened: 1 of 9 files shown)


app/views/pages/sandbox.html.slim
```diff
@@ -0,0 +1,46 @@
+- content_for :page_title_prefix do
+  | Sign in to Teaching Vacancies
```
cesidio, line 2: This is not relevant for this page, or maybe you want to use the correct content for this

app/views/pages/sandbox.html.slim
```diff
@@ -0,0 +1,46 @@
...
+
+- content_for :breadcrumbs do
+  = govuk_back_link text: t("buttons.back"), href: root_path
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    markdown:
```
cesidio, line 9: I would suggest not to use markdown here, if you have a look at this page, for me it's much more readable without it

https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/views/pages/job-application-preview.html.slim

but I'll leave this up to you. If you think this is more readable, I'll approve it like this

cesidio (approved):

PR #5433, opened 2022-09-22
Review fe tweaks

alexbowen:
3 tickets - see commits


adamsilver:
At the moment you put the class on the last li if all 3 salary options are checked. But they might not all be checked. The last one (even if it's the first and only one) need the margin-bottom class.

Screenshot 2022-09-23 at 10 10 52@2x

josephhull676 (approved):

PR #5509, opened 2022-10-19
[TEVA-4447] Add reset_benefits_details to resettable.rb

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4447

## Changes in this PR:

This method is added to reset the benefits_details field to nil if benefits is set to false. This ensures that it is not shown on the job listing page and that if the user changes benefits to true again, the old benefits_details value is not present.


bencmitchell:
@robyoung91 looks good, removed the "additional allowances" that were present initially and it was no longer displayed on the listing as expected.
image

spec/concerns/reset_vacancies_in_job_listings_spec.rb
```diff
@@ -224,4 +224,17 @@
         .from(previous_further_details).to(nil)
     end
   end
+
+  context "when changing benefits" do
+    subject(:update_benefits) { vacancy.update(benefits: false) }
+
+    let(:vacancy) { create(:vacancy, benefits: true, benefits_details: "test") }
```
josephhull676, line 231: Apparently, unless you need a record to be persisted, `build_stubbed` is the best option to go for. Makes tests quicker! https://thoughtbot.com/blog/use-factory-bots-build-stubbed-for-a-faster-test

josephhull676 (approved):

PR #5512, opened 2022-10-21
[TEVA-4344] Fix bug allowing empty fields to be submitted

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4344

## Changes in this PR:

For some of the fields on this form (`skills_and_experience`, `school_offer` and `job_advert`), users can add bullet points. Because of the way we implemented this, when an empty field was submitted, the field's value was not blank - it was set to a string containing some html tags. We now remove these tags when validating to ensure that an empty field or a field only containing bullet points fails validation.

### DIFF
--- a/app/form_models/publishers/job_listing/about_the_role_form.rb
+++ b/app/form_models/publishers/job_listing/about_the_role_form.rb
@@ -1,8 +1,8 @@
 class Publishers::JobListing::AboutTheRoleForm < Publishers::JobListing::VacancyForm
   validates :ect_status, inclusion: { in: Vacancy.ect_statuses.keys }, if: -> { vacancy&.job_role == "teacher" }
-  validates :job_advert, presence: true, if: -> { vacancy.job_advert.present? }
-  validate :about_school_must_not_be_blank, if: -> { vacancy.job_advert.present? }
-  validates :skills_and_experience, presence: true, unless: -> { vacancy.job_advert.present? }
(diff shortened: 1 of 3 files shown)


app/form_models/publishers/job_listing/about_the_role_form.rb
```diff
@@ -41,34 +41,48 @@ def organisation_type
...
   def school_offer_presence
-    return if school_offer.present?
+    return if remove_html_tags(school_offer).present?
 
-    if organisation_type == "schools"
-      errors.add(:school_offer, I18n.t("about_the_role_errors.schools_offer.blank", organisation: organisation_type))
-    else
-      errors.add(:school_offer, I18n.t("about_the_role_errors.school_offer.blank", organisation: organisation_type))
-    end
+    errors.add(:school_offer, :blank, organisation: organisation_type)
```
robyoung91, line 46: I'm wondering if we can get rid of `organisation_type` for the error message - the label originally changed depending on the organisation type but we changed it back to say `What does your school offer?` for all orgs - maybe we should do the same with the error message.

robyoung91 (approved):

app/form_models/publishers/job_listing/about_the_role_form.rb
josephhull676: Something for a content designer perhaps. Will drop a message to a content designer to see what they think.

PR #5514, opened 2022-10-21
[TEVA-4469] fix styling of editor generated lists so they are exactly govuk like

alexbowen:
public view

Screenshot 2022-10-21 at 15 39 20

review view

Screenshot 2022-10-21 at 15 39 11


josephhull676 (approved):
Not sure what [class] is though lol

PR #5525, opened 2022-10-25
[TEVA-4452] Display contract_type with duration on review page

josephhull676:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4452

### DIFF
--- a/app/presenters/vacancy_presenter.rb
+++ b/app/presenters/vacancy_presenter.rb
@@ -68,13 +68,15 @@ class VacancyPresenter < BasePresenter
   end
 
   def contract_type_with_duration
-    type = model.contract_type ? I18n.t("helpers.label.publishers_job_listing_contract_type_form.contract_type_options.#{model.contract_type}") : nil
+    return nil unless model.contract_type.present?
+
     duration = if model.fixed_term?
-                 "(#{model.fixed_term_contract_duration})"
+                 model.fixed_term_contract_duration
                elsif model.parental_leave_cover?
-                 "(#{model.parental_leave_cover_contract_duration})"
+                 model.parental_leave_cover_contract_duration
                end
-    [type, duration].compact.join(" ")
+
+    [I18n.t("helpers.label.publishers_job_listing_contract_type_form.contract_type_options.#{model.contract_type}"), duration].compact.join(" - ")
   end
 
   def school_group_names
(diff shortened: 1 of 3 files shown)


bencmitchell:
Hi @josephhull676 
Looks good ✔️ 

On the review page and also on the job listing
image
image
image

app/presenters/vacancy_presenter.rb
```diff
@@ -68,13 +68,13 @@ def readable_subjects
...
     duration = if model.fixed_term?
-                 "(#{model.fixed_term_contract_duration})"
+                 model.fixed_term_contract_duration
                elsif model.parental_leave_cover?
-                 "(#{model.parental_leave_cover_contract_duration})"
+                 model.parental_leave_cover_contract_duration
                end
-    [type, duration].compact.join(" ")
+
+    [I18n.t("helpers.label.publishers_job_listing_contract_type_form.contract_type_options.#{model.contract_type}"), duration].compact.join(" - ")
```
BroiSatse, line 77: What will happen if contract_type is nil? I'd guess this would throw an exception / return the translation key depending on environment? Should we add `default: nil`

josephhull676: In both places where we call `contract_type_with_duration` (`app/views/publishers/vacancies/vacancy_review_sections/_job_details.html.slim` and 
`app/views/vacancies/listing/_job_details.html.slim`) we check that it's not nil first before calling it, so this shouldn't be a problem.

Perhaps for safety I could just add a guard clause?

josephhull676: Then again, just checked... If `contract_type` is nil, this line returns an empty string:

`[I18n.t("helpers.label.publishers_job_listing_contract_type_form.contract_type_options.#{model.contract_type}"), duration].compact.join(" - ")`

so I guess there's no need for the guard clause?

BroiSatse: Tight, so behaviour of I18n on missing translation depends on rails configuration and it can be different between environments. At the moment it seems that `I18n.t('some.non.existing.key')` returns "translation_missing: en.some.non.existing.key") which could be displayed to the user.

If we always have to use this with a guard close I think we should move the guard inside the method - this would keep us safe in the case we ever changed the configuration of i18n in the future.

BroiSatse (approved):

alexbowen (approved):

PR #5534, opened 2022-10-26
Downgrade Redis Gem

ddippolito:
Temporary until we find out why it's erroring


Gemfile
```diff
@@ -55,7 +55,7 @@ gem "rack-attack"
 gem "rack-cors"
 gem "rails_semantic_logger"
 gem "recaptcha"
-gem "redis"
+gem 'redis', '4.8.0'
```
josephhull676, line 58: Double quotes

josephhull676 (approved):

Gemfile
```diff
@@ -55,7 +55,7 @@ gem "rack-attack"
 gem "rack-cors"
 gem "rails_semantic_logger"
 gem "recaptcha"
-gem "redis"
```
BroiSatse, line 58: Should we add a comment why we fix it?

PR #5542, opened 2022-11-01
Teva 4405 create the school profile page

BroiSatse:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4405

### Before

image
image

### After

image
image

### DIFF
--- a/app/views/layouts/_header.html.slim
+++ b/app/views/layouts/_header.html.slim
@@ -3,9 +3,9 @@
   - if publisher_signed_in?
     = header.navigation_item text: t("nav.manage_jobs"), href: organisation_path, active: manage_jobs_active?
     - if current_organisation.school_group?
-      = header.navigation_item text: t("nav.manage_settings"), href: publishers_schools_path, active: schools_in_your_trust_active?
+      = header.navigation_item text: t("nav.organisation_profile"), href: publishers_schools_path, active: schools_in_your_trust_active?
     - else
-      = header.navigation_item text: t("nav.manage_settings"), href: publishers_school_path(current_organisation), active: schools_in_your_trust_active?
+      = header.navigation_item text: t("nav.school_profile"), href: publishers_school_path(current_organisation), active: schools_in_your_trust_active?
     = header.navigation_item text: t("nav.notifications_html", count: current_publisher.notifications.unread.count), href: publishers_notifications_path
     = header.navigation_item text: t("nav.school_hiring_guides"), href: posts_path(section: "get-help-hiring")
     = header.navigation_item text: t("nav.sign_out"), href: destroy_publisher_session_path, options: { method: :delete }
--- a/app/views/publishers/schools/_organisation.html.slim
+++ b/app/views/publishers/schools/_organisation.html.slim
@@ -1,48 +1,38 @@
 = govuk_summary_list do |summary_list|
   - summary_list.row do |row|
-    - row.key text: t("schools.description")
-    - row.value text: organisation.description? ? truncate(organisation.description, length: 130) : "Not provided"
-    - row.action text: t("buttons.change"), href: edit_publishers_school_path(organisation), classes: "govuk-link--no-visited-state"
+    - row.key text: t(".name.label")
+    - row.value text: organisation.name
 
   - unless organisation.local_authority?
     - summary_list.row do |row|
-      - row.key text: t("schools.address")
+      - row.key text: t(".address.label")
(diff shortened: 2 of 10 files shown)


georgeschena (approved):

app/views/layouts/_header.html.slim
```diff
@@ -3,9 +3,9 @@
   - if publisher_signed_in?
     = header.navigation_item text: t("nav.manage_jobs"), href: organisation_path, active: manage_jobs_active?
     - if current_organisation.school_group?
-      = header.navigation_item text: t("nav.manage_settings"), href: publishers_schools_path, active: schools_in_your_trust_active?
+      = header.navigation_item text: t("nav.organisation_profile"), href: publishers_schools_path, active: schools_in_your_trust_active?
     - else
-      = header.navigation_item text: t("nav.manage_settings"), href: publishers_school_path(current_organisation), active: schools_in_your_trust_active?
+      = header.navigation_item text: t("nav.school_profile"), href: publishers_school_path(current_organisation), active: schools_in_your_trust_active?
```
josephhull676, line 8: In my experience, we consider both Schools and SchoolGroups as Organisations. Perhaps a bit nit picky, but perhaps `nav.school_group_profile` is better or something like that?

app/views/publishers/schools/_organisation.html.slim
```diff
@@ -1,48 +1,38 @@
...
-      - row.key text: t("schools.urn")
-      - row.value text: organisation.urn
-  - elsif organisation.trust?
     - summary_list.row do |row|
-      - row.key text: t("school_groups.uid")
-      - row.value text: organisation.uid
+      - row.key text: t(".school_age.label")
+      - row.value text: age_range(organisation)
+
+    - if (size = organisation.gias_data["NumberOfPupils"].presence)
```
josephhull676, line 25: Assuming we can use `school_size(organisation)` here.

Can be found in:

`app/helpers/organisations_helper.rb`

josephhull676: Rather than not showing this row if we don't have this data, should we also show "No information available" as we do here if we don't have this info:

`app/views/vacancies/listing/_school.html.slim`

spec/system/publishers_can_manage_settings_spec.rb
```diff
@@ -1,15 +1,18 @@
 require "rails_helper"
 
-RSpec.describe "Publishers can manage settings" do
+RSpec.describe "Publishers can manage organization/school profile" do
```
josephhull676, line 3: Think we tend to use British English spelling 🇬🇧 😂

josephhull676 (approved):

app/views/layouts/_header.html.slim
BroiSatse: Yeah, I can see the language is a bit ambiguous here. The way I see it here is that school is a specialised type of organisation, so when there is only one organisation a user can interact with, then the link says School profile. Otherwise, it says Organisation profile. I am not certain if school_group would work here, as it also covers LAs.

app/views/publishers/schools/_organisation.html.slim
BroiSatse: Good question, probably need to confirm with prod - the ticket states not to display education phase if the information is missing, so I want with the same logic here.

Actually it seems there's even a better helper there - `number_of_pupils(school)` :)

spec/system/publishers_can_manage_settings_spec.rb
BroiSatse: Ah, you mean correct English :P Yeah, my RubyMine keeps suggesting US spellings...

app/views/layouts/_header.html.slim
josephhull676: I _think_ we consider LAs to be "school groups" as well.

josephhull676: But I guess if it says "Organisation profile", organisation is fine 😅

app/views/publishers/schools/_organisation.html.slim
BroiSatse: It is now confirm that we should not display any information that is not provided from gias and users can't add.

app/views/layouts/_header.html.slim
BroiSatse: I'll mark this as resolved for now, but we definitely need more conversation around the language. :)

PR #5547, opened 2022-11-02
Add DfE Analytics gem

ddippolito:
The Data & Insights team have recently released this new gem to streamline the way teams in Teacher Services sends data to the analytical platform in BigQuery. This work migrates our custom bespoke streaming code to the gem built by the data team.

## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4467

## How to review this PR?

It's fairly straight forward but if you're lost you can review commit by commit

### DIFF
--- a/app/models/location_polygon.rb
+++ b/app/models/location_polygon.rb
@@ -13,6 +13,8 @@ class LocationPolygon < ApplicationRecord
   end
 
   def self.include?(location)
+    return super unless location.is_a? String
+
     ALL_IMPORTED_LOCATIONS.include?(mapped_name(location))
   end
 
--- a/config/initializers/dfe_analytics.rb
+++ b/config/initializers/dfe_analytics.rb
@@ -1,0 +1,49 @@
+DfE::Analytics.configure do |config|
+  # Whether to log events instead of sending them to BigQuery.
+  #
+  config.log_only = false
+
+  # Whether to use ActiveJob or dispatch events immediately.
+  #
+  config.async = false
+
+  # Which ActiveJob queue to put events on
+  #
+  # config.queue = :default
+
+  # The name of the BigQuery table we’re writing to.
+  #
+  config.bigquery_table_name = ENV.fetch("BIGQUERY_TABLE_NAME", nil)
+
+  # The name of the BigQuery project we’re writing to.
+  #
+  config.bigquery_project_id = ENV.fetch("BIGQUERY_PROJECT_ID", nil)
+
+  # The name of the BigQuery dataset we're writing to.
+  #
+  config.bigquery_dataset = ENV.fetch("BIGQUERY_DATASET", nil)
+
+  # Service account JSON key for the BigQuery API. See
+  # https://cloud.google.com/bigquery/docs/authentication/service-account-file
+  #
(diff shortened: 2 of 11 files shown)


app/models/location_polygon.rb
```diff
@@ -13,6 +13,8 @@ def self.with_name(location)
   end
 
   def self.include?(location)
+    return super unless location.is_a? String
+
```
BroiSatse, line 17: So yeah, that is a bit hacky. Would be nice if we could rename this method so that we don't override basic Module interface...

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,49 @@
+DfE::Analytics.configure do |config|
+  # Whether to log events instead of sending them to BigQuery.
+  #
+  config.log_only = false
+
+  # Whether to use ActiveJob or dispatch events immediately.
+  #
+  config.async = false
```
BroiSatse, line 8: Should we enable this?

config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,49 @@
...
+  # A proc which returns true or false depending on whether you want to
+  # enable analytics. You might want to hook this up to a feature flag or
+  # environment variable.
+  #
+  # config.enable_analytics = proc { true }
+
+  # The environment we’re running in. This value will be attached
+  # to all events we send to BigQuery.
+  #
+  config.environment = ENV.fetch("RAILS_ENV", "development")
```
BroiSatse, line 48: Just wondering if we could do `Rails.env` here?

spec/jobs/alert_mailer_job_spec.rb
```diff
@@ -39,7 +39,7 @@
     allow_any_instance_of(AlertMailerJob).to receive(:provider_job_id) { job_id }
 
     allow_any_instance_of(AlertMailerJob).to receive(:subscription) { subscription }
-    expect(AlertMailerJob.queue_adapter).to receive(:enqueue).twice.ordered
+    expect(AlertMailerJob.queue_adapter).to receive(:enqueue).exactly(3).times.ordered
```
BroiSatse, line 42: Ok, how the changes here changes this test?

BroiSatse (approved):

spec/jobs/alert_mailer_job_spec.rb
ddippolito: [BroiSatse](https://github.com/BroiSatse) have a look at the commit message for this line of code 75e22db667f8c5ed5aae14e35f580c8c50363ff9 👀

config/initializers/dfe_analytics.rb
ddippolito: We should!

app/models/location_polygon.rb
ddippolito: Yes I was thinking about that too. I wanted to separate the PRs. But good point

ddippolito: @BroiSatse there are a few cases of included where I'm not sure if we're passing a polygon or not. It needs a bit of investigation. I'll raise a separate PR for it

PR #5584, opened 2022-11-16
Displaying profile incomplete banners

BroiSatse:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4507

### Before

image

### After

image


app/views/publishers/organisations/_organisation.html.slim
```diff
@@ -29,12 +29,17 @@
 
   - summary_list.row do |row|
     - row.key text: t(".website.label")
-    - row.value text: organisation.url ? open_in_new_tab_link_to(organisation.url, organisation.url, class: "wordwrap") : t(".not_provided")
+    - row.with_value do
+      - link = open_in_new_tab_link_to(organisation.url, organisation.url, class: "wordwrap") if organisation.url.present?
+      = required_profile_info(value: link, missing_prompt: t(".website.missing_prompt"), missing_text: t(".not_provided"))
+
+    / - row.value text: organisation.url ? open_in_new_tab_link_to(organisation.url, organisation.url, class: "wordwrap") : t(".not_provided")
```
alexbowen, line 36: commented code?

app/assets/stylesheets/base/govuk/components/_inset-text.scss
```diff
@@ -1,11 +1,30 @@
 .govuk-inset-text {
-  .govuk-summary-list__value & {
+  .govuk-summary-list__value &, .link-list & {
```
alexbowen, line 2: im not sure i see why you need to create this custom class `.link-list` ?

BroiSatse: There is another component .link-list which I've added in a previous PR - it is in fact quite similar to summary-list, but it is used for displaying links to the schools below the school group page. Without that change, schools with incomplete profiles would have quite a lot of margin around them within that list.

alexbowen: yeah i checked the review app and it seems just spacing styles, so can they be applied with govuk classes and then we dont need `link-list` ?

BroiSatse: `link-list` adds spacing and borders, it has some special treatment for the first child as well and different spacing rules for headings. However, it is possible that we could just use summary-list for it - I remember I've tried that and had some spacing issues, but I might be able to resolve them. I'll have a go then :)

alexbowen: its nice to use govuk classes and styles if possible. let me know if u need a hand https://design-system.service.gov.uk/

alexbowen (approved):

PR #5596, opened 2022-11-21
Hide list of schools on LA profile

josephhull676:
## Changes in this PR:

We now hide the list of associated schools from a hiring staff user on a school group's profile if the school group is a local authority. This is because all schools are associated with the LA they are within, geographically. This means that all local authority users will be able to change the profiles of all schools within their geographic area.

We are unsure of the policy implications for this, so, for now, have decided to take the ability to do this away whilst we explore this issue. This PR only hides the list of schools on a school group's profile page if that school group is an LA. 

We should probably prevent LAs from being able to reach the edit page full stop (although it's quite unlikely that they would visit it as they would need the direct URL). This will be easier as a result of changes I will (hopefully) shortly introduce with a PR for https://dfedigital.atlassian.net/browse/TEVA-4476

### DIFF
--- a/app/views/publishers/organisations/show.html.slim
+++ b/app/views/publishers/organisations/show.html.slim
@@ -15,13 +15,12 @@
 
     = render "organisation", organisation: @organisation
 
-    - if @organisation.school_group?
-
+    - if current_organisation.trust?
       h2.govuk-heading-m class="govuk-!-margin-top-8"
         = t(".schools")
 
       = govuk_summary_list do |summary_list|
-        - @organisation.schools.not_closed.order(:name).partition(&:profile_complete?).reverse.flatten.each do |school|
+        - current_organisation.schools.not_closed.order(:name).partition(&:profile_complete?).reverse.flatten.each do |school|
           - summary_list.row do |row|
             - row.with_value do
               - if school.profile_complete?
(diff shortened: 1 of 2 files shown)


app/views/publishers/organisations/show.html.slim
```diff
@@ -15,13 +15,12 @@
...
 
-    - if @organisation.school_group?
-
+    - if current_organisation.trust?
       h2.govuk-heading-m class="govuk-!-margin-top-8"
         = t(".schools")
 
       = govuk_summary_list do |summary_list|
-        - @organisation.schools.not_closed.order(:name).partition(&:profile_complete?).reverse.flatten.each do |school|
+        - current_organisation.schools.not_closed.order(:name).partition(&:profile_complete?).reverse.flatten.each do |school|
```
alexbowen, line 23: seems like complicated ruby to have in a view but realise was there already, so your call if you do anything about it!

alexbowen (approved):

app/views/publishers/organisations/show.html.slim
josephhull676: I'll have a think and discuss with people and if we want to move this into the controller or something we can do that in a later PR I guess!

PR #5601, opened 2022-11-23
remove axe a11y testing due to lack of value

alexbowen:
was added a few year back as an experiment and a bit of due diligence towards the a11y audit

removing as it has bought very little value to the project and is just extra code to maintain


.github/workflows/test.yml
```diff
@@ -31,11 +31,11 @@ jobs:
             name: system - support users
             params: 'spec/system/support_users'
           -
-            name: system - other & a11y
-            params: '--exclude-pattern "spec/system/{jobseekers,publishers,support_users}, spec/system/{jobseekers,publishers}_*_spec.rb" spec/system spec/accessibility'
+            name: system - other
+            params: '--exclude-pattern "spec/system/{jobseekers,publishers,support_users}, spec/system/{jobseekers,publishers}_*_spec.rb" spec/system'
```
josephhull676, line 35: I'm not entirely sure how this works, but I see 'exclude pattern', so I imagine that any test with a path matching the pattern given is not run.

Does this mean `unit` and `system - other` are both running our unit tests?

Gemfile.lock
```diff
@@ -124,9 +124,6 @@ GEM
     axe-core-api (4.4.1)
       dumb_delegator
       virtus
```
josephhull676, line 126: Seems like it hasn't removed this?

.github/workflows/test.yml
alexbowen: not entirely sure, i sort of search and replaced. ill see on this run of PR actions...

josephhull676 (approved):

PR #5617, opened 2022-11-28
Change how completed_steps are added for non-build steps

josephhull676:
In our job creation journey we have two steps that live outside of the build process - application_form and documents. After noticing that the application_form step was not being correctly added to completed_steps I decided to look into what was causing this issue. I noticed that, in the case of a step being outside of the build controller / process, completed_steps needed to be set in the params_to_save, rather than being merged into the params before updating the vacancy (see app/controllers/publishers/vacancies/build_controller.rb:52).

There is no need for application_form_params(params) in wizardable.rb as this is only used to construct / configure the params for forms submitted for steps within the build process / controller. In the case of the application form, the configuration of params is done in the application_forms_contoller.  For this reason, it has been removed. I then did something similar with the documents controller and form to make it consistent.

I also removed completed_steps from the params to save for the important_dates and working patterns steps forms as the completed steps are added to the params in wizardable.rb.

### DIFF
--- a/app/controllers/publishers/vacancies/documents_controller.rb
+++ b/app/controllers/publishers/vacancies/documents_controller.rb
@@ -10,7 +10,7 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
         send_event(:supporting_document_created, document.original_filename, document.size, document.content_type)
       end
 
-      vacancy.update(completed_steps: completed_steps)
+      vacancy.update(documents_form.params_to_save)
 
       render :index
     else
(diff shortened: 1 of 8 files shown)


app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -1,6 +1,8 @@
 require "google/apis/drive_v3"
 
 class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseController
+  include Publishers::Wizardable
```
robyoung91, line 4: I don't think we need this as the `Publishers::Vacancies::BaseController` includes `Publishers::Wizardable`

robyoung91 (approved):

PR #5635, opened 2022-12-02
Summarise failed vacancy imports

georgeschena:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4567

## Changes in this PR:

When a vacancy is attempted to be imported we now store it in a separate table if it fails, alongside this we store the source of the vacancy and the errors associated with it. 

This will now allow us to go back and look at why a vacancy has failed, and do checks against batches of failed imports, which we can then send back to the source and ask them to make adjustments to then reduce these errors going forward.

We also don't save duplicate failed vacancies, thus reducing duplicate failures.

### DIFF
--- a/app/jobs/import_from_vacancy_sources_job.rb
+++ b/app/jobs/import_from_vacancy_sources_job.rb
@@ -15,7 +15,7 @@ class ImportFromVacancySourcesJob < ApplicationJob
             Rails.logger.info("Imported vacancy #{vacancy.id} from feed #{source_klass.source_name}")
           else
             report_validation_errors(source_klass, vacancy)
-            FailedImportedVacancy.create(source: source_klass.source_name, external_reference: vacancy.external_reference, import_errors: vacancy.errors.to_json)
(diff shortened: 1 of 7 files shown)


josephhull676 (approved):
This may be based on my ignorance of how the vacancy import feature works (and perhaps what you've done here - let me know 😅) but I wonder about this scenario:

If we receive a bunch of vacancies from a source one day and they all fail, for each vacancy that fails we store a record in this new table.

We send an email (as described in the ticket) to someone on the team with all of the vacancies that failed to import on that day. They then contact the source to let them know so they can fix them. 

They fix them and they all successfully import the next day.

Then the next day we receive a new bunch of vacancies that all fail to import.

We then need to send an email to someone in the team with all the failed vacancies so they can speak to the source again. 

How do we differentiate between these new failures and old failures? 

Could we have some sort of batch_id or perhaps we could remove vacancies that have been successfully imported at a later date?

ddippolito (approved):
Great!! 🚀

PR #5655, opened 2022-12-07
Fix flaky spec

robyoung91:


spec/jobs/import_from_vacancy_sources_job_spec.rb
```diff
@@ -133,7 +133,7 @@ def each(...)
           "slug" => "mallowpond-high-school",
           "start_date_type" => "specific_date",
           "starts_asap" => nil,
-          "starts_on" => "2023-12-06",
+          "starts_on" => Date.today,
```
BroiSatse, line 136: We're changing string to date here inside `eq`. How is it not breaking the test?

alexbowen (approved):

PR #5663, opened 2022-12-12
[TEVA-4539] Add safeguarding information to school profiles

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4539

## Changes in this PR:

This PR adds the `safeguarding_information` to the Organisation model, and allows users to update this field on the school/organisation "manage profile" page. The safeguarding information is then displayed on the school profile page. School groups can also update their schools' safeguarding information via the manage profile page.

### Before
"Manage profile"
image

"Preview profile"
image

"School profile (jobseeker perspective)"
image

### After
"Manage profile (before adding safeguarding info)"
image

"Safeguarding information form"
image

"Manage profile (after adding safeguarding info)"
image

"Preview profile"
image

"School profile (jobseeker perspective)"
image

### DIFF
--- a/app/controllers/publishers/organisations/safeguarding_information_controller.rb
+++ b/app/controllers/publishers/organisations/safeguarding_information_controller.rb
@@ -1,0 +1,26 @@
+class Publishers::Organisations::SafeguardingInformationController < Publishers::OrganisationsController
+  def edit
+    organisation
+    @safeguarding_information_form = Publishers::Organisation::SafeguardingInformationForm.new(safeguarding_information: organisation.safeguarding_information)
+  end
+
+  def update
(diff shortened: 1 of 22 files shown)


alexbowen:
This Pr looks very familiar - probably why there are a load on conflicts now!

alexbowen (approved):

bencmitchell:
Hi @robyoung91 
This might not have been explicit on the ticket, but with these new fields we'll need to update the check we run to determine if a profile is complete or not. This should trigger both the interstitial page and also the banner that we show when the profile is incomplete.
image

The page itself we need to update the styling on so that it matches that of this page on the proto (there'll be a separate ticket to bring the other pages around to this style)
If you use the proto page here as a guide, but with "Commitment to safeguarding" in place of "Organisation description" please that would be great.
image

PR #5681, opened 2022-12-29
[TEVA-4591] Iterate on schools name search

robyoung91:
## Jira ticket URL

https://dfedigital.atlassian.net/browse/TEVA-4591

## Changes in this PR:

- Adds the `pg_search` gem which allows us to use the postgres full text search feature for Organisations. A new `tsvector` column (and index), `searchable_content`, containing the parsed organisation names is added the organisations table (and updated when the organisation record is saved).

## Next steps:

- [ ] Run `Organisation.update_all_searchable_content!` on production and other environments to populate the `searchable_content` column for existing organisations.

[TEVA-4591]: https://dfedigital.atlassian.net/browse/TEVA-4591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


josephhull676 (commented):
Looking at the ticket I see: 

> Iterate the search so that searching for a school’s name with multiple words splits the words and searches for each word anywhere within the organisation name

So locally (without these changes) I searched "bexley" to see if it would find Bexleyheath Academy and it did. Do you have an example search I could use?

Also I see we've followed a different pattern to what we've done with `searchable_content` and full text searches on vacancies. I'd imagine this was to save time so we could get this out, but perhaps it'd be worth a refactor at some point. Perhaps we could make `app/models/concerns/database_indexable.rb` more abstract.

robyoung91:
> Looking at the ticket I see:
> 
> > Iterate the search so that searching for a school’s name with multiple words splits the words and searches for each word anywhere within the organisation name
> 
> So locally (without these changes) I searched "bexley" to see if it would find Bexleyheath Academy and it did. Do you have an example search I could use?

So the current implementation is quite crude and just checks the entire search string against the organisation names with `scope.where("organisations.name ILIKE ?", "%#{name}%")` - this means that, for example, `weydon multi academy trust` (or any substring of that) would match the organisation Weydon Multi Academy Trust, but `weydon trust` wouldn't.

> Also I see we've followed a different pattern to what we've done with `searchable_content` and full text searches on vacancies. I'd imagine this was to save time so we could get this out, but perhaps it'd be worth a refactor at some point. Perhaps we could make `app/models/concerns/database_indexable.rb` more abstract.

Yeah I think it's definitely worth having a go at making database_indexable more abstract - we may want to define `generate_searchable_content` on the organisation and vacancy models but the other two public methods I think should be straightforward. 

I decided to take a simpler approach to implementing full text search for organisations by using the `pg_search` gem for organisations as we don't require as much flexibility as the vacancy keyword searching (custom synonyms etc).

josephhull676:
> > Looking at the ticket I see:
> > > Iterate the search so that searching for a school’s name with multiple words splits the words and searches for each word anywhere within the organisation name
> > 
> > 
> > So locally (without these changes) I searched "bexley" to see if it would find Bexleyheath Academy and it did. Do you have an example search I could use?
> 
> So the current implementation is quite crude and just checks the entire search string against the organisation names with `scope.where("organisations.name ILIKE ?", "%#{name}%")` - this means that, for example, `weydon multi academy trust` (or any substring of that) would match the organisation Weydon Multi Academy Trust, but `weydon trust` wouldn't.
> 
> > Also I see we've followed a different pattern to what we've done with `searchable_content` and full text searches on vacancies. I'd imagine this was to save time so we could get this out, but perhaps it'd be worth a refactor at some point. Perhaps we could make `app/models/concerns/database_indexable.rb` more abstract.
> 
> Yeah I think it's definitely worth having a go at making database_indexable more abstract - we may want to define `generate_searchable_content` on the organisation and vacancy models but the other two public methods I think should be straightforward.
> 
> I decided to take a simpler approach to implementing full text search for organisations by using the `pg_search` gem for organisations as we don't require as much flexibility as the vacancy keyword searching (custom synonyms etc).

Sweet 👍 

Perhaps we can both have a go at abstracting it at some point!

josephhull676:
> > Looking at the ticket I see:
> > > Iterate the search so that searching for a school’s name with multiple words splits the words and searches for each word anywhere within the organisation name
> > 
> > 
> > So locally (without these changes) I searched "bexley" to see if it would find Bexleyheath Academy and it did. Do you have an example search I could use?
> 
> So the current implementation is quite crude and just checks the entire search string against the organisation names with `scope.where("organisations.name ILIKE ?", "%#{name}%")` - this means that, for example, `weydon multi academy trust` (or any substring of that) would match the organisation Weydon Multi Academy Trust, but `weydon trust` wouldn't.
> 
> > Also I see we've followed a different pattern to what we've done with `searchable_content` and full text searches on vacancies. I'd imagine this was to save time so we could get this out, but perhaps it'd be worth a refactor at some point. Perhaps we could make `app/models/concerns/database_indexable.rb` more abstract.
> 
> Yeah I think it's definitely worth having a go at making database_indexable more abstract - we may want to define `generate_searchable_content` on the organisation and vacancy models but the other two public methods I think should be straightforward.
> 
> I decided to take a simpler approach to implementing full text search for organisations by using the `pg_search` gem for organisations as we don't require as much flexibility as the vacancy keyword searching (custom synonyms etc).

Can see there's a conflict but I'll approve 🙂

josephhull676 (approved):

PR #5689, opened 2023-01-04
Create the candidate 'Your profile' page

georgeschena:
## Jira ticket & Trello URL

Jira: https://dfedigital.atlassian.net/browse/TEVA-4434
Trello: https://trello.com/c/mml6ix4X/30-create-the-candidate-your-profile-page

## Changes in this PR:

- Created the “Your profile” page 
- The section headings are displayed but not including any of the links to each section as part of this ticket. Essentially the bare bones to then create the connecting pages.

**Once the other pages are added you will need to update the `page_path` key with the url to that new page**

## Screenshots of UI changes:

Screenshot 2023-01-04 at 09 26 02

[TEVA-4434]: https://dfedigital.atlassian.net/browse/TEVA-4434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/jobseekers/profile/index.html.slim
+++ b/app/views/jobseekers/profile/index.html.slim
@@ -7,6 +7,6 @@
 
     - @sections.each do |section|
         = tag.h2(section[:title], class: "govuk-heading-m govuk-!-font-size-27 govuk-!-padding-top-6")
-        p = govuk_link_to section[:link_text], section[:page_path], class: "govuk-!-padding-bottom-6"
+        p = govuk_link_to section[:link_text], instance_exec(&section[:page_path]), class: "govuk-!-padding-bottom-6"
 
     p = t(".activate_profile_text")
--- a/app/views/jobseekers/profile/index.html.slim
+++ b/app/views/jobseekers/profile/index.html.slim
@@ -1,0 +1,12 @@
+- content_for :page_title_prefix, t(".page_title")
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl = t(".page_title")
+    p = t(".page_description")
+
+    - @sections.each do |section|
+        = tag.h2(section[:title], class: "govuk-heading-m govuk-!-font-size-27 govuk-!-padding-top-6")
+        p = govuk_link_to section[:link_text], section[:page_path], class: "govuk-!-padding-bottom-6"
+
+    p = t(".activate_profile_text")
--- a/app/views/jobseekers/profiles/show.html.slim
+++ b/app/views/jobseekers/profiles/show.html.slim
@@ -6,7 +6,8 @@
     p = t(".page_description")
 
     - @sections.each do |section|
-        = tag.h2(section[:title], class: "govuk-heading-m govuk-!-font-size-27 govuk-!-padding-top-6")
+        h2.govuk-heading-m class="govuk-!-font-size-27 govuk-!-padding-top-6"
+          = section[:title]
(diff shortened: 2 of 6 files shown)


BroiSatse:
One potential issue is that the `page_path` will be dynamic depending on the user (I assume) or at least **might** be dynamic. We could convert them into a proc then!

app/views/jobseekers/profile/index.html.slim
```diff
@@ -0,0 +1,12 @@
+- content_for :page_title_prefix, t(".page_title")
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl = t(".page_title")
+    p = t(".page_description")
+
+    - @sections.each do |section|
+        = tag.h2(section[:title], class: "govuk-heading-m govuk-!-font-size-27 govuk-!-padding-top-6")
```
josephhull676, line 9: No need to use the tag helper here I think. Could do:

```
h2.govuk-heading-m class="govuk-!-font-size-27 govuk-!-padding-top-6"
  = section[:title]
```

josephhull676: Is what we do elsewhere

georgeschena: Nice!

josephhull676:
> One potential issue is that the `page_path` will be dynamic depending on the user (I assume) or at least **might** be dynamic. Also, I don't think url helpers are avialable in the context of the controller class (only in the instance). We should then accept page_path as a proc to evaluate in the correct context and pass the user?

I guess it depends on whether we want to use this constant to generate these links in the future. My guess would be that we're not going to and that this is just been added to get this done so we can build this feature out from this page?

config/routes.rb
```diff
@@ -78,6 +78,8 @@
       end
     end
 
+    resources :profile, only: %i[index]
```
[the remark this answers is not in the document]
georgeschena: Its a weird one, I did think about this. I guess if we look at the jobseekers account page thats a show route, so maybe using show makes more sense here

BroiSatse (approved):

config/routes.rb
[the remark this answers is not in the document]
BroiSatse: Another option is to use `resource` rather than `resources` - so no index, just show that doesn't require id in param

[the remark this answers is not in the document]
josephhull676: Why would this be preferable?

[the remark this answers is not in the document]
georgeschena: Yeah I am making the change so its `resource :profile, only: %i[show]` we do this for `resource :account, only: %i[show]` anyway

app/views/jobseekers/profiles/show.html.slim
```diff
@@ -0,0 +1,13 @@
+- content_for :page_title_prefix, t(".page_title")
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl = t(".page_title")
+    p = t(".page_description")
+
+    - @sections.each do |section|
+        h2.govuk-heading-m class="govuk-!-font-size-27 govuk-!-padding-top-6"
```
alexbowen, line 9: seems a bit odd the have `govuk-!-font-size-27` class when `govuk-heading-m` is size 24 and the standard govuk size?

georgeschena: Good spot

PR #5715, opened 2023-01-16
Remove call to `@organisation.trust` in templates

robyoung91:
## Changes in this PR:

When we call the `trust` method, the `@organisation` object is a SchoolGroup, so we are in fact calling `Kernel#trust` rather than a method we have defined (like on the School model). `Kernel#trust` is removed in Ruby 3.2.0 and so will throw an error when we upgrade.

### DIFF
--- a/app/views/publishers/organisations/schools/preview.html.slim
+++ b/app/views/publishers/organisations/schools/preview.html.slim
@@ -1,4 +1,4 @@
-- content_for :page_title_prefix, t(".page_title", school_group_name: @organisation.trust.name)
+- content_for :page_title_prefix, t(".page_title", school_group_name: @organisation.name)
 
 - content_for :breadcrumbs do
   = govuk_back_link text: t(".exit_preview_link_text"), href: publishers_organisation_path(@organisation)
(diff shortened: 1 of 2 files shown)


app/views/publishers/organisations/schools/preview.html.slim
```diff
@@ -1,4 +1,4 @@
-- content_for :page_title_prefix, t(".page_title", school_group_name: @organisation.trust.name)
+- content_for :page_title_prefix, t(".page_title", school_group_name: @organisation.name)
```
josephhull676, line 1: Hmm, I wonder here... @organisation will be the trust (look at `app/controllers/publishers/organisations_controller.rb`), rather than the school who's profile you're previewing, right? 

Probably best to have the page title include the name of the school rather than the trust, no?

Perhaps we need to grab the school here: `app/controllers/publishers/organisations/schools_controller.rb` and pass the school to the view and use it instead?

robyoung91: @josephhull676 I think we use the very similarly named `app/views/publishers/organisations/preview.html.slim` for individual school profile previews, and the template I've changed in this PR is just for school groups.

josephhull676: Ahhh yeah sorry I'm thinking about the organisation school's index page. In that case, you'd want the trust's name in the title. Cool, approved!

josephhull676 (approved):

PR #5727, opened 2023-01-20
Test vacancy document downloaded events

ddippolito:


app/controllers/documents_controller.rb
```diff
@@ -33,4 +50,8 @@ def send_event
                             filename: file.filename)
     end
   end
+
+  def current_user
```
alexbowen, line 54: whats this for? i cant see it being used above?

alexbowen (approved):

PR #5737, opened 2023-01-24
Add Every vacancy source integration

ddippolito:
No tests is required as it's a simple copy of previous ATS integration. 
At this point It might worth considering ways to DRY up the integration code.

### DIFF
--- a/app/vacancy_sources/every_vacancy_source.rb
+++ b/app/vacancy_sources/every_vacancy_source.rb
@@ -102,12 +102,12 @@ class EveryVacancySource
     raise HTTParty::ResponseError, error_message unless response.success?
 
     parsed_response = JSON.parse(response.body)
-    raise FusionImportError, error_message if parsed_response["error"]
+    raise EveryImportError, error_message + parsed_response["error"] if parsed_response["error"]
 
     parsed_response
   end
 
   def error_message
-    "Something went wrong with Fusion Import"
+    "Something went wrong with Fusion Import. Response:"
   end
 end
--- a/app/vacancy_sources/every_vacancy_source.rb
+++ b/app/vacancy_sources/every_vacancy_source.rb
@@ -1,0 +1,105 @@
+class EveryVacancySource
+  FEED_URL = ENV.fetch("VACANCY_SOURCE_EVERY_FEED_URL").freeze
+  SOURCE_NAME = "every".freeze
+
+  class EveryImportError < StandardError; end
+
+  include Enumerable
+
+  def self.source_name
+    SOURCE_NAME
+  end
+
+  def each
+    results.each do |result|
+      v = Vacancy.find_or_initialize_by(
+        external_source: SOURCE_NAME,
+        external_reference: result["reference"],
+      )
+
+      # An external vacancy is by definition always published
+      v.status = :published
+      # Consider publish_on date to be the first time we saw this vacancy come through
+      # (i.e. today, unless it already has a publish on date set)
+      v.publish_on ||= Date.today
+
+      begin
+        v.assign_attributes(attributes_for(result))
+      rescue ArgumentError => e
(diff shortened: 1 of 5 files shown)


app/vacancy_sources/every_vacancy_source.rb
```diff
@@ -0,0 +1,113 @@
...
+  def results
+    feed["result"]
+  end
+
+  def feed
+    response = HTTParty.get(FEED_URL)
+    raise HTTParty::ResponseError, error_message unless response.success?
+
+    parsed_response = JSON.parse(response.body)
+    raise FusionImportError, error_message if parsed_response["error"]
```
BroiSatse, line 105: Should we add error `parsed_response['error']`to the message?

app/vacancy_sources/every_vacancy_source.rb
```diff
@@ -0,0 +1,113 @@
...
+    item["ect_suitable"] == "yes" ? "ect_suitable" : "ect_unsuitable"
+  end
+
+  def results
+    feed["result"]
+  end
+
+  def feed
+    response = HTTParty.get(FEED_URL)
+    raise HTTParty::ResponseError, error_message unless response.success?
```
BroiSatse, line 102: Would be great to have some more details to the exception here.

app/vacancy_sources/every_vacancy_source.rb
```diff
@@ -0,0 +1,113 @@
...
+  def job_role(item)
+    item["jobRole"].presence
+      &.gsub("headteacher", "senior_leader")
+      &.gsub("head_of_year", "middle_leader")
+      &.gsub("learning_support", "education_support")
+      &.gsub(/\s+/, "")
+  end
+
+  def ect_status_for(item)
+    return unless item["ect_suitable"].presence
```
BroiSatse, line 91: `return if ...blank?`

BroiSatse, line 91: Also double checking - most of the keys in item are camelCase, this is snakecase. Is this correct?

app/vacancy_sources/every_vacancy_source.rb
```diff
@@ -0,0 +1,113 @@
...
+      v.status = :published
+      # Consider publish_on date to be the first time we saw this vacancy come through
+      # (i.e. today, unless it already has a publish on date set)
+      v.publish_on ||= Date.today
+
+      begin
+        v.assign_attributes(attributes_for(result))
+      rescue ArgumentError => e
+        v.errors.add(:base, e)
+      end
```
BroiSatse, line 30: Wouldn't we rather see an exception here so we can fix it?

BroiSatse (approved):

app/vacancy_sources/every_vacancy_source.rb
ddippolito: Normally yes, but error handling and reporting is handled higher up here:
That vacancy will be yielded back to the caller which will report on it based on its validity.

https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/jobs/import_from_vacancy_sources_job.rb#L18

app/vacancy_sources/every_vacancy_source.rb
ddippolito: Good spot.

ddippolito: What if `item["ect_suitable"]` is nil. Wouldn't calling `.blank?` on it fail?

ddippolito: What's the difference? Is it because `if` reads better than `unless` ?

app/vacancy_sources/every_vacancy_source.rb
ddippolito: We can

PR #5764, opened 2023-02-07
[TRELLO-115] add QTS update section to jobseeker profiles

alexbowen:
Screenshot 2023-02-08 at 12 07 28
Screenshot 2023-02-08 at 12 07 21
Screenshot 2023-02-08 at 12 07 13


georgeschena (approved):

bencmitchell:
@alexbowen the review page doesn't currently match the prototype designs,  where the "Year QTS awarded" is in a separate field when present.
This is also displayed this way on the Profile page itself

image
image

alexbowen:
> @alexbowen the review page doesn't currently match the prototype designs, where the "Year QTS awarded" is in a separate field when present. This is also displayed this way on the Profile page itself

whoops, wont be difficult to regig

app/controllers/jobseekers/profiles_controller.rb
```diff
@@ -4,31 +4,31 @@ class Jobseekers::ProfilesController < Jobseekers::BaseController
...
       link_text: "Add personal details",
-      page_path: -> { "" },
+      page_path: -> { "/" },
     },
     {
       title: "Job preferences",
       key: "",
       link_text: "Add job preferences",
-      page_path: -> { "" },
+      page_path: -> { "/" },
```
josephhull676, line 13: Why this change sorry?

app/controllers/jobseekers/profiles/qualified_teaching_status_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Jobseekers::Profiles::QualifiedTeachingStatusController < Jobseekers::ProfilesController
+  def edit
+    @profile = JobseekerProfile.find_by(jobseeker_id: current_jobseeker.id)
```
josephhull676, line 3: Perhaps we should set @profile in the profiles controller and just use it in this controller.

In the profiles controller you'd have something like:
```

  private

  def profile
    @profile ||= JobseekerProfile.find_by(jobseeker_id: current_jobseeker.id)
  end
```

Then you can just use `profile` in the view and in the controller.

app/form_models/jobseekers/profile/qualified_teaching_status_form.rb
```diff
@@ -0,0 +1,10 @@
+class Jobseekers::Profile::QualifiedTeachingStatusForm
+  include ActiveModel::Model
```
josephhull676, line 2: Don't need to include this module if you inherit from BaseForm I think

app/views/jobseekers/profiles/about_you/show.html.slim
```diff
@@ -7,10 +7,6 @@
   .govuk-grid-column-two-thirds
     h1.govuk-heading-l = t(".page_title")
 
-    = govuk_summary_list do |summary_list|
-      - summary_list.row do |row|
-        - row.key text: t(".summary_title")
-        - row.value text: @profile.about_you
-        - row.action(text: t("buttons.change"), href: edit_jobseekers_profile_about_you_path)
+    == render(partial: "summary", locals: { profile: @profile, title: t(".summary_title") })
```
josephhull676, line 10: Why the need for a partial here?

app/views/jobseekers/profiles/qualified_teaching_status/edit.html.slim
```diff
@@ -0,0 +1,23 @@
...
+  .govuk-grid-column-two-thirds
+    = form_for @form, url: jobseekers_profile_qualified_teaching_status_path, method: "patch" do |f|
+      = f.govuk_error_summary
+
+      span.govuk-caption-l = t(".page_title")
+
+      = f.govuk_radio_buttons_fieldset :qualified_teaching_status, legend: { size: "l", tag: "h1" } do
+        = f.govuk_radio_button :qualified_teaching_status, "yes", link_errors: true
+          = f.govuk_text_field :qualified_teaching_status_year,
+            label: { text: "Year QTS awarded", class: "govuk-label--s" }
```
josephhull676, line 16: Think you can add this label to `activerecord.yml` instead. That's the pattern we follow elsewhere.

db/migrate/20230201151352_add_qualified_teaching_status_to_jobseeker_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddQualifiedTeachingStatusToJobseekerProfiles < ActiveRecord::Migration[7.0]
+  def change
+    add_column :jobseeker_profiles, :qualified_teaching_status, :string
```
josephhull676, line 3: I wonder if this should be an enum instead?

To do so, in the JobseekerProfile model you'd have something like:

`enum qualified_teacher_status: { yes: 0, no: 1, on_track: 2 }`

Then, the migration will look like this:

`add_column :vacancies, :qualified_teacher_status, :integer`

app/views/jobseekers/profiles/about_you/show.html.slim
alexbowen: so it can be used on review page and the main profile index page

app/controllers/jobseekers/profiles_controller.rb
alexbowen: i dunno, was from early on in coding task i think i change back to original

bencmitchell:
@alexbowen very minor, but this should be capitalised, this is the same with "no" and occurs on all pages currently. 
image

In the "On track" scenario, this should read "I'm on track to receive QTS" on the review and profile page 
image
****

app/form_models/jobseekers/profile/qualified_teaching_status_form.rb
alexbowen: changed about you form too

PR #5779, opened 2023-02-08
Jobseeker profile: Job preferences

BroiSatse:
Job Prefererence journey completed!

Things that are missing:
- More tests - feature tests are coming
- Proper translations

Questions:
- Locations are currently stored as a json array within the same model. This will eventually make it more difficult to use for searching, should we extract them to a separate table?


ddippolito:
As discussed offline. Happy to approve this PR.

As a follow up tasks we'll move away from having locations stored as json and move to relational db table instead.
CC: @georgeschena

ddippolito (approved):

alexbowen:
There are no unit tests for the multistep form library? it might be implicitly tested to a degree in a system spec but that is specific to job prefenreces and as is shared library code explicit tests needed surely?

alexbowen:
all these commits need squashing or fixing up too

BroiSatse:
>  all these commits need squashing or fixing up too

Why? The commits as written contain a story of how this was developed, squashing will result in one massive chunk of changes with no history. There's very little benefit in maintaining linear history and quite a lot of problems with it.

BroiSatse:
> There are no unit tests for the multistep form library

@alexbowen Yes, they definitely need to be added. The library is currently in a "proof of concept" phase, meaning that it's API can change diametrically at any point in which case all the unit tests would only slow the changes down. Once we have a few journeys using it and it's API becomes somewhat more stable, then it might be worth extracting it to a separate repository and thoroughly test it.

alexbowen:
> [@BroiSatse ](/BroiSatse)
> ` [rubocop](/DFE-Digital/teaching-vacancies/pull/5779/commits/0e19d3c4889f6d13147e2442aea03b12a13a4417)`

ok maybe all commits was an overstatement but personally i think commits like 'fix rubocop' are unhelpful in their own right

app/views/jobseekers/profiles/job_preferences/_summary.html.slim
```diff
@@ -0,0 +1,58 @@
...
+      join_with: ", ",
+    },
+    locations: {
+      if: ->(form) { form.locations.any? },
+      map_values: ->(value, _) { "#{value[:location]} (#{t('jobs.search.number_of_miles', count: value[:radius].to_i)})" },
+      join_with: "</br>",
+    },
+  }
+
+  list = govuk_summary_list do |summary_list|
```
alexbowen, line 30: from here down doesnt seem like the sort of thing that should be happening in a view template but rather extracted into a helper or similar. there is too much logic constucting the summary list

app/views/jobseekers/profiles/job_preferences/_summary.html.slim
```diff
@@ -0,0 +1,58 @@
+ruby:
+  multistep = Jobseekers::JobPreferencesForm.from_record(profile.job_preferences)
```
alexbowen, line 2: instantiating a class in a view also??

BroiSatse: @alexbowen - Yes, this view, as written, is very closely related to the particular implementation of the form and it is much healthier to instantiate it here rather than in a controller - especially since this is a partial which can be reused in any place without any need to modify the controller.

Once this is extracted as an abstract component, the form would be still instantiated in the partial to be passed to the component.

app/views/jobseekers/profiles/job_preferences/_summary.html.slim
BroiSatse: Yes, I'm planning on moving it into a component at some point. Helpers are evil

BroiSatse:
@alexbowen 
> and there shouldnt be merge commits

Well, that's a discussion I'm having quite a lot! :) There's really no benefit of maintaining linear, merge-free git history other than aesthetic and there are a lot of drawbacks - to name one, team enforcing linear history cannot safely branch off any other branch than master, which is a major issue collaboration-wise.

Let's take the Multistep branch which was extracted into a separate PR from this branch. AFAIK, it served as a base for 2 other branched. If at any point I decided to alter the git history, those other branched would be in trouble. If one of the other branches were squash merged, the remaining two people would have a fun series of conflicts to resolve, with commit duplicates in the history assigned to me but created by someone else.

As a developer I cannot know who is using commits that I've pushed to a remote, which is way I try to avoid any history changing pushes. Naturally, there are exceptions, but there should never be any rigid rule in a form of "never do xxx"

PR #5785, opened 2023-02-10
Reduce numbers of events sent to Big Query

ddippolito:


app/controllers/api/events_controller.rb
```diff
@@ -30,4 +31,16 @@ def frontend_event
   def bad_request
     head(:bad_request)
   end
+
+  def send_dfe_analytics_event(type, data)
+    fail_safe do
+      event = DfE::Analytics::Event.new
+        .with_type(type)
```
asatwal, line 38: We should add `with_user` here if it is available.

app/controllers/authentication_controller.rb
```diff
@@ -25,6 +25,19 @@ def clear_extra_publisher_session_entries
   end
 
   def trigger_jobseeker_sign_in_event(success_or_failure, errors = nil)
+    event = DfE::Analytics::Event.new
+      .with_type(:jobseeker_sign_in_attempt)
+      .with_request_details(request)
+      .with_response_details(response)
+      .with_user(current_jobseeker)
+      .with_data(
+        email_identifier: StringAnonymiser.new(params.dig(:jobseeker, :email)),
```
asatwal, line 34: Now that the DfE Analytics GEM is being used, if the email is in analytics_pii.yml, then we should use the DfE:Analytics anonymiser ie:
```Ruby
DfE::Analytics.anonymise(params.dig(:jobseeker, :email)) 
```

Otherwise the GEM anonymiser is likely to use a different hashing algorithm from a custom anonymiser, and joining up this value from database tables will not work as expected.

This is a general comment that is applicable throughout this PR.

app/controllers/authentication_controller.rb
```diff
@@ -34,6 +47,15 @@ def trigger_jobseeker_sign_in_event(success_or_failure, errors = nil)
   end
 
   def trigger_successful_publisher_sign_in_event(sign_in_type, publisher_oid = nil)
+    event = DfE::Analytics::Event.new
+      .with_type(:successful_publisher_sign_in_attempt)
+      .with_request_details(request)
+      .with_response_details(response)
+      .with_user(current_publisher)
+      .with_data(user_anonymised_publisher_id: StringAnonymiser.new(publisher_oid), sign_in_type: sign_in_type)
```
asatwal, line 55: Not sure if adding `user_anonymised_publisher_id` is required, as the `with_user` method can now be configured to anonymise the user identifier. Please see: https://github.com/DFE-Digital/dfe-analytics/pull/64

Also we must use either `id` or `oid` consistently for any given entity. For example, if the `analytics.yml` and `analytics_pii.yml` specify `publisher_id` the in the user_id field we must use `publisher.id`.


This is a general comment that is applicable throughout this PR.

app/controllers/authentication_controller.rb
```diff
@@ -50,10 +72,23 @@ def trigger_successful_support_user_sign_in_event(sign_in_type, oid = nil)
   end
 
   def trigger_failed_dsi_sign_in_event(sign_in_type, oid = nil)
+    event = DfE::Analytics::Event.new
+      .with_type(:failed_dsi_sign_in_attempt)
+      .with_request_details(request)
+      .with_response_details(response)
+      .with_user(current_user)
+      .with_data(user_anonymised_id: StringAnonymiser.new(oid), sign_in_type: sign_in_type)
```
asatwal, line 80: Please see comment about `id`, `oid` and `DfE:Analytics.anonymiser()` above.

app/controllers/authentication_controller.rb
```diff
@@ -50,10 +72,23 @@ def trigger_successful_support_user_sign_in_event(sign_in_type, oid = nil)
...
+
+    DfE::Analytics::SendEvents.do([event])
+
     request_event.trigger(
       :failed_dsi_sign_in_attempt,
       user_anonymised_id: StringAnonymiser.new(oid),
```
asatwal, line 86: Not sure if field `user_anonymised_id` is required and please see comment about id, oid and DfE:Analytics.anonymiser() above.

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -167,4 +168,17 @@ def vacancy
...
   end
+
+  def send_dfe_analytics_event
+    fail_safe do
+      event = DfE::Analytics::Event.new
+        .with_type(:vacancy_apply_clicked)
+        .with_request_details(request)
+        .with_response_details(response)
+        .with_user(current_jobseeker)
+        .with_data(vacancy_id: StringAnonymiser.new(vacancy.id))
```
asatwal, line 179: Not really sure if `vacancy_id` should be anonymised - Only need to do this if it is PII.

app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -72,6 +74,25 @@ def after_document_delete_path
...
 
+  def send_dfe_analytics_event(event_type, name, size, content_type)
+    fail_safe do
+      event = DfE::Analytics::Event.new
+        .with_type(event_type)
+        .with_request_details(request)
+        .with_response_details(response)
+        .with_user(current_publisher)
+        .with_data(
+          vacancy_id: StringAnonymiser.new(vacancy.id),
```
asatwal, line 85: Not really sure if vacancy_id should be anonymised - Only need to do this if it is PII.

asatwal (commented):
Please see comments below. Main points are:

 1. The identifier used must be consistent, ie for any given entity we should use either `id` or `oid`. For example if `analytics.yml` or `analytics_pii.yml` use `publisher_id`, then the user identifier used in the `user_id` field must be `publisher.id`
 2. Now that the DfE Analytics GEM is being used, any field that is anonymised should use the DfE:Analytics anonymiser ie: `DfE::Analytics.anonymise()`. Otherwise the GEM anonymiser is likely to use a different hashing algorithm from a custom anonymiser, and joining up this value from database tables will not work as expected.

PR #5800, opened 2023-02-15
[TRELLO-75] Add work history section of jobseeker profile

josephhull676:
## Jira ticket URL

https://trello.com/c/eb2BUOVG/75-create-the-work-history-section-of-the-candidate-profile

## Changes in this PR:

- Added UI to add, change and delete roles to their work history
- Made changes to employments table schema and model to allow for them to belong to either a job application or a jobseeker profile
- Added tests

## Important Notes

Changes to the employments table schema model will allow us to, when we pre-populate job applications with profile data, copy the profile's employment records and associate the new records with the job application. We may want to also set up some sort of self-referential association so we can link these copies to the original.

This will mean that pre-populated employment history entries on a job application will be a snapshot of their profile's employment history entries at the point in time when they begin the job application. This will prevent job application employment history entries from being changed after the submission of a job application (without these changes, this could happen if a jobseeker was to change an employment history entry on their profile after the submission of a job application).


app/views/jobseekers/profiles/employments/edit.html.slim
```diff
@@ -0,0 +1,17 @@
...
+- content_for :breadcrumbs do
+  = govuk_back_link text: t("buttons.back"), href: "#"
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_for form, url: jobseekers_profile_work_history_path(params[:id]), method: :patch do |f|
+      = f.govuk_error_summary
+
+      span.govuk-caption-l = t(".caption")
+      label.govuk-heading-l = t(".title")
```
alexbowen, line 12: and here

app/views/jobseekers/profiles/employments/_fields.html.slim
```diff
@@ -0,0 +1,20 @@
+= f.govuk_text_field :organisation, label: { size: "m" }
```
alexbowen, line 1: i chose to put this file in the shared partials minus the save and continue button which i think makes more sense but willing to discuss, but prob best we re consistent

app/controllers/jobseekers/profiles/employments_controller.rb
```diff
@@ -0,0 +1,62 @@
+class Jobseekers::Profiles::EmploymentsController < Jobseekers::ProfilesController
+  def new; end
```
robyoung91, line 2:
```suggestion
```

Not sure if we need this and the `edit` action to be explicitly defined

app/form_models/jobseekers/profile/employment_form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  attr_reader :started_on, :ended_on
+
+  validates :organisation, presence: true
+  validates :job_title, presence: true
+  validates :started_on, presence: true
+  validates :started_on, date: { before: :ended_on }, if: -> { ended_on.present? }
+  validates :current_role, inclusion: { in: %w[yes no] }
+  validates :ended_on, presence: true, unless: -> { current_role == "yes" }
+  validates :ended_on, date: { after: :started_on }, unless: -> { current_role == "yes" }
```
robyoung91, line 18: Maybe for `:started_on` and `:ended_on` we should also validate `on_or_before: :today` or something similar

app/views/jobseekers/employments/_employments.html.slim
```diff
@@ -0,0 +1,21 @@
+- employments.each do |employment|
+  = render DetailComponent.new(title: employment.job_title, classes: ["work_history_entry"]) do |detail_component|
+    - detail_component.body do
+      = govuk_summary_list do |summary_list|
+        - summary_list.row(html_attributes: { id: "organisation" }) do |row|
```
robyoung91, line 5: For these `id`s we'll end up with multiple elements on the same page with the same `id` - maybe interpolate with the index or something?

app/views/jobseekers/profiles/employments/_summary.html.slim
```diff
@@ -0,0 +1,6 @@
+- content_for :page_title_prefix, "Placeholder"
```
robyoung91, line 1: We need the copy for this page title

app/controllers/jobseekers/profiles/employments_controller.rb
alexbowen: they are not needed, i dunno joe said otherwise yesterday when looking at my task? ill leave them there for the moment

app/views/jobseekers/profiles/employments/_summary.html.slim
alexbowen: dont think its nessary anyway, have removed

app/views/jobseekers/profiles/employments/_fields.html.slim
alexbowen: oh this is because the subjects thing so leave it as is but would be good to resolve somepoint so profile and job apps are 100% aligned

app/views/jobseekers/employments/_employments.html.slim
alexbowen: the ids arent needed so removed alltogether

robyoung91 (approved):

PR #5834, opened 2023-03-02
[112] Pre-fill jobseeker profiles on first visit

elliotcm:
Copies over relevant data from a previous application (if present).  Duplicates qualifications and employment history records.


elliotcm:
[Pre-populate Candidate profile from past application data](https://trello.com/c/SuVrvKOs/112-pre-populate-candidate-profile-from-past-application-data)

PR #5835, opened 2023-03-02
Fixed jobseeker profile location scope and creation fixes

BroiSatse:


app/models/job_preferences/location.rb
```diff
@@ -8,16 +8,18 @@ class Location < ApplicationRecord
 
     before_create :set_area
 
+    scope :containing, ->(point) { where("ST_Within(ST_GeomFromText(?), area::geometry)", "SRID=4326;#{point.as_text}")}
```
elliotcm, line 11: Where does this one get used?

BroiSatse: It will be used in both mine and Davide's branches. We just paired to write it, though it might be useful to add.

PR #5837, opened 2023-03-03
[165] Pre-fill applications from jobseeker profiles if present

elliotcm:


elliotcm:
[Use Profile data to pre-fill an application form](https://trello.com/c/aZzfxcAp/165-use-profile-data-to-pre-fill-an-application-form)

elliotcm:
Updated privacy notice

## Before

Screenshot 2023-03-08 at 10 09 40

## After

Screenshot 2023-03-08 at 10 14 42

PR #5855, opened 2023-03-10
[185] Add key-stage acronym inflections

elliotcm:
This makes `humanize` correctly upcase them.

### Before

```
[1] pry(main)> "ks1".humanize
=> "Ks1"
```

### After

```
[1] pry(main)> "ks1".humanize
=> "KS1"
```


elliotcm:
[Add acronym inflectors for key stages](https://trello.com/c/4pUIHh6y/185-add-acronym-inflectors-for-key-stages)

PR #5856, opened 2023-03-10
Update Dfe Analytics Gem

ddippolito:
## Trello ticket URL

https://trello.com/c/HHdIYWQ0/120-migrate-all-bespoke-events-to-custom-events-in-dfe-analytics

## Changes in this PR:

Update of the DfE Analytics gem

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -29,7 +29,7 @@ class AuthenticationController < ApplicationController
       .with_type(:jobseeker_sign_in_attempt)
       .with_request_details(request)
       .with_response_details(response)
-      .with_user(current_user)
+      .with_user(current_jobseeker)
       .with_data(
         email_identifier: StringAnonymiser.new(params.dig(:jobseeker, :email)),
         success: success_or_failure == :success,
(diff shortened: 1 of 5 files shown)


app/controllers/authentication_controller.rb
```diff
@@ -51,7 +51,7 @@ def trigger_successful_publisher_sign_in_event(sign_in_type, publisher_oid = nil
       .with_type(:successful_publisher_sign_in_attempt)
       .with_request_details(request)
       .with_response_details(response)
-      .with_user(current_publisher)
+      .with_user(current_user)
```
elliotcm, line 54: I don't know that these changes are ideal, e.g. after logging in a jobseeker we should never send anything other than a jobseeker along with the event.

Although we do currently log out different devise scopes when another logs in, I'm not sure whether that's the case for all three scopes, and that could change in future.

Might want to leave these as-were.

elliotcm (changes requested):
We can't always rely on having only one of the user types logged in at once, so it's probably best to leave the specific user type helpers in the places where we're triggering events for that user type.

elliotcm (approved):

PR #5883, opened 2023-03-17
[213] Ensure original profile PII columns are unused (2/4)

elliotcm:
Step 2 of https://github.com/ankane/lockbox#migrating-existing-data


elliotcm:
[Encrypt PII in profiles table](https://trello.com/c/n1sJg3PZ/213-encrypt-pii-in-profiles-table)

PR #5894, opened 2023-03-22
[189] Allow jobseekers to hide their profile from organisations

elliotcm:
## Changes in this PR:

See the "Hide profile from schools or trusts" section of the jobseeker profile.

### DIFF
--- a/app/controllers/jobseekers/profiles/hide_profile_controller.rb
+++ b/app/controllers/jobseekers/profiles/hide_profile_controller.rb
@@ -1,0 +1,66 @@
+class Jobseekers::Profiles::HideProfileController < Jobseekers::ProfilesController
+  def show
+    @form = Jobseekers::Profile::HideProfileForm.new(hide_profile: profile.hide_profile)
+  end
+
+  def confirm_hide
+    form_params = params
+      .require(:jobseekers_profile_hide_profile_form)
+      .permit(:hide_profile)
+
+    if (@form = Jobseekers::Profile::HideProfileForm.new(form_params)).valid?
+      profile.update(hide_profile: @form.hide_profile)
+
+      if profile.hide_profile?
+        redirect_to add_jobseekers_profile_hide_profile_path
+      else
+        redirect_to review_jobseekers_profile_hide_profile_path
+      end
+    else
+      render :show
+    end
+  end
+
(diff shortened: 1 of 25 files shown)


elliotcm:
[Candidates can hide their profile from organisations](https://trello.com/c/w4mB4hzQ/189-candidates-can-hide-their-profile-from-organisations)

bencmitchell:
@elliotcm this all looks good to me 👍

app/controllers/jobseekers/profiles/hide_profile_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    @form = Jobseekers::Profile::HideProfileForm.new(hide_profile: profile.hide_profile)
+  end
+
+  def confirm_hide
+    form_params = params
+      .require(:jobseekers_profile_hide_profile_form)
+      .permit(:hide_profile)
+
+    if (@form = Jobseekers::Profile::HideProfileForm.new(form_params)).valid?
+      profile.update(hide_profile: @form.hide_profile)
```
BroiSatse, line 12: Nitpick: should we use `update!` here instead - non-banged version risks failed save without any feedback

BroiSatse:
I found one interesting scenario here - if a user selects "yes" on the first question, and then cancels it results in profile displaying "Yes" next to hide from organisations with no organisation listed. The same happens when the last organisation is removed from the list. It's a minor thing, not even sure if it is an issue, but I'd imagine it should change to "No" if the list is empty.

BroiSatse (approved):

elliotcm:
> I found one interesting scenario here - if a user selects "yes" on the first question, and then cancels it results in profile displaying "Yes" next to hide from organisations with no organisation listed. The same happens when the last organisation is removed from the list. It's a minor thing, not even sure if it is an issue, but I'd imagine it should change to "No" if the list is empty.

@BroiSatse Yeah the behaviour of that page when there's no orgs set (it's the same if you delete them all) is a bit weird but that's how the prototype works.

elliotcm:
> The only question I have is if you checked `i18n` phrases with someone from content. Other than that super solid! 🚀

@ddippolito All the copy came straight from the prototype.

PR #5897, opened 2023-03-23
[230] Don't trigger JS-only buttons when pressing enter in the quals form

elliotcm:
Implicit form submission will find the first "submit" button and `click` it.  We don't want either of these two buttons to submit the form, so turning them into `type="button"` prevents this.


elliotcm:
[Pressing enter when in focus state on ‘year qual award’ field on GCSE pattern deletes the latest subject that is added - a11y issue](https://trello.com/c/62sNwZ5A/230-pressing-enter-when-in-focus-state-on-year-qual-award-field-on-gcse-pattern-deletes-the-latest-subject-that-is-added-a11y-issue)

PR #5903, opened 2023-03-24
[244] Make schools scrollable/searchable for MATs on profiles index

elliotcm:
### Before
Screenshot 2023-03-24 at 14 39 46

### After
Screenshot 2023-03-24 at 14 42 41


bencmitchell:
Hi @elliotcm , not sure how to review this one given the number of schools in Weydon on the review app.
Looking at the images above it's hard to see if this filter is scrollable (in the same way that the Subject filter would be for example, is this the case? 
image

elliotcm:
@bencmitchell I've added a bunch more schools to weydon on that review app

bencmitchell:
> @bencmitchell I've added a bunch more schools to weydon on that review app

Thanks @elliotcm , looks great 👍

BroiSatse (approved):

PR #5904, opened 2023-03-24
[98] Don't show the applications reminder to local authorities

elliotcm:


bencmitchell:
Looks good @elliotcm, nothing to attach a screenshot of but interstitial no longer seen when creating a job listing as an LA user so all good from my side 👍

BroiSatse (approved):

PR #5905, opened 2023-03-24
Allowing trusts to invite candidated to apply for all their roles

BroiSatse:
https://trello.com/c/xrfhJtzo/246-trust-level-access-is-only-showing-candidates-that-match-with-head-office-not-constituent-academies-within-the-trust

### DIFF
--- a/app/form_models/publishers/invite_form.rb
+++ b/app/form_models/publishers/invite_form.rb
@@ -51,7 +51,7 @@ module Publishers
     def applicable_jobs
       existing_invitations = InvitationToApply.where(jobseeker_id: jobseeker_profile.jobseeker_id)
       @applicable_jobs ||= job_preferences
-        .vacancies(organisation.vacancies.active)
+        .vacancies(organisation.all_vacancies.active)
         .where.not(id: existing_invitations.select(:vacancy_id))
     end
 


spec/fixtures/vcr/Requesting_support/when_recaptcha_is_invalid/and_the_form_is_invalid/does_not_redirect_to_invalid_recaptcha_path.yml
```diff
@@ -200,4 +200,54 @@ http_interactions:
       encoding: UTF-8
       string: '{"id":"bf70e7b4c5eb436bafd25ae462de41af"}'
   recorded_at: Fri, 24 Mar 2023 13:00:47 GMT
+- request:
```
elliotcm, line 203: ?

elliotcm (dismissed):
I'm assuming we don't want this VCR change?

spec/fixtures/vcr/Requesting_support/when_recaptcha_is_invalid/and_the_form_is_invalid/does_not_redirect_to_invalid_recaptcha_path.yml
BroiSatse: It's automatically generated by VCR, I can only assume either the old request tape has expired or sth else happened causing this extra request to be recorded.

elliotcm: Looks like it's being generated by `spec/system/get_help_spec.rb`.

Don't think a new PR is required but it might be nice to put it in another commit for future devs.

BroiSatse: Actually, it might be generated because I have Sentry keys added in my .env file (which I needed for some previous work). After removing the env keys, the change is not showing up. I'll remove this from this PR

PR #5912, opened 2023-03-28
[81] Redo employment gap detection

elliotcm:
This is a fresh version of the logic with comprehensive unit tests.

This is hard to screenshot given all the variations.

### DIFF
--- a/app/services/jobseekers/job_applications/employment_gap_finder.rb
+++ b/app/services/jobseekers/job_applications/employment_gap_finder.rb
@@ -1,0 +1,64 @@
+class Jobseekers::JobApplications::EmploymentGapFinder
+  def initialize(job_application)
+    @job_application = job_application
+    @today = Time.zone.today
+  end
+
+  def significant_gaps(threshold: 3.months)
+    job_application.employments.each_with_object({}) do |employment, gaps|
+      next if employment.current_role?
+      next if gap_to_today_is_less_than_threshold?(employment, threshold)
+      next if overlapping_employment?(employment)
+      next if next_employment_started_within_threshold?(employment, threshold)
+
+      gaps[employment] = {
+        started_on: employment.ended_on + 1.day,
+        ended_on: gap_end_date(employment) || today,
+      }
+    end
+  end
+
+  private
+
+  attr_reader :job_application, :today
+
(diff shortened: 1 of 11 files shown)


elliotcm:
[Investigate/Refactor gap in employment](https://trello.com/c/8MBVRSv4/256-investigate-refactor-gap-in-employment)

app/services/jobseekers/job_applications/employment_gap_finder.rb
```diff
@@ -0,0 +1,64 @@
...
+  private
+
+  attr_reader :job_application, :today
+
+  def gap_to_today_is_less_than_threshold?(employment, threshold)
+    employment.ended_on + threshold >= today
+  end
+
+  def overlapping_employment?(employment)
+    job_application.employments.any? do |other|
```
josephhull676, line 30: Could you do something like `.excluding(employment)` here?

Could save on the extra line `next if employment == other`

app/services/jobseekers/job_applications/employment_gap_finder.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def next_employment_started_within_threshold?(employment, threshold)
+    return unless (next_start = next_employment_start(employment))
+
+    next_start <= employment.ended_on + threshold
+  end
+
+  def next_employment_start(employment)
+    job_application.employments
+      .reject { |other| other == employment }
```
josephhull676, line 49: And perhaps here? 🤷‍♂️

josephhull676 (approved):

app/services/jobseekers/job_applications/employment_gap_finder.rb
elliotcm: @josephhull676 This is not a bad idea, but having tried it just now the problem is that only works if all the records are saved to the database, whereas the current implementation will work on any array of employments.

bencmitchell:
This looks good to me 👍

app/services/jobseekers/job_applications/employment_gap_finder.rb
josephhull676: Ahh ok 👍

PR #5916, opened 2023-03-29
Update download_application button text

georgeschena:


bencmitchell:
Looks good 👍 
image

bencmitchell:
> Looks good. Just need to check with Alex if the fullstop is intended, as I've not seen that on buttons before image

@georgeschena could you remove the "." please, then this is good to go I think

bencmitchell:
Cheers @georgeschena , looks good 👍

elliotcm (approved):

PR #5918, opened 2023-03-31
[258] Only show full time and part time on subscription form

elliotcm:
NB: there is some tech debt here.  At some point we should change the enum of working pattern types in the `Vacancy` model to no longer have these, once we're sure they're not in use any more.

### Before
Screenshot 2023-03-31 at 10 08 57

### After
Screenshot 2023-03-31 at 10 13 08


elliotcm:
[Remove old education phases from the Job Alert form](https://trello.com/c/hjtc7T0G/258-remove-old-education-phases-from-the-job-alert-form)

bencmitchell:
All good 👍 
image

PR #5944, opened 2023-04-13
Removed long obsolete, hardcoded aria-describedby attribute from SearchableCollection component

BroiSatse:
The element matching hardcoded id has been removed in September last year at a291c5224ccaf0f18f5a0f2508d85da4caab628d. 

In any case, this attribute should be dynamic.


josephhull676 (approved):
Test failing but I assume it's flakey.

PR #5977, opened 2023-04-25
Send email events also the the new analytical platform

ddippolito:
## Trello ticket URL
https://trello.com/c/Wp0OMdII/310-tv-migration-blockers

## Changes in this PR:

When migrating to the new Analytical Platform we forgot to add mail events to the migrations. This is a temporary PR to make sure we can send events correctly. And will refine at a later stage

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -86,7 +86,7 @@ class SubscriptionsController < ApplicationController
 
   def trigger_create_job_alert_clicked_event
     request_event.trigger(:vacancy_create_job_alert_clicked, vacancy_id: StringAnonymiser.new(vacancy_id))
-    trigger_dfe_analytics_event(:vacancy_create_job_alert_clicked, { vacancy_id: StringAnonymiser.new(vacancy_id) })
+    trigger_dfe_analytics_event(:vacancy_create_job_alert_clicked, { vacancy_id: vacancy_id })
   end
 
   def trigger_dfe_analytics_event(type, data)
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -17,7 +17,7 @@ southampton_la = SchoolGroup.find_by!(local_authority_code: "852")
 users = [
   { email: "alex.lee@education.gov.uk", family_name: "Alex", given_name: "Lee" },
   { email: "alisa.ali@education.gov.uk", family_name: "Ali", given_name: "Alisa" },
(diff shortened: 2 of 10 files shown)


config/analytics_custom_events.yml
```diff
@@ -18,6 +19,9 @@ shared:
   - jobseeker_application_submitted
   - jobseeker_subscription_update
   - jobseeker_confirmation_instructions
+  - jobseeker_inactive_account
+  - jobseeker_account_closed
+  - jobseeker_job_listing_ended_early
```
scruti, line 24: Question: :raising_hand_man: 
Are these new events we want to send to send? 
Just trying to figure out if this is strictly related to the email events or just using the PR to add some extra info on top of triggering email events.

scruti (approved):

db/seeds.rb
```diff
@@ -17,6 +17,7 @@
 users = [
   { email: "alex.lee@education.gov.uk", family_name: "Alex", given_name: "Lee" },
   { email: "alisa.ali@education.gov.uk", family_name: "Ali", given_name: "Alisa" },
+  { email: "brandon.chan@education.gov.uk", family_name: "Chan", given_name: "Brandon" },
```
scruti, line 20: I think Brandon email is: `brandon1.chan@education.gov.uk`. It shows like this on Slack, Outlook profile...

app/controllers/subscriptions_controller.rb
```diff
@@ -86,7 +86,7 @@ def notify_new_subscription(subscription)
 
   def trigger_create_job_alert_clicked_event
     request_event.trigger(:vacancy_create_job_alert_clicked, vacancy_id: StringAnonymiser.new(vacancy_id))
-    trigger_dfe_analytics_event(:vacancy_create_job_alert_clicked, { vacancy_id: StringAnonymiser.new(vacancy_id) })
+    trigger_dfe_analytics_event(:vacancy_create_job_alert_clicked, { vacancy_id: StringAnonymiser.new(vacancy_id).to_s })
```
scruti, line 89: Is the Vacancy ID a PII?  We are removing the anonymiser calls on it further down the PR.

db/seeds.rb
brandonchan1: Correct, I have brandon1.chan@education.gov.uk

app/controllers/subscriptions_controller.rb
brandonchan1: Agree with @scruti. It's just id so does not need to be anonymised

scruti (approved):
Re-checked and LGTM!

PR #6008, opened 2023-05-04
Re-enable analytics check task

scruti:
## Trello card URL

https://trello.com/c/prlthggR

## Changes in this PR:
- Removed the overriding of `dfe:analytics:check` task. Re-enabling it to run for every test run.
- Based on the check task errors, did run a `dfe:analytics:rengenerate_blocklist` and moved the new records from the `blocklist.yml` to `analytics.yml` and `analytics_pii.yml` as needed. 
- Removed a one-off backfilling job. 
  - It isn't needed since it was originally run for the initial event backfill, and its tests raise an exception due to the naming used by one of the models recently added in `analytics.yml`. Not worth the hassle of fixing unused code.

### Output:
A clean state where all the DB attributes are effectively included-excluded from DFE-Analytics.

### To do after release:

- [ ] Execute a DFE Analytics import to backfill the following entities:
  - [ ] invitation_to_applies
  - [ ] job_preferences_locations
  - [ ] jobseeker_profiles
  - [ ] jobseeker_profile_excluded_organisations
  - [ ] publishers


config/analytics_blocklist.yml
```diff
@@ -67,11 +71,6 @@
   - geopoint
   - created_at
   - updated_at
-  :jobseeker_profiles:
-  - first_name
-  - last_name
-  - phone_number_provided
-  - phone_number
```
scruti, line 74: these fields were removed from the `jobseeker_profiles` table.

config/analytics_blocklist.yml
```diff
@@ -152,21 +151,19 @@
...
   :job_preferences:
   - id
-  - jobseeker_id
-  - jobseeker_profile_id
   - roles
   - phases
   - key_stages
   - subjects
   - working_patterns
-  - locations
```
scruti, line 168: not present in the table

config/application.rb
```diff
@@ -16,8 +16,6 @@
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
-fields = DfE::Analytics::Fields
-def fields.check!; end
```
scruti, line 20: This was overriding the `check` rake method used by the `dfe:analytics:check` rake task.
As a consequence, the task wasn't running the expected checks when running the tests, causing multiple fields/tables to be missing from `analytics/analytics_blocklist` yaml files.

KyleMacPherson (approved):

PR #6009, opened 2023-05-05
Replace deprecated methods

KyleMacPherson:
https://trello.com/c/SMoYZ3RF/322-upgrade-govuk-components-gem-from-330-400

## Changes in this PR:

Version 3.0.0 of the view-component gem removes the deprecated slots setter method, now we need to use with_SLOT_NAME instead. See here: https://github.com/ViewComponent/view_component/releases/tag/v3.0.0.

We need to make this change in order to be able to [Bump govuk-components from 3.3.0 to 4.0.0](https://github.com/DFE-Digital/teaching-vacancies/pull/5993). I have made these changes in a separate PR in order to keep the dependabot PR clean and understandable.

NOTE: I found the uses of the deprecated methods by running the specs while using the `warn_on_deprecated_slot_setter` flag set in our components. There may be deprecated methods used on pages that are not touched by the tests that will still break if/when we upgrade to govuk-components 4.0.0. 

## Next steps:

### DIFF
--- a/app/components/account_survey_link_component.rb
+++ b/app/components/account_survey_link_component.rb
@@ -1,4 +1,4 @@
-class AccountSurveyLinkComponent 
+class AccountSurveyLinkComponent
   def initialize(origin:)
     @origin = origin
   end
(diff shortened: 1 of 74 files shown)


KyleMacPherson:
[Upgrade govuk-components gem from 3.3.0 -> 4.0.0](https://trello.com/c/SMoYZ3RF/322-upgrade-govuk-components-gem-from-330-400)

PR #6048, opened 2023-05-22
Enforce document presence when submitting form

scruti:
## Trello card URL

- https://trello.com/c/S2NOImgj/338-bug-exception-when-submitting-empty-upload-additional-documents
- Bug in Sentry: https://teaching-vacancies.sentry.io/issues/3868404686

## Changes in this PR:

The condition allowing to submit an empty "additional document" form as long as another document was already present in the vacancy was causing an exception down the controller.

We had two options:
- Allow submitting the empty form while guarding the controller code against the exception.
- Enforce the presence of a document on form submission through validation. Users can click "back" or "cancel" if they don't want to attach a document. **Team decided to go with this option**.

### Before
image

### After
Screenshot_20230522_193525

### DIFF
--- a/app/form_models/publishers/job_listing/documents_form.rb
+++ b/app/form_models/publishers/job_listing/documents_form.rb
@@ -38,6 +38,6 @@ class Publishers::JobListing::DocumentsForm < Publishers::JobListing::VacancyFor
     return unless vacancy.include_additional_documents
     return if documents.present?
 
-    errors.add(:documents, :blank) unless vacancy.supporting_documents.any?
+    errors.add(:documents, :blank)
   end
 end


app/form_models/publishers/job_listing/documents_form.rb
```diff
@@ -38,6 +38,6 @@ def document_presence
     return unless vacancy.include_additional_documents
     return if documents.present?
 
-    errors.add(:documents, :blank) unless vacancy.supporting_documents.any?
```
scruti, line 41: This condition allowed publishers to submit the form without `documents` attached as long as the vacancy already had other `supporting_documents` attached to it. We are removing this possibility

scruti:
[Bug: Exception when submitting empty "Upload additional documents"](https://trello.com/c/S2NOImgj/338-bug-exception-when-submitting-empty-upload-additional-documents)

scruti:
@bencmitchell this bugfix is ready for a review (the review app will show up in a few minutes)

bencmitchell:
Looks good @scruti 
image

PR #6057, opened 2023-05-25
Fix documents upload issues

scruti:
## Trello card URL
- https://trello.com/c/S2NOImgj

## Changes in this PR:

Follows up on:
- https://github.com/DFE-Digital/teaching-vacancies/pull/6048
- https://github.com/DFE-Digital/teaching-vacancies/pull/6056

The condition allowing to submit an empty "additional document" form as long as another document was already present in the vacancy was causing an exception down the controller.

We had two options:

Allow submitting the empty form while guarding the controller code against the exception.
Enforce the presence of a document on form submission through validation. Users can click "back" or "cancel" if they don't want to attach a document. The team decided to go with this option.

### Before
image

### After
Screenshot_20230522_193525

### DIFF
--- a/app/controllers/publishers/vacancies/base_controller.rb
+++ b/app/controllers/publishers/vacancies/base_controller.rb
@@ -37,15 +37,19 @@ class Publishers::Vacancies::BaseController < Publishers::BaseController
     )
   end
 
+  def redirect_to_job_summary
+    if vacancy.published?
+      redirect_to organisation_job_path(vacancy.id), success: t("publishers.vacancies.show.success")
+    else
+      redirect_to organisation_job_review_path(vacancy.id)
+    end
+  end
+
   def redirect_to_next_step
     if save_and_finish_later?
       redirect_to organisation_job_path(vacancy.id), success: t("publishers.vacancies.show.success")
     elsif all_steps_valid?
-      if vacancy.published?
-        redirect_to organisation_job_path(vacancy.id), success: t("publishers.vacancies.show.success")
-      else
-        redirect_to organisation_job_review_path(vacancy.id)
-      end
+      redirect_to_job_summary
     else
       redirect_to organisation_job_build_path(vacancy.id, next_invalid_step)
     end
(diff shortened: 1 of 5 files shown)


app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -29,12 +29,12 @@ def destroy
...
   def confirm
-    if confirmation_form.valid?
-      return redirect_to_next_step unless uploading_more_documents?
+    return render :index unless confirmation_form.valid?
 
+    if uploading_more_documents?
       redirect_to new_organisation_job_document_path(vacancy.id)
     else
-      render :index
+      redirect_to_job_summary
```
scruti, line 37: These changes fix the `redirect_to_next_step` wrongly sending the user to the "upload an additional document" page even when they answer "NO" to the "do you want to upload an additional document?" question.

We cannot rely on its logic for this case, as the documents form without a document is no longer valid, so we do an explicit redirection to the right step instead.

app/controllers/publishers/vacancies/base_controller.rb
```diff
@@ -37,15 +37,19 @@ def form_sequence
     )
   end
 
+  def redirect_to_job_summary
+    if vacancy.published?
+      redirect_to organisation_job_path(vacancy.id), success: t("publishers.vacancies.show.success")
+    else
+      redirect_to organisation_job_review_path(vacancy.id)
+    end
+  end
```
scruti, line 46: This bit is called in 2 different places so extracted it into an easier to understand/follow method.

KyleMacPherson (approved):
Thanks for the comments @scruti, really helped me to understand the reasoning behind the changes. I will try to do this more in my future PRs!

scruti:
[Bug: Exception when submitting empty "Upload additional documents"](https://trello.com/c/S2NOImgj/338-bug-exception-when-submitting-empty-upload-additional-documents)

PR #6084, opened 2023-06-06
Fix documents upload issues (third attempt!)

scruti:
## Trello card URL
- https://trello.com/c/S2NOImgj
## Changes in this PR:

Context of previous attempts:
- First attempt: https://github.com/DFE-Digital/teaching-vacancies/pull/6048
- Second attempt:  https://github.com/DFE-Digital/teaching-vacancies/pull/6057

### Context 
The validation bypass on the form was causing an exception on the service, as users were selecting "yes" to upload an additional document question and then submitting the form without any file attached.

This resulted in an exception being raised: https://teaching-vacancies.sentry.io/issues/3868404686

This bypass on the documents form validation was introduced to force the vacancy flow control method `all_steps_valid?` to be successful so the flow can move to the next step after choosing "no" when enquired about wanting to add additional documents or to allow to publish the job vacancy draft.

### What are we doing
- We are removing the validation bypass so forms cannot be submitted without an attached document. 
  - This resolves the original bug.
- We are aligning the `DocumentsForm` attribute naming with `Vacancy` documents attribute naming. 
  - This allows us to align this step with the rest regarding the `FormSequence#all_steps_valid?` validations. With the renaming from `documents` to `supporting_documents` and exposing it as a field, the FormSequence will be able to automatically map the Vacancy supporting documents into the DocumentForm supporting documents and identify the step as valid/already completed.
  - This way, we won't need to bypass or ignore the step validation to check if all the vacancy steps are valid and the vacancy can be progressed/published.

### Before
image

### After
image

### After
image

### DIFF
--- a/app/controllers/publishers/vacancies/documents_controller.rb
+++ b/app/controllers/publishers/vacancies/documents_controller.rb
@@ -5,7 +5,7 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
 
   def create
     if documents_form.valid?
-      documents_form.documents.each do |document|
+      documents_form.supporting_documents.each do |document|
         vacancy.supporting_documents.attach(document)
         send_event(:supporting_document_created, document.original_filename, document.size, document.content_type)
         send_dfe_analytics_event(:supporting_document_created, document.original_filename, document.size, document.content_type)
@@ -50,7 +50,7 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
 
   def documents_form_params
     (params[:publishers_job_listing_documents_form] || params)
-      .permit(documents: [])
+      .permit(supporting_documents: [])
       .merge(completed_steps: completed_steps)
   end
 
--- a/app/controllers/publishers/vacancies/documents_controller.rb
+++ b/app/controllers/publishers/vacancies/documents_controller.rb
@@ -68,7 +68,10 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
 
   def after_document_delete_path
     if vacancy.supporting_documents.none?
-      new_organisation_job_document_path(vacancy.id, back_to_review: params[:back_to_review], back_to_show: params[:back_to_show])
+      organisation_job_build_path(vacancy.id,
+                                  :include_additional_documents,
+                                  back_to_review: params[:back_to_review],
+                                  back_to_show: params[:back_to_show])
     else
       organisation_job_documents_path(vacancy.id, back_to_review: params[:back_to_review], back_to_show: params[:back_to_show])
     end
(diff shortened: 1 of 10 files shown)


app/form_models/publishers/job_listing/documents_form.rb
```diff
@@ -38,6 +38,6 @@ def document_presence
     return unless vacancy.include_additional_documents
     return if documents.present?
 
-    errors.add(:documents, :blank) unless vacancy.supporting_documents.any?
+    errors.add(:documents, :blank)
```
scruti, line 41: This was the cause of the original bug, allowing users to submit a form without any attached document.

app/form_models/publishers/vacancy_form_sequence.rb
```diff
@@ -63,6 +63,8 @@ def next_incomplete_step_subjects?
   end
 
   def not_validatable_steps
-    %i[subjects review].freeze
+    %i[subjects review].tap do |steps|
+      steps << :documents if @vacancy.supporting_documents.any?
+    end
```
scruti, line 68: By setting `documents` step as one of the non-validatable steps in the vacancy form sequence, when checking if all steps are valid, the now "invalid" documents form won't interfere with it.

spec/system/publishers_can_add_aditional_documents_to_a_vacancy_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(current_path).to eq(organisation_job_documents_path(vacancy.id))
+
+    # Once decided not to include additional documents, can continue to the next step
+    answer_include_additional_documents(false)
+    expect(current_path).to eq(organisation_job_review_path(vacancy.id))
+    expect(page).to have_content(vacancy.job_role.humanize)
+
+    # Can publish the job listing
+    click_on I18n.t("publishers.vacancies.show.heading_component.action.publish")
+    expect(current_path).to eq(organisation_job_summary_path(vacancy.id))
```
scruti, line 46: This bit was the cause of the last attempt having to be reverted. It is now covered with the system specs.

app/controllers/publishers/vacancies/documents_controller.rb
```diff
@@ -68,7 +68,10 @@ def uploading_more_documents?
 
   def after_document_delete_path
     if vacancy.supporting_documents.none?
-      new_organisation_job_document_path(vacancy.id, back_to_review: params[:back_to_review], back_to_show: params[:back_to_show])
+      organisation_job_build_path(vacancy.id,
+                                  :include_additional_documents,
+                                  back_to_review: params[:back_to_review],
+                                  back_to_show: params[:back_to_show])
```
scruti, line 74: Unrelated to the bug:
We are changing the redirection path when the last document is deleted from the vacancy documents list. Our BA proposed this during review so we quickly changed it.

app/form_models/publishers/job_listing/documents_form.rb
```diff
@@ -36,8 +36,8 @@ def params_to_save
 
   def document_presence
     return unless vacancy.include_additional_documents
-    return if documents.present?
+    return if supporting_documents.present?
 
-    errors.add(:documents, :blank) unless vacancy.supporting_documents.any?
+    errors.add(:supporting_documents, :blank)
```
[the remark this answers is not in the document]
KyleMacPherson: Nice!

PR #6095, opened 2023-06-12
Split 2 of the jobseeker profile preference roles

scruti:
Two of the jobseeker profiles preferrence roles are bein split into multiple roles.
- "Headteacher, deputy or assistant headteacher" split into 3 roles:
  - "Headteacher"
  - "Deputy headteacher"
  - "Assistant headteacher"
- "Head of year, department, curriculum or phase" split into 2 roles:
  - "Head of year or phase"
  - "Head of department or curriculum"

We are backfilling any existing user preference selecting the replaced broad roles with all the more granular roles they were split into.

## Changes in this PR:

## Next steps:

### DIFF
--- a/app/form_models/jobseekers/job_preferences_form.rb
+++ b/app/form_models/jobseekers/job_preferences_form.rb
@@ -4,7 +4,8 @@ module Jobseekers
   class JobPreferencesForm
     include Multistep::Form
 
-    ROLES = %i[teacher senior_leader middle_leader teaching_assistant higher_level_teaching_assistant education_support sendco].freeze
+    ROLES = %i[teacher head_of_year head_of_department headteacher_assistant headteacher_deputy headteacher
+               teaching_assistant higher_level_teaching_assistant education_support sendco].freeze
     PHASES = %i[nursery primary middle secondary through].freeze
     WORKING_PATTERNS = %i[full_time part_time].freeze
 
@@ -22,7 +23,7 @@ module Jobseekers
       validate :validate_roles
 
       def options
-        ROLES.to_h { |opt| [opt.to_s, I18n.t("helpers.label.publishers_job_listing_job_role_form.job_role_options.#{opt}")] }
+        ROLES.to_h { |opt| [opt.to_s, I18n.t("helpers.label.jobseekers_job_preferences_form.role_options.#{opt}")] }
       end
 
       def validate_roles
--- a/config/locales/forms.yml
+++ b/config/locales/forms.yml
@@ -369,6 +369,18 @@ en:
         employment_history_section_completed_options:
           false: No, I'll come back to it later
           true: Yes, I've completed this section
+      jobseekers_job_preferences_form:
+        role_options:
+          education_support: Learning support, cover supervisor or tutor
+          head_of_year: Head of year or phase
+          head_of_department: Head of department or curriculum
+          headteacher: Headteacher
+          headteacher_assistant: Assistant headteacher
+          headteacher_deputy: Deputy headteacher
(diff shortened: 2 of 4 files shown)


spec/models/job_preferences/job_scope_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      let(:jobseeker_phases) { %w[primary secondary] }
+
+      it "returns any vacancies with a phase matching the jobseeker phases" do
+        expect(job_scope.call).to contain_exactly(vacancy1, vacancy2, vacancy3, vacancy4)
+      end
+    end
+
+    context "with some of the jobseeker phases matching vacancy phases" do
+      let(:jobseeker_phases) { %w[primary middle] }
+
```
KyleMacPherson, line 115: I know this isn't related to this PR but shouldn't this show `vacancy1` and `vacancy2` as well as they both include `primary`?

spec/models/job_preferences/job_scope_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+      it "returns any vacancies without any key stages or with all their key stages included in the jobseeker key stages" do
+        expect(job_scope.call).to contain_exactly(vacancy1, vacancy2, vacancy3, vacancy4)
+      end
+    end
+
+    context "with some of the jobseeker key stages matching vacancy key stages" do
+      let(:jobseeker_key_stages) { %w[ks1 ks3] }
+
+      it "returns any vacancies without any key stages or with all their key stages included in the jobseeker key stages" do
+        expect(job_scope.call).to contain_exactly(vacancy1, vacancy2)
```
KyleMacPherson, line 145: Like the above comment, not related to this change but I would expect `vacancy4` to show here. I may well be misunderstanding though!

KyleMacPherson (approved):

spec/models/job_preferences/job_scope_spec.rb
scruti: Thanks for trying to understand every assertion here :clap: .

This is why I decided/needed to test the whole `JobScope` behaviour. The scoping logic is mainly done through `PostgreSQL` array methods and is not immediately obvious.

For the phases scoping, the `JobScope` applies this query:
```
.where("phases <@ ARRAY[?]::int[]", Vacancy.phases.values_at(*job_preferences.phases))
```

From [PostgreSQL Array functions page](https://www.postgresql.org/docs/current/functions-array.html)

```
anyarray <@ anyarray → boolean

Is the first array contained by the second?

ARRAY[2,2,7] <@ ARRAY[1,7,4,2,6] → 
```

It filters out any vacancy with multiple phases if all the vacancy phases are not selected in the jobseeker preferences. 
As `vacancy1` and `vacancy2` have 2 phases: `["primary", "secondary"]` and the jobseeker preferences are `["primary", "middle"]`, vacancies are excluded from the results due to jobseeker preferences not including `"secondary"`.

The behaviour you expect corresponds to the scopes using `&&` operator instead of `<@`. EG: The `subjects` scoping is done that way.

spec/models/job_preferences/job_scope_spec.rb
scruti: As mentioned above, is due to the query in `JobScope` using `<@` operator. `vacancy4` has `ks2` key_stage and is missing from the `ks1, ks3` jobseeker ones. So it doesn't match the vacancy.

scruti:
[Splitting out filters on candidate profiles](https://trello.com/c/EKCbPxos/366-splitting-out-filters-on-candidate-profiles)

PR #6096, opened 2023-06-12
Add extra data to failed validation form event

ddippolito:
This pull requests fixes the `form_validation_failed` custom event sent to Big Query.
This event needs to use the`Event` class originally built before the DfE Analytics gem was released. This is because, differently from other events we migrated, the `form_validation_failed` event does not have access to sessions and current users data, as the form objects don't inherit from `ApplicationController`.

Unfortunately, this PR adds a bit of duplication, but this is only a temporary measure as when we fully migrate to the new platform we should be able to remove a considerable amount of legacy code including the duplication

### DIFF
--- a/app/form_models/base_form.rb
+++ b/app/form_models/base_form.rb
@@ -17,8 +17,10 @@ class BaseForm
   end
 
   def send_errors_to_big_query
-    EventContext.trigger_event(:form_validation_failed, event_data) if errors.any?
-    EventContext.trigger_for_dfe_analytics(:form_validation_failed, event_data) if errors.any?
+    return if errors.none?
+
+    EventContext.trigger_event(:form_validation_failed, event_data)
+    EventContext.trigger_for_dfe_analytics(:form_validation_failed, event_data)
   end
 
   private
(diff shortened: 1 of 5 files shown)


app/form_models/base_form.rb
```diff
@@ -17,13 +17,8 @@ def self.optional?
...
   def send_errors_to_big_query
-    event = DfE::Analytics::Event.new
-      .with_type(:form_validation_failed)
-      .with_data(event_data)
-
-    DfE::Analytics::SendEvents.do([event])
-
     EventContext.trigger_event(:form_validation_failed, event_data) if errors.any?
+    EventContext.trigger_for_dfe_analytics(:form_validation_failed, event_data) if errors.any?
   end
```
scruti, line 22: Suggestion:
```ruby
  def send_errors_to_big_query
    return if errors.none?

    EventContext.trigger_event(:form_validation_failed, event_data)
    EventContext.trigger_for_dfe_analytics(:form_validation_failed, event_data)
  end
  ```

scruti (approved):

PR #6107, opened 2023-06-15
Filter by org

KyleMacPherson:
https://trello.com/c/QT6UH2CY/331-add-an-organisation-type-filter

## Changes in this PR:

This PR adds a filter to the jobseekers jobs and schools page allowing them to filter by organisation type. The previous PR was reverted [here](https://github.com/DFE-Digital/teaching-vacancies/pull/6105) due to [this bug](https://teaching-vacancies.sentry.io/issues/4247141352/?project=6212514&query=firstRelease%3A907efda2c98b69ce1c68445013ac84ebf9f608dc&referrer=issue-stream&sort=freq&stream_index=0).

I have addressed this issue by using a subquery rather than `distinct` when implementing the new feature.

## Next steps:


app/queries/vacancy_filter_query.rb
```diff
@@ -33,6 +33,24 @@ def call(filters) # rubocop:disable Metrics/AbcSize
...
+    selected_school_types = []
+
+    if filters[:organisation_types].include?("Academy")
+      selected_school_types.push("Academy", "Academies", "Free schools", "Free school")
+    end
+
+    if filters[:organisation_types].include?("Local authority maintained schools")
+      selected_school_types << "Local authority maintained schools"
+    end
+
```
KyleMacPherson, line 48: I have used this subquery to avoid using distinct here which fixes the error we are getting, although I think it will make the query slightly more expensive.

scruti (approved):

PR #6126, opened 2023-06-21
Redesign and extend cookies banner and preferences

scruti:
## Trello card URL
- https://trello.com/c/HwgS3j2n
## Changes in this PR:
- Redesign of the cookie banner.
- Redesign the cookies preferences page, informing about marketing cookies on top of the essential and analytical cookies.
- The user can accept/reject additional cookies (including analytical and marketing purposes) on the cookies preferences page. No option will be selected by default.
- Force all users to accept/reject cookies with new wording, regardless of having accepted analytical cookies before this change release.
- Display a confirmation banner after setting the cookies preferences.

### Banner content/buttons redesign:
image

### Cookies preferences page redesign:
image
image

### Confirmation banner:
image


app/components/cookies_banner_component.rb
```diff
@@ -1,13 +1,13 @@
 class CookiesBannerComponent < ApplicationComponent
-  attr_reader :create_path, :reject_path, :preferences_path
+  attr_reader :accept_path, :reject_path, :preferences_path
```
scruti, line 2: This is a non-functional change. I decided to replace `create` with `accept` to reflect better the user action of accepting cookies.

app/controllers/application_controller.rb
```diff
@@ -41,7 +41,7 @@ def not_found(_error = nil)
   private
 
   def cookies_preference_set?
-    cookies["consented-to-cookies"].present?
+    cookies["consented-to-analytics-cookies"].present? && cookies["consented-to-marketing-cookies"].present?
```
scruti, line 44: Users see the cookies banner unless analytics and marketing cookies consent is answered.
This will make existing users who `consented-to-cookies` before this PR answer again for their cookies settings.

app/controllers/cookies_preferences_controller.rb
```diff
@@ -2,16 +2,28 @@ class CookiesPreferencesController < ApplicationController
   before_action :set_previous_url_in_session
 
   def new
-    @cookies_preferences_form = CookiesPreferencesForm.new({ cookies_consent: cookies["consented-to-cookies"] })
+    # TODO: Remove "consented-to-cookies" check after Dec 2023 (6 months from this comment). That cookie
+    # won't be valid anymore.
+    cookies_analytics_consent =
+      if cookies["consented-to-analytics-cookies"] == "yes" || cookies["consented-to-cookies"] == "yes"
```
scruti, line 8: The current existing cookies agreement will set the `consented-to-analytics-cookies` option.

app/helpers/application_helper.rb
```diff
@@ -39,7 +39,7 @@ def footer_links
   end
 
   def include_google_tag_manager?
-    cookies["consented-to-cookies"] == "yes" && Rails.configuration.app_role.production?
+    cookies["consented-to-analytics-cookies"] == "yes" && Rails.configuration.app_role.production?
```
scruti, line 42: This area is under discussion:
I expect we will need to make changes on Google Tag manager side (and probably on this PR) regarding what to include/exclude depending on user consenting/not consenting to analytics/marketing cookies.

KyleMacPherson (approved):

KyleMacPherson:
Nice work Marc! This part in CookiesPreferencesController (which is not a change you made in this PR!) took me a while to understand:

```
  def cookies_preference_set?
    true
  end
```

I think it is to ensure we don't show a cookies banner to the user while they're viewing their cookies preferences? If so I think we could maybe change the method name to `def viewing_cookie_preferences?` and add a method like this to the application controller which we would use instead of `cookies_preferences_set?` to determine whether or not to show the cookies banner:

```
  def show_cookies_banner
    cookies_preference_set? || viewing_cookies_preferences?
  end
```

KyleMacPherson (approved):

PR #6131, opened 2023-06-23
Out of scope schools

KyleMacPherson:
## Trello card URL

https://trello.com/b/1QyLnb9W/teaching-vacancies-sprint-board

## Changes in this PR:

Prior to this change some schools were being shown on our schools page which are out of scope for the teaching vacancies app. This change stops these types or organisation which are out of scope for us from showing on the schools page.

## Next steps:

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -47,7 +47,7 @@ class Organisation < ApplicationRecord
       .or(where(id: SchoolGroupMembership.select(:school_id).where(school_group_id: registered_organisations)))
   end)
 
-  scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES) }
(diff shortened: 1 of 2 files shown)


app/models/organisation.rb
```diff
@@ -46,7 +47,9 @@ class Organisation < ApplicationRecord
       .or(where(id: SchoolGroupMembership.select(:school_id).where(school_group_id: registered_organisations)))
   end)
 
-  scope :visible_to_jobseekers, -> { schools.not_closed.or(Organisation.trusts).registered_for_service }
+  scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES).or(where(detailed_school_type: nil)) }
+
```
KyleMacPherson, line 51: I had to add in `or(where(detailed_school_type: nil)` because I found that rails where not queries do not return instances that have nil values in the specified field which is pretty strange. Can check it out here if you are interested: https://thoughtbot.com/blog/activerecord-s-where-not-and-nil

KyleMacPherson:
[Fix issue causing out of scope schools to show up on the service](https://trello.com/c/G6xKhQpE/376-fix-issue-causing-out-of-scope-schools-to-show-up-on-the-service)

scruti (approved):

PR #6136, opened 2023-06-27
Changes to email confirmation flow

KyleMacPherson:
https://trello.com/c/4wOtzUFz/378-functionality-to-re-send-verification-link-for-jobseeker-users

## Changes in this PR:

This PR changes the flow when users confirm their account after signing up.

## Next steps:

### DIFF
--- a/app/controllers/jobseekers/confirmations_controller.rb
+++ b/app/controllers/jobseekers/confirmations_controller.rb
@@ -1,8 +1,4 @@
 class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
-  def show
-    super
-  end
-
   protected
 
   def after_confirmation_path_for(_resource_name, resource)
--- a/app/controllers/jobseekers/confirmations_controller.rb
+++ b/app/controllers/jobseekers/confirmations_controller.rb
@@ -5,7 +5,7 @@ class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
 
   protected
 
-  def after_confirmation_path_for(resource_name, resource)
+  def after_confirmation_path_for(_resource_name, resource)
     sign_in(resource)
     request_event.trigger(:jobseeker_email_confirmed)
     flash.delete(:notice)
--- a/app/controllers/jobseekers/confirmations_controller.rb
+++ b/app/controllers/jobseekers/confirmations_controller.rb
@@ -1,6 +1,6 @@
 class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
   def show
-    super if request.method == "POST"
+    super
   end
 
   protected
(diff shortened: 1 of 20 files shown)


app/controllers/jobseekers/confirmations_controller.rb
```diff
@@ -1,15 +1,15 @@
 class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
   def show
-    super if request.method == "POST"
+    super
   end
 
   protected
 
-  def after_confirmation_path_for(resource_name, resource)
+  def after_confirmation_path_for(_resource_name, resource)
```
KyleMacPherson, line 8: We now want to redirect to the confirmation_jobseekers_account page every time that a user confirms their account, I double checked this behaviour with Alex to make sure!

spec/system/jobseekers_can_create_a_job_alert_from_a_search_spec.rb
```diff
@@ -68,28 +68,6 @@
         end
       end
     end
-
```
KyleMacPherson, line 71: I have removed a lot of tests that assert that we redirect to where the user previously was after they confirm their account as it is no longer the desired behaviour.

scruti: I was puzzled by this. It makes sense. Thanks for the comment!

app/controllers/jobseekers/confirmations_controller.rb
```diff
@@ -1,15 +1,15 @@
 class Jobseekers::ConfirmationsController < Devise::ConfirmationsController
   def show
-    super if request.method == "POST"
+    super
   end
```
scruti, line 4: Question :question: 
Now that `#show` does not override any logic from the parent, could we remove this definition? Or is needed?

KyleMacPherson: Ah yep, I'm we can just remove it 👍

app/views/layouts/_flash_messages.html.slim
```diff
@@ -1,4 +1,7 @@
 - flash.each do |variant, message|
   - next unless variant.in? t("banners").keys.map(&:to_s)
   = govuk_notification_banner title_text: t("banners.#{variant}"), classes: "govuk-notification-banner--#{variant} govuk-!-margin-top-3 govuk-!-margin-bottom-0" do |notification_banner|
-    - notification_banner.with_heading(text: message)
```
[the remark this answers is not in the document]
scruti: I was trying to figure out how to keep this area generic while implementing changes only in the affected area (the Jobseeker session controller, model, etc).
So, following the typical devise tweaking approach, I search what we need to override from Devise :smile: 

In this case, the message is built in `Devise::Models::Confirmable#inactive_message`. So I attempted redefining it in `Jobseeker#inactive_message` and then building up the text with the link there.

But with my approach, we have different issues: 
- We must declare the text as HTML safe in the banner layout code. 
- We are embedding HTML in the notification banner helper text parameter instead of using the parameters created for this specific purpose.
- We need to load route helpers for rails (for getting the URL) in a rails model.

Given those downsides, I give up and think your code has fewer downsides than what I've attempted.
We may not like it, but is simple enough.

scruti (approved):
LGTM. Thanks for they key comments.

PR #6187, opened 2023-07-12
Fix: Avoid using sequence for vacancy in prod env

scruti:
Since FactoryBot is used in Production environment we have found ourselves in quite a mess.
 Temporal fix until we stop using FB in production.

The fix keeps the `sequence` usage for Vacancy titles in test environment, while defaulting to the previous `sample` approach only for production environment.

We have added a ticket in the Tech Debt backlog to remove the usage of FactoryBot in production, but it will take time since it is used to generate a sample job application in the vacancy creation flow.

Previous attempt to fix this: https://github.com/DFE-Digital/teaching-vacancies/pull/6177

Where this is used in production (screenshot from review app after fix):
image

### DIFF
--- a/spec/factories/vacancies.rb
+++ b/spec/factories/vacancies.rb
@@ -1,18 +1,22 @@
 FactoryBot.define do
-  factory :vacancy do
-    job_titles = [
-      "Tutor of Science",
-      "PROGRESS LEADER (HEAD OF DEPARTMENT)",
-      "Tutor of Music (Part time, permanent)",
-      "Tutor of Maths MPS",
-      "Tutor of PE (male)",
-      "Games Design Tutor",
-      "Team Leader of Maths",
-      "KEY STAGE 2 Tutor",
-      "Lead in Health and Social Care",
-      "Director of Learning - Science",
-    ].freeze
+  job_titles = [
+    "Tutor of Science",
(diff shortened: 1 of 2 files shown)


spec/factories/vacancies.rb
```diff
@@ -200,9 +204,5 @@
       external_reference { "J3D1" }
       external_advert_url { "https://example.com/jobs/123" }
     end
-
-    trait :non_sequential_job_title do
-      job_title { factory_sample(job_titles) }
-    end
```
scruti, line 206: This was my previous attempt not to call `sequence` in production. I thought it worked, but the error came back today.

spec/factories/vacancies.rb
```diff
@@ -44,7 +48,7 @@
     expires_at { 6.months.from_now.change(hour: 9, minute: 0, second: 0) }
     hired_status { nil }
     include_additional_documents { false }
-    sequence :job_title, job_titles.cycle
+    job_title { Rails.env.production? ? factory_sample(job_titles) : generate(:job_title) }
```
scruti, line 51: Only calling the sequence in non-prod envs.
So the current production  job application example through the factory bot will keep looking "pretty", while testing will generate sequential job titles to avoid ambiguous matching errors.

KyleMacPherson (approved):

PR #6211, opened 2023-07-21
Record "Start Date" coming from vacancy sources

scruti:
## Trello card URL

- https://trello.com/c/Z4dHUUis/442-enable-start-dates-from-ats-providers-to-appear-on-job-listings

## Changes in this PR:

- `Fusion` and `MyNewTerm` feeds include `startDate` values in the vacancies. This PR ensures they get recorded.
  - When possible, we take it like a literal date and record it as `Vacancy#starts_on` 
  - When coming as a text, extra wording, etc. a literal parse is prone to errors.  In that case, we record it as a text on `Vacancy#other_start_date_details`. 
    - This matches the Hiring staff vacancy creation flow when they don't enter an exact date but fill in an input text for the Start Date form.
  - I haven't introduced it for the `UnitedLearning` feed, as non of the vacancies in their feed include the field (not the value, but not even the field name). So it seems their integration is not exposing the StartDate.

- Namespace the whole `VacancySource` area.
  This helps us wrap the different parts of the `VacancySource` codebase without mixing everything in the same folder. For example, we have introduced the concept of `VacancySource::Parser` used by a `VacancySource::Source`.
   There is room for further improvements, as we have repeated code around (including using the `StartDate` parser) and other things that could be extracted (`JobRole` parser shared between all sources comes to mind). But I tried to keep the PR to a reasonable size and only made the changes that would allow me to introduce the `VacancySource::Parser::StartDate` concept and accommodate it in a coherent codebase around it.
   
image

### DIFF
--- a/app/jobs/import_from_vacancy_sources_job.rb
+++ b/app/jobs/import_from_vacancy_sources_job.rb
@@ -1,5 +1,7 @@
 class ImportFromVacancySourcesJob < ApplicationJob
-  SOURCES = [UnitedLearningVacancySource, FusionVacancySource, MyNewTermVacancySource].freeze
+  SOURCES = [VacancySource::Source::UnitedLearning,
+             VacancySource::Source::Fusion,
+             VacancySource::Source::MyNewTerm].freeze
 
   queue_as :default
 
--- a/app/vacancy_sources/vacancy_source/parser/start_date.rb
+++ b/app/vacancy_sources/vacancy_source/parser/start_date.rb
@@ -1,5 +1,5 @@
 # Parses, when possible, the start date values coming from the external vacancy sources
-class VacancySourceStartDate
+class VacancySource::Parser::StartDate
   # EG: "04-09-2023", "4-9-23", "04/09/2023", "4/9/23", "04.09.2023"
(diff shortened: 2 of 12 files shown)


KyleMacPherson (approved):

scruti:
@ddippolito that is a good point. And thanks for pointing out a particular gem.

I've checked Chronic with some sample data from the integrations and works brilliantly parsing exact dates in natural language.

The problem is that in our incoming data, we don't always want to record an exact date in the vacancy, as usually comes with some added text.  Examples:
```
Chronic.parse  "September 2023"
=> 2023-09-01

Chronic.parse "1st September 2023 or as soon as possible afterwards"
=> 2023-09-01

Chronic.parse "September 2023 or sooner if available"
=> 2023-09-1 

Chrinic.parse "Autumn Term 2023"
=> 2022-09-23 
```
When that info comes, I think we should store it as free text (like when the hiring staff introduces those sentences in the text input for the start date during the vacancy creation flow).

If we were parsing it as well as Chronic does, we would be setting very specific dates on our site, while the job posting date intended to mention a period or be more vague, and I think this would be a problem.

PR #6220, opened 2023-07-24
Add ActiveRecord DB consistency tools - Plus fix missing Foreign Keys

scruti:
## Trello card URL

- Part of the tasks described in https://trello.com/c/ZFVIvvfF

## Changes in this PR:
This starts the journey of fixing inconsistencies & issues between our ActiveRecord models and the DB schema.

- Introduces gems in the development group to identify issues with our ActiveRecord and Schema definitions.
  - [ActiveRecord doctor ](https://github.com/gregnavis/active_record_doctor)
  - [Database Consistency](https://github.com/djezzzl/database_consistency)
  - The reason for introducing both of them is that, at the moment, they're identifying lots of issues. And each tool identifies some issues the other doesn't. So, for the time being, let's add both. We can decide on one or another (or both!) when adding them to CI.
- Add missing Foreign Keys in DB. Ensuring data consistency in `belongs_to` ActiveRecord definitions.


.database_consistency.todo.yml
```diff
@@ -0,0 +1,251 @@
+---
```
scruti, line 1: This is like a Rubocop ToDo file. It is whitelisting current violations so we can remove them from here and fix them bit by bit.

.database_consistency.yml
```diff
@@ -0,0 +1,9 @@
+# Ignore false positive from Rails' ActionText and ActiveStorage
```
scruti, line 1: This is the actual configuration file used by `database_consistency` tool.

db/migrate/20230724180457_add_organisations_id_school_group_memberships_school_id_fk.rb
```diff
@@ -0,0 +1,8 @@
+class AddOrganisationsIdSchoolGroupMembershipsSchoolIdFk < ActiveRecord::Migration[7.0]
```
scruti, line 1: Non-blocking migrations (`online_migrations`  gem helping a lot here).

KyleMacPherson (approved):

PR #6226, opened 2023-07-25
Vacancy search: few improvements for performance gains

scruti:
## Trello card URL

- https://trello.com/c/MI8uwJte

## Review APP to check that things didn't break
- https://teaching-vacancies-review-pr-6226.london.cloudapps.digital/

## Changes in this PR:

Few bits to improve the performance of the Vacancy Search.
- Reduce the calls to DB for the whole search query when possible. 
  - Mainly by memoising the query results count on the first call and reusing this number instead of triggering new counts. 
- Reduce loading all the search results in memory when possible.
  - A few spots were loading all the search results and operating over them. We could avoid loading the whole search results into memory by directly filtering in the DB query.
- Sends only current page results ids to Analytics (To confirm this was a bug before releasing it). 

## What I discarded:
- I attempted to be smart by preloading organisations and the logo attachment retrieval, reducing the number of queries. But then I realised that:
  - Given that the results are paginated, the N+1 we're avoiding has a magnitude of... 10. Not big gains here.
  - Preloading gets all the results into memory, not the paginated subset. When run locally with a few vacancies was fast, but in production with thousands of vacancies, this probably will have worse performance than the 10 individual queries triggered per page.
   
## What I've learnt:
- `Search::WiderSuggestionsBuilder.new(search_criteria).suggestions`, called when no results were found for the desired location, triggers multiple search queries (one per radius option above the current radius). This very expensive call causes the slowest searches (as does the original search plus 1 to 9 further searches!).

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -56,7 +56,7 @@ class VacanciesController < ApplicationController
         search_criteria: form.to_hash,
         sort_by: form.sort.by,
         page: params[:page] || 1,
-        total_count: @vacancies_search.total_count,
+        total_count: vacancy_ids.size,
         vacancies_on_page: vacancy_ids,
         location_polygon_used: polygon_id,
         landing_page: params[:landing_page_slug],
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -49,7 +49,7 @@ class VacanciesController < ApplicationController
 
   def trigger_search_performed_event
     fail_safe do
-      vacancy_ids = @vacancies_search.vacancies.pluck(:id)
+      vacancy_ids = @vacancies_search.vacancies.unscope(:order).pluck(:id)
       polygon_id = DfE::Analytics.anonymise(@vacancies_search.location_search.polygon.id) if @vacancies_search.location_search.polygon
 
       event_data = {
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -4,7 +4,7 @@ class VacanciesController < ApplicationController
 
   def index
     @vacancies_search = Search::VacancySearch.new(form.to_hash, sort: form.sort)
-    @pagy, @vacancies = pagy(@vacancies_search.vacancies)
+    @pagy, @vacancies = pagy(@vacancies_search.vacancies, count: @vacancies_search.total_count)
   end
 
   def show
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -49,7 +49,7 @@ class VacanciesController < ApplicationController
 
   def trigger_search_performed_event
     fail_safe do
-      vacancy_ids = @vacancies_search.vacancies.map(&:id)
+      vacancy_ids = @vacancies_search.vacancies.pluck(:id)
       polygon_id = DfE::Analytics.anonymise(@vacancies_search.location_search.polygon.id) if @vacancies_search.location_search.polygon
 
       event_data = {
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -49,14 +49,14 @@ class VacanciesController < ApplicationController
 
   def trigger_search_performed_event
     fail_safe do
-      vacancy_ids = @vacancies_search.vacancies.unscope(:order).pluck(:id)
+      vacancy_ids = @vacancies.pluck(:id)
       polygon_id = DfE::Analytics.anonymise(@vacancies_search.location_search.polygon.id) if @vacancies_search.location_search.polygon
 
       event_data = {
         search_criteria: form.to_hash,
         sort_by: form.sort.by,
         page: params[:page] || 1,
-        total_count: vacancy_ids.size,
+        total_count: @vacancies_search.total_count,
         vacancies_on_page: vacancy_ids,
         location_polygon_used: polygon_id,
         landing_page: params[:landing_page_slug],
--- a/app/services/search/similar_jobs.rb
+++ b/app/services/search/similar_jobs.rb
@@ -24,9 +24,9 @@ class Search::SimilarJobs
       Search::VacancySearch
         .new(criteria)
         .vacancies
+        .excluding(vacancy)
         .limit(NUMBER_OF_SIMILAR_JOBS * 3) # Fetch more than we need in case some expire while being cached
-        .reject { |job| job.id == vacancy.id }
-        .map(&:id)
+        .pluck(:id)
     end
   end
 end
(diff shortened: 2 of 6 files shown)


app/controllers/vacancies_controller.rb
```diff
@@ -4,7 +4,7 @@ class VacanciesController < ApplicationController
 
   def index
     @vacancies_search = Search::VacancySearch.new(form.to_hash, sort: form.sort)
-    @pagy, @vacancies = pagy(@vacancies_search.vacancies)
+    @pagy, @vacancies = pagy(@vacancies_search.vacancies, count: @vacancies_search.total_count)
```
scruti, line 7: It prevents `Pagy` from triggering an expensive extra full query with COUNT internally for the pagination menu. When `count` is given, use the given value.

In this case, we pass a `total_count` value that has been memoised so that further calls won't trigger the query again.

app/controllers/vacancies_controller.rb
```diff
@@ -49,14 +49,14 @@ def set_headers
 
   def trigger_search_performed_event
     fail_safe do
-      vacancy_ids = @vacancies_search.vacancies.map(&:id)
+      vacancy_ids = @vacancies_search.vacancies.unscope(:order).pluck(:id)
```
scruti, line 52: The idea of sending ALL the results to analytics defeats the pagination point.
The memory impact of mapping through all the unpaginated results here is massive for the request lifespan.

We use `pluck` to mitigate it, so we don't need to instantiate all the search result objects but directly get a list of IDs from DB.

This `pluck` call causes issues with the order criteria, but as we don't need ordered results here, we can remove the ordering and achieve a way more efficient way to get the result IDs.

app/controllers/vacancies_controller.rb
```diff
@@ -49,14 +49,14 @@ def set_headers
...
       polygon_id = DfE::Analytics.anonymise(@vacancies_search.location_search.polygon.id) if @vacancies_search.location_search.polygon
 
       event_data = {
         search_criteria: form.to_hash,
         sort_by: form.sort.by,
         page: params[:page] || 1,
-        total_count: @vacancies_search.total_count,
+        total_count: vacancy_ids.size,
```
scruti, line 59: This triggered the whole search again in DB to then count it.
With the changes on the `total_count` adding memoisation, this should be already memorised now, but having an array with all the result ids, we can pass the array size and ensure we won't be triggering a DB query.

app/services/search/vacancy_search.rb
```diff
@@ -50,7 +50,7 @@ def vacancies
   end
 
   def total_count
-    vacancies.count
+    @total_count ||= vacancies.count
```
scruti, line 53: Memoise so it avoids triggering the `vacancies` scope query each time we retrieve its size.

app/services/search/similar_jobs.rb
```diff
@@ -24,9 +24,9 @@ def similar_job_ids
       Search::VacancySearch
         .new(criteria)
         .vacancies
+        .excluding(vacancy)
         .limit(NUMBER_OF_SIMILAR_JOBS * 3) # Fetch more than we need in case some expire while being cached
-        .reject { |job| job.id == vacancy.id }
-        .map(&:id)
+        .pluck(:id)
```
scruti, line 29: Execute everything in DB, saving the instantiation of all the search results to iterate over them.

app/services/search/vacancy_search.rb
```diff
@@ -36,21 +36,21 @@ def location_search
   end
 
   def wider_search_suggestions
-    return unless vacancies.empty? && search_criteria[:location].present?
+    return unless total_count.zero? && search_criteria[:location].present?
```
scruti, line 39: This was triggering the scope so another new query into DB.

app/views/vacancies/index.html.slim
```diff
@@ -9,7 +9,8 @@
 h1 class="govuk-heading-l" role="heading" aria-level="1"
   = t("jobs.search_result_heading", count: number_with_delimiter(@vacancies_search.total_count))
 
-- if @vacancies_search.active_criteria? && @vacancies.count.positive?
+- any_vacancies = @vacancies.any?
```
scruti, line 12: memoising the paginated results count saves us multiple DB queries. I believe we could even use the already memoised `@vacancies_search.total_count` here and not trigger a new single query.
TODO: Check this.

app/views/vacancies/index.html.slim
```diff
@@ -9,7 +9,8 @@
 h1 class="govuk-heading-l" role="heading" aria-level="1"
   = t("jobs.search_result_heading", count: number_with_delimiter(@vacancies_search.total_count))
 
-- if @vacancies_search.active_criteria? && @vacancies.count.positive?
+- any_vacancies = @vacancies_search.total_count.positive?
+- if @vacancies_search.active_criteria? && any_vacancies
```
scruti, line 13: avoids triggering further counts to DB.

app/controllers/vacancies_controller.rb
```diff
@@ -49,7 +49,7 @@ def set_headers
 
   def trigger_search_performed_event
     fail_safe do
-      vacancy_ids = @vacancies_search.vacancies.map(&:id)
+      vacancy_ids = @vacancies.pluck(:id)
```
scruti, line 52: This change:
- following pagination, sends the current page vacancies instead of all the vacancies. 
- avoids loading the vacancies in memory to loop over to extract the ID. I've locally tested that the `pluck` call works without issues when called over the paginated vacancies.

app/services/search/wider_suggestions_builder.rb
```diff
@@ -15,16 +15,15 @@ def initial_radius
...
   def suggestions
-    RADIUS_OPTIONS
-      .select { |r| r > initial_radius }
-      .map { |radius| [radius.to_s, wider_results(radius)] }
-      .uniq(&:second)
-      .reject { |options| options.second.zero? }
+    @suggestions ||= RADIUS_OPTIONS.select { |r| r > initial_radius }
+                                   .map { |radius| [radius.to_s, wider_results_count(radius)] }
+                                   .uniq(&:second)
+                                   .reject { |options| options.second.zero? }
```
scruti, line 21: Costly computation. Computes the search between 1 and 9 times (depending on initial_radius).

app/services/search/vacancy_search.rb
```diff
@@ -36,21 +36,21 @@ def location_search
...
 
-    Search::WiderSuggestionsBuilder.new(search_criteria).suggestions
+    @wider_search_suggestions ||= Search::WiderSuggestionsBuilder.new(search_criteria).suggestions
```
scruti, line 41: Costly operation. (Does between 1 and 9 search queries to DB, each with a different Radius).

PR #6242, opened 2023-07-27
Add missing association indexes in DB

scruti:
## Trello card URL
- https://trello.com/c/ZFVIvvfF

## Changes in this PR:
Add missing indexes corresponding to active record associations. Replace existing indexes for unique ones when appropriate (for has_one associations).

### DIFF
--- a/db/migrate/20230725115018_add_personal_details_jobseeker_profile_id_index.rb
+++ b/db/migrate/20230725115018_add_personal_details_jobseeker_profile_id_index.rb
@@ -1,0 +1,8 @@
+class AddPersonalDetailsJobseekerProfileIdIndex < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :personal_details, ["jobseeker_profile_id"], name: :index_personal_details_jobseeker_profile_id, unique: true, algorithm: :concurrently
+    remove_index :personal_details, name: :index_personal_details_on_jobseeker_profile_id, algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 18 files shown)


db/migrate/20230725115018_add_personal_details_jobseeker_profile_id_index.rb
```diff
@@ -0,0 +1,8 @@
+class AddPersonalDetailsJobseekerProfileIdIndex < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :personal_details, ["jobseeker_profile_id"], name: :index_personal_details_jobseeker_profile_id, unique: true, algorithm: :concurrently
+    remove_index :personal_details, name: :index_personal_details_on_jobseeker_profile_id, algorithm: :concurrently
```
scruti, line 6: There was a non-unique index for the same attribute. Replaced it.

spec/models/personal_details_spec.rb
```diff
@@ -50,14 +50,6 @@
         end
       end
 
-      context "when there's an existing profile" do
-        before { create(:jobseeker_profile, jobseeker:) }
-
-        it "does not set completed steps" do
-          expect(personal_details.completed_steps).to be_empty
-        end
-      end
```
scruti, line 59: This test setup made nonsense with the `Jobseeker has_one jobseeker_profile` association. The introduction of a unique index made this raise an exception. Removing this test as is bases in a non-valid test setup.

PR #6243, opened 2023-07-28
Add missing uniqueness validations

scruti:
## Trello card URL
- https://trello.com/c/ZFVIvvfF

## Changes in this PR:

Adding a uniqueness validation in model attributes used by DB unique indexes. The DB unique index grants these records cannot be duplicated in DB, but this will add a proper validation error on the model to back it up.

Only adding it when is appropriate, so ignored multiple alerts:
- When generated from internal tool/gem (Devise tokens, slugs from friendly_id).
- When the data is an id/similar imported from GIAS or DFE **Sign-in.**


.database_consistency.yml
```diff
@@ -7,3 +7,50 @@ ActiveStorage::Blob:
   enabled: false
 ActiveStorage::VariantRecord:
   enabled: false
+
```
scruti, line 10: This is like a Rubocop Ignore rule list.
Added comments explaining why we're ignoring these rule breaks.

KyleMacPherson (approved):

PR #6276, opened 2023-08-07
Add IP blocking to Rack Attack

scruti:
The list of IPs to block is configurable through an ENV variable. Any IP listed there will have all its incoming traffic to the service blocked through Rack Attack.

Adding this capability to block a spider bot flooding our services with continuous requests (multiple per millisecond) 24/7 and hasn't modified its request rate even when we added extended throttling periods.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -5,6 +5,11 @@ class Rack::Attack
       @remote_ip ||= (env["action_dispatch.remote_ip"] || ip).to_s
     end
   end
+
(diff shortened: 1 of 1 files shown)


config/initializers/rack_attack.rb
```diff
@@ -5,6 +5,11 @@ def remote_ip
       @remote_ip ||= (env["action_dispatch.remote_ip"] || ip).to_s
     end
   end
+
+  BLOCKED_IPS = ENV.fetch("RACK_ATTACK_BLOCKED_IPS", "").split(",").map(&:strip).freeze # Array of IPs to block
```
scruti, line 9: I have extracted it to a constant so it computes the blocked IP list once when starting the app and doesn't do the computation again for each request.

PR #6285, opened 2023-08-11
Update title for accessibility purposes to reflect which sort by meth…

KyleMacPherson:
…od is used

## Changes in this PR:

## Next steps:


app/views/vacancies/index.html.slim
```diff
@@ -7,7 +7,7 @@
 = render "vacancies/search/page_title_and_description", landing_page: @landing_page
 
 h1 class="govuk-heading-l" role="heading" aria-level="1"
-  = t("jobs.search_result_heading", count: number_with_delimiter(@vacancies_search.total_count))
+  = t("jobs.search_result_heading.sorted_by_#{@vacancies_search.sort.by}", count: number_with_delimiter(@vacancies_search.total_count))
```
scruti, line 10: :raising_hand: I don't think this change matches the ticket description.  From what I understood, the ticket requests to change the page title only, not the displayed HTML header in the page content.

I would expect the `<title>` tags to contain the extra info while the `H1` content stays the same.

scruti (approved):

PR #6294, opened 2023-08-15
Allow downloading job alert and general feedback data by date range

KyleMacPherson:
https://trello.com/c/WHuTu8TM/504-fix-download-function-on-supportal

## Changes in this PR:

Allow users to download general and job alert feedback using the date ranges provided.

## Next steps:

### DIFF
--- a/spec/system/support_users/feedback_spec.rb
+++ b/spec/system/support_users/feedback_spec.rb
@@ -56,7 +56,7 @@ RSpec.describe "Feedback supportal section" do
       comment: "Highly satisfactory",
       occupation: "Student",
       rating: "highly_satisfied",
-      email: "faketestingemail@someemail.com",
+      email: "goodnews@someemail.com",
       user_participation_response: "interested",
       origin_path: "/jobs",
       created_at: "2022-03-16 10:00",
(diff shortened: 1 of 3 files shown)


spec/system/support_users/feedback_spec.rb
```diff
@@ -96,6 +120,15 @@ def testid_for(time)
       expect(csv.second).to eq [other_feedback.created_at.to_s, "identified", "jobseeker", other_feedback.feedback_type, other_feedback.origin_path, other_feedback.email, other_feedback.occupation, other_feedback.rating, other_feedback.comment, other_feedback.category]
     end
 
+    it "allows user to download data for a specific date range" do
```
scruti, line 123: Suggestion: I think the tests would be more solid if they explicitly test that there is no third row in the CSV (or any other way that the other feedback is not included) here. 
Same for the other test without a date range filter (it shouldn't include feedback older than the current month, right?).

scruti (approved):

KyleMacPherson:
[Fix download function on supportal](https://trello.com/c/WHuTu8TM/504-fix-download-function-on-supportal)

PR #6308, opened 2023-08-22
Switch to use job roles across service

scruti:
## Trello card URL

https://trello.com/c/voHMJtsB

## Review App:
https://teaching-vacancies-review-pr-6308.london.cloudapps.digital/

## Changes in this PR:

**TLDR:**
Replaces the usage of `job_role: String` across the service, using `job_roles: [String]` instead.

**Detailed:**
We move the service from populating `Vacancy#job_role` value to setting `Vacancy#job_roles` instead.

Before, the user flow was setting `job_role`, and when the role was `senior_leader`, we were also storing it as `[headteacher, deputy_headteacher, assistant_headteacher]` `job_roles`. Equivalent mapping was done for `middle_leader`.

Now we have inverted the mapping, so when the user flow sets `job_roles`, and when the `job_roles` value is either `headteacher`, `deputy_headteacher` or `assistant_headteacher`, we map it to `senior_leader` in `job_role`. Again, equivalent mapping was done for `middle_leader` roles.

**Heads up:** 
I've defaulted in some areas, when calculating/displaying something based on their `job_role`, to use the first of the `job_roles` values. 
  - For any vacancy created after this change, `job_roles` will have a single value.
  - For existing vacancies that were `senior_leader` or `middle_leader`, `job_roles` contains all the granular senior/middle leadership values. In that case, we cannot choose what is appropriate to use, so we default to the first element.
  
**Bugfix**: We have removed some legacy role landing page that clashed with the new roles enum values and is currently linking to the wrong vacancies.

## Screenshots of UI changes:
image

image

image

image


app/controllers/publishers/vacancies_controller.rb
```diff
@@ -118,9 +118,7 @@ def show_new_features_page?
   end
 
   def publisher_has_used_feature?
-    current_publisher.vacancies.where(enable_job_applications: true).any? do |vacancy|
-      vacancy.job_role == "education_support"
-    end
+    current_publisher.vacancies.where(enable_job_applications: true).with_job_roles("education_support").any?
```
scruti, line 121: `with_job_roles(...)` is a scope method provided by `array_enum` gem.

app/models/job_preferences/job_scope.rb
```diff
@@ -24,27 +18,11 @@ def call
...
-    # EG:
-    # - Jobseeker role preferences: ["headteacher", "assistant_headteacher"]
-    # - Method output: ["headteacher", "assistant_headteacher", "senior_leader"]
-    # - Vacancies with role "senior_leader" will be returned.
-    # - When vacancies are migrated to the new granular roles, vacancies wih roles "headteacher" and/or "assistant_headteacher"
-    #   will be returned.
-    def roles
-      job_preferences.roles.tap do |roles|
-        ROLE_MAPPINGS.each { |generic_role, granular_roles| roles << generic_role if granular_roles.intersect?(roles) }
+    def apply_job_roles(scope)
```
scruti, line 21: `with_any_of_job_roles` is a scope method provided by `array_enum` gem. It raises an exception if there is no values passed (`[]`) to th e call. So had to add this `if/else`  to ensure we cover both situations while returning a valid scope.

app/models/vacancy.rb
```diff
@@ -100,7 +94,7 @@ class Vacancy < ApplicationRecord
                   if: proc { |vacancy| vacancy.listed? }
 
   before_save :on_expired_vacancy_feedback_submitted_update_stats_updated_at
-  before_save :on_job_role_set_job_roles
+  before_save :on_job_roles_set_job_role
```
scruti, line 97: We went from populating `job_roles` on DB when the `job_role` is set from user flow to doing the opposite: 
Now the user flow will populate `job_roles`, and the service will map it to `job_role` in DB.

app/queries/vacancy_filter_query.rb
```diff
@@ -18,10 +18,8 @@ def call(filters) # rubocop:disable Metrics/AbcSize
     built_scope = built_scope.where("vacancies.subjects && ARRAY[?]::varchar[]", filters[:subjects]) if filters[:subjects].present?
 
     # General filters
-    built_scope = built_scope.where(job_role: job_roles(filters[:job_roles])) if job_roles(filters[:job_roles]).present?
+    built_scope = built_scope.with_any_of_job_roles(filters[:job_roles]) if job_roles(filters[:job_roles]).present?
     built_scope = built_scope.ect_suitable if filters[:ect_statuses]&.include?("ect_suitable") || filters[:job_roles]&.include?("ect_suitable")
-    # TODO: Remove this scope when we do not have any more live SEND responsible jobs
-    built_scope = built_scope.where(":job_roles = ANY (job_roles)", job_roles: 2) if filters[:job_roles]&.include?("send_responsible")
```
scruti, line 24: Removing legacy code that is incompatible with new `job_roles` values.

app/vacancy_sources/vacancy_source/source/every.rb
```diff
@@ -42,7 +42,7 @@ def attributes_for(item)
       salary: item["salary"],
       expires_at: Time.zone.parse(item["expiresAt"]),
       external_advert_url: item["advertUrl"],
-      job_role: job_role(item),
+      job_roles: job_roles(item),
```
scruti, line 45: This ATS source is still not running in production. Added the mapping to correspond with [the PR introducing Every integration](https://github.com/DFE-Digital/teaching-vacancies/pull/5799). Will adapt the tests in that PR (or here if 5799 gets merged first).

app/views/home/_landing_page_links.html.slim
```diff
@@ -9,7 +9,6 @@
         - group.with_landing_page "teaching-assistant-jobs"
         - group.with_landing_page "sendco-jobs"
         - group.with_landing_page "education-support-jobs"
-        - group.with_landing_page "send-responsible-jobs"
```
scruti, line 12: removing legacy landing page that matched the wrong vacancy roles.

config/landing_pages.yml
```diff
@@ -17,24 +17,30 @@ shared:
   # criteria is correct and doesn't break the search engine).
 
   ### Job roles
-  send-responsible-jobs:
```
scruti, line 20: legacy landing page that returns wrong vacancies. :axe:

config/locales/jobs.yml
```diff
@@ -34,9 +34,6 @@ en:
       job_filters: Job Filters
       job_roles: Job role
       ect_suitable: Suitable for early career teachers
-      send_responsible: SEND responsibilities
-      send_responsible_option: Has SEND responsibilities (special educational needs and disabilities)
-      send_responsible_only: With SEND responsibilities
```
scruti, line 39: legacy/unused :axe:

config/routes/legacy_redirects.rb
```diff
@@ -23,8 +23,6 @@
 get "teaching-jobs-for-education_support", to: redirect("education-support-jobs")
 get "teaching-jobs-for-education-support", to: redirect("education-support-jobs")
 get "teaching-jobs-for-sendco", to: redirect("sendco-jobs")
-get "teaching-jobs-for-send_responsible", to: redirect("send-responsible-jobs")
-get "teaching-jobs-for-send-responsible", to: redirect("send-responsible-jobs")
```
scruti, line 27: legacy/unused :axe:

spec/helpers/vacancies_helper_spec.rb
```diff
@@ -106,7 +106,7 @@
 
   describe "#vacancy_breadcrumbs" do
     subject { vacancy_breadcrumbs(vacancy).to_a }
-    let(:vacancy) { build_stubbed(:vacancy, :teacher, job_title: "A Job") }
```
scruti, line 109: This `:teacher` trait wasn't defined in the factory :shrug:

spec/queries/vacancy_filter_query_spec.rb
```diff
@@ -131,12 +131,5 @@
       }
       expect(subject.call(filters)).to contain_exactly(vacancy2)
     end
-
-    it "transforms legacy job role filters to new ones" do
-      filters = {
-        job_roles: %w[sen_specialist],
-      }
-      expect(subject.call(filters)).to contain_exactly(vacancy3)
-    end
```
scruti, line 140: Legacy :axe:

spec/system/application_sitemap_spec.rb
```diff
@@ -9,7 +9,7 @@
       document = Nokogiri::XML::Document.parse(body)
       nodes = document.search("url")
 
-      expect(nodes.count).to eq(292)
+      expect(nodes.count).to eq(294)
```
scruti, line 12: This is a magic number. The latest number was changed when a new role was introduced. I understood that introducing new job roles makes the sitemap to include the extra nodes, and... went with it! 
I personally believe this test probably could be axed, but prefer to kick the can forward as was easy to just increase the number as needed :shrug:

app/views/publishers/vacancies/activity_log/show.html.slim
```diff
@@ -12,5 +12,5 @@
           = vacancy_activity_log_item(attribute, diff.last, organisation_type_basic(vacancy.organisation))
 
       p.govuk-body-s class="govuk-!-margin-top-1"
-        | By #{version.actor.papertrail_display_name} on #{version.created_at}
+        | By #{version.actor&.papertrail_display_name.presence || "system"} on #{version.created_at}
```
scruti, line 15: Unrelated bugfix. Noticed the issue on the review app, so fixed it on the same PR.

app/vacancy_sources/vacancy_source/source/every.rb
```diff
@@ -79,12 +79,20 @@ def main_organisation(item)
...
+    return [] if role.blank?
+
+    # Translate legacy senior/middle leader into all the granular roles split from them
+    return Vacancy::SENIOR_LEADER_JOB_ROLES if role.include? "senior_leader"
+    return Vacancy::MIDDLE_LEADER_JOB_ROLES if role.include? "middle_leader"
+
+    Array.wrap(role.gsub("deputy_headteacher_principal", "deputy_headteacher")
+                   .gsub("assistant_headteacher_principal", "assistant_headteacher")
+                   .gsub("headteacher_principal", "headteacher")
+                   .gsub(/head_of_year_or_phase|head_of_year/, "head_of_year_or_phase")
```
KyleMacPherson, line 93: [Minor] I think `head_of_year_or_phase` is redundant here

app/vacancy_sources/vacancy_source/source/fusion.rb
```diff
@@ -93,14 +93,17 @@ def main_organisation(item)
...
-    item["jobRole"]
-      .gsub(/deputy_headteacher|assistant_headteacher|headteacher/, "senior_leader")
-      .gsub(/head_of_year_or_phase|head_of_department_or_curriculum|head_of_year/, "middle_leader")
-      .gsub("learning_support", "education_support")
-      .gsub(/\s+/, "")
+    # Translate legacy senior/middle leader into all the granular roles split from them
+    return Vacancy::SENIOR_LEADER_JOB_ROLES if role.include? "senior_leader"
+    return Vacancy::MIDDLE_LEADER_JOB_ROLES if role.include? "middle_leader"
+
+    Array.wrap(role.gsub(/head_of_year_or_phase|head_of_year/, "head_of_year_or_phase")
```
KyleMacPherson, line 104: [Minor] I think `head_of_year_or_phase` is redundant here

app/vacancy_sources/vacancy_source/source/my_new_term.rb
```diff
@@ -92,14 +92,20 @@ def find_schools(item)
...
-     .gsub(/learning_support|other_support|science_technician/, "education_support")
-     .gsub(/\s+/, "")
+    # Translate legacy senior/middle leader into all the granular roles split from them
+    return Vacancy::SENIOR_LEADER_JOB_ROLES if role.include? "senior_leader"
+    return Vacancy::MIDDLE_LEADER_JOB_ROLES if role.include? "middle_leader"
+
+    Array.wrap(role.gsub("deputy_headteacher_principal", "deputy_headteacher")
+                   .gsub("assistant_headteacher_principal", "assistant_headteacher")
+                   .gsub("headteacher_principal", "headteacher")
+                   .gsub(/head_of_year_or_phase|head_of_year/, "head_of_year_or_phase")
```
KyleMacPherson, line 106: [Minor] I think `head_of_year_or_phase` is redundant here

app/vacancy_sources/vacancy_source/source/ventrus.rb
```diff
@@ -85,14 +85,20 @@ def find_schools(item)
...
-    .gsub(/learning_support|other_support|science_technician/, "education_support")
-    .gsub(/\s+/, "")
+    # Translate legacy senior/middle leader into all the granular roles split from them
+    return Vacancy::SENIOR_LEADER_JOB_ROLES if role.include? "senior_leader"
+    return Vacancy::MIDDLE_LEADER_JOB_ROLES if role.include? "middle_leader"
+
+    Array.wrap(role.gsub("deputy_headteacher_principal", "deputy_headteacher")
+                   .gsub("assistant_headteacher_principal", "assistant_headteacher")
+                   .gsub("headteacher_principal", "headteacher")
+                   .gsub(/head_of_year_or_phase|head_of_year/, "head_of_year_or_phase")
```
KyleMacPherson, line 99: [Minor] I think `head_of_year_or_phase` is redundant here

app/views/publishers/vacancies/vacancy_review_sections/_job_details.html.slim
```diff
@@ -17,7 +17,8 @@ h2 class="govuk-heading-m govuk-!-margin-bottom-4"
     - row.key
       = t("jobs.job_role")
     - row.value
-      = t("helpers.label.publishers_job_listing_job_role_form.job_role_options.#{vacancy.job_role}") if vacancy.job_role?
+      - if vacancy.job_roles.any?
```
KyleMacPherson, line 20: [Minor] I think you might be able to use the `readable_job_roles` method here

KyleMacPherson (changes requested):
Looks great so far, I left some very minor comments but none of them are blocking.

One blocking issue is that I think we need to update the filter mapping found in `config/search/keyword_filter_mappings.yml` where we still have mappings to the old job role such as:

`"head teacher":
  job_roles:
    - senior_leader`

I think this is what is causing the review app to crash when we enter these kind of search terms on the vacancy search page.

app/vacancy_sources/vacancy_source/source/every.rb
scruti: This is tricky to see in the code and easier to understand with a breaking test (that is how I noticed :sweat: ):
When you have `"head_of_year_or_phase".gsub(/head_of_year/, "head_of_year_or_phase")` the output is `head_of_year_or_phase_or_phase`, because it replaces the `head_of_year` substring.

With this regexp `/head_of_year_or_phase|head_of_year/`, the `head_of_year` matching is a fallback, so it doesn't apply if the left side (`head_of_year_or_phase`) matches, avoiding the issue explained above.

We could also achieve this by improving the matcher, to be fair. Maybe we could clean the spaces as a first step (instead of the current last step) and then do a `.gsub(/^head_of_year$/,...)` regexp to be more precise about not replacing it if it is a substring?

KyleMacPherson: Ahhhhh, I see, that makes sense! I don't think we need to add anymore complicated regex, I just didn't think of this scenario 😅

app/views/publishers/vacancies/vacancy_review_sections/_job_details.html.slim
scruti: Good catch, thanks for pointing it out!

KyleMacPherson (approved):

PR #6328, opened 2023-08-29
Distance sorting

KyleMacPherson:
https://trello.com/c/afFGrJ0P/468-adding-distance-as-a-sort-by-function

## Changes in this PR:

This PR allows users to sort by distance if they enter a location in the City, county or postcode text box. By default the results will be sorted by distance if they enter a location, but the user can choose to select other sort options.

Note: brakeman raised warnings about potential sql injection vulnerabilities. After investigating the warnings I have moved them into the brakeman.ignore file with a note for each explaining why I don't believe that they're a cause for concern.

## Next steps:

### DIFF
--- a/app/queries/location_query.rb
+++ b/app/queries/location_query.rb
@@ -8,7 +8,7 @@ class LocationQuery < ApplicationQuery
 
   private
 
-  def call(field_name, location_query, radius_in_miles, sort_by_distance = false)
+  def call(field_name, location_query, radius_in_miles, sort_by_distance: false)
     normalised_query = normalise_query(location_query)
     radius = convert_miles_to_metres(radius_in_miles.to_i)
 
@@ -47,7 +47,7 @@ class LocationQuery < ApplicationQuery
 
   def handle_coordinates(field_name, query, radius, sort_by_distance)
     coordinates = Geocoding.new(query).coordinates
-    
+
     # TODO: Geocoding class currently returns this on error, it should probably raise a
     # suitable error instead. Refactor later!
     return scope.none if coordinates == [0, 0]
@@ -56,7 +56,7 @@ class LocationQuery < ApplicationQuery
     @scope = scope.where("ST_DWithin(#{field_name}, ?, ?)", point, radius)
 
     sort_by_coordinates_distance(field_name, point) if sort_by_distance
-    
+
     scope
   end
 
--- a/app/queries/location_query.rb
+++ b/app/queries/location_query.rb
@@ -8,7 +8,7 @@ class LocationQuery < ApplicationQuery
 
   private
 
-  def call(field_name, location_query, radius_in_miles, sort_by_distance)
+  def call(field_name, location_query, radius_in_miles, sort_by_distance = false)
     normalised_query = location_query&.strip&.downcase
     radius = convert_miles_to_metres(radius_in_miles.to_i)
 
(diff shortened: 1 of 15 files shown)


app/queries/location_query.rb
```diff
@@ -8,31 +8,65 @@ class LocationQuery < ApplicationQuery
...
+    return scope.none if coordinates == [0, 0]
+
+    point = "POINT(#{coordinates.second} #{coordinates.first})"
+    @scope = scope.where("ST_DWithin(#{field_name}, ?, ?)", point, radius)
+
+    sort_by_coordinates_distance(field_name, point) if sort_by_distance
+
+    scope
+  end
+
```
KyleMacPherson, line 62: I have included the sorting logic in this file because the sorting queries rely on all of the same logic we use to filter by distance and I wanted to avoid duplicating a lot of this logic across multiple classes

app/services/search/vacancy_sort.rb
```diff
@@ -1,16 +1,35 @@
...
+  def default_sort_option
+    if location.present?
+      distance_option
+    else
+      publish_on_desc_option
+    end
+  end
+
+  def by_db_column?
+    return true if sort_by == "publish_on_non_default"
```
KyleMacPherson, line 25: this is to allow "publish_on_non_default" to functionally behave like "publish_on"

spec/services/search/radius_builder_spec.rb
```diff
@@ -35,6 +35,14 @@
           end
         end
       end
+
+      context "when radius is not an integer" do
+        let(:radius) { "; DROP TABLE Vacancies; " }
```
KyleMacPherson, line 40: Asserting that user input is sanitised by the Search::RadiusBuilder class to avoid sql injection

KyleMacPherson:
[Adding distance as a sort by function](https://trello.com/c/afFGrJ0P/468-adding-distance-as-a-sort-by-function)

app/queries/location_query.rb
```diff
@@ -8,31 +8,65 @@ class LocationQuery < ApplicationQuery
...
+  def sort_by_polygon_distance(field_name)
+    @scope = scope.select("vacancies.*, ST_Distance(#{field_name}, ST_Centroid(location_polygons.area)) AS distance")
+                  .order(Arel.sql("ST_Distance(#{field_name}, ST_Centroid(location_polygons.area))"))
+  end
+
+  def sort_by_coordinates_distance(field_name, point)
+    @scope = scope.select("vacancies.*, ST_Distance(#{field_name}, '#{point}') AS distance")
+                  .order(Arel.sql("ST_Distance(#{field_name}, '#{point}')"))
```
scruti, line 70: Nice work :clap:

app/services/search/vacancy_search.rb
```diff
@@ -50,18 +49,35 @@ def vacancies
...
+    if sort.by == "publish_on_non_default"
+      "publish_on"
+    else
+      sort.by
+    end
+  end
+
+  def order_scope(scope, sort_by_distance)
+    # if sort_by_distance is true then the sorting is handled by the search_by_filter method so we do not re-order here.
+    return scope if sort_by_distance
```
scruti, line 77: Suggestion: Given that we already have a guard clause checking for the `sort.by` value, I would personally find this more readable if we avoid passing a second parameter and use the `sort` object directly. But this is just how I personally feel about it, so is not a request!
```suggestion
  def order_scope(scope)
    # if sort_by_distance is true then the sorting is handled by the search_by_filter method so we do not re-order here.
    return scope if sort&.by == "distance"
```

scruti (approved):
Great work! This was tricky!

PR #6376, opened 2023-09-13
Update cf dependant app start command

scruti:
## Trello card URL
- https://trello.com/c/gBFEak7S
## Changes in this PR:

The way the deployment command only runs DB migrations on the first DB instance was dependent on CloudFoundry commands.
We replace it with a non-platform dependent solution that works for both CF and AKS:
It will attempt to run the migration in all the instances but will ignore any concurrent migration errors. So the original migration will run while the other instances will be able to run.

We got this approach from Registree Trainee Teachers Service [Link](https://github.com/DFE-Digital/register-trainee-teachers/blob/597443aaeccc8f5d79e3d097bed0bede72f2f439/lib/tasks/migrate_swallowing_concurrent_migration_exceptions.rake)

## How we tested it

1 - Aded a migration with a new field for Vacancies in the PR.
2 - Modified the number of web app instances for review in the PR.
3 - Checked that the Review App was created with the correct number of instances running.
4 - Connected to a Rails console in the Review App, and checked that the new attribute was available for Vacancies (migration ran successfully).
5 - Removed the web app instances/migration changes from the branch.

### DIFF
--- a/documentation/terraform.md
+++ b/documentation/terraform.md
@@ -85,7 +85,7 @@ Occasionally we see `terraform plan` output like this
 ```
 # module.paas.cloudfoundry_app.web_app will be updated in-place
   ~ resource "cloudfoundry_app" "web_app" {
-        command                    = "bundle exec rake cf:on_first_instance db:migrate && rails s"
+        command                    = "bundle exec rake db:migrate:ignore_concurrent_migration_exceptions && rails s"
         disk_quota                 = 1024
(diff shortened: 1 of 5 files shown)


saliceti:
Should we add it to the Dockerfile as well? https://github.com/DFE-Digital/teaching-vacancies/blob/main/Dockerfile#L69C1-L69C56

PR #6414, opened 2023-09-22
Fix: Do not override vacancy source parsing errors

scruti:
## Trello card URL
- https://trello.com/c/YvKSN890

## Changes in this PR:
When the source parsing adds an error to a Vacancy, this custom error was getting overridden with the blind attempt to save the vacancy from the orchestrator class in a later stage.

The attempt to save triggers its own validation that erases any existing error message attached to the vacancy object.

Example:
A vacancy containing an invalid job "FooBar" would get an error attached as "base: ['FooBar is not a valid job_roles']" when trying to assign the import source attributes to a new Vacancy object.

Then, the field itself would become empty due to the failed assignation: "vacancy.job_roles -> []"

If we try to do "vacancy.save" , it fails validation but tells us that job_roles and any other following field in the assignation are Empty.

So the consequences are:

1 - We get a "cannot be blank" error for a field that wasn't blank but
  contained wrong information. Missing essential information for its
  debugging.
2 - WE get a bunch of other field errors for the following fields that were
  present but didn't get assigned due to the first error. These are not
  real errors and only add noise and confusion.

Solution:
If a Vacancy comes from the parsing stage with existing errors, do not attempt to save it and just create a FailedImportedVacancy object with the incoming error. This error will highlight the issue with the source data.

### DIFF
--- a/spec/vacancy_sources/vacancy_source/source/broadbean_spec.rb
+++ b/spec/vacancy_sources/vacancy_source/source/broadbean_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe VacancySource::Source::Broadbean do
   let(:response_body) { file_fixture("vacancy_sources/broadbean.xml").read }
   let(:response) { double("BroadbeanHttpResponse", success?: true, body: response_body) }
 
-  let!(:school1) { create(:school, name: "Test School", urn: "111111", phase: :primary) }
+  let!(:school1) { create(:school, name: "Test School", urn: "111111", phase: :primary, created_at: 1.week.ago) }
   let!(:school_group) { create(:school_group, name: "E-ACT", uid: "12345", schools: schools) }
   let(:schools) { [school1] }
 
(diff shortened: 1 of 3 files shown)


spec/vacancy_sources/vacancy_source/source/broadbean_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:response_body) { file_fixture("vacancy_sources/broadbean.xml").read }
   let(:response) { double("BroadbeanHttpResponse", success?: true, body: response_body) }
 
-  let!(:school1) { create(:school, name: "Test School", urn: "111111", phase: :primary) }
+  let!(:school1) { create(:school, name: "Test School", urn: "111111", phase: :primary, created_at: 1.week.ago) }
```
scruti, line 7: Fixes an unrelated flaky test (ensuring we get the correct order on school creation/association).

KyleMacPherson (approved):

PR #6416, opened 2023-09-22
Fix phase mapping for all sources

scruti:
## Trello card URL

- https://trello.com/c/w9rQOj3T

## Changes in this PR:

Maps the incoming `phase` from all the sources to ensure that the specified values from our TV integration document match the correct value in the DB.

Currently, this is failing in multiple sources, where, for example, they send "through_school" as specified in the integration document. Still, it fails to be imported due to the value in our system only being accepted as `through`.

This mapping resolves this and similar issues.

### Before
image

### DIFF
--- a/app/vacancy_sources/vacancy_source/source/united_learning.rb
+++ b/app/vacancy_sources/vacancy_source/source/united_learning.rb
@@ -60,9 +60,7 @@ class VacancySource::Source::UnitedLearning
       subjects: item["Subjects"].presence&.split(","),
       working_patterns: item["Working_patterns"].presence&.split(","),
       contract_type: item["Contract_type"].presence,
-      # TODO: This is coming through unexpectedly in the feed - the parameterize call can be removed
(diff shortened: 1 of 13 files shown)


app/vacancy_sources/vacancy_source/source/united_learning.rb
```diff
@@ -60,9 +60,7 @@ def attributes_for(item)
       subjects: item["Subjects"].presence&.split(","),
       working_patterns: item["Working_patterns"].presence&.split(","),
       contract_type: item["Contract_type"].presence,
-      # TODO: This is coming through unexpectedly in the feed - the parameterize call can be removed
-      #       when the correct values are coming through
-      phases: [organisations_for(item).first&.readable_phase],
+      phases: phase_for(item),
```
scruti, line 63: Don't know precisely what this comment refers to, but the feed contains the Phase now. So using it.

KyleMacPherson (approved):

PR #6441, opened 2023-09-28
Add warning for applicants without right to work

ddippolito:
## Trello card URL
https://trello.com/c/FY75ESch/520-add-warning-for-applicants-without-right-to-work-applying-for-jobs-without-visa-sponsorship

## Changes in this PR:

This PR adds a warning page and a small warning section for applications without a permit to work in the UK applying for jobs not sponsoring visas.

## Screenshots of UI changes:
warning 1
warning 2

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -27,7 +27,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
 
   # rubocop:disable Style/GuardClause
   def about_your_application
-    if profile.nil? || profile&.personal_details&.right_to_work_in_uk?
+    if profile.nil? || profile&.personal_details&.right_to_work_in_uk? || vacancy.visa_sponsorship_available?
       redirect_to new_quick_apply_jobseekers_job_job_application_path(vacancy.id)
     end
   end


spec/requests/jobseekers/job_applications_spec.rb
```diff
@@ -48,7 +48,7 @@
 
           it "redirects to `new_quick_apply_jobseekers_job_job_application_path`" do
```
KyleMacPherson, line 49: Think we need to change this as it is redirecting to `about_your_application_jobseekers_job_job_application_path` now. Could add a test to assert that it redirects to `new_quick_apply_jobseekers_job_job_application_path` if the jobseeker has the right to work in the uk

KyleMacPherson (approved):
Looks good to me, I think it might be good to add some feature tests to assert that the new warning concept page and the hint text to warn jobseekers appears as expected though

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -25,6 +25,14 @@ def review
...
   end
 
+  # rubocop:disable Style/GuardClause
+  def about_your_application
+    if profile.nil? || profile&.personal_details&.right_to_work_in_uk?
+      redirect_to new_quick_apply_jobseekers_job_job_application_path(vacancy.id)
+    end
+  end
+  # rubocop:enable Style/GuardClause
+
```
scruti, line 35: Question :raising_hand_man: wouldn't this cause a double redirection when the `about_your_application` step is not needed? 
`new -> about_your_application -> new_quick_apply...`

In that case, have you considered grouping all the redirection logic together in the `#new` method?
It would either redirect to `A` or `B` instead of redirecting to `A` and then redirecting to `B` if `A` wasn't necessary.
```ruby
  def new
    request_event.trigger(:vacancy_apply_clicked, vacancy_id: vacancy.id)
    send_dfe_analytics_event

    return unless quick_apply?

    if profile.nil? || profile&.personal_details&.right_to_work_in_uk?
      redirect_to new_quick_apply_jobseekers_job_job_application_path(vacancy.id)
    else
      redirect_to about_your_application_jobseekers_job_job_application_path(vacancy.id)
    end
  end
```

scruti: Fair enough!

scruti (approved):

PR #6446, opened 2023-09-29
Add AKS commands shortcuts to makefile

scruti:
Since we are migrating our services to `AKS`, this PR extends our AKS documentation explaining where our TV environments are hosted and adds `kubectl` commands aliases/shortcuts to facilitate common developer tasks:

- Opening a rails console
- Opening a shell
- Executing a rake task

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,10 +289,10 @@ define get_app_first_pod
 	$(eval POD_NAME=$(shell kubectl -n $(AZURE_NAMESPACE) get pods | grep -- teaching-vacancies-$(env)-[0-9] | awk '{print $$1}'))
 endef
 
-# make review pr_id=5432 ssh
-# make qa ssh
-.PHONY: ssh
-ssh:
+# make review pr_id=5432 shell
+# make qa shell
+.PHONY: shell
+shell:
 	$(call get_app_first_pod)
 	kubectl -n $(AZURE_NAMESPACE) exec -ti $(POD_NAME) -- sh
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -284,6 +284,33 @@ test-cluster:
 get-cluster-credentials: set-azure-account
 	az aks get-credentials --overwrite-existing -g ${CLUSTER_RESOURCE_GROUP_NAME} -n ${CLUSTER_NAME}
 
+define get_app_first_pod
```
saliceti, line 287: I think this is not required as you can do: `kubectl exec -ti deployment/<deployment_name>` and it will automatically take the first pod

Makefile
```diff
@@ -284,6 +284,33 @@ test-cluster:
...
+  $(if $(env), , $(error Missing <env>. Usage: "make <env> command"))
+	$(eval POD_NAME=$(shell kubectl -n $(AZURE_NAMESPACE) get pods | grep -- teaching-vacancies-$(env)-[0-9] | awk '{print $$1}'))
+endef
+
+# make review pr_id=5432 ssh
+# make qa ssh
+.PHONY: ssh
+ssh:
```
saliceti, line 295: It's not really ssh though. Maybe call it `shell:` ?

global_config/production.sh
```diff
@@ -2,4 +2,5 @@ CONFIG=production
 var_file=production
 CONFIG_SHORT=pd
 AZURE_SUBSCRIPTION=s189-teacher-services-cloud-production
+AZURE_NAMESPACE=tv-production
```
saliceti, line 5: The namespace is set in the terraform configuration json: https://github.com/DFE-Digital/teaching-vacancies/blob/main/terraform/workspace-variables/production.tfvars.json#L44
You could remove duplication by reading from the json using `jq`

Makefile
scruti: So much better. Thanks!

Makefile
scruti: So much better. Thanks!

saliceti (approved):

PR #6449, opened 2023-09-29
[709] Fix "Host is taken" on review apps

saliceti:
## Trello card URL
https://trello.com/c/O9goNKOS

## Changes in this PR:
Th review apps tried to create routes and ingresses for external domains but they don't use CDN. We now pass an empty list for the zones so it doesn't try to create them.

## How to review
https://teaching-vacancies-review-pr-6447.test.teacherservices.cloud/


saliceti:
[BUGFIX: TV host is taken on review apps](https://trello.com/c/O9goNKOS/709-bugfix-tv-host-is-taken-on-review-apps)

bj2k61 (approved):

PR #6453, opened 2023-10-02
Update search criteria for subscriptions

ddippolito:
This removes outdated working patterns from searches

## Trello card URL
https://trello.com/c/uwkAJaLs/527-remove-legacy-search-filters-from-subscriptions

## Changes in this PR:
I suspect there's no need to change subscription data in the database as, at some point, they'll become inactive.

### DIFF
--- a/lib/tasks/update_subscription_legacy_search_criteria.rake
+++ b/lib/tasks/update_subscription_legacy_search_criteria.rake
@@ -1,0 +1,47 @@
+# rubocop:disable Metrics/BlockLength
+namespace :subscriptions do
+  desc "Update working_patterns and job_roles in search_criteria for all subscriptions"
+  task update_legacy_search_criteria: :environment do
+    puts "Starting to update working_patterns and job_roles in search_criteria..."
+
+    target_working_patterns = %w[flexible term_time job_share]
+    senior_leader_replacements = %w[headteacher deputy_headteacher assistant_headteacher]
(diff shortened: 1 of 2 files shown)


lib/tasks/update_subscription_legacy_search_criteria.rake
```diff
@@ -0,0 +1,47 @@
...
+namespace :subscriptions do
+  desc "Update working_patterns and job_roles in search_criteria for all subscriptions"
+  task update_legacy_search_criteria: :environment do
+    puts "Starting to update working_patterns and job_roles in search_criteria..."
+
+    target_working_patterns = %w[flexible term_time job_share]
+    senior_leader_replacements = %w[headteacher deputy_headteacher assistant_headteacher]
+    middle_leader_replacements = %w[head_of_year_or_phase head_of_department_or_curriculum]
+
+    Subscription.find_each do |subscription|
```
scruti, line 11: Suggestion: 
Given that performance is a concern, instead of going one by one through all the existing subscriptions (114k), we could iterate and update only the affected subset (27k).

Took some time to read the Postgres Jsonb docs and found the way to query subscriptions that had any of the given values as search criteria job roles or search criteria working patterns:

```ruby
 Subscription.where("search_criteria->'job_roles' ?| array['senior_leader', 'middle_leader']")
  .or(Subscription.where("search_criteria->'working_patterns' ?| array['flexible', 'term_time', 'job_share']"))
```

scruti (approved):

PR #6467, opened 2023-10-06
Validate email address for jobseekers

KyleMacPherson:
## Changes in this PR:

Currently, in production, there is no email validation on the sign up or edit email page. This means that users entering `foo@bar` for example can successfully register an account with this invalid email (or change their email to this invalid email after signing up), although they obviously can't confirm their email address and will see this page:

Screenshot 2023-10-06 at 11 46 15

This results in us (devise) attempting to send a confirmation email resulting in the error `ValidationError: email_address Not a valid email address (Notifications::Client::BadRequestError)` which we have been seeing in sentry. I believe it will also result in this error occurring 5 days later when we attempt to send them another confirmation email/reminder.

This PR adds email_address validation on the jobseeker model which stops a user being able to register an account with an email such as foo@bar which stops the initial email and error being sent, as well as stopping the additional email attempt 4 days later. Using an invalid email will result in this after this change:

Screenshot 2023-10-06 at 12 54 12

## Next steps:

### DIFF
--- a/app/jobs/delete_jobseekers_with_incorrect_emails.rb
+++ b/app/jobs/delete_jobseekers_with_incorrect_emails.rb
@@ -1,0 +1,13 @@
+class DeleteJobseekersWithIncorrectEmails < ApplicationJob
+  queue_as :default
+
+  def perform
+    client = Notifications::Client.new(ENV.fetch("NOTIFY_KEY"))
+
+    failed_email_addresses = client.get_notifications(template_type: "email", status: "permanent-failure").collection.map(&:email_address)
+
+    failed_email_addresses.each do |failed_email|
+      Jobseeker.find_by(email: failed_email)&.destroy
+    end
+  end
+end
--- a/app/jobs/delete_jobseekers_with_incorrect_emails_job.rb
+++ b/app/jobs/delete_jobseekers_with_incorrect_emails_job.rb
@@ -1,13 +1,11 @@
-class DeleteJobseekersWithIncorrectEmails < ApplicationJob
-  queue_as :default
+class DeleteJobseekersWithIncorrectEmailsJob < ApplicationJob
+  queue_as :low
 
   def perform
     client = Notifications::Client.new(ENV.fetch("NOTIFY_KEY"))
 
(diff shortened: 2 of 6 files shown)


scruti (approved):

app/jobs/delete_jobseekers_with_incorrect_emails.rb
```diff
@@ -0,0 +1,13 @@
...
+  queue_as :default
+
+  def perform
+    client = Notifications::Client.new(ENV.fetch("NOTIFY_KEY"))
+
+    failed_email_addresses = client.get_notifications(template_type: "email", status: "permanent-failure").collection.map(&:email_address)
+
+    failed_email_addresses.each do |failed_email|
+      Jobseeker.find_by(email: failed_email)&.destroy
+    end
```
scruti, line 11: :raising_hand_man: 2 suggestions in this block:
1. If I read the code right it seems we could turn it into a single DB query instead of looping over the list (`where...destroy_all`)
2. As we are destroying the Jobseeker maybe we want to make sure that this only happens to unconfirmed users? Just to play safe :crossed_fingers:  (`confirmed_at: nil`).
```suggestion
    Jobseeker.where(email: failed_email_adresses, confirmed_at: nil).destroy_all
```

scruti (approved):

app/jobs/delete_jobseekers_with_incorrect_emails_job.rb
```diff
@@ -0,0 +1,11 @@
+class DeleteJobseekersWithIncorrectEmailsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    client = Notifications::Client.new(ENV.fetch("NOTIFY_KEY"))
+
+    failed_email_addresses = client.get_notifications(template_type: "email", status: "permanent-failure").collection.map(&:email_address)
+
+    Jobseeker.where(email: failed_email_adresses, confirmed_at: nil).destroy_all
```
scruti, line 9:
```suggestion
    Jobseeker.where(email: failed_email_addresses, confirmed_at: nil).destroy_all
```

PR #6468, opened 2023-10-06
Amend flaky broadbean test

ddippolito:
There is an annoying flaky test popping up occasionally in CI.
Does what it says on the tin and fixes it.


scruti (approved):
Thanks. So annoying!

PR #6495, opened 2023-10-23
Ensure that a jobseekers work history is displayed with the most rece…

KyleMacPherson:
…nt job first

## Trello card URL

https://trello.com/c/GLu1nwaG/589-bug-fix-work-history-ordering

## Changes in this PR:

The main change in this PR is reversing the ordering of the work history of a jobseeker as displayed to a jobseeker themselves when viewing or previewing their applications/profiles and to a publisher via the jobseeker application.

As part of this ticket I found the following areas which I felt like I could refactor:
- a misnamed i18n variable. It is named add_another_break but it actually refers to the "Add reason for the break" button. I renamed this in this PR.
- using employment dates, rather than the calculated break dates to display breaks in employment to the user.
- miscalculating break dates.

## Next steps:

### DIFF
--- a/app/views/jobseekers/job_applications/build/employment_history.html.slim
+++ b/app/views/jobseekers/job_applications/build/employment_history.html.slim
@@ -62,7 +62,7 @@
             p.govuk-body
               = govuk_link_to t(employments.job.none? ? "buttons.add_job" : "buttons.add_another_job"), new_jobseekers_job_application_employment_path(job_application)
               = " or "
-              = govuk_link_to t("buttons.add_another_break"), new_jobseekers_job_application_break_path(job_application, started_on: employment.ended_on, ended_on: (employments[index + 1]&.started_on || Date.current))
+              = govuk_link_to t("buttons.add_reason_for_break"), new_jobseekers_job_application_break_path(job_application, started_on: gap[:started_on], ended_on: (gap[:ended_on] || Date.current))
 
       = govuk_button_link_to t(employments.job.none? ? "buttons.add_job" : "buttons.add_another_job"), new_jobseekers_job_application_employment_path(job_application), class: "govuk-button--secondary"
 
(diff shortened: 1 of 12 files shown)


app/views/jobseekers/job_applications/build/employment_history.html.slim
```diff
@@ -62,7 +62,7 @@
             p.govuk-body
               = govuk_link_to t(employments.job.none? ? "buttons.add_job" : "buttons.add_another_job"), new_jobseekers_job_application_employment_path(job_application)
               = " or "
-              = govuk_link_to t("buttons.add_another_break"), new_jobseekers_job_application_break_path(job_application, started_on: employment.ended_on, ended_on: (employments[index + 1]&.started_on || Date.current))
+              = govuk_link_to t("buttons.add_reason_for_break"), new_jobseekers_job_application_break_path(job_application, started_on: gap[:started_on], ended_on: (gap[:ended_on] || Date.current))
```
KyleMacPherson, line 65: here we were calculating the employment gap but not using this calculation to populate the gap info on the page. This caused the employment gap to be miscalculated when we changed the order of the employment history.

config/locales/forms.yml
```diff
@@ -2,7 +2,7 @@ en:
   buttons:
     accept_all_cookies: Accept all cookies
     accept_and_continue: Accept and continue
-    add_another_break: add a reason for this break
```
KyleMacPherson, line 5: this variable name made me think that this link was for adding an additional break in work history, rather than adding a reason for the break in work history. I changed the name to better reflect what it was for (imo).

app/views/jobseekers/job_applications/build/employment_history.html.slim
```diff
@@ -12,7 +12,7 @@
     p.govuk-body = t(".description2")
 
     - if employments.any?
-      - employments.each_with_index do |employment, index|
+      - employments.reverse_each do |employment|
```
scruti, line 15: Didn't know this one :open_mouth:

scruti (approved):
LGTM! Thanks for fixing the unrelated bits.

PR #6535, opened 2023-11-27
Enabling integrations to send visa sponsorship

KyleMacPherson:
## Trello card URL

https://trello.com/c/aCgjtcYK/634-enabling-integrations-to-send-visa-sponsorship

## Changes in this PR:

This PR adds the ability for ATS providers to send visa_sponsorship_available data, defaulting it to false if no visa sponsorship information data is sent.

## Next steps:


spec/fixtures/files/vacancy_sources/ark.xml
```diff
@@ -34,6 +34,7 @@
         https://arkcareers.engageats.co.uk/Vacancies/W/1972/0/247720/19522/tutor-fellow-primary-or-secondary
       </engAts:rssFeedVacancyUrl>
       <pubDate>Tue, 09 Mar 2021 00:00:00 GMT</pubDate>
+      <visaSponsorshipAvailable>true</visaSponsorshipAvailable>
```
scruti, line 37: I have strong suspicions that, if coming, the field will come as part of the `engAts` fields. I think I recall when the value is a custom value added for the Schools to fill outside the standard provided fields, it comes as `<engAts:field>`, but we cannot know 100% until they send it (if they decide to).

scruti (approved):
LGTM. I left one comment on Ark fixture regarding how I think the field will (if) come. But we cannot know so... :shrug:

spec/fixtures/files/vacancy_sources/ark.xml
KyleMacPherson: Ah I see. I am happy to go with your suspicions!

PR #6544, opened 2023-11-29
Move away from Rails Secrets file

scruti:
## Trello card URL

- https://trello.com/c/NkolkFWM

## Changes in this PR:

The Rails secrets file was deprecated with Rails 5.2 in favour of an encrypted credentials file with a master key.

In our service, the secrets file only contained the secret_key_base value.

Moving to Rails 7.1, the usage of the Rails Secrets file became incompatible with the upgrade.

The good news is that, as in our service, we use Environment Variables stored on AWS Parameter Store instead of Secrets/Credentials, stop utilising the secrets.yml file is as easy as removing the file.

`Rails.application.secret_key_base` method looks up for the value defined as ENV variable first, then in the credentials and finally in the secrets.

- https://api.rubyonrails.org/classes/Rails/Application.html#method-i-secret_key_base

This allows us not to change any infrastructure configuration, as our non-test/dev environments already had this value set up as an env variable.

Given we use env variables for configuring our services, we do not need to generate a credentials file for the time being.


lib/message_encryptor.rb
```diff
@@ -20,7 +20,7 @@ def crypt
   end
 
   def secret
-    salt = Rails.application.secrets.secret_key_base
+    salt = Rails.application.secret_key_base
```
scruti, line 23: This method looks up for the secret key base first in the env variable, second in the credentials file and last in the secrets file.
So we can fully move to use ENV vars as done to configure the rest of secrets, and need neither secrets or credentials file.

config/secrets.yml
```diff
@@ -1,35 +0,0 @@
...
-# Do not keep production secrets in the unencrypted secrets file.
-# Instead, either read values from the environment.
-# Or, use `bin/rails secrets:setup` to configure encrypted secrets
-# and move the `production:` environment over there.
-
-production:
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
-
-staging:
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
scruti, line 35: As this env variable was already defined in AWS Parameter store, we don't need to change any infra config and `Rails.application.secret_key_base` will directly pick it up.

KyleMacPherson (approved):

config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  # config.secret_key = 'b5218e6e7ad3bc9c848548a018081ec533c780d875c736798353c0d08edb287daa99e33ad6921a88d6099dea52f78917a37a783373efb9d279796f787b094cc2'
+  config.secret_key = Rails.application.secret_key_base
```
scruti, line 15: Devise kept calling `Rails.secrets.secret_key_base` internally, which would still cause issues with Rails 7.1 upgrade (tested it locally). Setting this value fixes it as it doesn't try to look up for it anymore.

PR #6545, opened 2023-11-30
Dont import from out of scope schools

KyleMacPherson:
## Trello card URL

https://trello.com/c/D1agxlnj/638-remove-vacancies-from-independent-schools

## Changes in this PR:

This PR blocks the import of out of scope schools from the my new term feed and trashes existing vacancies, across all providers, that come from out of scope schools.

Note: The next step will be to block the import of out of scope schools from all ATS providers, but as it is only an issue with MNT currently I have submitted this ticket as a first step to fix the out of scope school problem as quickly as possible.

## Next steps:

### DIFF
--- a/lib/tasks/trash_specific_vacancies.rake
+++ b/lib/tasks/trash_specific_vacancies.rake
@@ -1,15 +1,25 @@
 namespace :vacancies do
   desc "Trash vacancies with specific external references"
   task trash_specific: :environment do
-    external_references_to_trash = %w[
-      10e5719e-6685-46e6-9393-e231ff905fd8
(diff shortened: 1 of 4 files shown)


lib/tasks/trash_specific_vacancies.rake
```diff
@@ -1,15 +1,25 @@
 namespace :vacancies do
   desc "Trash vacancies with specific external references"
   task trash_specific: :environment do
```
scruti, line 3:
```suggestion
  desc "Trash vacancies belonging to non-accepted school types"
  task trash_non_accepted_school_types: :environment do
```

scruti (approved):

PR #6550, opened 2023-12-04
Stop out of scope imports

KyleMacPherson:
## Trello card URL
https://trello.com/c/D1agxlnj/638-block-out-of-scope-vacancies-from-coming-via-integrations

## Changes in this PR:

This PR stops us from importing vacancies from ATS providers which we deem out of scope based on the detailed_school_type. I already added similar functionality [here](https://github.com/DFE-Digital/teaching-vacancies/pull/6545) for the my new term import and this PR extends it across the other providers.

## Next steps:


app/models/concerns/vacancy_source_shared.rb
```diff
@@ -0,0 +1,20 @@
+module VacancySourceShared
```
scruti, line 1: :clap:

scruti (approved):

PR #6555, opened 2023-12-07
Visa sponsorship available filter

KyleMacPherson:
## Trello card URL

https://trello.com/c/5lc334l8/623-visa-sponsorship-filter

## Changes in this PR:

This PR adds:
- filtering by visa_sponsorship_available
- keyword mapping to visa sponsorship filter

## Next steps:

### DIFF
--- a/app/models/concerns/database_indexable.rb
+++ b/app/models/concerns/database_indexable.rb
@@ -39,6 +39,7 @@ module DatabaseIndexable
         organisations.map(&:local_authority_within).reject(&:blank?).uniq,
         organisations.map(&:county).reject(&:blank?).uniq,
         organisations.map(&:region).reject(&:blank?).uniq,
+        VacancyPresenter.new(self).readable_visa_sponsorship_availability,
       ],
     ).tsvector
   end
--- a/lib/tasks/update_serachable_content_for_vacancies_with_visa_sponsorship.rake
+++ b/lib/tasks/update_serachable_content_for_vacancies_with_visa_sponsorship.rake
@@ -1,7 +1,7 @@
 namespace :vacancy do
   desc "Update searchable content for vacancies with visa sponsorship"
   task update_searchable_content: :environment do
-    Vacancy.where(visa_sponsorship_available: true).find_each do |vacancy|
+    Vacancy.visa_sponsorship_available.find_each do |vacancy|
       vacancy.update!(searchable_content: vacancy.generate_searchable_content)
     end
   end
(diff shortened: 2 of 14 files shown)


scruti (approved):
Neat! :clap:

app/models/concerns/database_indexable.rb
```diff
@@ -39,6 +39,7 @@ def generate_searchable_content # rubocop:disable Metrics/AbcSize,Metrics/Method
         organisations.map(&:local_authority_within).reject(&:blank?).uniq,
         organisations.map(&:county).reject(&:blank?).uniq,
         organisations.map(&:region).reject(&:blank?).uniq,
+        VacancyPresenter.new(self).readable_visa_sponsorship_availability,
```
KyleMacPherson, line 42: There was an interesting bug on the review app where when a user entered "visa" as a keyword, the visa_sponsorship_available filter was selected but 0 vacancies were found. When the visa_sponsorship_available was ticket without any keyword, 2 vacancies were found.

This was due to the full text search not finding any mention of the words "visa" in these jobs, despite the fact that the field visa_sponsorship_available was set to true.

In order to fix this behaviour, I added the readable_visa_sponsorship_availability method to the vacancy presenter class and added this to the searchable_content so that any vacancies with visa_sponsorship_available set to true would have "visa sponsorship" added to the searchable_content field.

lib/tasks/update_serachable_content_for_vacancies_with_visa_sponsorship.rake
```diff
@@ -0,0 +1,8 @@
+namespace :vacancy do
+  desc "Update searchable content for vacancies with visa sponsorship"
+  task update_searchable_content: :environment do
+    Vacancy.where(visa_sponsorship_available: true).find_each do |vacancy|
+      vacancy.update!(searchable_content: vacancy.generate_searchable_content)
+    end
+  end
+end
```
KyleMacPherson, line 8: I added this rake task to update searchable_content for each vacancy which had visa_sponsorship_available set to true because otherwise they would not have "visa sponsorship" in their searchable_content until they were edited and would not show up in the filtered results when a user entered "visa" in the keyword search box.

lib/tasks/update_serachable_content_for_vacancies_with_visa_sponsorship.rake
```diff
@@ -0,0 +1,8 @@
+namespace :vacancy do
+  desc "Update searchable content for vacancies with visa sponsorship"
+  task update_searchable_content: :environment do
+    Vacancy.where(visa_sponsorship_available: true).find_each do |vacancy|
```
scruti, line 4: I think you forgot to use the Vacancy scope you defined above:

```suggestion
    Vacancy.visa_sponsorship_available.find_each do |vacancy|
```

PR #6583, opened 2024-01-04
Remove references to recaptcha_reply.

KyleMacPherson:
We recently removed verify_recaptcha which [seems to populate recaptcha_reply](https://github.com/ambethia/recaptcha#recaptcha_reply) so recaptcha_reply is currently always set to nil which is causing us [this issue](https://teaching-vacancies.sentry.io/issues/4806870173/?alert_rule_id=10370581&alert_type=issue&environment=production&notification_uuid=7a4987ab-598e-425e-bf50-5a6a588f16b6&project=6212514&referrer=slack)

This change uses dig to reference the recaptcha_score from the recaptcha_reply to fix the issue.

## Changes in this PR:

## Next steps:

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -81,6 +81,7 @@ class SubscriptionsController < ApplicationController
 
   def notify_new_subscription(subscription)
     # subscription.update(recaptcha_score: recaptcha_reply&.dig("score"))
+    subscription.save!
     Jobseekers::SubscriptionMailer.confirmation(subscription.id).deliver_later
     trigger_subscription_event(:job_alert_subscription_created, subscription)
   end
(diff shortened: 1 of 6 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -80,7 +80,8 @@ def destroy
   private
 
   def notify_new_subscription(subscription)
-    subscription.update(recaptcha_score: recaptcha_reply&.dig("score"))
+    # subscription.update(recaptcha_score: recaptcha_reply&.dig("score"))
+    subscription.save!
```
KyleMacPherson, line 84: `subscription` was being persisted by the update method. As I have commented it out here I needed to call save on the subscription to persist it.

PR #6584, opened 2024-01-04
Dont allow non uk locations

KyleMacPherson:
## Trello card URL

https://trello.com/c/cMb8ZhA5/714-invalidating-location-entries-for-outside-the-uk

## Changes in this PR:

- Stop users from being able to enter locations outside of the UK in job preferences and job alerts
- Adds additional hint text and error messaging around locations in job preferences and job alerts

## Next steps:

### DIFF
--- a/app/form_models/jobseekers/job_preferences_form.rb
+++ b/app/form_models/jobseekers/job_preferences_form.rb
@@ -138,7 +138,7 @@ module Jobseekers
 
       def location_within_united_kingdom
         unless Geocoder.search(location).map(&:country).include?("United Kingdom")
-          errors.add(:location, 'Location is not within United Kingdom')
+          errors.add(:location, I18n.t('activemodel.errors.models.jobseekers/job_preferences_form/location_form.attributes.location.blank'))
         end
       end
     end
--- a/app/form_models/jobseekers/job_preferences_form.rb
+++ b/app/form_models/jobseekers/job_preferences_form.rb
@@ -137,9 +137,9 @@ module Jobseekers
       private
 
       def location_within_united_kingdom
-        unless Geocoder.search(location).map(&:country).include?("United Kingdom")
-          errors.add(:location, I18n.t('activemodel.errors.models.jobseekers/job_preferences_form/location_form.attributes.location.blank'))
-        end
+        return if Geocoder.search(location).map(&:country).include?("United Kingdom")
+
+        errors.add(:location, I18n.t("activemodel.errors.models.jobseekers/job_preferences_form/location_form.attributes.location.blank"))
       end
     end
 
--- a/app/form_models/jobseekers/job_preferences_form.rb
+++ b/app/form_models/jobseekers/job_preferences_form.rb
@@ -127,19 +127,11 @@ module Jobseekers
       attribute :location
       attribute :radius
 
-      validates :location, :radius, presence: true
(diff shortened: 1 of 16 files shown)


app/form_models/jobseekers/job_preferences_form.rb
```diff
@@ -128,10 +128,19 @@ class LocationForm
...
         [0, 1, 5, 10, 15, 20, 25, 50, 100, 200].map { |radius| [radius, I18n.t("jobs.search.number_of_miles", count: radius)] }
       end
+
+      private
+
+      def location_within_united_kingdom
+        return if Geocoder.search(location).map(&:country).include?("United Kingdom")
+
+        errors.add(:location, I18n.t("activemodel.errors.models.jobseekers/job_preferences_form/location_form.attributes.location.blank"))
+      end
```
scruti, line 143: Suggestion :raising_hand: 

If I did read this right, both custom validation methods in the `JobPreferencesForm` and the `SubscriptionForm` do the same and present the same validation error.  

Have you considered extracting this as a custom Validator? It looks like I would be a good fit for it.

EG:
```ruby
class WithinUnitedKingdomValidator< ActiveModel::EachValidator
  def validate_each(record, attribute, value)
    return if value.blank?

    unless Geocoder.search(value).map(&:country).include?("United Kingdom")
      record.errors.add attribute, (blablabla)
    end
  end
end

class JobPreferencesForm
  ...
  validates :location, :radius, presence: true, within_united_kingdom: true
  ...
end
```

spec/requests/subscriptions_spec.rb
```diff
@@ -4,6 +4,11 @@
 RSpec.describe "Subscriptions" do
   let(:vacancy) { create(:vacancy, organisations: [build(:school)]) }
 
+  before do
+    mock_response = [double(country: "United Kingdom")]
+    allow(Geocoder).to receive(:search).and_return(mock_response)
+  end
```
scruti, line 10: :raising_hand_man: Suggestion.

To avoid all this boilerplate on the "default" case of within UK locations, maybe this setup could be extracted to the [rails spec helper](https://github.com/DFE-Digital/teaching-vacancies/blob/0d724e4b0f82c4e9ce3c2ed46a46f46b35c7edd9/spec/rails_helper.rb#L56-L81)?

This way could be invoked on the needed tests with a RSpec tag or even be always defined by default :shrug:

spec/rails_helper.rb
```diff
@@ -77,6 +80,7 @@
   end
 
   config.before(:each, geocode: true) do
+    allow(Geocoder).to receive(:search).and_call_original
```
KyleMacPherson, line 83: I needed to unstub the search method in order for the geocode tests to run as expected!

spec/system/jobseekers/jobseekers_can_manage_their_job_alerts_from_the_dashboard_spec.rb
```diff
@@ -20,9 +20,10 @@
       end
 
       context "when editing a job alert" do
-        before { create(:location_polygon, name: "london") }
-
-        before { click_on I18n.t("jobseekers.subscriptions.index.link_manage") }
+        before do
```
scruti, line 23: :clap:

scruti (approved):
LGTM!

PR #6585, opened 2024-01-05
Exclude non valid schools from org publisher preferences listed schools

scruti:
## Trello card URL
- https://trello.com/c/7DbtpTeY

## Changes in this PR:

It ensures schools belonging to not accepted school types in Teaching Vacancies are not available for selection on the Local Authority publisher preferences.

It includes some refactoring on the original code used to exclude these school types from the external Vacancy Source Imports, so it accommodates extending this inclusion to the rest of the application.


app/vacancy_sources/vacancy_source/source/united_learning.rb
```diff
@@ -50,10 +50,6 @@ def each
 
   private
 
-  def vacancy_listed_at_excluded_school_type?(school)
-    EXCLUDED_DETAILED_SCHOOL_TYPES.include?(school.detailed_school_type) if school
-  end
-
```
scruti, line 56: From what I see, this only differed from the original method in the fact that operates over an individual school as input instead of a collection. 
Opted to use the original method, passing a collection with a single school as an argument.

KyleMacPherson (approved):
Very clean! 👏

PR #6608, opened 2024-01-16
Fix lack of default ActiveStorage service in dev

scruti:
Since Rails 7.1 this config is required. Local development environment was failing due to this being missing. So we aligned it with the rest of environments configurations.


scruti:
If this is the worst that would happen after a Rails upgrade I would sign for it every single time :sweat_smile:

PR #6613, opened 2024-01-17
Re-enable Google recaptcha

scruti:
## Trello card URL
Two tickets in one PR:
- https://trello.com/c/8jalfCz0
- https://trello.com/c/wSpTxvaZ

## Changes in this PR:
Recaptcha was disabled due to lots of false positives after a mailing list communication. Restoring it back.

We also noticed that every visit to the "recaptcha failed" page caused a new error being logged into Sentry.
If you're on the page and reload it, it sends a new "recaptcha error" to Sentry.
This causes lots of false positives due to bots visiting the page.

I resolved it by sending the error to Sentry only on the source of the error when the check was done and failed.

As part of this work, we extracted all the recaptcha-related controller code to a controller concern, and only included it on the relevant controllers that use recaptcha checks on their forms.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,8 +2,6 @@ class ApplicationController < ActionController::Base
   include Pagy::Backend
   include DfE::Analytics::Requests
 
-  SUSPICIOUS_RECAPTCHA_THRESHOLD = 0.5
-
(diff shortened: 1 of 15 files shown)


KyleMacPherson (approved):
Very nice!

PR #6615, opened 2024-01-19
Update gov uk frontend

ddippolito:
## Trello card URL
https://trello.com/c/GkZ0XPQT/726-upgrade-to-govuk-components-v5

## Changes in this PR:
This PR provides a major version upgrade to:

`govuk-frontend` (node package)
`govuk-components` (Ruby gem)


app/notifications/publishers/job_application_data_expiry_notification.rb
```diff
@@ -1,6 +1,7 @@
 class Publishers::JobApplicationDataExpiryNotification < Noticed::Base
   include ActionView::Helpers::UrlHelper
   include GovukLinkHelper
+  include GovukVisuallyHiddenHelper
```
peteryates, line 4: I'm not sure it's necessary to explicitly include this, [I _think_ it should include itself by default](https://github.com/x-govuk/govuk-components/blob/main/app/helpers/govuk_visually_hidden_helper.rb#L15).

peteryates (approved):
Looks great to me 👌🏽 

Left one comment on explicit inclusion of modules, I think they should automatically include themselves (both `GovukLinkHelper` and `GovukVisuallyHiddenHelper`).

app/notifications/publishers/job_application_data_expiry_notification.rb
peteryates: Ah sorry, of course 👌🏽

scruti (approved):

PR #6630, opened 2024-01-27
Temporally disable ImportFromVacancySources job

scruti:
The times this job runs correlate with our server outage downtime times. 
I'm temporarily disabling it to see if it resolves our outage issues.


scruti:
Merging this without dev approval as this is to resolve a server outage, and it just disables one of our scheduled jobs. Want to check if, after releasing this, we stop having DB CPU usage spikes/outages.

PR #6638, opened 2024-01-29
Import from vacancy sources performance improvements

scruti:
## Trello card URL
- https://trello.com/c/CnIP5aPf

## Changes in this PR:
Two main changes in how we import from external vacancy sources, aiming to improve the performance of the service:

#### 1. Parallelise the different imports for each source into independent jobs.
The job used to sequentially go one by one through all the external sources and import their vacancies. 
This takes a long time for a single running job for the worker.

We are keeping the current Job as a scheduled orchestrator. Upon its run, it will queue an independent job for each of the external sources we integrate with.

#### 2. Stop overwriting in DB unchanged vacancies from the feed
The system was importing and re-writing every single Vacancy found in the feed. This was causing thousands of DB writes in big feeds (EG: MyNewTerm).

Then, it used the `not updated since the start of the last run` as a way to identify removed Vacancies from the feed to mark as "removed" in DB.

We are moving to a less DB writing process:
1. Only write the imported vacancy from the feed into DB if is either new or updates the existing vacancy.
2. To identify vacancies from DB to be marked as "removed", we build a list of external reference ids coming from the feed, compare it against the external reference ids for the source live vacancies in our system, and remove from our system the ones that are not present in the feed anymore.

We aim to have less DB heavy importing process.


app/jobs/import_from_vacancy_source_job.rb
```diff
@@ -0,0 +1,84 @@
...
+    report_validation_errors(total_count, errors)
+    mark_removed_vacancies_from_source(source_external_references)
+  end
+
+  def mark_removed_vacancies_from_source(source_external_references)
+    Vacancy.live
+           .where(external_source: @source_name)
+           .where.not(external_reference: source_external_references)
+           .update_all(status: :removed_from_external_system)
+  end
```
scruti, line 46: We mark the records as deleted with a single UPDATE query without validations, callbacks or instantiating the objects in memory. It will simply update the state to `:removed_from_exteranl_system` at DB level.

app/jobs/import_from_vacancy_source_job.rb
```diff
@@ -0,0 +1,84 @@
...
+    report_validation_errors(total_count, errors)
+    mark_removed_vacancies_from_source(source_external_references)
+  end
+
+  def mark_removed_vacancies_from_source(source_external_references)
+    Vacancy.live
+           .where(external_source: @source_name)
+           .where.not(external_reference: source_external_references)
+           .update_all(status: :removed_from_external_system, updated_at: Time.zone.now)
+  end
```
scruti, line 46: We mark the records as deleted with a single UPDATE query without validations, callbacks or instantiation of objects in memory. It will simply update the state to ` :removed_from_exteranl_system` and the `:updated_at` timestamp at DB level.

app/jobs/import_from_vacancy_source_job.rb
```diff
@@ -0,0 +1,84 @@
...
+    total_count = 0
+    source_external_references = []
+    errors = {}
+
+    source_klass.new.each do |vacancy|
+      total_count += 1
+      source_external_references << vacancy.external_reference
+
+      # Avoids unnecessary and expensive DB updates for vacancies that have not changed since they were imported.
+      next if vacancy.persisted? && !vacancy.changed?
```
scruti, line 23: This is the main performance update. It doesn't even try to import it if the assigned attributes into the existing vacancy didn't change.

PR #6641, opened 2024-01-31
Normalise external sources vacancy values before comparing them with stored values

scruti:
## Trello card URL
- https://trello.com/c/CnIP5aPf

## Changes in this PR:
Hundreds of vacancies keep getting re-updated in the DB with every pull from the feeds when they have no changes in the feed from when they were created/updated. 

This is causing the DB CPU to spike while importing external sources vacancies.

While debugging the issue, we noticed an [ActiveRecord callback normalising some data before saving the vacancies](https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/models/concerns/resettable.rb#L4-L27).

This callback, only being applied on a `before :save` call, makes the content from the vacancy loaded from DB to appear to be different to the vacancy values parsed from the feed.

But this is a false positive, as once we attempt to save the feed again, the ActiveRecord callback will normalise the changed attributes, and the result will be the same as the one already stored.

Resolved by normalising the incoming data before comparing it with the DB record for any changes.


app/jobs/import_from_vacancy_source_job.rb
```diff
@@ -4,23 +4,24 @@ class ImportFromVacancySourceJob < ApplicationJob
...
+    @source_klass = source_klass
     @source_name = source_klass.source_name
-    import_from_source(source_klass)
+    @vacancies_count = 0
+    @external_references = []
+    @errors = {}
+
+    import_from_source
+    report_validation_errors
+    mark_removed_vacancies_from_source
```
scruti, line 15: This is a refactor to make `import_from_source` more readable (or at least I think so) and `perform` commands the steps one by one. I find this easier to follow now, but let me know if you think otherwise!

app/jobs/import_from_vacancy_source_job.rb
```diff
@@ -30,18 +31,27 @@ def import_from_source(source_klass)
...
 
-  def mark_removed_vacancies_from_source(source_external_references)
+  # Avoids unnecessary and expensive DB updates for vacancies that have not changed since they were imported.
+  #
+  # Before checking for changes it calls a method that sets to 'nil' multiple invalid field values.
+  # This method is called as an AR callback on the vacancy before it is saved.
+  # Calling it here avoids 'changed?' returning true when, after being normalised by the callback and saved,
+  # the DB value wouldn't change.
+  def existing_vacancy_unchanged?(vacancy)
+    vacancy.reset_dependent_fields
```
scruti, line 47: This little f*er called in a `before :save` from a model concern (even more fun to find!) was what caused the `nil` subjects from the feed to turn into `[]` in DB sometimes, plus vacancies from the feed coming with subjects that were never saved  (and the same in other attributes!)

KyleMacPherson (approved):

PR #6645, opened 2024-01-31
Fix Ark integration org association to vacancies

scruti:
There is a vacancy from Ark associated with 1902 schools because of a bug on this bit.

The bug is caused by searching Orgs by URN even when the URN is nil, which effectively associates the vacancy with any org in the system without a URN (1902 at the moment).

Applying the same Org/Trust associations rules as recently brought to other integrations.


app/vacancy_sources/vacancy_source/source/ark.rb
```diff
@@ -171,11 +171,14 @@ def organisation_fields(schools)
 
   def find_schools(item)
     multi_academy_trust = SchoolGroup.trusts.find_by(uid: TRUST_UID)
-    urn = item.supp_value
+    school_urn = item.supp_value
 
-    multi_academy_trust&.schools&.where(urn: urn).presence ||
-      Organisation.where(urn: urn).presence ||
```
scruti, line 177: `Organisation.where(urn: nil).presence` => 1902 completely unrelated orgs!

KyleMacPherson (approved):

PR #6655, opened 2024-02-07
Reduce LocationPolygons DB queries per search

scruti:
Multiple bits that build up the vacancies search results call over and over the LocationPolygons DB search for the same location.

Reducing this to a minimum by:
- Memoizing the search for polygon from the first time it gets queried.
- Injecting this polygon into subsequent search bits, so it doesn't need to be queried from DB again.

### Before: 7 DB calls
image

### After: 3 DB calls
image


KyleMacPherson (approved):
Wow very nice work 🙇

PR #6683, opened 2024-02-22
Resolve flaky tests

scruti:
## Trello card URL
- https://trello.com/c/kSImWKUs

### Avoid Schools with repeated names being generated
This is causing ambiguous matches over Capybara tests, when factory bot is creating schools. Sometimes they have repeated names and actions over clicking links fail due to the name being present twice.

### Reorder assertions to resolve flaky test
In case the error is happening for asserting over the result ordering before the order has been
applied & the results reloaded.

## Test results
Re-ran the tests about 10 times without any flaky test errors.


KyleMacPherson:
🙇 tyvm

PR #6739, opened 2024-03-19
Update profiles from job application submissions

scruti:
## Trello card URL
https://trello.com/c/QupKHOe3

## Changes in this PR:
When jobseekers submit job applications, they're asked if they want to update their qualifications and/or work history in their profile using the job application data they've just provided.

If they choose so, it will replace their profile info with the job application one.

### On the submission form
image

### After submission
image

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -56,6 +56,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
     raise ActionController::RoutingError, "Cannot submit non-draft application" unless job_application.draft?
 
     if review_form.valid? && all_steps_valid?
+      update_jobseeker_profile!(job_application, review_form)
       job_application.submit!
       @application_feedback_form = Jobseekers::JobApplication::FeedbackForm.new
     else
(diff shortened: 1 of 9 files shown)


spec/models/jobseeker_profile_spec.rb
```diff
@@ -46,4 +46,36 @@
       end
     end
   end
+
+  describe "#replace_qualifications!" do
```
KyleMacPherson, line 50: My anxiety would make me create another qualification that was not associated with this profile so that we can assert that only the correct qualifications were deleted!

spec/system/jobseekers/jobseekers_can_update_their_profile_from_job_applications_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(page).to have_content(I18n.t("jobseekers.job_applications.submit.panel.title"))
+
+    click_link "Your profile"
+
+    expect(page).to have_css("h3.govuk-summary-card__title", text: "Original profile qualification")
+    expect(page).not_to have_css("h3.govuk-summary-card__title", text: "Application qualification")
+
+    expect(page).to have_css("h3.govuk-summary-card__title", text: "Application employment")
+    expect(page).not_to have_css("h3.govuk-summary-card__title", text: "Original profile employment")
+  end
```
KyleMacPherson, line 49: [Minor] I would add an assertion to check that if the user does not check the update profile checkboxes then their employments/qualifications in the profile are not overwritten.

KyleMacPherson (approved):

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -98,6 +99,12 @@ def withdraw
 
   private
 
+  def update_jobseeker_profile!(job_application, form)
+    profile = job_application.jobseeker.jobseeker_profile
```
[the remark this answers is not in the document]
scruti: Very good point! Probably we wouldn't even want to display the question there is no profile set for the jobseeker. Going to guard against it and also raise it with the BA.

PR #6740, opened 2024-03-19
Extend permitted params to include teaching support jobs, non teaching support jobs and visa options

ddippolito:
## Trello card URL
https://trello.com/c/DWTrlcwE/894-job-alerts-bug

## Changes in this PR:

This addresses a bug in the subscription form where only the `teaching_job_roles` were being passed in the parameters of the form, causing `teaching_support_job_roles` and `non_teaching_support_job_roles` to be omitted.
A similar issue was also happening with `visa_sposorship` filter.


KyleMacPherson (commented):
Looks good to me so far but I think adding some tests could be helpful.

Also on the ticket it looks like there is another requirement around the way job roles is showing on the job alerts page (i.e ["teaching"])

KyleMacPherson (approved):

PR #6747, opened 2024-03-21
Enable to skip the status check when manually deleting review apps

scruti:
## Trello card URL
https://trello.com/c/meN4h0BX

## Changes in this PR:
There are dangling review apps stuck for weeks that cannot be deleted because their status check fails.

Adding an option to skip the status check as a requirement to delete the review app allows us to manually delete those, while still checking the status by default.

## Screenshots of UI changes:
It adds an optional checkbox to skip the status check when required.
image

Successfully deletes a stuck review app that previously failed to be deleted due to the status check failing
image


KyleMacPherson:
That's cool, didn't know you could do that type of thing 👌

PR #6755, opened 2024-03-25
Introduce Recaptcha v3 check with v2 fallback

scruti:
## Trello card URL
- https://trello.com/c/cw5A5nw9/910-spike-set-recaptcha-v2-on-forms-after-recaptcha-v3-check-fails

## Changes in this PR:

Recaptcha v3 is used as a first check on the form submissions using Recaptcha.
When the minimum score threshold is not achieved, instead of ignoring it or blocking the user, we will display a Recaptcha v2 check on the form next to the submission button, asking the user to complete it.

Then, the user may be requested to resolve a Recaptcha puzzle to be able to submit the form.

#### Controller level
We are adding a method for the controllers that automates the recaptcha checking the typical flow and speeds up adding recaptcha into different places:

```ruby
recaptcha_protected(form: whatever_form) do
  # code to execute when recaptcha passes
end
```

Using this wrap, there will be a default behaviour when `recaptcha` fails:
- Sets the flag to display the Recaptcha V2 puzzle in the view.
- Adds a validation error to the form, if a form has been given.
- Renders the fallback template, that defaults to `:new` but can be defined to anything on the `recaptcha_protected` call.

If we need different behaviour, we can still use the lower abstraction level `recaptcha_is_valid?` (and related) methods in our controllers and manually build the behaviour as before. But this default with `recaptcha_protected` was good enough to cover all the existing uses of Recaptcha in our service.

#### On the views
Replaced the helper with a shared partial that, depending on a flag being present or not, it will load recaptcha v2 or v3 code on the form page.

It styles Recaptcha V2 to match the Error summary linking as a field and shows the red line style to highlight the need to complete this check for submission.

```
= render "shared/recaptcha", form: f
```

### Originally it doesn't show anything on the form
image
----

### On submission, if (and only if) the Recaptcha V3 check fails, it shows the form again with a "validation" error and asks the user to resolve Recaptcha V2.

image
----

### Upon clicking on the error, it sends the user to the Recaptcha check

image
----

### When clicking the Recaptcha V2 check, it may pass the check or ask the user to resolve a problem.
image
image
----

### Once resolved the recaptcha and re-submitted it does the expected submission/redirection.

image


app/controllers/concerns/recaptcha_checking.rb
```diff
@@ -1,24 +1,35 @@
...
   end
 
-  def log_invalid_recaptcha(form: nil, score: nil)
-    form_name = form.class.name.gsub("::", "").underscore.humanize if form.present?
-    Sentry.with_scope do |scope|
-      scope.set_tags("form.name": form_name, "recaptcha.score": score)
-      Sentry.capture_message("Invalid recaptcha", level: :warning)
-    end
+  def recaptcha_v3_is_valid?(model = nil)
+    return false
```
scruti, line 25: For enforcing V2 recaptcha on review testing. Remove before releasing.

app/helpers/application_helper.rb
```diff
@@ -25,10 +25,6 @@ def meta_description
     end
   end
 
-  def recaptcha
-    recaptcha_v3(action: controller_name, nonce: request.content_security_policy_nonce)
-  end
```
scruti, line 30: Turned into a partial view `shared/recaptcha`

app/controllers/errors_controller.rb
```diff
@@ -33,10 +33,4 @@ def internal_server_error
       format.json { render json: { error: "Internal server error" }, status: :internal_server_error }
     end
   end
-
-  def invalid_recaptcha
-    respond_to do |format|
-      format.html { render status: :unauthorized }
-    end
-  end
```
scruti, line 41: Became unreachable as we won't send users here anymore.

app/controllers/concerns/recaptcha_checking.rb
```diff
@@ -1,24 +1,33 @@
...
+  def recaptcha_is_valid?
+    recaptcha_v3_is_valid? || recaptcha_v2_is_valid?
   end
 
-  def log_invalid_recaptcha(form: nil, score: nil)
-    form_name = form.class.name.gsub("::", "").underscore.humanize if form.present?
-    Sentry.with_scope do |scope|
-      scope.set_tags("form.name": form_name, "recaptcha.score": score)
-      Sentry.capture_message("Invalid recaptcha", level: :warning)
-    end
```
scruti, line 22: Now we fallback into a puzzle instead of recording/sending the failed score to Sentry.

KyleMacPherson (approved):
Nice work @scruti 🙇

PR #6781, opened 2024-04-11
change aria-owns to aria-controls

KyleMacPherson:
## Trello card URL

https://trello.com/c/EPRpNbM5/923-a11y-412-name-role-value-required-aria-attributes-must-be-provided

## Changes in this PR:

This PR changes aria-owns "subjects__listbox" to aria-controls "subjects__listbox" to more accurately represent the relationship between the subjects search box and the checkboxes

### Before

Screenshot 2024-04-11 at 16 05 40

### After
Note: I think the additional 2 issues are due to the review app banner.
Screenshot 2024-04-11 at 16 05 49

## Next steps:

### DIFF
--- a/app/components/searchable_collection_component/searchable_collection_component.html.slim
+++ b/app/components/searchable_collection_component/searchable_collection_component.html.slim
@@ -2,7 +2,7 @@
   div class="#{border_class} #{scrollable_class}"
     - if searchable?
       .searchable-collection-component__search
-        input.govuk-input.icon.icon--left.icon--search.js-action data-searchable-collection-target="input" data-action="input->searchable-collection#input" placeholder=@text[:placeholder] aria-label=@text[:aria_label] aria-expanded="true" aria-describedby=@text[:aria_describedby] role="combobox"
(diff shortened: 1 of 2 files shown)


app/components/searchable_collection_component/searchable_collection_component.html.slim
```diff
@@ -2,7 +2,7 @@
   div class="#{border_class} #{scrollable_class}"
     - if searchable?
       .searchable-collection-component__search
-        input.govuk-input.icon.icon--left.icon--search.js-action data-searchable-collection-target="input" data-action="input->searchable-collection#input" placeholder=@text[:placeholder] aria-label=@text[:aria_label] aria-expanded="true" aria-describedby=@text[:aria_describedby] aria-owns="subjects__listbox" role="combobox"
+        input.govuk-input.icon.icon--left.icon--search.js-action data-searchable-collection-target="input" data-action="input->searchable-collection#input" placeholder=@text[:placeholder] aria-label=@text[:aria_label] aria-expanded="true" aria-describedby=@text[:aria_describedby] aria-controls="subjects__listbox" role="combobox"
```
scruti, line 5: Apparently, this is neither valid as introduces a new issue:
image

scruti: Oh, this is fun. The issue is detected because the `subjects__listbox` id doesn't exist until a value is introduced on the subjects searchbox. Then the `id` gets added dynamically and the accessibility issue disappears:

image

 How can we avoid it? :thinking:

PR #6782, opened 2024-04-12
 Add 'id' to checkbox list

ddippolito:
The `aria-owns` attribute in the searchable collection's input referenced an `id` that didn't exist in the DOM structure. This resulted in an invalid association between the input field and the checkbox list, which was spotted in our accessibility audit and marked as an error.

This commit added an 'id' attribute to the `div` containing the checkboxes to resolve this.

These changes significantly improve the accessibility of the searchable collection. By creating a valid association between the containing input and the list of checkboxes, assistive technologies can now better understand the relationship. This results in a more seamless user experience for individuals relying on these technologies.


scruti (approved):
LGTM!

PR #6793, opened 2024-04-17
Add aria landmarks

ddippolito:
## Trello card URL
https://trello.com/c/TB1y7xDC/925-a11y-131-info-and-relationships-page-regions-not-identified-with-aria-landmarks

## Changes in this PR:

This PR adds a few aria landmarks throughout the site. The govuk components backlink helper doesn't provide a way to use roles and aria labels so we're changing to plain markup.

### DIFF
--- a/app/components/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
+++ b/app/components/vacancy_form_page_heading_component/vacancy_form_page_heading_component.html.slim
@@ -1,6 +1,5 @@
 - content_for :breadcrumbs do
-  a.govuk-back-link aria-label="Back navigation" role="navigation" href=back_path
-    = t("buttons.back")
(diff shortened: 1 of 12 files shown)


scruti:
Hi @ddippolito . 
Regarding the back-links. Govuk components backlink helper allows to pass html attributes:
https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/back_link_component.rb#L4

I tested it locally and seems to work: 
```slim
= govuk_back_link text: t("app.back_to_homepage"), href: root_path, html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
```
Translates to:
```html
<a class="govuk-back-link" aria-label="Back navigation" role="navigation" href="/">Return to homepage</a>
```

scruti (approved):
LGTM!

PR #6797, opened 2024-04-17
[1654] Add infra team statuscake contact group

saliceti:
## Trello card URL
https://trello.com/c/Z6eliKhq

## Changes in this PR:
We have created a separate infra group to avoid duplication and added it to all services


scruti (approved):
Thanks @saliceti

PR #6801, opened 2024-04-18
a11y: Combine first name last name into a singular row on profile page

KyleMacPherson:
## Trello card URL

https://trello.com/c/jvCZD72y/951-a11y-244-link-purpose-in-context-redundant-links-present-name-in-profiles

## Changes in this PR:

Prior to this change we had accessibility issues stemming from the "Change first name" and "Change last name" links going to the same page. This change addresses the issue by combining both names under a "Name" row so that there is only one link to change name for a user on the jobseeker profile.

Note: I didn't make any changes to the forms/fields in the database as I don't think this is necessary for a purely presentational issue but I am open to other opinions on that!

### Before
Screenshot 2024-04-19 at 12 20 20

### After
Screenshot 2024-04-19 at 12 19 46

## Next steps:

### DIFF
--- a/app/views/jobseekers/profiles/personal_details/_summary.html.slim
+++ b/app/views/jobseekers/profiles/personal_details/_summary.html.slim
@@ -2,16 +2,19 @@ ruby:
   form = Jobseekers::Profile::PersonalDetailsForm.from_record(profile.personal_details)
 
   attributes = {
-                 first_name: nil,
-                 last_name: nil,
                  phone_number_provided: { map_values: ->(value) { value.nil? ? "" : t("helpers.label.personal_details_form.phone_number_provided_options.#{value}") } },
                  phone_number: nil,
                  right_to_work_in_uk: { map_values: ->(value) { value.nil? ? "" : t("jobseekers.profiles.personal_details.work.options.#{value}") } },
                }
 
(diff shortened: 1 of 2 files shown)


app/views/jobseekers/profiles/personal_details/_summary.html.slim
```diff
@@ -11,6 +11,7 @@ ruby:
 
   summary = govuk_summary_list(html_attributes: { id: "personal_details" }) do |summary_list|
     attributes.each do |attribute, options|
+      next if attribute == :last_name
```
scruti, line 14: Non-blocking suggestion: 
Instead of having this conditional logic. Could the first row for the name be built separately as the `email` row is done at the end?

From a shallow look, it seems we could remove `first_name` and `last_name` from the `attributes` hash and have something like:
```ruby
summary_list.with_row do |row|
  row.with_key text: "Name"
  row.with_value text: "#{form.first_name} #{form.last_name}"
  row.with_action(text: t("buttons.change"), href: edit_personal_details_jobseekers_profile_path(step: "hardcoded_step_name"), visually_hidden_text: " name")
  
attributes.each do |attribute, options|
# Code without conditionals or skipping
...
```

For your consideration, but not blocking!

PR #6810, opened 2024-04-22
a11y: Move cookies banner to top of page

KyleMacPherson:
## Trello card URL

https://trello.com/c/DdQozyE9/963-a11y-2411-focus-not-obscured-minimum-user-interface-components-are-obscured-when-receiving-keyboard-focus

## Changes in this PR:

During the accessibility audit it was observed that while the cookie banner is in show, the user can still tab to the background content meaning that user interface components are obscured by this banner. It would be expected that, even though the cookie banner is on screen, the user can still see all content on the page when tabbed to. Moving the cookie banner to the top of the page should fix this issue.

## Next steps:

### DIFF
--- a/app/views/layouts/_skip_links.html.slim
+++ b/app/views/layouts/_skip_links.html.slim
@@ -1,4 +1,2 @@
-- if show_cookies_banner?
-  = govuk_skip_link(href: "#cookie-consent", text: t("app.skip_link.cookie"))
 = govuk_skip_link(href: "#main-content", text: t("app.skip_link.main"))
 = content_for :skip_links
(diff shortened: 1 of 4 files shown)


app/views/layouts/_skip_links.html.slim
```diff
@@ -1,4 +1,2 @@
-- if show_cookies_banner?
-  = govuk_skip_link(href: "#cookie-consent", text: t("app.skip_link.cookie"))
```
scruti, line 2: If not used anywhere else, could we remove the locale definition? `"app.skip_link.cookie"`

scruti (approved):

PR #6819, opened 2024-04-24
a11y: Reorganising html elements for accessibility purposes

KyleMacPherson:
## Trello card URL

https://trello.com/c/FNiiPNGK/973-a11y-412-name-role-value-incorrect-announcements-candidate-profile-page-hrs-in-the-heading-elements

## Changes in this PR:

Prior to this change, qualifications and work history entries on jobseeker profiles had all the details nested within an h2 tag which caused issues for users which need to use screen readers. This change reorganises these elements in a more accessible way.

## Next steps:


scruti (approved):
LGTM!

PR #6820, opened 2024-04-25
a11y: Visual changes to jobseeker profile layout as viewed by publisher

KyleMacPherson:
## Trello card URL

https://trello.com/c/V89fYh6H/962-a11y-144-resize-text-rendering-issue-observed-zooming-text-only

## Changes in this PR:

In the recent accessibility report it was flagged that when zooming in on the publishers jobseeker profile page the user loses some of the context of the buttons. This change aims to rectify this issue by putting the buttons in a full width column.

## Next steps:

### DIFF
--- a/app/views/publishers/jobseeker_profiles/_layout.html.slim
+++ b/app/views/publishers/jobseeker_profiles/_layout.html.slim
@@ -4,7 +4,7 @@
   = govuk_back_link text: t("buttons.back"), href: publishers_jobseeker_profiles_path, html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
 
 .govuk-grid-row
-  .govuk-grid-column-two-thirds
+  .govuk-grid-column-full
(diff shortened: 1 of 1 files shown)


app/views/publishers/jobseeker_profiles/_layout.html.slim
```diff
@@ -21,4 +21,5 @@
             data-clipboard-target="button"
             class="js-action govuk-button govuk-button--secondary govuk-!-margin-bottom-2 clipboard-copy"]
         | Copy email
+  .govuk-grid-column-two-thirds
```
scruti, line 24: Non-blocking question: Shouldn't this 2/3 column be nested under a separate/new row?

In the original layout, we had everything under a 2/3 column in a single row. Now we have 2/3 column and a full-size column together within the same row.
Combining 2/3 + Full in a single row seems contradictory.  So given that this is going to be displayed below of the full-size column content, it seems to belong to a different row?

scruti (approved):
Left a question/suggestion for the layout structuring.

PR #6835, opened 2024-05-02
Add Training and CPD section to job application

KyleMacPherson:
## Trello card URL

https://trello.com/c/pWFL7Dx1/918-add-cpd-section-application

## Changes in this PR:

This PR adds a Training and CPD section to the application form which allows jobseekers to enter details about training and continuing professional development courses they have completed and displays this data to the hiring staff.

## Next steps:


app/models/jobseeker_profile.rb
```diff
@@ -70,6 +70,13 @@ def replace_employments!(new_employments)
     end
   end
 
+  def replace_training_and_cpds(new_training_and_cpds)
```
scruti, line 73:
```suggestion
  def replace_training_and_cpds!(new_training_and_cpds)
```

scruti (approved):
Approved with a question/name convention suggestion!

app/models/jobseeker_profile.rb
KyleMacPherson: Ah yeah, good call. Will make that change!

Yeah it's used here! https://github.com/DFE-Digital/teaching-vacancies/pull/6835/files#diff-fa61ca8687a1914d46866fd3c81eddf807b27cb0076442bdf1b96d03100966f7R109

scruti: Thanks for checking!

PR #6845, opened 2024-05-08
Replace editor bullet point button with hint text

scruti:
## Trello card URL
- https://trello.com/c/WsMIQ6N8

## Changes in this PR:

The text editor "toggle bullet point" button was highlighted as an accessibility issue in our recent accessibility assessment.

The team decided it was not worth keeping it and we can display a hint text instead.

### Before
image

### After
image

### DIFF
--- a/app/components/editor_component/editor_component.html.slim
+++ b/app/components/editor_component/editor_component.html.slim
@@ -2,12 +2,9 @@
   .editor-component__controls
     label class="#{label_classes}" data-action="click->editor#focus" for="#{@field_name}-field" id=@label[:id]
       = @label[:text]
-    - if @hint.present?
-      .govuk-hint id="#{@field_name}-hint"
-        = @hint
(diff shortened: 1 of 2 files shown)


app/components/editor_component/editor_component.html.slim
```diff
@@ -2,12 +2,9 @@
   .editor-component__controls
     label class="#{label_classes}" data-action="click->editor#focus" for="#{@field_name}-field" id=@label[:id]
       = @label[:text]
-    - if @hint.present?
-      .govuk-hint id="#{@field_name}-hint"
-        = @hint
+    .govuk-hint id="#{@field_name}-hint"
+      = @hint.presence || "You can copy and paste any information and keep bullet point formatting."
     .editor-component__toolbar data-editor-target="toolbar"
```
scruti, line 7: Removing this line breaks the input behaviour. So I ended up leaving it while removing the toolbar button only.

PR #6854, opened 2024-05-14
Refactor DFE Sign In related code

scruti:
This work is part of an overhaul of how we structure/namespace the code related to external integrations before adding new integrations to the service.

This is a refactor, so the behaviour should stay the same. The principles behind the refactoring are:
- Define clear boundaries between usage of "lib" and "app/services" project folders:
  - The library folder is left to domain-agnostic code that could be extracted into a gem and doesn't depend on our service domain. In this case, the library folder is exclusively used for  DFE Sign In API-related code.
  - The services folder is used for code that depends on our domain. In this case, we use code relating to parsing, exporting and writing into DB the DSI user's info obtained from the API calls.

- Structure the DSI-related code following Object Oriented Programming principles.
  - Instead of having a collection of methods around, we grouped and extracted different DSI-related objects with different responsibilities.
  - Used modules/classes to namespace and group the DSI code.
  - Normalised the service objects public interface methods to `#call`.

## Before
image

## After
image
image


app/jobs/export_dsi_users_to_big_query_job.rb
```diff
@@ -1,13 +1,10 @@
-require "export_dsi_users_to_big_query"
-require "export_dsi_approvers_to_big_query"
```
scruti, line 2: This code is domain-dependant and has been moved from library to service objects. There is no need to require it anymore.

app/jobs/export_dsi_users_to_big_query_job.rb
```diff
@@ -1,13 +1,10 @@
...
 class ExportDSIUsersToBigQueryJob < ApplicationJob
   queue_as :low
 
   def perform
     return if DisableExpensiveJobs.enabled?
 
-    ExportDSIUsersToBigQuery.new.run!
-    ExportDSIApproversToBigQuery.new.run!
+    Publishers::DfeSignIn::BigQueryExport::Users.new.call
+    Publishers::DfeSignIn::BigQueryExport::Approvers.new.call
```
scruti, line 8: - Used modules/classes to structure and namespace the code. 
- Normalised the service objects public interface to `.call`

app/services/publishers/dfe_sign_in/big_query_export/base.rb
```diff
@@ -0,0 +1,26 @@
+require "dfe_sign_in/api"
+require "google/cloud/bigquery"
+
+module Publishers::DfeSignIn::BigQueryExport
+  class Base
+    include DfeSignIn::API
+    include Publishers::DfeSignIn::Parsing
```
scruti, line 7: Using composition through modules, we can combine library code (API calls) with the domain-related parsing module.

app/services/publishers/dfe_sign_in/update_users_in_db.rb
```diff
@@ -0,0 +1,49 @@
+require "dfe_sign_in/api"
+
+module Publishers::DfeSignIn
+  class UpdateUsersInDb
+    include DfeSignIn::API
+    include Publishers::DfeSignIn::Parsing
```
scruti, line 6: This class has nothing to do with BigQ exports. 
Both share the DSI API and Parsing code through composition while keeping the BigQ export-related code exclusively within its namespace/classes.

lib/dfe_sign_in/api.rb
```diff
@@ -0,0 +1,43 @@
+require "dfe_sign_in/api/request"
+require "dfe_sign_in/api/response"
+
+module DfeSignIn
+  module API
```
scruti, line 5: The original API code has been completely restructured into different modules/classes.

lib/dfe_sign_in/api/response.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def number_of_pages
+        raise(error_message) if response["numberOfPages"].nil?
+
+        response["numberOfPages"]
+      end
+
+      def users
+        raise(DfeSignIn::API::Response::NilUsersError, error_message) if users_nil_or_empty?
```
scruti, line 19: As this error was always raised when iterating over the response users on a higher abstraction level, I have moved it directly to the `#users` call to improve cohesion and reduce the coupling from the class calling for the results.

KyleMacPherson (approved):

PR #6896, opened 2024-06-05
Fix Sentry rule ignoring Safari parse JSON errors

scruti:
There is a well-known issue with Safari browsers parsing JSON information. This triggers a JS browser error that doesn't affect the user navigation but causes tens of thousands of errors consuming our Sentry quota.

We had a rule to ignore these so our quota doesn't get hit, but the wording of the error slightly changed, causing the ignore rule to be bypassed.

image

image

image

### DIFF
--- a/app/assets/javascript/application.js
+++ b/app/assets/javascript/application.js
@@ -34,7 +34,7 @@ Sentry.init({
   release: window.sentryConfig.release,
   integrations: [],
   tracesSampleRate: 0, // Disable tracing (performance monitoring, doesn't impact errors)
-  ignoreErrors: [/'Object\.prototype\.hasOwnProperty\.call\(o,"telephone"\)'/],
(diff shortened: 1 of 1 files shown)


app/assets/javascript/application.js
```diff
@@ -34,7 +34,7 @@ Sentry.init({
   release: window.sentryConfig.release,
   integrations: [],
   tracesSampleRate: 0, // Disable tracing (performance monitoring, doesn't impact errors)
-  ignoreErrors: [/'Object\.prototype\.hasOwnProperty\.call\(o,"telephone"\)'/],
+  ignoreErrors: [/'Object\.prototype\.hasOwnProperty\.call\([eo],"telephone"\)'/],
```
scruti, line 37: Now the message comes with `e, "telephone"` instad of `o, "telephone"`. It will ignore both notations now.

KyleMacPherson (approved):

PR #6933, opened 2024-06-24
Schedule export service to DWP find a job

scruti:
## Trello card URL
- https://trello.com/c/MGIXSJIZ

## Changes in this PR:

- schedule daily export of vacancies to DWP Find a Job service.
- extend the documentation with information about the schedule and how to debug/test the integration.


app/jobs/export_vacancies_published_and_updated_since_yesterday_to_dwp_find_a_job_service_job.rb
```diff
@@ -0,0 +1,9 @@
+class ExportVacanciesPublishedAndUpdatedSinceYesterdayToDwpFindAJobServiceJob < ApplicationJob
```
scruti, line 1: I know these class names are very verbose/long but my main goal is that someone can easily see what the job does without having to jump into deeper code.

"The job exports vacancies published and updated since yesterday to DWP Find a Job service"

PR #6937, opened 2024-06-26
Application data

KyleMacPherson:
## Trello card URL

https://trello.com/c/9mSB17Pd/1048-application-data-saving

## Changes in this PR:

This PR changes the logic of how we pre-populate quick apply job applications. Prior to this change, if a profile existed we used that to populate the application, and only used a previous application if no profile existed. This PR switches that logic around to pre-populate using a previous application where possible, and only use a jobseeker profile as back up.

It also adds in some additional changes:
- we now pre-populate personal statements
- we now use an imported tag to display data that has been imported from a previous application (but not a jobseeker profile)
- adds in some warning data to let users know that they cannot re-submit a withdrawn application.
- refactors the code in the quick apply controller and the quick apply class to switch the logic out of the controller and simplify the logic.

## Next steps:

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -1,4 +1,3 @@
-# rubocop:disable Metrics/ClassLength
 class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseController
   include Jobseekers::QualificationFormConcerns
 
@@ -42,7 +41,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
     raise ActionController::RoutingError, "Cannot quick apply if there's no profile or non-draft applications" unless quick_apply?
 
     new_job_application = prefill_job_application_with_available_data
-    
+
     redirect_to jobseekers_job_application_review_path(new_job_application)
   end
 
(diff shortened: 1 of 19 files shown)


app/services/jobseekers/job_applications/quick_apply.rb
```diff
@@ -7,17 +7,72 @@ def initialize(jobseeker, vacancy)
...
 
   def job_application
-    new_job_application.assign_attributes(recent_job_application.slice(*attributes_to_copy).merge(completed_steps: completed_steps, in_progress_steps: in_progress_steps))
-    copy_qualifications
-    copy_employments
-    copy_references
-    copy_training_and_cpds
+    return new_job_application unless previously_submitted_application? || jobseeker_profile
+
+    source = previously_submitted_application? ? recent_job_application : jobseeker_profile
```
scruti, line 12: Suggestion :raising_hand_man: 
I think you could avoid the extra query triggered by `previously_submitted_application?` If you directly use `recent_job_application` (that may be `nil` if no recent applications exist).
```ruby
    source = recent_job_application || jobseeker_profile
    return new_job_application unless source
```

Then you could get rid of `previously_submitted_application?` definition.

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -208,52 +205,6 @@ def send_dfe_analytics_event
     end
   end
 
-  def prefill_application(application)
```
scruti, line 211: Great choice removing all this code from the controller. Much better :clap:

scruti (approved):
LGTM! :clap:

PR #6946, opened 2024-07-04
Remove DfE::Analytics::SendEvents from logs

scruti:
These jobs are constantly run and clog our Logit.io instances due to the high volume of updates. 
We do not need them Logit as doesn't give us any valuable info/context.

We're filtering them out through Semantic logger.


saliceti (approved):

saliceti:
[Remove dfe analytics logs from Logit.io](https://trello.com/c/fxSQSbA6/1907-remove-dfe-analytics-logs-from-logitio)

PR #6962, opened 2024-07-16
Redirects for user guides

KyleMacPherson:
## Trello card URL

https://trello.com/c/XZyMNJAQ/1090-tv-user-guide-redirects-campaign-site-merger

## Changes in this PR:

We recently changed the paths to our user guides, however over 60 other services still link to the previous paths and we have recently sent out an email campaign with using the old links, which are currently broken. 

This change redirects to the appropriate pages in order to ensure that the links on other services and the email campaign work as expected.. I made the changes in the routes rather than add logic to the controller to avoid bloating it with code related to legacy links. I have hardcoded the redirects because we already have some constraints in the routes file which match jobseeker-guides and get-help-hiring.

## Next steps:

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,6 +12,40 @@ Rails.application.routes.draw do
 
   get "check" => "application#check"
 
+  get "/get-help-hiring/how-to-approve-access-for-hiring-staff", to: redirect { |_params, _request|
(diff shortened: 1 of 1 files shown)


rjlynch (approved):
I think this approach is nice and straight forward, much more direct than doing this in the controller 👍

EllieNodder (approved):

PR #6974, opened 2024-07-24
change email copy for you haven't subscribed

joegibb:


scruti (changes requested):
We also need to remove `mail_to` argument from the places where `not_requested.body` is used/called.

Adding it in a bit...

scruti (approved):
Removed the mail address argument from the call and tested it on the review app. All good!

image

PR #6999, opened 2024-08-13
Install dfe-frontend and use card component

ddippolito:
## Changes in this PR:
https://trello.com/c/Oj2nPI4R/1116-add-in-full-with-column-to-new-jobseeker-guides-page

### DIFF
--- a/app/views/posts/index.html.slim
+++ b/app/views/posts/index.html.slim
@@ -12,7 +12,7 @@
 .dfe-grid-container
   - @subcategories.each do |subcategory|
     .dfe-card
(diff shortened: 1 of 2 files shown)


KyleMacPherson (approved):

scruti (changes requested):
:raising_hand: Accessibility issues:
- There is a jump from `h1` to `h3` when it should be `h1` -> `h2`
image

Also the `alt` description of the images is redundant:
image

I recall the consultancy telling us to remove alt text from decorative images (like this case).

scruti (approved):

PR #7004, opened 2024-08-14
Fix filter bug

KyleMacPherson:
## Changes in this PR:

This change fixes a bug that originated in [this pr](https://github.com/DFE-Digital/teaching-vacancies/pull/6966) which was causing the working patterns filter to not work (apart from job share). This managed to get into production because we had no tests for the working pattern filter (apart from the one added for job share in the above PR) so I have added some working pattern filter tests to ensure this will not happen again.

## Next steps:

### DIFF
--- a/app/queries/vacancy_filter_query.rb
+++ b/app/queries/vacancy_filter_query.rb
@@ -112,7 +112,7 @@ class VacancyFilterQuery < ApplicationQuery
       built_scope = built_scope.where(is_job_share: true)
     end
 
-    built_scope.with_any_of_working_patterns(working_patterns) if working_patterns.present?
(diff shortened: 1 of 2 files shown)


app/queries/vacancy_filter_query.rb
```diff
@@ -112,7 +112,7 @@ def add_working_patterns_filters(working_patterns, built_scope)
       built_scope = built_scope.where(is_job_share: true)
     end
 
-    built_scope.with_any_of_working_patterns(working_patterns) if working_patterns.present?
+    built_scope = built_scope.with_any_of_working_patterns(working_patterns) if working_patterns.present?
```
scruti, line 115: Oh silly little thing :face_exhaling:

scruti (approved):
LGTM. Thanks for extending the tests!

PR #7008, opened 2024-08-15
Adding return to teaching adviser video transcript

EllieNodder:
## Changes in this PR:

On our return to teaching page, we have a link to a video transcript currently still hosted on the campaign site.

This PR is to move this transcript across to the main TV site.

## Next steps:


scruti (changes requested):
Had a conversation with @EllieNodder reviewing this. The lack of image breaks the display of this post as a card and displays an error page.

@EllieNodder when you have some time could you confirm if the original intention was that the new transcript information not be listed as a post and close the PR if so?

EllieNodder:
> Had a conversation with @EllieNodder reviewing this. The lack of image breaks the display of this post as a card and displays an error page.
> 
> @EllieNodder when you have some time could you confirm if the original intention was that the new transcript information not be listed as a post and close the PR if so?

Sorry @scruti - completely missed your comment! Yes, think this has been picked up in a different PR now so happy to close this one.

PR #7046, opened 2024-09-06
Logout after all login specs

starswan:
## Changes in this PR:

Ensure that all tests with a login before block do a logout in the after block


scruti (approved):
LGTM! Thanks for looking into this :crossed_fingers:

PR #7056, opened 2024-09-09
Add job profile fields to job application

ddippolito:
## Trello card URL
https://trello.com/c/AX8cK5ov/1131-update-statutory-induction-period-add-trn-application
https://trello.com/c/YGPrSLt6/1132-add-statutory-induction-period-trn-profiles

## Changes in this PR:

This PR adds the Teacher Reference Number (TRN) to the jobseeker profile and job application. The `job_applications` table in the database already has a field for `teacher_reference_number`, which predates the `jobseekers_profiles`. This PR allows the new TRN form field under the professional status section to be saved in the profiles table (even though the underlying form model is a job application). 

This can be seen as added complexity, but there are reasons behind it:

1. The job application database table is bloated. Historically, whenever a new field needed to be introduced to the application, it was added to the table itself without normalising it into multiple tables when required.
2. The TRN does not change from one application to another and belongs to the jobseeker profile semantically
3. Jobseekers can currently manage their jobseeker profiles from a dedicated area on the website where they can change the TRN and other information as they prefer. If we were to leave the TRN in a job application, we would have had the same problem in reverse. Furthermore, users can submit multiple applications with different TRNs, which opens up a lot of problems if we were to use the job application table to store the field

All future fields that belong to the profile will follow the same approach.

#### Note:

The current state of the job application table is to be considered technical debt, and we have created a ticket to understand the pros and cons of removing some fields from it and moving it to the jobseeker profile:
https://trello.com/c/zcGPrRxk/1040-investigate-separation-of-job-application-fields-from-the-jobapplications-table

https://github.com/DFE-Digital/teaching-vacancies/blob/main/db/schema.rb#L192-L250

## Screenshots of UI changes:
Screenshot 2024-09-09 at 20 12 20
Screenshot 2024-09-09 at 20 13 09

### DIFF
--- a/lib/tasks/update_qualified_teacher_status_for_profiles.rake
+++ b/lib/tasks/update_qualified_teacher_status_for_profiles.rake
@@ -1,10 +1,6 @@
 namespace :jobseeker_profile do
   desc "Update qualified_teacher_status from 'non_teacher' to 'no'"
   task update_non_teacher_status: :environment do
-    profiles_to_update = JobseekerProfile.where(qualified_teacher_status: :non_teacher)
-
-    profiles_to_update.each do |profile|
-      profile.update(qualified_teacher_status: :no)
-    end
+    JobseekerProfile.where(qualified_teacher_status: :non_teacher).update_all(qualified_teacher_status: :no)
   end
 end


app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -38,20 +39,27 @@ def back_path
   end
 
   def form
-    @form ||= form_class.new(form_attributes)
+    @form ||= begin
+      attributes = form_attributes
+      attributes.merge!(jobseeker_profile_attributes) if step == :professional_status
```
scruti, line 44: Suggestion: 

If I'm reading this right. The `form_attributes` is already merging a set of params only if the step is professional_status  [here](https://github.com/DFE-Digital/teaching-vacancies/pull/7056/files#diff-6fe827802e214e8b154f45038828bf184bafafb495542a1b0bb111b48659b01fR61):
```ruby
attributes.merge!(trn_params) if step == :professional_status
```

This line also conditionally merges the jobseeker profile attributes if the step is professional status.

I personally would find this logic easier to read/follow if both `merge!` calls were together under a single `if step == :professional_status` condition.

I'm leaving it as a suggestion you may want to explore. But as long as it works... this is just something I noticed that maybe would be simpler :shrug:

app/helpers/job_applications_helper.rb
```diff
@@ -40,6 +40,15 @@ def job_application_qualified_teacher_status_info(job_application)
     end
   end
 
+  def job_application_jobseeker_profile_info(job_application)
+    profile = job_application&.jobseeker&.jobseeker_profile
+    if profile&.has_teacher_reference_number.present? && profile&.has_teacher_reference_number == "yes"
```
scruti, line 45: The second condition also returns false if the `profile&.has_teacher_reference_number` is not present.
```suggestion
    if profile&.has_teacher_reference_number == "yes"
```

lib/tasks/update_qualified_teacher_status_for_profiles.rake
```diff
@@ -0,0 +1,10 @@
+namespace :jobseeker_profile do
+  desc "Update qualified_teacher_status from 'non_teacher' to 'no'"
+  task update_non_teacher_status: :environment do
+    profiles_to_update = JobseekerProfile.where(qualified_teacher_status: :non_teacher)
+
+    profiles_to_update.each do |profile|
+      profile.update(qualified_teacher_status: :no)
+    end
```
scruti, line 8: Suggestion:

Unless there is a need to instantiate the profiles in memory one by one we could do this with a single SQL query that would be way faster.

```suggestion
    JobseekerProfile.where(qualified_teacher_status: :non_teacher)
                    .update_all(qualified_teacher_status: :no)
```

spec/system/jobseekers/jobseekers_can_add_professional_status_to_their_profile_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+        choose "No"
+      end
+
+      click_on "Save and continue"
+
+      expect_page_to_have_professional_status_information("no", nil, "", nil)
+    end
+  end
+
+  def expect_page_to_have_professional_status_information(qts, year, trn, statutory_induction_complete)
```
scruti, line 120: suggestion: 

Given the number of parameters, maybe we could convert them into named params for easier reading:
eg: `expect_page_to_have_professional_status_information(qts: "yes", year: "2000", trn: "1234567", statutory_induction_complete: "on_track")`
vs 
`expect_page_to_have_professional_status_information("yes", "2000", "1234567", "on_track")`

scruti (approved):
LGTM!

Left a few suggestions for your consideration.

PR #7066, opened 2024-09-11
Validate email addresses

starswan:
## Trello card URL

https://trello.com/c/cybo2kgi/1205-bounced-job-alerts-auto-deletion

## Changes in this PR:

1. Use stricter validation via valid_email2 to prevent gmail.vom type addresses.
2. Fix bug in both email status jobs to scan all bouncing emails not just first 250

The 'invalid email address' message now seems a bit inaccurate - the address can be the 'correct format' but still wrong.

## Next steps:


app/services/gov_uk_notify_status_client.rb
```diff
@@ -0,0 +1,18 @@
...
+    opts = options.merge(template_type: "email")
+    notifications = @client.get_notifications(opts).collection
+
+    Enumerator.new do |yielder|
+      while notifications.any?
+        notifications.each { |n| yielder << n }
+
+        notifications = @client.get_notifications(opts.merge(older_than: notifications.last.id)).collection
+      end
+    end
```
scruti, line 16: I am struggling to understand what this code does.

Is this the way to `Fix bug in both email status jobs to scan all bouncing emails not just first 250` forcing to retrieve older than a default limit?

starswan: Yes - that's what the API docs say, and neither (existing) implementation does that.

This is just a generator to lazily retrieve all the email notifications that match the filter provided by the caller

spec/rails_helper.rb
```diff
@@ -20,6 +20,9 @@
 Geocoder::DEFAULT_STUB_COORDINATES = [51.1789, -1.8262].freeze
 Geocoder::DEFAULT_LOCATION = "TE5 T1NG".freeze
 
+# https://stackoverflow.com/questions/1368163/is-there-a-standard-domain-for-testing-throwaway-email
+TEST_EMAIL_DOMAIN = "contoso.com".freeze
```
scruti, line 24: Thanks for the context!

scruti (approved):
LGTM!

PR #7076, opened 2024-09-13
Merge account data

KyleMacPherson:
## Trello card URL

https://trello.com/c/zoc33VKe/1143-manual-account-merger-journey-import-data-from-tv-account-to-new-gol-sign-in

## Changes in this PR:

This PR introduces functionality for a user to request to transfer data from an old account they have previously set up. They get sent a confirmation email to their new account, containing a code they can use to confirm the transfer.

Notes:
- I have not implemented the hard delete of the old account. Is this necessary? It seems like it increases risk without a large benefit.
- This feature has been written as part of the migration to gov.uk one login, but does not rely on any one login functionality. There will be no links to the request account transfer pages on our service but users will be able to use them if they find them, I have not put this behind a feature flag or limited the feature to certain users. I don't think this is an issue but maybe it's something we want to consider?

## Next steps:

### DIFF
--- a/app/form_models/jobseekers/account_transfer_form.rb
+++ b/app/form_models/jobseekers/account_transfer_form.rb
@@ -1,0 +1,20 @@
+class Jobseekers::AccountTransferForm < BaseForm
+  attr_accessor :email, :account_merge_confirmation_code
+
+  validates :email, presence: true
+  validates :email, email_address: true, if: -> { email.present? }
+  validates :account_merge_confirmation_code, presence: true
+  validate :validate_jobseeker_email_and_code_match
+
+  private
+
+  def validate_jobseeker_email_and_code_match
+    jobseeker = Jobseeker.find_by(email: email)
+    
+    if jobseeker.nil? || jobseeker.account_merge_confirmation_code != account_merge_confirmation_code
+      errors.add(:confirmation_code, :confirmation_code_mismatch, message: "Confirmation code does not match.")
+    elsif jobseeker.account_merge_confirmation_code_generated_at < 1.hour.ago
+      errors.add(:confirmation_code, :confirmation_code_expired, message: "Confirmation code has expired. Please request a new code.")
+    end
+  end
+end
--- a/app/form_models/jobseekers/account_transfer_form.rb
+++ b/app/form_models/jobseekers/account_transfer_form.rb
@@ -1,8 +1,7 @@
 class Jobseekers::AccountTransferForm < BaseForm
   attr_accessor :email, :account_merge_confirmation_code
-
   validates :email, presence: true
-  validates :email, email_address: true, if: -> { email.present? }
+  validates :email, email_address: true
   validates :account_merge_confirmation_code, presence: true
   validate :validate_jobseeker_email_and_code_match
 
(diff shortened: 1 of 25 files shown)


app/form_models/jobseekers/account_transfer_form.rb
```diff
@@ -0,0 +1,20 @@
+class Jobseekers::AccountTransferForm < BaseForm
+  attr_accessor :email, :account_merge_confirmation_code
+
+  validates :email, presence: true
+  validates :email, email_address: true, if: -> { email.present? }
```
scruti, line 5: This validation is being removed in @starswan's PR: https://github.com/DFE-Digital/teaching-vacancies/pull/7066
It seems whichever goes in first, the other PR will need to be updated!

app/form_models/jobseekers/account_transfer_form.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  private
+
+  def validate_jobseeker_email_and_code_match
+    jobseeker = Jobseeker.find_by(email: email)
+
+    if jobseeker.nil? || jobseeker.account_merge_confirmation_code != account_merge_confirmation_code
+      errors.add(:account_merge_confirmation_code, :confirmation_code_mismatch, message: "Confirmation code does not match.")
+    elsif jobseeker.account_merge_confirmation_code_generated_at < 1.hour.ago
+      errors.add(:account_merge_confirmation_code, :confirmation_code_expired, message: "Confirmation code has expired. Please request a new code.")
```
scruti, line 17: Suggestion:
Moving the error messages to the locales file.

app/form_models/jobseekers/request_account_transfer_email_form.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :email, email_address: true, if: -> { email.present? }
+  validate :validate_recent_code_request, if: -> { email.present? }
+
+  def validate_recent_code_request
+    jobseeker = Jobseeker.find_by(email: email)
+    return if jobseeker.nil?
+
+    return unless jobseeker.account_merge_confirmation_code_generated_at.present? && jobseeker.account_merge_confirmation_code_generated_at > 1.minute.ago
+
+    errors.add(:email, :recent_code_request, message: "Please wait 1 minute before requesting another code.")
```
scruti, line 14: Suggestion:
Moving the error message to the locales file.

config/analytics.yml
```diff
@@ -190,6 +190,8 @@ shared:
     - created_at
     - updated_at
     - account_closed_on
+    - account_merge_confirmation_code
+    - account_merge_confirmation_code_generated_at
```
scruti, line 194: If we need this info in analytics, I think we should only send the timestamp (so we can trace how many accounts have a pending merge requested?), not the code itself (that acts as a kind of a password).

app/services/jobseekers/account_transfer.rb
```diff
@@ -0,0 +1,55 @@
...
+  def call
+    raise AccountNotFoundError, "Account with email not found" unless account_to_transfer
+
+    ActiveRecord::Base.transaction do
+      transfer_profile
+      transfer_feedbacks
+      transfer_job_applications
+      transfer_saved_jobs
+      update_subscriptions
+    end
```
scruti, line 19: Regarding the PR description comment:
> I have not implemented the hard delete of the old account. Is this necessary? It seems like it increases risk without a large benefit.

I think not deleting the old account is problematic:
- This service removes the profile, feedbacks, job_applications, saved_jobs and subscriptions from the old account. All that is left is an empty shell, with no value to the service or the user.
- The old accounts would count in the total of "active" accounts in the service, misrepresenting the number of real users.
- The old accounts would also receive communications if we send them to all active users, campaigns...
- The user may get confused and use their old credentials to find themselves with an empty account, no profile... and maybe even start executing actions from this account.

What are the possible risks on hard deleting the account once we have removed all their info?

spec/services/jobseekers/account_transfer_spec.rb
```diff
@@ -0,0 +1,101 @@
+require "rails_helper"
+
+RSpec.describe Jobseekers::AccountTransfer do
```
scruti, line 3: Nice tests! :clap:

PR #7078, opened 2024-09-13
HOTFIX - Create a jobseeker profile when applying for a new job if the user does not have one

ddippolito:
An issue in production has revealed that users encounter an error when trying to apply for a job without creating a profile first. This pull request fixes the error by requiring the creation of a job seeker profile when applying for a job, in case the user doesn't have one. 

https://teaching-vacancies.sentry.io/issues/5836838716/?alert_rule_id=10370581&alert_type=issue&environment=production&notification_uuid=0b8ce9c6-3864-49e7-af6c-43bc9ec97676&project=6212514&referrer=slack

### DIFF
--- a/spec/system/jobseekers/jobseekers_can_add_professional_status_to_their_job_application_spec.rb
+++ b/spec/system/jobseekers/jobseekers_can_add_professional_status_to_their_job_application_spec.rb
@@ -48,4 +48,25 @@ RSpec.describe "Jobseekers can add details about their qualified teacher status 
 
     expect(page).not_to have_css("h2", text: "There is a problem")
   end
+
+  it "creates a jobseeker profile if the jobseeker does not have one" do
+    visit jobseekers_job_application_build_path(job_application, :professional_status)
+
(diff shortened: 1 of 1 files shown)


KyleMacPherson (approved):

app/controllers/jobseekers/job_applications/build_controller.rb
```diff
@@ -134,10 +134,17 @@ def trn_params
...
   end
 
-  def update_jobseeker_profile
-    current_jobseeker.jobseeker_profile.update(
-      teacher_reference_number: form_params[:teacher_reference_number],
-      has_teacher_reference_number: form_params[:has_teacher_reference_number],
-    )
+  def update_or_create_jobseeker_profile
+    if current_jobseeker.jobseeker_profile.nil?
+      current_jobseeker.create_jobseeker_profile(
```
starswan, line 139: I think this needs a ! at the end - is there enough data here to create a profile?

starswan: I'm not sure. Both personal_details and job_preferences are has_one w/o optional - I *think* that rails will insist on them?

PR #7087, opened 2024-09-16
Fix omniauth config for allowing both publishers and jobseekers

scruti:
## What this PR does:
- Fixes and extends Govuk One Login request spec to cover new business logic about where to redirect jobseekers after login.
- Fixes linting errors.
- Changes the Omniauth Jobseekers Onelogin configuration approach to be compatible with Publishers Omniauth (DFE Sign-in)

## What this PR does not do:
- Fixing existing tests based in Jobseekers Devise user/password approach. This will be tackled in a separate PR.

## Long explanation

Originally we were using Devise "omniauthable" module on the jobseekers.

This was breaking the system tests for the Publisher users (based on
Omniauth DFE Sign-in service).

We have found that "omniauthable" is not compatible with multiple models
using Omniauth. So we have to remove the "omniauthable" wrapper/config
over plain Omniauth for the Jobseeker, and configure it directly through
the Omniauth configuration, as done with DFE Sign-in Publisher users.

Devise documentation and instructions:
https://github.com/heartcombo/devise/wiki/OmniAuth-with-multiple-models

### DIFF
--- a/app/models/jobseeker.rb
+++ b/app/models/jobseeker.rb
@@ -1,16 +1,16 @@
 class Jobseeker < ApplicationRecord
   has_encrypted :last_sign_in_ip, :current_sign_in_ip
 
-  devise(*%I[confirmable
-             database_authenticatable
-             lockable
-             omniauthable
-             recoverable
-             registerable
-             timeoutable
-             trackable
-             validatable],
-         omniauth_providers: [:openid_connect])
+  devise(*%I[
+    confirmable
+    database_authenticatable
+    lockable
+    recoverable
+    registerable
+    timeoutable
+    trackable
+    validatable
+  ])
 
   has_many :feedbacks, dependent: :destroy, inverse_of: :jobseeker
   has_many :job_applications, dependent: :destroy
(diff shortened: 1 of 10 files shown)


app/models/jobseeker.rb
```diff
@@ -1,16 +1,16 @@
...
-  devise(*%I[confirmable
-             database_authenticatable
-             lockable
-             omniauthable
-             recoverable
-             registerable
-             timeoutable
-             trackable
-             validatable],
-         omniauth_providers: [:openid_connect])
```
scruti, line 13: This was breaking DFE Sign-in (and hence all those tests errors) as the `omniauthable` module is not compatible with other models using Omniauth :shrug:

app/services/jobseekers/govuk_one_login.rb
```diff
@@ -3,7 +3,7 @@ module Jobseekers::GovukOneLogin
   CALLBACKS_BASE_URL = "#{ENV.fetch('DOMAIN').include?('localhost') ? 'http' : 'https'}://#{ENV.fetch('DOMAIN')}".freeze
 
   CALLBACKS = {
-    login: "#{CALLBACKS_BASE_URL}/jobseekers/auth/openid_connect/callback",
+    login: "#{CALLBACKS_BASE_URL}/jobseekers/auth/govuk_one_login/callback",
```
scruti, line 6: This is a cosmetic refactor. Thought is better pattern to name things (as done with `/auth/dfe/`)

KyleMacPherson (approved):

PR #7090, opened 2024-09-17
Onelogin no account found page

KyleMacPherson:
## Trello card URL

https://trello.com/c/QITtyko2/1146-gol-to-tv-journey-2-new-sign-in-with-no-email-match-account-found

## Changes in this PR:

This page creates a page for jobseekers to explain that an account was not found matching their gov.uk one login details and explains some helpful information. It also redirects users to this page after they log in unless they were applying for a quick apply job.

## Next steps:

### DIFF
--- a/app/controllers/jobseekers/govuk_one_login_callbacks_controller.rb
+++ b/app/controllers/jobseekers/govuk_one_login_callbacks_controller.rb
@@ -2,21 +2,21 @@ class Jobseekers::GovukOneLoginCallbacksController < Devise::OmniauthCallbacksCo
   include Jobseekers::GovukOneLogin::Errors
   # Devise redirects response from Govuk One Login to this method.
   # The request parameters contain the response from Govuk One Login from the user authentication through their portal.
+
+  
   def openid_connect
     if (govuk_one_login_user = Jobseekers::GovukOneLogin::UserFromAuthResponse.call(params, session))
       session[:govuk_one_login_id_token] = govuk_one_login_user.id_token
+      
+      jobseeker = Jobseeker.find_by('lower(email) = ?', govuk_one_login_user.email.downcase)
 
-      if existing_jobseeker_first_sign_in_via_one_login(govuk_one_login_user)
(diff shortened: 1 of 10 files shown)


app/controllers/jobseekers/govuk_one_login_callbacks_controller.rb
```diff
@@ -10,6 +10,11 @@ def openid_connect
         session[:user_exists_first_log_in] = { value: "true", path: "/", expires: 1.hour.from_now }
       end
 
+      if no_existing_jobseeker_with_email_exists?(govuk_one_login_user)
+        session[:newly_created_user] = { value: "true", path: "/", expires: 1.hour.from_now }
+      end
+
+
       jobseeker = Jobseeker.find_or_create_from_govuk_one_login(email: govuk_one_login_user.email,
                                                                 govuk_one_login_id: govuk_one_login_user.id)
```
scruti, line 19: From what I see here, we are at the point of having 3 different `Jobseeker.find_by(email:  govuk_one_login_user.email)` calls happening sequentially for every connection, while at least 2 of those calls (the session setting ones) are based on "A" or "B" logic exclusive from each other.

Could we refactor this to query the jobseeker once by email (is the `email: govuk_one_login_user.email` non case sensitive by default?), and then, based on the retrieved `jobseeker` action on the session & update/create the jobseeker?

app/views/jobseekers/accounts/account_not_found.html.slim
```diff
@@ -0,0 +1,9 @@
+- content_for :page_title_prefix, t(".page_title")
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-l = t(".page_title")
+
+    p.govuk-body = t(".could_not_find_account")
+    p.govuk-body = t(".continue_to_profile", link: govuk_link_to(t(".profile_link_text"), jobseekers_profile_path)).html_safe
+    p.govuk-body = t(".transfer_account")
```
scruti, line 9: Shouldn't this be a link to the new account transfer flow?

scruti (approved):
LGTM!

PR #7099, opened 2024-09-20
Improve pdf download

ddippolito:
## Trello card URL
https://trello.com/c/rMwjrQcg/1075-improve-pdf-download-printed-formatting-of-tv-applications

## Changes in this PR:

We receive many complaints from Hiring Staff about the poor readability of the PDF export of job applications. The primary use case seems to be printing the PDF and adding notes manually. 
This change introduces Prawn as a PDF building library to generate the Job Application PDF, allowing us to fine-tune the control of page breaks and the PDF's appearance.

## Screenshots of UI changes:
Screenshot 2024-09-23 at 09 47 48

### DIFF
--- a/app/helpers/pdf_helper.rb
+++ b/app/helpers/pdf_helper.rb
@@ -52,7 +52,7 @@ module PdfHelper
 
   def add_professional_status(pdf)
     professional_status = [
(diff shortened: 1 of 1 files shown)


scruti (approved):
LGTM. Nice approach.

The only downside I see on "drawing" the pdf ad-hoc is that we will need to update this every time we change the job application fields. So it is prone to get outdated :sweat: 

But is an issue I don't think we can avoid if we want something nicer than the previous "print the page" approach.

PR #7101, opened 2024-09-20
Delete unconfirmed accounts on all bounces

starswan:
## Trello card URL

https://trello.com/c/cybo2kgi/1205-bounced-job-alerts-auto-deletion

## Changes in this PR:

Is there anything I might have missed with this plan? If a user confirm email bounces for any reason, presumably it's ok to delete the account as it's probably not a real person. We re-send the confirm email every 5 days, does that need altering too in some way? Should we delete unconfirmed accounts if they don't respond after a certain time period?

## Next steps:


starswan:
Closing as this is all going away in November 2024 when we release OneLogin support

PR #7120, opened 2024-09-26
Allow jobseekers to update their email from govuk one login

scruti:
## Trello card URL
- https://trello.com/c/SYaQ2Q7C

## Changes in this PR:

Users can change their email addresses using GovUK OneLogin.

This will cause the same user ID to come with a different email address in the OneLogin callback/user info payload.

To allow this to work as intended:
- When possible, we will match the OneLogin user with the TV jobseeker using their OneLogin ID.

- Matching by email will only be used to identify possible jobseekers in TV who are signing in to OneLogin for the first time and still have no OneLogin ID associated with their TV user.

- We need to disable Devise "reconfirmable" behaviour. 
  That logic is now dealt with in GovUK OneLogin site. We do not need to confirm email changes in-house anymore and was breaking the updates coming from OneLogin unless disabled.

### DIFF
--- a/spec/system/jobseekers/jobseekers_can_change_email_spec.rb
+++ b/spec/system/jobseekers/jobseekers_can_change_email_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe "Jobseekers can change email" do
   after { logout }
 
   describe "updating email and confirming change" do
-    it "validates and submits the form, sends emails, redirects to check your email page, confirms the change, updates the email and the subscriptions associated with the previous email and redirects to saved_jobs page" do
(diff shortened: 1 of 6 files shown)


spec/system/jobseekers/jobseekers_can_change_email_spec.rb
```diff
@@ -14,7 +14,7 @@
   after { logout }
 
   describe "updating email and confirming change" do
-    it "validates and submits the form, sends emails, redirects to check your email page, confirms the change, updates the email and the subscriptions associated with the previous email and redirects to saved_jobs page" do
+    xit "validates and submits the form, sends emails, redirects to check your email page, confirms the change, updates the email and the subscriptions associated with the previous email and redirects to saved_jobs page" do
```
scruti, line 17: Left this file as a reminder we need to remove the whole Devise flow logic.

PR #7128, opened 2024-09-30
Sign-out Jobseekers through Govuk OneLogin

scruti:
We had already built the endpoint and uri helpers to use the GovUK OneLogin /logout endpoint, but the "Sign Out" link in our navigation bar was still directly destroying the session through Devise.

The behaviour caused the jobseeker to be signed out from our service, but still signed-in in GovUK OneLogin. This caused the next sign-in attempt to directly redirect back the jobseeker signed in our service without having been prompted to any security user/password check by OneLogin service.

We have fixed this by:
- Replacing the navigation Sign Out button to send the user to the /logout endpoint in GovUK One Login. This signs them out from OneLogin.
- Ensuring OneLogin redirects the user back to our internal session destroy endpoint. So the user session in our service is also signed out.

To achieve this, we had to enable a new route, to accept GET requests to the session-destroying endpoint (defaults to DESTROY).

### The navigation "Sign Out" link now points to GovUK OneLogin URL
Screenshot From 2024-09-30 17-56-37

Screenshot From 2024-09-30 17-56-55

### After following the link the user gets sent back and signed out from our service

Screenshot From 2024-09-30 17-57-12

### DIFF
--- a/app/views/layouts/_header.html.slim
+++ b/app/views/layouts/_header.html.slim
@@ -12,7 +12,7 @@
     = header.with_navigation_item text: t("nav.create_a_job_alert"), href: new_subscription_path
     = header.with_navigation_item text: t("nav.your_profile"), href: jobseekers_profile_path
     = header.with_navigation_item text: t("nav.your_account"), href: jobseekers_account_path, active: your_account_active?
-    = header.with_navigation_item text: t("nav.sign_out"), href: destroy_jobseeker_session_path, options: { method: :delete }
+    = header.with_navigation_item text: t("nav.sign_out"), href: jobseeker_logout_uri.to_s
   - elsif support_user_signed_in?
     = header.with_navigation_item text: t("nav.support_user_dashboard"), href: support_user_root_path, active: current_page?(support_user_root_path)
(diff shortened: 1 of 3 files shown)


app/views/layouts/_header.html.slim
```diff
@@ -12,7 +12,7 @@
     = header.with_navigation_item text: t("nav.create_a_job_alert"), href: new_subscription_path
     = header.with_navigation_item text: t("nav.your_profile"), href: jobseekers_profile_path
     = header.with_navigation_item text: t("nav.your_account"), href: jobseekers_account_path, active: your_account_active?
-    = header.with_navigation_item text: t("nav.sign_out"), href: destroy_jobseeker_session_path, options: { method: :delete }
+    = header.with_navigation_item text: t("nav.sign_out"), href: jobseeker_logout_uri.to_s
```
scruti, line 15: This is an existing helper previously built for OneLogin. It builds the URL for onelogin `/logout` endpoint with the callback `/sign_out` endpoint in our side to be redirected back to.

KyleMacPherson (approved):

PR #7137, opened 2024-10-03
Add new campaign bespoke landing page

ddippolito:
## Trello card URL
https://trello.com/c/kqe2MWBV/1253-create-bespoke-landing-pages-for-new-teachers-digital-activity

## Changes in this PR:
We're adding a new campaign bespoke landing page. Users will be able to access this temporary landing page through links in QR codes within physical booklets or through emails. The page has a different design from the orginal job search page but we can re-use some of the components.

## Screenshots of UI changes:

Screenshot 2024-10-03 at 11 44 38

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -85,9 +85,9 @@ class VacanciesController < ApplicationController
 
   def extract_working_patterns(campaign_params)
     working_patterns = []
-    working_patterns << "full_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_fulltime)) == "true"
-    working_patterns << "part_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_parttime)) == "true"
-    working_patterns << "job_share" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_jobshare)) == "true"
+    working_patterns << "full_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_fulltime))
+    working_patterns << "part_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_parttime))
+    working_patterns << "job_share" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_jobshare))
     working_patterns
   end
 
--- a/documentation/landing-pages.md
+++ b/documentation/landing-pages.md
@@ -1,3 +1,11 @@
+## Introduction
+
+Teaching Vacancies uses dynamic landing pages to enhance the user experience by offering tailored job searches based on various criteria such as location, organisation, and specific job roles. 
+These landing pages are predefined with unique search filters and localised content, allowing users to find relevant job opportunities quickly and efficiently.
+
+This document outlines how landing pages are structured, how the system retrieves them, and how the search process works in conjunction with these pages. 
+It also covers the configuration of landing pages through YAML files and explains how to further customise their behavior to suit specific needs.
+
 1. **Routes**  
    Dynamic routes are defined for different types of landing pages:
     - Location-based (`teaching-jobs-in-:location_landing_page_name`)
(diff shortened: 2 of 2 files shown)


app/views/vacancies/campaign_landing_page.html.slim
```diff
@@ -0,0 +1,63 @@
...
+- if @vacancies_search.active_criteria? && any_vacancies
+  p = t("subscriptions.link.help_text_html", link: govuk_link_to(t("subscriptions.link.text"), new_subscription_path(search_criteria: @vacancies_search.active_criteria, coordinates_present: @vacancies_search.point_coordinates.present?)))
+
+= form_for @form, as: "", url: jobs_path, method: :get, html: { data: { controller: "form" }, role: "search" } do |f|
+  .custom-banner-background
+    .govuk-width-container
+      .custom-banner
+        .govuk-grid-row
+          .govuk-grid-column-one-third-at-desktop
+            h1.govuk-heading-l class="govuk-!-margin-bottom-4" = "#{@jobseeker_name}, find your primary teacher job"
```
starswan, line 22: It looks like we've only implemented the 'primary teacher' landing page here. Do we not need to parameterise this so we can do secondary teachers (and others)?

app/controllers/vacancies_controller.rb
```diff
@@ -48,6 +59,24 @@ def set_headers
...
+          .tap do |campaign_params|
+      campaign_params[:location] = campaign_params.delete(:email_location)
+      campaign_params[:radius] = campaign_params.delete(:email_radius)
+      campaign_params[:teaching_job_roles] = [campaign_params.delete(:email_jobrole)].compact
+      campaign_params[:subjects] = [campaign_params.delete(:email_subject)].compact
+      campaign_params[:phases] = [campaign_params.delete(:email_phase)].compact
+      working_patterns = []
+      working_patterns << "full_time" if campaign_params.delete(:email_fulltime) == "true"
+      working_patterns << "part_time" if campaign_params.delete(:email_parttime) == "true"
+      working_patterns << "job_share" if campaign_params.delete(:email_jobshare) == "true"
```
starswan, line 74: should we not use ActiveModel::Type::Boolean.new.cast(value) here so that we cope with values of "1" (and Yes?) as well as (lowercase) true?

app/controllers/vacancies_controller.rb
```diff
@@ -48,6 +59,24 @@ def set_headers
     response.set_header("X-Robots-Tag", "noarchive")
   end
 
+  def campaign_search_params
+    params.permit(:email_name, :email_postcode, :email_location, :email_radius, :email_jobrole, :email_subject,
+                  :email_phase, :email_ECT, :email_fulltime, :email_parttime, :email_jobshare, :email_contact)
+          .tap do |campaign_params|
+      campaign_params[:location] = campaign_params.delete(:email_location)
+      campaign_params[:radius] = campaign_params.delete(:email_radius)
```
starswan, line 67: I think we are supposed to pre-select 15 mile radius if it isn't specified (which it probably won't be?)

app/views/vacancies/campaign_landing_page.html.slim
EllieNodder: yes, could we make it so it pulls through either 'primary' or 'secondary', depending on their info in Mailchimp?

app/services/campaign_page.rb
```diff
@@ -0,0 +1,24 @@
+class CampaignPage
+  attr_reader :utm_content_code, :criteria, :banner_image
+
+  def self.exists?(utm_content)
+    Rails.application.config.campaign_pages.key?(utm_content.to_sym)
+  end
+
+  def self.[](utm_content_code)
+    raise "No such campaign page: '#{utm_content_code}'" unless exists?(utm_content_code)
```
scruti, line 9: Shouldn't we gracefully treat this so users don't get a service error when the param value is wrong?
I would expect a user to get a `404` page or get redirected to the generic vacancies page.

app/controllers/vacancies_controller.rb
```diff
@@ -48,6 +61,36 @@ def set_headers
...
+    campaign_params[:teaching_job_roles] = [campaign_params.delete(:email_jobrole)].compact
+    campaign_params[:subjects] = [campaign_params.delete(:email_subject)].compact
+    campaign_params[:phases] = [campaign_params.delete(:email_phase)].compact
+  end
+
+  def extract_working_patterns(campaign_params)
+    working_patterns = []
+    working_patterns << "full_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_fulltime)) == "true"
+    working_patterns << "part_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_parttime)) == "true"
+    working_patterns << "job_share" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_jobshare)) == "true"
```
scruti, line 90: I suspect these checks would never pass, since the boolean casting will return `true` instead of `"true"`. So `true == "true"` doesn't match.

I think we don't even need the comparison:

```suggestion
    working_patterns << "full_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_fulltime))
    working_patterns << "part_time" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_parttime))
    working_patterns << "job_share" if ActiveModel::Type::Boolean.new.cast(campaign_params.delete(:email_jobshare))
```

config/campaign_pages.yml
```diff
@@ -0,0 +1,134 @@
+shared:
+  whentoapplybespoke+PRIMARY:
+    banner_image: "campaigns/primary_teacher.jpg"
```
scruti, line 3: :clap:

PR #7188, opened 2024-10-18
Raise exception if vacancy cannot be created

scruti:
Trying to debug this [Sentry error](https://teaching-vacancies.sentry.io/issues/5998549447/events/d258d563044b47db9af2b4b7423ec22d/?project=6212514) noticed that the flow continued even when the vacancy failed to be created.

Instead of having an indirect exception down the flow due to a validation failure causing the vacancy not to be created and then a further call causing an error, we raise an exception directly if something fails during the creation of the vacancy.

### DIFF
--- a/app/controllers/publishers/vacancies_controller.rb
+++ b/app/controllers/publishers/vacancies_controller.rb
@@ -25,7 +25,7 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   end
 
   def create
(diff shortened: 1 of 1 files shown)


KyleMacPherson (approved):

starswan (commented):
This should have been protected with a rubocop rule - have we disabled it somehow?

 [Looks like we don't use rubocop-rails at all, which would have caught this (and the bug I found last week about date validation). I did a (very very) quick spike on it, and it appears that there are no more errors of this specific type in the codebase (Rails/SaveBang rule)]

PR #7198, opened 2024-10-23
Update rake task for TRN migration

ddippolito:
## Trello card URL
https://trello.com/c/rYYbUnVN/1345-delete-job-application-trns-from-the-database-and-migrate-missing-trns-to-jobseeker-profiles

## Changes in this PR:

Currently, we’re saving TRNs on job applications (that’s how it has always been done historically) and jobseeker profiles. The correct place is jobseeker profiles semantically and from a tech perspective because we ask TRNs to jobseekers when they complete their profiles. When they apply for new jobs, we pull the TRN from the profiles table in the database.

Before deleting the field from the job applications table, we first need to create a new profile for all the job seekers who submitted a job application providing a trn and migrate it to this newly created profile.

### DIFF
--- a/lib/tasks/migrate_trn_to_profiles.rake
+++ b/lib/tasks/migrate_trn_to_profiles.rake
@@ -3,42 +3,32 @@ namespace :jobseeker_profile do
   desc "Migrate Teacher Reference Number (TRN) from job_applications to jobseeker_profiles, creating profiles if necessary"
   task migrate_teacher_reference_number_to_profiles: :environment do
     batch_size = 500
+    failed_log = File.open("failed_migrations.log", "w")
 
-    namespace :jobseeker_profile do
-      desc "Migrate Teacher Reference Number (TRN) from job_applications to jobseeker_profiles, creating profiles if necessary"
-      task migrate_teacher_reference_number_to_profiles: :environment do
-        batch_size = 500
+    latest_application_ids = JobApplication
+                               .where.not(teacher_reference_number_ciphertext: [nil, ""])
+                               .select("DISTINCT ON (jobseeker_id) id")
+                               .order(:jobseeker_id, created_at: :desc)
+                               .pluck(:id)
 
-        latest_application_ids = JobApplication
-                                   .where.not(teacher_reference_number_ciphertext: [nil, ""])
-                                   .select("DISTINCT ON (jobseeker_id) id")
-                                   .order(:jobseeker_id, created_at: :desc)
-                                   .pluck(:id)
+    JobApplication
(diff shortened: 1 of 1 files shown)


lib/tasks/migrate_trn_to_profiles.rake
```diff
@@ -1,18 +1,26 @@
...
-        latest_job_application = JobApplication
-                                   .where(jobseeker_id: profile.jobseeker_id)
-                                   .order(created_at: :desc)
-                                   .first
+    JobApplication
+      .select("DISTINCT job_applications.jobseeker_id, job_applications.teacher_reference_number_ciphertext")
+      .where.not(teacher_reference_number_ciphertext: [nil, ""])
+      .joins("LEFT JOIN jobseeker_profiles ON jobseeker_profiles.jobseeker_id = job_applications.jobseeker_id")
+      .where("jobseeker_profiles.teacher_reference_number_ciphertext IS NULL OR jobseeker_profiles.teacher_reference_number_ciphertext = ''")
+      .find_in_batches(batch_size: batch_size) do |applications|
```
starswan, line 11: would it not be safer to write this in Active Record rather than SQL? Do we not want the *last* job application (by date) rather than an arbitrary one?

small point: you can use find_each rather than find_in_batches - it takes the same parameters and saves you an extra loop.

lib/tasks/migrate_trn_to_profiles.rake
```diff
@@ -1,18 +1,26 @@
...
+      applications.each do |application|
+        profile = JobseekerProfile.find_or_initialize_by(jobseeker_id: application.jobseeker_id)
 
-        if latest_job_application&.teacher_reference_number.present?
-          profile.update(teacher_reference_number: latest_job_application.teacher_reference_number)
-          puts "Migrated TRN for profile ID: #{profile.id}"
+        puts "Creating profile for jobseeker ID: #{application.jobseeker_id}" if profile.new_record?
+
+        profile.teacher_reference_number_ciphertext = application.teacher_reference_number_ciphertext
```
starswan, line 17: This line makes me a little nervous - should we not just be using teacher_reference_number?

lib/tasks/migrate_trn_to_profiles.rake
```diff
@@ -1,18 +1,26 @@
...
+        profile.has_teacher_reference_number = "yes"
+
```
starswan, line 19: profile.update(teacher_refernce_number: trn, has_teacher_reference_number: "yes") is a little more idiomatic than assignment and save IMHO

scruti (approved):

PR #7200, opened 2024-10-23
Remove email validation from Organisation  model

scruti:
The validation causes existing Schools not to be able to publish vacancies because their stored email is invalid.

The form's validation should cover schools trying to introduce or change a wrong email address.

Related Sentry error blocking a publisher from creating vacancies: [Here](https://teaching-vacancies.sentry.io/issues/6009418670/tags/user/?environment=production&project=6212514&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=19)

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -51,8 +51,6 @@ class Organisation < ApplicationRecord
 
   scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts).registered_for_service }
 
-  validates :email, email_address: true
-
   alias_attribute :data, :gias_data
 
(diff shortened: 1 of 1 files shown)


premiumsolutions11:
If you remove email validation from the Organization model, it may lead to issues with data integrity. Allowing invalid or incorrectly formatted emails could cause problems down the line, especially when trying to send communications or verify contacts. A better approach might be to modify the validation rules to be less restrictive or make email optional, depending on the use case. However, it's generally good practice to maintain at least some level of validation to ensure data accuracy.

PR #7201, opened 2024-10-23
Do not validate Org when creating a Vacancy

scruti:
The `has_many` association validates the associated object by default.

This is causing issues with some existing Organisations that have a pre-existing invalid email since we added email validation to the Org. model. Now they cannot create new vacancies.

Disabling the check on Org validity when creating a Vacancy will allow the publishers to still publish new vacancies.

[Related issue on Sentry](https://teaching-vacancies.sentry.io/issues/6009418670/tags/user/?environment=production&project=6212514&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=19)

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -66,7 +66,12 @@ class Vacancy < ApplicationRecord
   has_many :job_applications, dependent: :destroy
   has_one :equal_opportunities_report, dependent: :destroy
   has_many :organisation_vacancies, dependent: :destroy
-  has_many :organisations, through: :organisation_vacancies, dependent: :destroy, after_add: :refresh_geolocation, after_remove: :refresh_geolocation
+  has_many :organisations,
+           through: :organisation_vacancies,
+           dependent: :destroy,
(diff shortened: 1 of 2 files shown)


app/models/vacancy.rb
```diff
@@ -66,7 +66,7 @@ class Vacancy < ApplicationRecord
   has_many :job_applications, dependent: :destroy
   has_one :equal_opportunities_report, dependent: :destroy
   has_many :organisation_vacancies, dependent: :destroy
-  has_many :organisations, through: :organisation_vacancies, dependent: :destroy, after_add: :refresh_geolocation, after_remove: :refresh_geolocation
+  has_many :organisations, through: :organisation_vacancies, dependent: :destroy, validate: false, after_add: :refresh_geolocation, after_remove: :refresh_geolocation
```
starswan, line 69: looks good - but personally I would have added a test and/or comment to explain. It's hard to spot in this long line

scruti: Added both @starswan

KyleMacPherson (approved):

PR #7212, opened 2024-10-28
Make emergency login keys polymorphic

KyleMacPherson:
## Trello card URL

Work leading up to being able to complete this card: https://trello.com/c/EDeonNSy/1351-onelogin-set-fallback-sign-in-by-email-option-for-jobseekers

## Changes in this PR:

This PR makes the emergency_login_key model polymorphic. Emergency login keys are used for the fallback login method we use for hiring staff currently when we don't have access to DfE sign in (mainly used for testing on review apps).

I am making the emergency log in keys polymorphic so that we can introduce a similar fallback login flow for jobseekers. I think it makes sense for jobseekers also to be able to own emergency login keys in this scenario.

Note: I am destroying all existing emergency log in keys in the migration as they are currently only used by team members to sign in on review apps, are useless after 10 mins and are destroyed after use. Happy to change this to backfill the new column for existing emergency log in keys if the reviewer prefers, but I prefer this approach.

## Next steps:

### DIFF
--- a/db/migrate/20241028151455_add_owner_to_emergency_login_keys.rb
+++ b/db/migrate/20241028151455_add_owner_to_emergency_login_keys.rb
@@ -6,7 +6,9 @@ class AddOwnerToEmergencyLoginKeys < ActiveRecord::Migration[7.1]
     # Make publisher_id nullable to avoid NOT NULL constraint issues before deleting this column
     change_column_null :emergency_login_keys, :publisher_id, true
     # Add polymorphic owner reference
+    # rubocop:disable Rails/NotNullColumn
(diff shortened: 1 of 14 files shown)


db/migrate/20241028151455_add_owner_to_emergency_login_keys.rb
```diff
@@ -0,0 +1,24 @@
+class AddOwnerToEmergencyLoginKeys < ActiveRecord::Migration[7.1]
+  def up
+    # Emergency login keys are useless after 10 mins so no need to backfill them
```
scruti, line 3: Well thought

scruti (approved):

PR #7240, opened 2024-11-07
Fix account merge bug

KyleMacPherson:
## Changes in this PR:

There is currently an account merge bug which allows current jobseeker to delete their active, logged in account. This is happening because after merging data, the account from which the data was merged is then deleted. Prior to this change it was possible for a jobseeker to be logged in as "user@example.com" and request to merge data from "user@example.com" which then would delete their account, logging them out and losing all their data.

This PR fixes this issue by raising an error if the jobseeker tries to merge the account they are logged in as currently into itself.

## Next steps:


scruti (approved):
Thanks for fixing this so quikly

PR #7241, opened 2024-11-07
Fix account transfer edge issues

scruti:
- Do not even start transferring data if from the same user as signed in.
- Fix exceptions when logging errors for non-Activerecord objects.
- Add system test for error flow.


scruti:
"Emergency" merging this after hours to fix production issue that may cause data loss.

PR #7242, opened 2024-11-08
Add API clients management section to supportal

ddippolito:
## Trello card URL
https://trello.com/c/BGZc9dco/1327-add-api-clients-management-to-supportal

## Changes in this PR:
This adds a new section to the support for managing API clients. 
We will shortly build a new vacancies API and migrate the existing ATS integrations towards it.

## Screenshots of UI changes:
Screenshot 2024-11-08 at 10 20 06
Screenshot 2024-11-08 at 10 43 53
Screenshot 2024-11-08 at 10 20 42
Screenshot 2024-11-08 at 10 45 20


app/models/api_client.rb
```diff
@@ -0,0 +1,14 @@
+class ApiClient < ApplicationRecord
+  before_validation :generate_api_key, on: :create
+
+  validates :name, presence: true
+  validates :api_key, presence: true
+
+  def generate_api_key
+    self.api_key ||= SecureRandom.hex(20)
```
starswan, line 8: should this not set last_rotated_at? Then it can be non-null

starswan (approved):

scruti (changes requested):
The PR approach is great. Just what we needed and will make the ATS API management so much better.

My only request is to make the wording more specific.

Since this is a very specific API and subset of users and we have another API around, talking about "API" and "API Clients" seems quite generic and prone to confusion.

I would change it for `Publishers ATS API` or `Publishers API`. So it is clear what we are talking about.

`API Clients` 

image

app/controllers/support_users/api_clients_controller.rb
```diff
@@ -0,0 +1,36 @@
+class SupportUsers::ApiClientsController < ApplicationController
```
scruti, line 1: This is publicly accessible and doesn't require authentication. 

image

PR #7244, opened 2024-11-08
Remove devise functionality from jobseekers

scruti:
## Trello card URL
- https://trello.com/c/yEAtzluZ/1391-govuk-one-login-post-launch-removal-of-devise-unused-code-db-fields

## Changes in this PR:
Remove the code related to Jobseeker functionality that was moved to GovUK One Login:

registration, confirmation, email/password update, account locking, and all the references and recording of passwords.

Also, remove tasks/jobs that depend on confirmed/unconfirmed jobseekers.

Finally, we ignore the related DB table attributes prior to a further PR that will drop these attributes.


config/locales/jobseekers.yml
```diff
@@ -563,57 +558,12 @@ en:
       withdraw:
         success: You have withdrawn your application for %{job_title}
     registrations:
```
starswan, line 560: can line 560 go too?

app/models/jobseeker.rb
```diff
@@ -1,12 +1,20 @@
 class Jobseeker < ApplicationRecord
   has_encrypted :last_sign_in_ip, :current_sign_in_ip
 
+  self.ignored_columns += %w[encrypted_password
```
starswan, line 4: maybe a comment would be nice here? I don't know off-hand what this does

starswan (approved):

PR #7263, opened 2024-11-11
ATS imports skip vacancies without schools

scruti:
There is a persistent issue when some ATS sources list vacancies with a school URN that doesn't match any school in our system.

Sometimes those vacancies are being imported and, when listed, cause errors for the jobseekers.

Instead of relying on the validation attempt before recording the vacancy, we will completely skip the vacancy from being even built unless it is associated with an existing school.


app/services/vacancies/import/sources/ark.rb
```diff
@@ -42,6 +42,7 @@ def self.source_name
   def each
     items.each do |item|
       schools = find_schools(item)
+      next if schools.blank?
```
scruti, line 45: This is the actual fix for all the sources.

app/services/vacancies/import/sources/ark.rb
```diff
@@ -186,11 +187,13 @@ def organisation_fields(schools)
...
+
     school_urn = item.supp_value
+    schools = Organisation.where(urn: school_urn) if school_urn.present?
 
-    return [] if multi_academy_trust.blank? && school_urn.blank?
-    return Organisation.where(urn: school_urn) if multi_academy_trust.blank?
-    return Array(multi_academy_trust) if school_urn.blank?
+    return [] if multi_academy_trust.blank? && schools.blank?
+    return schools if multi_academy_trust.blank?
+    return Array(multi_academy_trust) if schools.blank?
```
scruti, line 196: This is more of a refactor to avoid triggering DB queries unless the URN to search for is present.

scruti:
@ddippolito good point. Going to get sure we have a ticket for it on the tech debt backlog.

PR #7292, opened 2024-11-18
Upgrade defaults ror 7.1

ddippolito:
## Trello card URL
https://trello.com/c/2FJobHjP/1401-rails-71-defaults-not-loaded-need-to-run-rails-appupdate-and-migrate-settings-forwards

## Changes in this PR:
Add rails 7.1 defaults

### DIFF
--- a/spec/configuration/filter_parameter_logging_spec.rb
+++ b/spec/configuration/filter_parameter_logging_spec.rb
@@ -7,7 +7,10 @@ RSpec.describe "Filter parameter logging configuration" do
   specify "all anonymised analytics fields should be filtered from logs" do
     analytics_hidden_pii.each_value do |shared|
       shared.each do |field|
-        expect(filter_params).to include(field.to_sym)
+        matched = filter_params.any? do |pattern|
+          pattern.is_a?(Regexp) ? pattern.match?(field.to_s) : pattern == field.to_sym
+        end
+        expect(matched).to be(true), "Expected #{field} to be included in filter parameters"
       end
     end
   end


config/application.rb
```diff
@@ -28,6 +28,8 @@ module TeachingVacancies
   class Application < Rails::Application
     config.load_defaults 7.0
```
starswan, line 29: should this not be 7.1 now?

app/models/subscription.rb
```diff
@@ -10,27 +10,31 @@ class Subscription < ApplicationRecord
...
 
-    ActiveSupport::MessageEncryptor.new(key_generator)
+    ActiveSupport::MessageEncryptor.new(key_generator, serializer: serializer)
   end
 
   def self.find_and_verify_by_token(token)
-    data = encryptor.decrypt_and_verify(token)
-    find(data[:id])
+    data = begin
+      encryptor(serializer: :json_allow_marshal).decrypt_and_verify(token)
```
starswan, line 26: do we need this change? The config seems to suggest that this is the default behaviour?

starswan, line 26: do we need this change? It looks this would get even more complicated with Rails 7.2. Are there any down-sides to just using the default?

app/models/subscription.rb
```diff
@@ -10,27 +10,31 @@ class Subscription < ApplicationRecord
...
+    end
+    find(data.symbolize_keys[:id])
   rescue ActiveSupport::MessageEncryptor::InvalidMessage
     raise ActiveRecord::RecordNotFound
   end
 
   def token
     token_values = { id: id }
-    self.class.encryptor.encrypt_and_sign(token_values)
+    self.class.encryptor(serializer: :json_allow_marshal).encrypt_and_sign(token_values)
```
starswan, line 37: having defaulted the encryptor to :json_allow_marshal, we don't need to make this change IMHO

app/models/subscription.rb
starswan: Think a comment (as per your slack message) would be useful here

starswan: I meant in the code...

starswan (approved):

app/controllers/publishers/candidate_profiles_interstitials_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Publishers::CandidateProfilesInterstitialsController < Publishers::BaseController
   skip_before_action :check_candidate_profiles_interstitial_acknowledged, only: %i[show]
-  skip_before_action :check_terms_and_conditions, only: %i[show update]
+  skip_before_action :check_terms_and_conditions, only: %i[show]
```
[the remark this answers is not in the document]
scruti: Oh awesome. Thanks for the clarification!

PR #7300, opened 2024-11-19
Namespace ATS-API and and extend its documentation

scruti:
## Trello card URL
- https://trello.com/c/1q8pnJVz/1386-ats-api-specification-errors-namespacing

## Notes for reviewers
- To check the API docs: https://teaching-vacancies-review-pr-7300.test.teacherservices.cloud/ats-api-docs/index.html
- The rswagger request specs for this API are failing. This is ok as they're used for generating documentation. We will make them pass when implementing the API.

### Move the ATS API route from `api/v2` to `ats-api/v1`
The original API in the service is an API that the public can use to get information about the posted vacancies.

The new API will be private and only accessible to the ATS so they can publish their vacancies in our system.

To avoid confusion and aiming for developers to identify the scope of the API through its namespace, we extracted the new API definition under the following path in our project:

"app/controller/publishers/ats_api/v1/...."

To avoid a long convoluted API endpoint for our clients, we will expose it under `/ats-api/v1` path.

### Changes on Swagger documentation generation on CI
- Extend the cache key to invalidate the cache and regenerate the docs if either the specs, controller or swagger helper definitions are modified.
- Rearrange documentation generation steps so they do not run when the cache is restored.

### Remove API version from required params.
Instead of requiring a parameter on every request, make it part of the API endpoint URL.

### Declare authorization format
API key through the request header.

### Add error responses with examples.
From standard errors like "Unauthorized", "Not found"... to a validation error.

### Extract common Vacancy schema.

So all the responses that return a Vacancy object have the same schema/examples.

### Before
image
image

### After
https://teaching-vacancies-review-pr-7300.test.teacherservices.cloud/ats-api-docs

image
image
image
image

#### Run generating the docs without restoring the cache and saving the cache after:
image

#### Run restoring cache and skipping docs generation:
image


spec/swagger_helper.rb
```diff
@@ -30,6 +30,248 @@
...
+                  properties: {
+                    error: { type: "string", example: "unauthorized" },
+                    message: { type: "string", example: "Invalid API key" },
+                  },
+                  required: %w[error message],
+                },
+              },
+            },
+          },
+          not_found_error: {
```
scruti, line 224: I duplicated all these schemas with the same structure to have individualized examples per error.

I believe once the API gets implemented and the request specs properly set, a single error schema would cover these cases, and the autogenerated examples for each error response could be set in the test file.

spec/swagger_helper.rb
```diff
@@ -30,6 +30,280 @@
...
+            type: "object",
+            properties: {
+              errors: {
+                type: "array",
+                items: {
+                  type: "object",
+                  properties: {
+                    error: { type: "string", example: "Not found" },
```
starswan, line 248: do we need both the 'error' and the 'message' in these schemas? It might be confusing to caller to know which one to look at.

scruti: Good point. We have declared the specific HTTP error codes, so the "error" key is redundant, you're right.

I was leaving room for multiple errors being returned. But this is not the case (besides on validations) since they're mutually exclusive. 

I'm going to update these schemas/examples. Thanks for bringing this up!

starswan (approved):

PR #7304, opened 2024-11-20
Upgrade rubocop-govuk gem to latest version

scruti:
- Brings a fix from rubocop on the recent exceptions on StringIO calls.
- Fixes new version rule violations detected in the project for both Rubocop and Slim-lint.


app/form_models/base_form.rb
```diff
@@ -34,4 +34,16 @@ def send_errors_to_big_query
...
+    number_of_words = remove_html_tags(attribute)&.split&.length
+
+    number_of_words&.>(number)
+  end
+
+  def remove_html_tags(field)
+    regex = /<("[^"]*"|'[^']*'|[^'">])*>/
+
+    field&.gsub(regex, "")
+  end
```
scruti, line 48: It was duplicated across multiple forms. So I have extracted it to the common parent definition at the same time that fixed the style violation.

starswan (approved):

PR #7310, opened 2024-11-22
Fix transcript bug on guidance pages

KyleMacPherson:
## Trello card URL
https://trello.com/c/frDPPOcM/1417-do-not-show-breadcrumbs-for-transcript-pages-on-guides-posts

## Changes in this PR:

- Remove breadcrumbs from transcripts guides/posts
- Show not found page if someone navigates to the transcripts category of posts

## Next steps:

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -1,6 +1,7 @@
 class PostsController < ApplicationController
   def index
     return not_found if params[:section] == "transcripts"
+    
     @subcategories = MarkdownDocument.all_subcategories(params[:section])
   end
 
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -1,7 +1,7 @@
 class PostsController < ApplicationController
   def index
+    return not_found if params[:section] == "transcripts"
     @subcategories = MarkdownDocument.all_subcategories(params[:section])
-    not_found if params[:section] == "transcripts"
   end
 
   def subcategory
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -1,6 +1,7 @@
 class PostsController < ApplicationController
   def index
     @subcategories = MarkdownDocument.all_subcategories(params[:section])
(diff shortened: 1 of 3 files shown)


app/views/posts/show.html.slim
```diff
@@ -2,12 +2,13 @@
...
-                                       params[:section].titleize.capitalize => posts_path(section: params[:section]),
-                                       formatted_subcategory => subcategory_path(section: params[:section], subcategory: params[:subcategory]),
-                                       @post.title => "" }
+- unless params[:section] == "transcripts"
+  - content_for :breadcrumbs do
+    nav aria-label="breadcrumb" role="navigation"
+      = govuk_breadcrumbs breadcrumbs: { t("breadcrumbs.home") => root_path,
+                                        params[:section].titleize.capitalize => posts_path(section: params[:section]),
+                                        formatted_subcategory => subcategory_path(section: params[:section], subcategory: params[:subcategory]),
+                                        @post.title => "" }
```
starswan, line 11: I'm slightly confused - this is in the 'show' action, but the controller change is in the 'index' action...?

app/controllers/posts_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PostsController < ApplicationController
   def index
     @subcategories = MarkdownDocument.all_subcategories(params[:section])
+    not_found if params[:section] == "transcripts"
```
starswan, line 4: should this not trigger earlier to avoid calling (the presumably invalid) MarkdownDocument.all_subcategories("transcripts") ?

app/views/posts/show.html.slim
KyleMacPherson: @starswan this is to remove breadcrumbs when we show an individual piece of guidance, the index action change is to ensure we don't show an overview page for the transcripts as a collection

app/controllers/posts_controller.rb
KyleMacPherson: Sure, can do!

scruti (approved):
LGTM!

PR #7318, opened 2024-11-28
Speed alert emails

starswan:
## Trello card URL

https://trello.com/c/w1KM928B/1414-refactor-daily-weeklyalerts-job-for-a-high-volume-of-subscriptions

## Changes in this PR:

Currently each (of 100k) subscriptions triggers an SQL query for the same (small) subset of recent vacancies. This change turns that on its head - if there are no keyword or location filters, the filtering is now all done in memory on the 'recent' subset (around 2000 for the 'weekly' email). 

## Next steps:


spec/rails_helper.rb
```diff
@@ -104,9 +109,12 @@
...
-  config.before(:each, geocode: true) do
-    allow(Geocoder).to receive(:search).and_call_original
-    allow(Rails.application.config).to receive(:geocoder_lookup).and_return(:default)
+  config.before do
+    allow(DisableExpensiveJobs).to receive(:enabled?).and_return(false)
+  end
+
+  config.before(:each, disable_expensive_jobs: true) do
+    allow(DisableExpensiveJobs).to receive(:enabled?).and_return(true)
   end
```
scruti, line 118: Very nice :clap:

app/jobs/alert_email/base.rb
```diff
@@ -1,19 +1,20 @@
 class AlertEmail::Base < ApplicationJob
+  MAXIMUM_RESULTS_PER_RUN = 500
```
scruti, line 2: Is this still needed? I only see [this removed call](https://github.com/DFE-Digital/teaching-vacancies/pull/7318/files#diff-bad11ea43382a817d4981b22e8afb2e5e790dd4c339d39866443649920bd107eL46)

app/jobs/alert_email/base.rb
```diff
@@ -1,19 +1,20 @@
...
   def perform
     return if DisableExpensiveJobs.enabled?
 
-    subscriptions.find_each do |subscription|
-      next if subscription.alert_run_today?
+    # The intent here is that if we don't have keyword or location searches, then this operation can all be done in memory
+    # really fast (1 week's worth of vacancies is around 2000, so not worth leaving on disk for each of 100k daily subscriptions
+    default_scope = Vacancy.includes(:organisations).live.order(publish_on: :desc).search_by_filter(from_date: from_date, to_date: Date.current)
+
+    already_run_ids = AlertRun.for_today.map(&:subscription_id)
```
scruti, line 11: I think we can avoid a step here an directly query the ids from DB:
```suggestion
    already_run_ids = AlertRun.for_today.pluck(:subscription_id)
```

app/jobs/alert_email/base.rb
```diff
@@ -1,19 +1,20 @@
...
 
-      vacancies = vacancies_for_subscription(subscription)
-      next if vacancies.blank?
+    subscriptions.find_each.reject { |sub| already_run_ids.include?(sub.id) }.each do |subscription|
+      vacancies = subscription.vacancies_matching default_scope
```
scruti, line 14: This seems to be looping twice over the subscriptions batch: once for the reject, and a second for finding/alerting.
We could avoid the first loop:

```suggestion
    subscriptions.find_each.each do |subscription|
       next if already_run_ids.include?(subscription.id)
       
      vacancies = subscription.vacancies_matching default_scope
```

scruti (approved):
Great work! Left a few comments for possible optimizations, but nothing blocking!

app/jobs/alert_email/base.rb
starswan: It was - I've out the usage back.

PR #7336, opened 2024-12-05
Fix "data" arg format for all analytics events

scruti:
## Trello card URL
- https://trello.com/c/6BDuNhsQ/1460-investigate-analytics-trackedlinkclicked-not-reaching-bigq
## Changes in this PR:

dfe-analytics v1.13 introduced a unannounced (and seems unexpected) breaking change on how the "data" argument must be formatted for the events.

Since then, we adapted our calls in some instances when we wanted/needed to use the new 'hidden_data' option.

The problem is, as much as in their documentation state that you can call a custom event with `.with_data(some: "info")`, in reality all the events instantiated like that get their data dropped unless are wrapped within a `data:` key: `.with_data(data: { some: "info" })`

- Ref: https://github.com/DFE-Digital/dfe-analytics/issues/182

### New custom test matcher

As part of these changes, I've copied and extended the `DfE::Analytics` `have_enqued_as_analytics_events` rspec matcher.

Our `have_enqued_as_analytics_event` matcher will be able to assert against the data contained in the event created by the gem, as
- First option. Assert that the event type was enqueued:
  `expect(:event_type).to have_been_enqueued_as_analytics_event`

- Second option. Assert that the event type was enqueued with specific data:
  `expect(:event_type).to have_been_enqueued_as_analytics_event(with_data: { key: "value" })`

- Third option. Assert that the event type was enqueued with specific data keys (without asserting against the values):
  `expect(:event_type).to have_been_enqueued_as_analytics_event(with_data: [:key1, :key2])`

## Screenshots of UI changes:

Screenshot From 2024-12-09 18-11-22
Screenshot From 2024-12-09 19-36-13

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -146,6 +146,11 @@ Style/HashSyntax:
 # This is incompatible with our other migration checks
 Rails/BulkChangeTable:
   Enabled: false
+
+# Incompatible with DfE-Analytics custom matcher we have extended (have_been_enqueued_as_analytics_event)
+RSpec/ExpectActual:
+  Enabled: false
+
 ##
 # Tech Debt
 # TODO: `rubocop-govuk` previously had all these metrics disabled as they are "just heuristics".
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -147,10 +147,6 @@ Style/HashSyntax:
 Rails/BulkChangeTable:
   Enabled: false
 
-# Incompatible with DfE-Analytics custom matcher we have extended (have_been_enqueued_as_analytics_event)
-RSpec/ExpectActual:
-  Enabled: false
-
 ##
 # Tech Debt
 # TODO: `rubocop-govuk` previously had all these metrics disabled as they are "just heuristics".
--- a/spec/system/publishers/publishers_can_extend_a_deadline_spec.rb
+++ b/spec/system/publishers/publishers_can_extend_a_deadline_spec.rb
@@ -90,7 +90,7 @@ RSpec.describe "Publishers can extend a deadline" do
       end
 
       it "sends an event to analytics" do
-        expect(:publisher_vacancy_relisted).to have_been_enqueued_as_analytics_event(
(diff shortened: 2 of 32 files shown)


.rubocop.yml
```diff
@@ -146,6 +146,11 @@ Style/HashSyntax:
 # This is incompatible with our other migration checks
 Rails/BulkChangeTable:
   Enabled: false
+
+# Incompatible with DfE-Analytics custom matcher we have extended (have_been_enqueued_as_analytics_event)
```
starswan, line 150: could we mark these as exceptions instead of having this in the global config?

spec/system/publishers/publishers_can_extend_a_deadline_spec.rb
```diff
@@ -90,7 +90,11 @@
       end
 
       it "sends an event to analytics" do
-        expect(:publisher_vacancy_relisted).to have_been_enqueued_as_analytics_events
+        expect(:publisher_vacancy_relisted).to have_been_enqueued_as_analytics_event(
+          with_data: {
```
starswan, line 94: This looks different from the others - are we sure that this is correct? (That the data is stringified inside the relist-form key)

starswan (approved):

spec/system/publishers/publishers_can_extend_a_deadline_spec.rb
scruti: Yes, this comes from the gem.
When the given value for a data key is a hash, [the gem calls `to_json` over it](https://github.com/DFE-Digital/dfe-analytics/blob/main/lib/dfe/analytics/event.rb#L105-L106). 

It seems needed as `data --> value` field in the BigQ table schema is defined as a STRING.

.rubocop.yml
scruti: I was hesitant to do this as they're a lot of them, but lets give it a try :eyes:

PR #7339, opened 2024-12-09
Summary page task list

starswan:
## Trello card URL

https://trello.com/c/X2IDrHKp/1433-application-summary-page

## Changes in this PR:

## Next steps:


.github/workflows/lint.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: bundle exec rubocop
 
       - name: Run Slim-Lint with PR annotations
```
scruti, line 27: From the change below looks like we're removing "report with PR annotations" flag. 
Adapting the description:
```suggestion
      - name: Run Slim-Lint
```

scruti (approved):

app/components/job_application_review_component/section.rb
```diff
@@ -21,10 +22,11 @@ def heading_text
...
 
-  def build_list
-    list = nil
-    govuk_summary_list { |l| list = l }
-    list
+  def edit_link
+    title = heading_text
+    text = "Change"
+    href = error_path
+    govuk_link_to text, href, aria: { label: "#{text} #{title}" }, classes: "govuk-!-display-none-print" if text && href
```
scruti, line 29: If I read this right the variable declarations and conditional link are  redundant? `text` and `href` will always be present, right?

Also... the "change" link is hardcoded to an "error_path"? Is this intentional? I have the feeling I am missing a bigger picture than this little bit of code :sweat_smile: 

```suggestion
    govuk_link_to "Change", error_path, aria: { label: "#{text} #{heading_text}" }, classes: "govuk-!-display-none-print"
```

starswan: error path isn't always set. This piece of code is old, its just re-located into here

scruti (approved):

PR #7352, opened 2024-12-16
Add Awarding Body to qualification result

starswan:
## Trello card URL

https://trello.com/c/BKrMfPoz/1438-add-awarding-body-and-design-changes-to-gcse-page-application-and-profiles

## Changes in this PR:

### Before
QualliesBefore

### After
QualliesAfter

## Next steps:


app/controllers/jobseekers/profiles/qualifications_controller.rb
```diff
@@ -1,24 +1,33 @@
 class Jobseekers::Profiles::QualificationsController < Jobseekers::ProfilesController
   include Jobseekers::QualificationFormConcerns
 
-  helper_method :category, :form, :jobseeker_profile, :qualification, :secondary?, :qualification_form_param_key
+  helper_method :jobseeker_profile, :qualification, :qualification_form_param_key
+
+  before_action :set_form_and_category, except: %i[review confirm_destroy destroy select_category submit_category]
```
scruti, line 6: Personal preference suggestion. But I would understand this better if it was an `only: %i[new create update]` (I believe these are the only methods it applies to?) rather than a long exclusion list.

scruti (commented):
Could we add some test coverage on this new field filling?

app/controllers/jobseekers/profiles/qualifications_controller.rb
starswan: it was a growing line during the story - so thanks, much better plan to just specify what's included

scruti (approved):
LGTM!

PR #7353, opened 2024-12-17
Enable federated auth for big query

vipin-dfe:
## Trello card URL
https://trello.com/c/yqj8hYRR

## Context
All services will be migrated to the workload identity federation for improved security.

## Changes in this PR:
Enable federated authentication for Google Big query

## Next steps:


neillturner (commented):
i think you need to set the 
"dataset_name" for all the workspace variables tfvars.json because the fix to enable null is not in the stable terraform module branch

terraform/app/modules/paas/variables.tf
```diff
@@ -96,6 +96,15 @@ variable "enable_logit" {}
 
 variable "aks_web_app_memory" {}
 
+variable "enable_dfe_analytics_federated_auth" {
+  description = "Create the resources in Google cloud for federated authentication and enable in application"
+  default     = false
```
saliceti, line 101: Can we remove the variable defaults at the module level? Defaults are already set at the root level

vipin-dfe: Removed.

neillturner (approved):
lgtm. i assume all the gcp entity names are correct

shaheislamdfe (commented):
lgtm

PR #7368, opened 2025-01-06
Ensure flexi_working column is populated with nil if the flexi_working form value only includes html tags

KyleMacPherson:
## Trello card URL

https://trello.com/c/27mBZozy/1474-flexi-working-box-stop-paras-ps-from-being-saved

## Changes in this PR:

Prior to this change the flexi_working column of vacancies could be populated with only html tags, even if no actual text was entered into this box. This change fixes that issue by adding a params_to_save method, following a pattern used in many other forms in our service, which ensures that the flexi_working column is only populated if the user enters text into the text box.

## Next steps:


scruti (approved):
LGTM!

PR #7374, opened 2025-01-08
Ats api implementation

ddippolito:
## Changes in this PR:

## Next steps:

### DIFF
--- a/app/controllers/publishers/ats_api/v1/vacancies_controller.rb
+++ b/app/controllers/publishers/ats_api/v1/vacancies_controller.rb
@@ -5,7 +5,7 @@ class Publishers::AtsApi::V1::VacanciesController < Api::ApplicationController
   rescue_from StandardError, with: :render_server_error
   rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
   rescue_from ActionController::ParameterMissing, with: :render_bad_request
-  rescue_from Publishers::AtsApi::CreateVacancyService::InvalidOrganisationError, with: :render_bad_request
+  rescue_from Publishers::AtsApi::OrganisationFetcher::InvalidOrganisationError, with: :render_unprocessable_entity
 
   def index
     @pagy, @vacancies = pagy(vacancies.where(publisher_ats_api_client: client), items: 100)
@@ -14,7 +14,7 @@ class Publishers::AtsApi::V1::VacanciesController < Api::ApplicationController
   end
 
   def show
-    @vacancy = Vacancy.find(params[:id])
+    @vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
 
     respond_to(&:json)
   end
@@ -26,7 +26,8 @@ class Publishers::AtsApi::V1::VacanciesController < Api::ApplicationController
   end
 
   def update
-    vacancy = Vacancy.find(params[:id])
+    vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
+
     result = Publishers::AtsApi::UpdateVacancyService.call(vacancy, permitted_vacancy_params)
 
     if result[:success]
@@ -38,7 +39,7 @@ class Publishers::AtsApi::V1::VacanciesController < Api::ApplicationController
   end
 
   def destroy
-    vacancy = Vacancy.find(params[:id])
+    vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
     vacancy.destroy!
 
     head :no_content
@@ -103,5 +104,9 @@ class Publishers::AtsApi::V1::VacanciesController < Api::ApplicationController
 
   def render_bad_request(exception = nil)
     render json: { error: exception&.message.presence || "Request body could not be read properly" }, status: :bad_request
+  end
+
+  def render_unprocessable_entity(exception)
+    render json: { error: exception.message }, status: :unprocessable_entity
   end
 end
--- a/app/services/publishers/ats_api/organisation_fetcher.rb
+++ b/app/services/publishers/ats_api/organisation_fetcher.rb
@@ -1,6 +1,8 @@
 module Publishers
   module AtsApi
     module OrganisationFetcher
+      InvalidOrganisationError = Class.new(StandardError)
+
       def fetch_organisations(school_params)
         return [] unless school_params
 
@@ -12,7 +14,7 @@ module Publishers
         multi_academy_trust = SchoolGroup.trusts.find_by(uid: trust_uid)
         schools = ::Organisation.where(urn: school_urns) if school_urns.present?
 
-        return [] if multi_academy_trust.blank? && schools.blank?
+        raise InvalidOrganisationError, "No valid organisations found" if multi_academy_trust.blank? && schools.blank?
         return schools.to_a if multi_academy_trust.blank?
         return Array(multi_academy_trust) if schools.blank?
 
--- a/app/services/publishers/ats_api/update_vacancy_service.rb
+++ b/app/services/publishers/ats_api/update_vacancy_service.rb
@@ -18,7 +18,6 @@ module Publishers
 
         def sanitised_params(params)
           organisations = fetch_organisations(params[:schools])
-          raise ActiveRecord::RecordNotFound, "No valid organisations found" if organisations.blank?
 
           params.except(:schools).merge(organisations: organisations)
         end
--- a/spec/requests/publishers/ats_api/v1/vacancies_spec.rb
+++ b/spec/requests/publishers/ats_api/v1/vacancies_spec.rb
@@ -425,7 +425,7 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
   path "/ats-api/v1/vacancies/{id}" do
     parameter name: "id", in: :path, type: :string, description: "id of the vacancy"
 
-    let(:id) { create(:vacancy, :external).id }
+    let(:id) { create(:vacancy, :external, publisher_ats_api_client: client).id }
 
     get("show vacancy") do
       tags "Vacancies"
@@ -467,10 +467,8 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
       response(500, "Internal server error") do
         schema "$ref" => "#/components/schemas/internal_server_error"
 
-        let(:id) { "123" }
-
         before do
-          allow(Vacancy).to receive(:find).and_raise(StandardError.new("Internal server error"))
+          allow(Vacancy).to receive(:find_by!).and_raise(StandardError.new("Internal server error"))
         end
 
         run_test!
@@ -737,7 +735,7 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
       response(422, "Validation error") do
         schema "$ref" => "#/components/schemas/validation_error"
 
-        let(:id) { create(:vacancy, :external).id }
+        let(:id) { create(:vacancy, :external, publisher_ats_api_client: client).id }
 
         let(:vacancy) do
           {
@@ -767,8 +765,6 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
       response(500, "Internal server error") do
         schema "$ref" => "#/components/schemas/internal_server_error"
 
-        let(:id) { "123" }
-
         let(:vacancy) do
           {
             vacancy: {
@@ -793,7 +789,7 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
         end
 
         before do
-          allow(Vacancy).to receive(:find).and_raise(StandardError.new("Internal server error"))
+          allow(Vacancy).to receive(:find_by!).and_raise(StandardError.new("Internal server error"))
         end
 
         run_test!
@@ -829,10 +825,8 @@ RSpec.describe "ats-api/v1/vacancies", openapi_spec: "v1/swagger.yaml" do
       response(500, "Internal server error") do
         schema "$ref" => "#/components/schemas/internal_server_error"
 
-        let(:id) { "123" }
-
         before do
-          allow(Vacancy).to receive(:find).and_raise(StandardError.new("Internal server error"))
+          allow(Vacancy).to receive(:find_by!).and_raise(StandardError.new("Internal server error"))
         end
 
         run_test!
--- a/app/services/publishers/ats_api/create_vacancy_service.rb
+++ b/app/services/publishers/ats_api/create_vacancy_service.rb
@@ -3,8 +3,6 @@ module Publishers
     class CreateVacancyService
       extend OrganisationFetcher
 
-      InvalidOrganisationError = Class.new(StandardError)
-
       class << self
         def call(params)
           vacancy = Vacancy.new(sanitised_params(params))
@@ -24,7 +22,6 @@ module Publishers
 
         def sanitised_params(params)
           organisations = fetch_organisations(params[:schools])
-          raise InvalidOrganisationError, "No valid organisations found" if organisations.blank?
 
           params[:publish_on] ||= Time.zone.today.to_s
           params.except(:schools).merge(organisations: organisations)
--- a/spec/services/publishers/ats_api/create_vacancy_service_spec.rb
+++ b/spec/services/publishers/ats_api/create_vacancy_service_spec.rb
@@ -70,9 +70,9 @@ RSpec.describe Publishers::AtsApi::CreateVacancyService do
     context "when organisations are invalid" do
       let(:school_urns) { { school_urns: [9999] } }
 
-      it "raises ActiveRecord::RecordNotFound" do
+      it "raises Publishers::AtsApi::OrganisationFetcher::CreateVacancyService" do
         expect { create_vacancy_service }.to raise_error(
-          Publishers::AtsApi::CreateVacancyService::InvalidOrganisationError,
+          Publishers::AtsApi::OrganisationFetcher::InvalidOrganisationError,
           "No valid organisations found",
         )
       end
--- a/spec/services/publishers/ats_api/update_vacancy_service_spec.rb
+++ b/spec/services/publishers/ats_api/update_vacancy_service_spec.rb
@@ -41,8 +41,11 @@ RSpec.describe Publishers::AtsApi::UpdateVacancyService do
     context "when organisations are invalid" do
       let(:school_urns) { { school_urns: [9999] } }
 
-      it "raises ActiveRecord::RecordNotFound" do
-        expect { udpate_vacancy_service }.to raise_error(ActiveRecord::RecordNotFound, "No valid organisations found")
+      it "raises Publishers::AtsApi::CreateVacancyService::InvalidOrganisationError" do
+        expect { udpate_vacancy_service }.to raise_error(
+          Publishers::AtsApi::OrganisationFetcher::InvalidOrganisationError,
+          "No valid organisations found",
+        )
       end
     end
 


app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,83 @@
...
 
+  rescue_from StandardError, with: :render_server_error
+  rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+  rescue_from ActionController::ParameterMissing, with: :render_bad_request
+
   def index
-    @pagy, @vacancies = pagy(vacancies, items: 100)
+    @pagy, @vacancies = pagy(vacancies.where(publisher_ats_api_client: client), items: 100)
 
-    respond_to(&:json)
```
starswan, line 7: I'm not sure what the motivation was to stop using jbuilder?

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,83 @@
...
-
-    respond_to do |format|
-      format.json do
-        if @vacancy.persisted?
-          render status: :created
-        else
-          render status: :bad_request
-        end
-      end
+    result = Publishers::AtsApi::V1::CreateVacancyService.new(permitted_vacancy_params).call
```
starswan, line 23: don't really understand this 'create service object' pattern. Could it not just be a function call?

app/models/vacancy.rb
```diff
@@ -137,6 +138,16 @@ def organisation
     organisations.find(&:trust?) || publisher_organisation || organisations.first&.school_groups&.first
   end
 
+  def schools
```
starswan, line 141: This feels like its in the wrong place. It's called 'schools' but it doesn't return schools which feels very confusing for a model method.

config/analytics_blocklist.yml
```diff
@@ -86,18 +95,28 @@
   - in_progress_steps
   - employment_history_section_completed
   - qualifications_section_completed
+  - following_religion
```
starswan, line 98: some stuff has clearly got mixed up in this PR...

db/migrate/20241217145653_add_publisher_ats_api_client_reference_to_vacancy_table.rb
```diff
@@ -0,0 +1,16 @@
+class AddPublisherAtsApiClientReferenceToVacancyTable < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
```
starswan, line 2: I'm not sure if this is the right strategy? The index create will be trivial (as there are no references), so I suspect it would be safer to keep the DDL transactions switched on and tell our library that we know what we're doing in this case

spec/requests/publishers/ats_api/v1/create_vacancy_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      let(:job_advert) { nil }
+      let(:job_roles) { [] }
+      let(:working_patterns) { [] }
+
+      let(:expected_response) do
+        {
+          status: :unprocessable_entity,
+          json: {
+            errors: [
+              { error: "job_title: can't be blank" },
```
starswan, line 79: do we need 'error' as the key here? I might have just expected the keys to be field names, and the values be the errors - in fact I thought that's what the swagger spec said?

spec/requests/publishers/ats_api/v1/update_vacancy_service_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    context "when organisations are invalid" do
+      let(:school_urns) { { school_urns: [9999] } }
+
+      it "raises ActiveRecord::RecordNotFound" do
+        expect { service.call }.to raise_error(ActiveRecord::RecordNotFound, "No valid organisations found")
+      end
+    end
+
+    context "when the vacancy fails validation" do
```
starswan, line 49: This feels strange - effectively the service is the controller. Would this code not be better in the controller?

spec/requests/publishers/ats_api/v1/vacancies_spec.rb
```diff
@@ -140,15 +168,15 @@
                 required: %i[trust_uid school_urns],
                 properties: {
                   trust_uid: {
-                    type: :integer,
-                    example: 12_345,
+                    type: :string,
```
starswan, line 171: I thought trust_uid was an integer?

spec/requests/publishers/ats_api/v1/vacancies_spec.rb
```diff
@@ -382,21 +513,220 @@
...
+                     working_patterns
+                     contract_type
+                     phases
+                     schools],
+        properties: {
+          external_advert_url: { type: :string, example: "https://example.com/jobs/123" },
+          publish_on: { type: :string, format: :date },
+          expires_at: { type: :string, format: :date },
+          job_title: { type: :string, example: "Teacher of Geography" },
+          job_advert: { type: :string, example: "Teacher of Geography" },
```
starswan, line 538: oops typo

spec/requests/publishers/ats_api/v1/vacancies_spec.rb
```diff
@@ -1,8 +1,7 @@
 require "swagger_helper"
 
-# rubocop:disable RSpec/EmptyExampleGroup
```
starswan, line 3: This is normal for swagger  specs

spec/swagger_helper.rb
```diff
@@ -64,11 +64,11 @@
...
               skills_and_experience: { type: :string, example: "We're looking for a dedicated Teacher of Geography" },
               salary: { type: :string, example: "£12,345 to £67, 890" },
-              benefits_details: { type: :string, example: "TLR2a" },
-              starts_on: { type: :string, example: "Summer Term" },
-              external_reference: { type: :string, example: "123GTZY" },
-              visa_sponsorship_available: { type: :boolean },
-              is_job_share: { type: :boolean },
+              benefits_details: { type: :string, example: "TLR2a", nullable: true },
+              starts_on: { type: :string, example: "Summer Term", nullable: true },
+              external_reference: { type: :string, example: "123GTZY", nullable: true },
```
starswan, line 69: external reference should be mandatory?

spec/swagger_helper.rb
```diff
@@ -64,11 +64,11 @@
...
+              visa_sponsorship_available: { type: :boolean, nullable: true },
```
starswan, line 70: IMHO these shouldn't be nullable - they are just optional properties. Otherwise we have 2 'null' states - missing and present null.

spec/swagger_helper.rb
```diff
@@ -92,15 +92,15 @@
...
+                        example: "12345",
                       },
                       school_urns: {
                         type: :array,
                         minItems: 0,
                         items: {
-                          type: :integer,
-                          example: 12_345,
+                          type: :string,
+                          example: "12345",
```
starswan, line 103: isn't URN a number?

spec/swagger_helper.rb
```diff
@@ -265,4 +277,5 @@
   # the key, this may want to be changed to avoid putting yaml in json files.
   # Defaults to json. Accepts ':json' and ':yaml'.
   config.openapi_format = :yaml
+  config.openapi_no_additional_properties = false # Allow additional properties
```
starswan, line 280: not sure we want to allow this?

spec/swagger_helper.rb
```diff
@@ -205,6 +205,18 @@
...
             },
           },
+          create_vacancy_response: {
+            type: :object,
+            properties: {
+              id: { type: :string, example: "9d8f5715-2e7c-4e64-8e34-35f510c12e66" },
+              errors: {
+                type: :array,
+                items: { type: :string },
+                nullable: true,
```
starswan, line 215: again don't want a nullable array

spec/swagger_helper.rb
```diff
@@ -205,6 +205,18 @@
...
+              },
+            },
+            required: %w[id],
```
starswan, line 218: if errors are possible, then id can't be required. Maybe this should be oneOf id an errors?

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,81 @@
...
+          params[:working_patterns] ||= []
+          params[:phases] ||= []
+
+          params.except(:schools, :trust_uid).merge(organisations: organisations)
+        end
+
+        def fetch_organisations(school_params)
+          return [] unless school_params
+
+          if school_params[:trust_uid].present?
```
starswan, line 41: do we want to check that this client can use these URNs?

app/controllers/publishers/ats_api/v1/update_vacancy_service.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        private
+
+        attr_reader :vacancy, :params
+
+        def permitted_params
+          organisations = fetch_organisations(params[:schools])
+          raise ActiveRecord::RecordNotFound, "No valid organisations found" if organisations.blank?
+
+          params[:publish_on] ||= Time.zone.today.to_s
```
starswan, line 26: if this were done after the merge, we would be altering a copy rather than the original?

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,83 @@
...
+
+    if result[:headers]
```
starswan, line 25: is this idiomatic to alter the headers hash in this way? Could we have not just put an error in the JSON instead? It might be hard to document this behaviour, so clients might miss it?

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,83 @@
...
+      contract_type
+      phases
+      schools
+    ]
   end
 
-  def vacancy
-    FactoryBot.build(:vacancy, :external)
+  def permitted_vacancy_params
+    missing_keys = required_vacancy_keys - params.fetch(:vacancy, {}).keys.map(&:to_sym)
```
starswan, line 67: I think we came up with a way of validating the input request against the JSON schema in the swagger  in my last project - https://github.com/ministryofjustice/cfe-civil/blob/d700c947da8474c45323675aaa39583022cca185/app/controllers/v6/assessments_controller.rb#L3

spec/jobs/import_from_vacancy_source_job_spec.rb
```diff
@@ -101,7 +101,7 @@ def each(...)
         expect(FailedImportedVacancy.first.external_reference).to eq("J3D1")
         expect(FailedImportedVacancy.first.import_errors.first).to eq("job_title:[can't be blank]")
         expect(FailedImportedVacancy.first.import_errors.last).to eq("phases:[can't be blank]")
-        expect(FailedImportedVacancy.first.vacancy).to eq(
+        expect(FailedImportedVacancy.first.vacancy).to include(
```
starswan, line 104: this looks like it doesn't belong here. eq is the correct check

app/controllers/api/application_controller.rb
```diff
@@ -9,4 +9,16 @@ def set_headers
...
+    render json: { error: "Internal server error", message: exception.message }, status: :internal_server_error
+  end
+
+  def render_not_found
+    render json: { error: "The given ID does not match any vacancy for your ATS" }, status: :not_found
+  end
+
+  def render_bad_request
+    render json: { error: "Request body could not be read properly" }, status: :bad_request
+  end
```
scruti, line 23: If these responses are only used by (and bind to) the ATS API, could we move the method definitions to the ATS API Vacancies controller?

db/migrate/20241217145653_add_publisher_ats_api_client_reference_to_vacancy_table.rb
starswan: maybe it needs a comment?

spec/requests/publishers/ats_api/v1/vacancies_spec.rb
starswan: job_advert example is 'Teacher of Geography' - is this mean to be a URL, or something else?

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
starswan: There are lots of ways that an API client could not conform to the swagger spec. We did it (above) to make our lives easier and ensure that all incoming requests were validated in line with our schema. Then the client gets a good experience, and we don't have to worry about validating the request

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
starswan: I was meaning a 'function' rather than a class. I often see these 'function' objects become hard to change as all methods have access to the 'state' (which isn't real). IMHO this service object could do a little less (e.g. not return the specific controller status)

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
starswan: https://stackoverflow.com/questions/9742380/link-headers-vs-link-elements-for-restful-json suggests that 'link' headers are not widely adopted - I hadn't heard of them until now. I'm just not sure how we would document the existence of such a header (compared to putting it in the JSON body) so that the client knows to look for it

config/analytics_blocklist.yml
```diff
@@ -86,18 +95,27 @@
   - in_progress_steps
   - employment_history_section_completed
   - qualifications_section_completed
+  - religious_reference_type
```
starswan, line 98: this is still in a tangle

spec/requests/publishers/ats_api/v1/update_vacancy_service_spec.rb
starswan: I think it does a little too much - returning controller statuses like 422 look like they belong in the controller

spec/swagger_helper.rb
starswan: is it possible that we've just mapped it wrongly in our DB? @scruti I'm not sure, but the data I've seen suggests that these are numbers

spec/swagger_helper.rb
starswan: oneOf - an object with an 'id' property or an object with an 'errors' property

config/analytics_blocklist.yml
```diff
@@ -86,18 +95,26 @@
   - in_progress_steps
   - employment_history_section_completed
   - qualifications_section_completed
+  - faith
```
starswan, line 98: This is still an issue?

config/analytics_blocklist.yml
```diff
@@ -106,19 +123,14 @@
...
-    - last_rotated_at
-    - created_at
-    - updated_at
   :vacancies:
   - readable_job_location
   - readable_phases
   - searchable_content
   - google_index_removed
   - expired_vacancy_feedback_email_sent_at
+  - religion_type
```
starswan, line 132: here too

app/controllers/publishers/ats_api/v1/update_vacancy_service.rb
```diff
@@ -0,0 +1,42 @@
...
+
+          private
+
+          attr_reader :vacancy, :params
+
+          def permitted_params(params)
+            organisations = fetch_organisations(params[:schools])
+            raise ActiveRecord::RecordNotFound, "No valid organisations found" if organisations.blank?
+
+            params.except(:schools, :trust_uid).merge(organisations: organisations)
```
starswan, line 22: is trust_uid passed into this method? If so then this ignores it...

app/controllers/publishers/ats_api/v1/update_vacancy_service.rb
```diff
@@ -0,0 +1,42 @@
...
+            else
+              { success: false, errors: format_errors(vacancy.errors.messages) }
+            end
+          end
+
+          private
+
+          attr_reader :vacancy, :params
+
+          def permitted_params(params)
```
starswan, line 18: I found this method name confusing given that the controller does the 'permitted params' filtering

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,77 @@
...
+            raise InvalidOrganisationError, "No valid organisations found" if organisations.blank?
+
+            params[:publish_on] ||= Time.zone.today.to_s
+            params[:working_patterns] ||= []
+            params[:phases] ||= []
+
+            params.except(:schools, :trust_uid).merge(organisations: organisations)
+          end
+
+          def fetch_organisations(school_params)
```
starswan, line 35: could we remove the duplication of this method between the create and the update?

app/controllers/publishers/ats_api/v1/update_vacancy_service.rb
```diff
@@ -0,0 +1,42 @@
...
+            return [] unless school_params
+
+            if school_params[:trust_uid].present?
+              SchoolGroup.trusts.find_by(uid: school_params[:trust_uid]).schools.where(urn: school_params[:school_urns])
+            else
+              School.where(urn: school_params[:school_urns])
+            end
+          end
+
+          def format_errors(errors)
```
starswan, line 35: error formatting appears to be duplicated.

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,77 @@
...
+            if vacancy.save
+              success_response(vacancy)
+            else
+              validation_error_response(vacancy)
+            end
+          end
+
+          private
+
+          def permitted_params(params)
```
starswan, line 24: I find the (almost) duplication of this between create and update a little confusing. This one throws InvalidOrganisationError, the other throws RecordNotFound.

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,77 @@
...
+            organisations = fetch_organisations(params[:schools])
+            raise InvalidOrganisationError, "No valid organisations found" if organisations.blank?
+
+            params[:publish_on] ||= Time.zone.today.to_s
+            params[:working_patterns] ||= []
+            params[:phases] ||= []
+
+            params.except(:schools, :trust_uid).merge(organisations: organisations)
```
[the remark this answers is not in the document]
starswan: but it's not used which looks strange?

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
starswan: IMHO this is suggesting that we have 1 service with 2 entry points (create and update). My challenge with it is that is had already diverged, potentially creating strange 'bugs' e.g.404 on update with invalid org, 400 on create with invalid org.

app/controllers/publishers/ats_api/v1/update_vacancy_service.rb
starswan: see comment above. Its just a function - so maybe these things belong together?

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,77 @@
...
+          end
+
+          private
+
+          def sanitised_params(params)
+            organisations = fetch_organisations(params[:schools])
+            raise InvalidOrganisationError, "No valid organisations found" if organisations.blank?
+
+            params[:publish_on] ||= Time.zone.today.to_s
+            params[:working_patterns] ||= []
```
starswan, line 29: blank working patterns is an invalid vacancy - so I'm not sure what the intent of this line is?

spec/requests/publishers/ats_api/v1/create_vacancy_service_spec.rb
scruti: Yeah happy to change this! 
I don't recall why I chose this but I think I was fighting swagger specs somehow and ended with this format.

spec/swagger_helper.rb
scruti: Looks like the URN and UID we import from the GIAS CSV come as numbers, yes. Was directly introduced like strings 4-5 years ago with the creation of the organisation table.

Not sure if there is any benefit in migrating them to numbers since they're "ids" not used for operations/calculations.

app/controllers/publishers/ats_api/v1/create_vacancy_service.rb
```diff
@@ -0,0 +1,66 @@
...
+          def sanitised_params(params)
+            organisations = fetch_organisations(params[:schools])
+            raise InvalidOrganisationError, "No valid organisations found" if organisations.blank?
+
+            params[:publish_on] ||= Time.zone.today.to_s
+            params.except(:schools).merge(organisations: organisations)
+          end
+
+          def conflict_vacancy(vacancy)
+            Vacancy.find_by(external_reference: vacancy.external_reference)
```
scruti, line 35: There is a chance (little, but possible) that multiple clients use the same `external_reference` within their internal systems. 
EG: 1, 2, 3...
We should also use the `publisher_ats_api_client` combined with the `external_reference` to ensure that we find a conflicting vacancy for the current client.

config/analytics_blocklist.yml
```diff
@@ -86,18 +95,25 @@
...
   - employment_history_section_completed
   - qualifications_section_completed
+  - place_of_worship
+  - religious_referee_name
+  - religious_referee_address
+  - religious_referee_role
+  - religious_referee_email
+  - religious_referee_phone
+  - baptism_address
+  - baptism_date
```
scruti, line 105: Some issue with the rebasing/branching? This should already be present on the base branch instead of being introduced by this branch.

spec/requests/publishers/ats_api/v1/update_vacancy_service_spec.rb
scruti: The main issue I see here is that the Service Object lives in the controllers folder and is tested as a request. 

Is it a controller or a service object? If it cannot be placed within the service objects folder and unit tested wouldn't that hint that we're dealing with the wrong abstraction?

app/controllers/publishers/ats_api/v1/organisation_fetcher.rb
```diff
@@ -0,0 +1,17 @@
...
+    module V1
+      module OrganisationFetcher
+        def fetch_organisations(school_params)
+          return [] unless school_params
+
+          if school_params[:trust_uid].present?
+            SchoolGroup.trusts.find_by(uid: school_params[:trust_uid]).schools&.where(urn: school_params[:school_urns]) || []
+          else
+            School.where(urn: school_params[:school_urns])
+          end
```
scruti, line 12: I think this misses one of the current behaviours for the existing integrations:
Allowing to publish vacancies for the trust itself when no school urns are provided but the trust id is given.

Commented the missing behaviours from the org matching in the current codebase (comments inlined):
```
  def find_schools(item)
    multi_academy_trust = SchoolGroup.trusts.find_by(uid: item["trustUID"]) if item["trustUID"].present?

    school_urns = item["schoolUrns"]&.split(",")
    schools = Organisation.where(urn: school_urns) if school_urns.present?

    return [] if multi_academy_trust.blank? && schools.blank?
    return schools if multi_academy_trust.blank?
    return Array(multi_academy_trust) if schools.blank? # THIS BIT SEEMS MISSING 

    # When having both trust and schools, only return the schools that are in the trust if any. Otherwise, return the trust itself.
    multi_academy_trust.schools.where(urn: school_urns).order(:created_at).presence || Array(multi_academy_trust) # THIS DEFAULTING TO THE TRUST IS ALSO MISSING. DESIRED THOUGH?
  end
```

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,79 @@
...
   def index
-    @pagy, @vacancies = pagy(vacancies, items: 100)
+    @pagy, @vacancies = pagy(vacancies.where(publisher_ats_api_client: client), items: 100)
 
     respond_to(&:json)
   end
 
   def show
-    @vacancy = vacancy
+    @vacancy = Vacancy.find(params[:id])
```
scruti, line 17: This seems to allow any API client to fetch any vacancy in the system (theirs or not). 
Should be limited to their own vacancies:
```suggestion
    @vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
```

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,79 @@
...
+
+    render result.slice(:json, :status)
   end
-  # rubocop:enable Rails/SaveBang
 
   def update
-    @vacancy = vacancy
-
-    respond_to(&:json)
+    vacancy = Vacancy.find(params[:id])
```
scruti, line 29: This seems to allow any API client to update any vacancy in the system (theirs or not). 
Should be limited to their API published  vacancies:
```suggestion
    vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
```

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,79 @@
...
+      @vacancy = vacancy
+      render :show
+    else
+      render json: { errors: result[:errors] }, status: :unprocessable_entity
+    end
   end
 
-  def destroy; end
+  def destroy
+    vacancy = Vacancy.find(params[:id])
```
scruti, line 41: This seems to allow any API client to destroy any vacancy in the system (theirs or not). 
Should be limited to their API published  vacancies:
```suggestion
    vacancy = Vacancy.find_by!(publisher_ats_api_client: client, id: params[:id])
```

app/controllers/publishers/ats_api/v1/vacancies_controller.rb
```diff
@@ -1,58 +1,79 @@
...
+      visa_sponsorship_available
+      external_reference
+      is_job_share
+      job_roles
+      working_patterns
+      contract_type
+      phases
+      schools
+    ]
   end
```
scruti, line 65: This seems the ideal candidate to be turned into a `REQUIRED_VACANCY_KEYS` constant.

config/analytics_blocklist.yml
```diff
@@ -93,11 +102,10 @@
   :emergency_login_keys:
   - id
   - not_valid_after
-  - publisher_id
```
scruti, line 96: related to this PR? Unrelated? Heads up in case we're bringing an unexpected change to `main` branch.

app/services/publishers/ats_api/update_vacancy_service.rb
```diff
@@ -0,0 +1,32 @@
...
+          end
+        end
+
+        private
+
+        attr_reader :vacancy, :params
+
+        def sanitised_params(params)
+          organisations = fetch_organisations(params[:schools])
+          raise ActiveRecord::RecordNotFound, "No valid organisations found" if organisations.blank?
```
scruti, line 21: :raising_hand_man: The response behaviour between the create & update when the wrong/no org is provided is inconsistent.

The `create` service raises an `InvalidOrganisationError`, that turns into a `:render_bad_request` response. Should we align both? I would rather have the `create` approach than a `404` when passing a wrong/missing value in a `PUT/PATCH` action. Would leave the `404` exclusively for the wrong vacancy id.

In an ideal scenario, I think this should be added in both `create/update` as a `:unprocessable_entity, errors: "trust_uid/school_urns: No valid organisations found"`, to match the validation error approach.

app/services/publishers/ats_api/organisation_fetcher.rb
```diff
@@ -0,0 +1,25 @@
...
+
+        trust_uid = school_params[:trust_uid]
+        school_urns = school_params[:school_urns]
+
+        # When having both trust and schools, only return the schools that are in the trust if any.
+        # Otherwise, return the trust itself.
+        multi_academy_trust = SchoolGroup.trusts.find_by(uid: trust_uid)
+        schools = ::Organisation.where(urn: school_urns) if school_urns.present?
+
+        raise InvalidOrganisationError, "No valid organisations found" if multi_academy_trust.blank? && schools.blank?
```
scruti, line 17: oh nice refactor :clap:

scruti (approved):
LGTM!

PR #7378, opened 2025-01-09
Add legacy support role filter name

starswan:
## Changes in this PR:

Fix crash on production due to legacy active filter name


app/models/subscription.rb
```diff
@@ -11,6 +11,8 @@ class Subscription < ApplicationRecord
   FILTERS = {
     teaching_job_roles: ->(vacancy, value) { vacancy.job_roles.intersect?(value) },
     support_job_roles: ->(vacancy, value) { vacancy.job_roles.intersect?(value) },
+    # support_job_roles used to be called this, and there are still actuve subscriptions with this name
```
scruti, line 14:
```suggestion
    # support_job_roles used to be called this, and there are still active subscriptions with this name
```

scruti (approved):

PR #7402, opened 2025-01-13
convert todays subscription alert run list to a set for performance

starswan:
## Changes in this PR:

convert todays subscription alert run list to a set for performance

The alert runs for today list could be large (>10_000 records) so convert to a set to improve lookup performance


scruti (approved):
Nice! I've learnt something new today :)

PR #7405, opened 2025-01-13
Delete alert runs longer than a week old

scruti:
## Trello card URL

- https://trello.com/c/FrndbHGR/1482-add-new-subject-utms-to-landing-pages-for-campaign-emails

## Changes in this PR:

- Refactor the Cron jobs schedule file: 
  Making it easier to see the current job schedules and to add new jobs in the right place.
- Add & Schedule to delete any AlertRun older than 7 days:
  We only need the alert runs to check if the alert was already sent today. Keeping 16M records has a performance/space hit and brings no benefit to the system. 7 days is a safe cutout to be able to check recent runs without storing unnecessary historical data and improving querying performance.

### DIFF
--- a/app/jobs/delete_old_alert_runs_job.rb
+++ b/app/jobs/delete_old_alert_runs_job.rb
@@ -1,0 +1,7 @@
+class DeleteOldAlertRunsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    AlertRun.where(run_on: ...1.week.ago).destroy_all
+  end
+end
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -1,141 +1,153 @@
+#
+# MONTHLY CRON JOBS
+#
 clear_emergency_login_keys:
-  cron: '30 0 1 * *'
+  cron: '30 0 1 * *' # "At 00:30 on day-of-month 1."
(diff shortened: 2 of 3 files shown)


app/jobs/delete_old_alert_runs_job.rb
```diff
@@ -0,0 +1,7 @@
+class DeleteOldAlertRunsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    AlertRun.where(run_on: ...1.week.ago).destroy_all
```
starswan, line 5: Having looked at historical alert run data, is it worth keeping a month of this?

starswan (approved):
Have left 1 comment, feel free to ignore if you don't agree. Personally I would like to get rid of this table completely if possible.

scruti:
Completely agree intending to get rid of this table if possible. For the time being, lets reduce it to a minimum to reduce any DB performance impact. And lets see if we can refactor the subscription alert runs to not depend on this.

app/jobs/delete_old_alert_runs_job.rb
scruti: I would like to push for making the table as small as we can (and a week seems plenty given the code only checks for "today") so we can ensure a significant impact with the change.

PR #7418, opened 2025-01-15
Precompute Centroid for polygon areas

scruti:
## Trello ticket
https://trello.com/c/cf3f17rU/1528-use-precomputed-centroid-of-locationpolygons-areas

## Description

The centroid is used to calculate the distances on the vacancy searches.

The centroid for a polygon area doesn't change unless the area changes too.

As we dynamically calculate it for every location polygon area against every vacancy during the searches, if we pre-compute it we save the DB to execute again the centroid calculation on every search.

**This is the first step in pre-computing and storing it in DB.**
Further work will follow:
- Use the pre-computed values on the location searches.
- Fix possible issues when searching over polygons without area/centroid.
- Legacy Polygons maintenance.

### DIFF
--- a/app/services/ons_data_import/base.rb
+++ b/app/services/ons_data_import/base.rb
@@ -18,17 +18,26 @@ class OnsDataImport::Base
         geometry = feature["geometry"].to_json
 
         Rails.logger.info("Persisting new area data for '#{name}' (#{type})")
-        ActiveRecord::Base.connection.exec_update("
-          UPDATE location_polygons
-          SET area=ST_GeomFromGeoJSON(#{ActiveRecord::Base.connection.quote(geometry)}),
-              location_type=#{ActiveRecord::Base.connection.quote(type)}
-          WHERE id='#{location_polygon.id}'
-        ")
+        set_area_data(location_polygon, geometry, type)
       end
     end
   end
 
   private
+
+  def set_area_data(location_polygon, geometry, type)
+    ActiveRecord::Base.connection.exec_update("
+      WITH geom AS (
+        SELECT ST_GeomFromGeoJSON(#{ActiveRecord::Base.connection.quote(geometry)}) AS geo
+      )
+      UPDATE location_polygons
+      SET area=geom.geo,
+          location_type=#{ActiveRecord::Base.connection.quote(type)},
+          centroid=ST_Centroid(geom.geo)
+      FROM geom
+      WHERE id='#{location_polygon.id}'
+    ")
+  end
 
   def arcgis_features(offset)
     params = [
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -116,29 +116,6 @@
       "note": ""
     },
     {
-      "warning_type": "SQL Injection",
-      "warning_code": 0,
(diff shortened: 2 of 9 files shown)


config/brakeman.ignore
```diff
@@ -1,39 +1,5 @@
 {
   "ignored_warnings": [
-    {
```
scruti, line 3: During the update of Brakeman ignore file, I went through deprecated alerts that weren't relevant anymore and removed them from the ignore file. Hence these blocks of removed code.

app/services/ons_data_import/base.rb
```diff
@@ -18,18 +18,27 @@ def call
         geometry = feature["geometry"].to_json
```
starswan, line 18: do we have to run this by hand? I couldn't find anywhere where we're running this script. I'm also worried about the 99 polygons in current set vs 247 ish in production - it looks like old ones are never removed - although of course some are needed as they are linked to by landing pages

starswan: The API name is out of date (for Cities) - it should be Major_Towns_and_Cities_Dec_2015_Boundaries_V2_2022 IMHO

scruti: @starswan good points.

On the when is this running? It is [scheduled to run every Sunday](https://github.com/DFE-Digital/teaching-vacancies/blob/main/config/schedule.yml#L19) with the `ImportPolygonDataJob`.

Also noticed the current set is way smaller and reached the same conclusion: Old ones are never removed (as the whole script only "finds or creates".
I am thinking of adding a rake task to compute the centroid for those with an area and missing centroid, what should cover the legacy existing polygons.

Seems the `4326`  gets added automatically by Rails when adding the migration into the schema. Will try to see if I find any option where this can be tweaked: `add_column :location_polygons, :centroid, :st_point, geographic: true`


I'm going to extend a bit the scope of this ticket and clarify:
- What do we want to do with legacy location polygons? (This may need product team input)
- In case of keeping legacy polygons, need to add a task to compute their area centroid.
- Over the new 98 polygons or so on the API calls, there are over 20 without area data (what do we do with those? do we want them at all?).
- Can we tweak the `srid` in the DB fields migrations? 
- Update the out of date polygon API data sources.

starswan: weird the composite polygons have neither area nor centroid, so I'm not sure how that is going to work...

starswan: Description: This file contains the digital vector boundaries for the Major Towns and Cities in England and Wales, as at December 2015. Version 2 includes centroid data in the attributes table. [yummy!]

scruti: @starswan I've been studying migrating all the endpoints to the ones that provide `LAT/LONG` data for their centroid.

While this is accurate (compared them with our computed ones) for the Major Towns and Cities endpoint, for counties and regions endpoints the LAT/LONG is rounded up to the closes non-fractional number, turning the provided centroid as heavily inaccurate (multiple counties get the same centroid info, for example).

So I'm going to discard the attempt to pull this info from the APIs and keep computing it each time we update the location polygons, as it happens in the background weekly and turned out to be more precise than the available data on the ONS.

starswan (approved):

PR #7422, opened 2025-01-16
Set option for faster geography evaluation

scruti:
## Changes in this PR:

By setting the `use_spheroid = false` flag in the `ST_DWithin` and `ST_Distance` PostGis over geography data, the operation benefits from a faster spherical calculation rather than the default computing on the spheroid determined by the SRID.

The downside is a slightly less accurate computation. But seems a good tradeoff since we're experiencing performance issues on the distance calculations and our distances are not that big to make a huge difference on the search results.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,36 +1,48 @@
 {
   "ignored_warnings": [
     {
-      "warning_type": "Cross-Site Scripting",
-      "warning_code": 2,
-      "fingerprint": "24062c57b5ebf41e370cdc934e79a6465da62a505547bc540b364ce04ab2fd4c",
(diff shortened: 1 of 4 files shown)


config/brakeman.ignore
```diff
@@ -1,36 +1,48 @@
 {
   "ignored_warnings": [
     {
-      "warning_type": "Cross-Site Scripting",
```
scruti, line 4: As I had to add the modified SQL queries to the Brakeman ignore file, I also did an automated cleanup of legacy ignored warnings that are no longer needed. Hence all the code removal.

This is automated by `brakeman -I` tool.

starswan (approved):

PR #7429, opened 2025-01-20
Update API docs

ddippolito:
## Trello card URL
https://trello.com/c/c89aJ4g8/1447-documentation-of-api-for-users

## Changes in this PR:
Add API documentation


spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
...
         title: "Teaching Vacancies ATS API",
         version: "v1",
+        description: <<~DESCRIPTION,
+          This document outlines the API’s key features, the data required for each listing,
+          and the steps needed to set up an integration.
+
+          This documentation describes each endpoint’s request parameters, response formats, and possible errors,
+          so you can seamlessly integrate Teaching Vacancies into your ATS workflow.
+
+          ## Introduction
```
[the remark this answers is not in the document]
scruti: I've tested this locally and it works. Resized it to 25% of the original image while maintaining the proportions.
```suggestion
          ## Introduction
          
          <img src="/teaching_vacancies_api_diagram.jpg" alt="diagram" width="960" height="540">
 
```

spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
...
+          The **Teaching Vacancies ATS API** enables you to manage job listings on behalf of schools or trusts
+          through your own Applicant Tracking System (ATS) or HR software.
+
+          By calling this API, you can:
+
+          - **List** all your active vacancies (with pagination)
+          - **Retrieve** details for a single vacancy
+          - **Create** new vacancies
+          - **Update** existing vacancies
+          - **Delete** vacancies when they’re no longer needed
```
scruti, line 39: Suggestion up for discussion: I wonder if the `"they're no longer needed"` wording will make clients believe that they need to remove each vacancy at the end of their life.

Maybe something more explicit? EG:
```suggestion
          - **Delete** vacancies if they need to be removed before their expiration date
```

spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
...
+
+          ## Authentication
+
+          Each request requires a valid API key in the `X-Api-Key` header to ensure only authorised
+          clients can manage vacancies.
+
+          Include this key in the X-Api-Key header of each request.
```
scruti, line 46:
```suggestion
          Include this key in the `X-Api-Key` header of each request.
```

spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
...
+          If the key is missing or invalid, the API will respond with an HTTP 401 (Unauthorized) status.
```
scruti, line 47:
```suggestion
          If the key is missing or invalid, the API will respond with an `HTTP 401` (Unauthorized) status.
```

spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
       info: {
         title: "Teaching Vacancies ATS API",
         version: "v1",
+        description: <<~DESCRIPTION,
```
scruti, line 21: Nice! :clap:

spec/swagger_helper.rb
```diff
@@ -18,6 +18,44 @@
...
+          ## Authentication
+
+          Each request requires a valid API key in the `X-Api-Key` header to ensure only authorised
+          clients can manage vacancies.
+
+          Include this key in the X-Api-Key header of each request.
+          If the key is missing or invalid, the API will respond with an HTTP 401 (Unauthorized) status.
+
+          This ensures that only approved clients can create, update, or remove job listings.
+          If you ever need a new or replacement key, let us know, and we’ll assist you with the process.
```
scruti, line 50: Should we add an email address linking to `let us know` ? There was a shared inbox email (that I don't have access to ATM) that we could use here. Will check tomorrow with the team.

scruti:
Nice. I left some suggestions for the current changes. Besides that there are 2 bits I would like to see added to the docs:

## 1. Put the docs behind `user/password` access.
So it doesn't get indexed by Google once it goes into production. 
This can be configured on `config/initializers/rswag_ui.rb`  (code to do so is commented). We need to set some user/password env variables for this config.

## 2. Changes in the `vacancy` API schema.

### 2.1 Extended descriptions/examples.
The `vacancy` schema needs to be extended so that every attribute in the schema has an example value & description.

Ideally, we would bring the "Purpose & Restrictions" info from the current V2 specification document as a "description" on the API schema fields.
The idea is to be sure that clients will understand what is the purpose of each field in plain English, and any restriction not reflected in the schema format.

This description will also be useful to indicate when we need "Choose one" or "One or more values from this list" besides the pure schematic formatting.

EG (the text is some random stuff for example):
`expires_at: { type: :string, format: :date, example: "2025-03-13", description: "The end date of the vacancy. Must be after the start date." }`
image

### 2.2 Schema inaccuracies.
Going through the fields, I noticed the schema needs to be updated to reflect our real restrictions.

Out of the scope of this PR, but we cannot start testing until we fix these. So happy to either fix them here or create a further ticket to address these:

So far noticed this one, but worth a double-check by multiple people in case I missed others:
- `expires_at` needs to be moved to `date-time` format. 
  An oversight when we defined the schema for the first time, but allowing them to set an expiration time for the vacancies is a functionality we currently provide and probably want to keep.

scruti (approved):
Great improvements on the docs. LGTM!

PR #7433, opened 2025-01-21
Add professional body memberships to jobseeker profile

KyleMacPherson:
## Trello card URL

https://trello.com/c/Lb2qV41G/1497-add-new-professional-bodies-membership-section-profiles

## Changes in this PR:

This PR adds a new section called professional body membership to the jobseeker profile page which allows jobseekers to outline their professional body memberships.

## Next steps:


app/controllers/jobseekers/profiles/professional_body_memberships_controller.rb
```diff
@@ -0,0 +1,55 @@
+class Jobseekers::Profiles::ProfessionalBodyMembershipsController < Jobseekers::ProfilesController
+  helper_method :jobseeker_profile, :form, :professional_body_membership
```
starswan, line 2: I'm nit a fan of this helper_method pattern. Can we not just instantiate the form in a controller variable? Stops it being used everywhere in the view w/o the @ to show that we're using a  variable. WDYT?

app/controllers/jobseekers/profiles/professional_body_memberships_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  def destroy
+    professional_body_membership.destroy!
+    redirect_to review_jobseekers_profile_professional_body_memberships_path, success: t(".success")
+  end
+
+  def form
+    @form ||= Jobseekers::ProfessionalBodyMembershipForm.new(form_attributes)
+  end
+
+  def form_attributes
```
starswan, line 36: I have been trying to take this pattern out of the codebase. IMHO it's really hard to read, and collapses into simple code when taken out - @form = Form.new(params) is much easier to read?

config/locales/forms.yml
```diff
@@ -610,17 +610,22 @@ en:
...
       jobseekers_training_and_cpd_form:
         name: Name of course or training
         provider: Training provider
         grade: Grade (optional)
         year_awarded: Date completed
         year_completed_hint: For example, 1998
         course_length: Course length
-
-
+      jobseekers_professional_body_membership_form:
```
starswan, line 620: I have been breaking these changes out into individual yml files - personally these files are scary big and really hard to work with - which section is which? (hint, legend and label). I think we can slowly tame them, but I'm finding it much nicer (and much quicker) than hacking 3  1k line translation files

db/migrate/20250117112840_create_professional_body_membership.rb
```diff
@@ -0,0 +1,15 @@
+class CreateProfessionalBodyMembership < ActiveRecord::Migration[7.2]
+  # rubocop:disable Rails/ThreeStateBooleanColumn
```
starswan, line 2: I think in this case the 3 state boolean is wrong - exam_taken is a true/false radio, so it is always known and cannot be null

spec/factories/professional_body_memberships.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :professional_body_membership do
+    name { "Teachers Union" }
+    membership_type { "Platinum" }
+    membership_number { "100" }
+    date_membership_obtained { "2020" }
+    exam_taken { "Yes" }
```
starswan, line 7: this should be 'true'?

db/migrate/20250117112840_create_professional_body_membership.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :professional_body_memberships, id: :uuid do |t|
+      t.timestamps
+      t.string :name
+      t.string :membership_type
+      t.string :membership_number
+      t.string :date_membership_obtained
```
starswan, line 9: This appears to be a number on the form (the year) so maybe this should be  a number and stored as 'year membership obtained'?

app/models/professional_body_membership.rb
```diff
@@ -0,0 +1,3 @@
+class ProfessionalBodyMembership < ApplicationRecord
+  belongs_to :jobseeker_profile, optional: true
```
starswan, line 2: is this correct? I would expect it to always be attached to a profile? Isn't that the idea?

app/form_models/jobseekers/professional_body_membership_form.rb
```diff
@@ -0,0 +1,8 @@
+class Jobseekers::ProfessionalBodyMembershipForm
+  include ActiveModel::Model
+
+  attr_accessor :name, :membership_type, :membership_number, :date_membership_obtained, :exam_taken
```
starswan, line 4: do we need this abstraction, or could we attach the validations directly to the model?

starswan (approved):

PR #7437, opened 2025-01-22
Downcase non language subjects in campaign pages

ddippolito:
## Changes in this PR:

Downcase non language subjects in campaign pages

### DIFF
--- a/app/services/campaign_page.rb
+++ b/app/services/campaign_page.rb
@@ -1,6 +1,7 @@
 class CampaignPage
   attr_reader :utm_content_code, :criteria, :banner_image, :hidden_filters
 
+  # Language subjects need to be uppercase.
   LANGUAGE_SUBJECTS = %w[English Spanish German French].freeze
 
   def self.exists?(utm_content)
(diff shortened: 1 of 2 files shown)


app/services/campaign_page.rb
```diff
@@ -1,6 +1,8 @@
 class CampaignPage
   attr_reader :utm_content_code, :criteria, :banner_image, :hidden_filters
 
+  LANGUAGE_SUBJECTS = %w[English Spanish German French].freeze
```
starswan, line 4: it would be nice to add a comment explaining why we're using this in this way?

spec/services/campaign_page_spec.rb
```diff
@@ -35,7 +35,13 @@
 
   describe "banner_title" do
     it "returns the banner title" do
-      expect(campaign_page.banner_title("Severus")).to eq("Severus, this is a fake campaign")
+      expect(campaign_page.banner_title("Severus", "Geography")).to eq("Severus, find the right geography fake job for you")
+    end
+
+    context "when the subject is in the language subjects" do
+      it "returns the banner title" do
```
starswan, line 42: ...without lower casing...?

starswan (approved):

app/services/campaign_page.rb
```diff
@@ -20,6 +22,7 @@ def initialize(utm_content_code, criteria)
   end
 
   def banner_title(name, subject = nil)
```
starswan, line 24: I know its not in this PR, but if you don't pass 'subject' here then don't you get an extra space in the text 'find the right  job for you'...?

PR #7449, opened 2025-01-27
Make Google API Client authentication compatible with Analytics Federated Auth

scruti:
## Trello card URL
- https://trello.com/c/Axm4qiA1/1549-fix-issue-in-google-index-drive-apis-authentication-vs-analytics-wif-auth

## Changes in this PR:

For security reasons, we need to enable Analytics Federated Auth. And the changes come from DevOps team across multiple services.
Enabling this option made our existing `GOOGLE_API_JSON_KEY` (belonging to a Service Account) invalid on the GoogleApi Client calls (for web search index updates and google drive calls). It seems the service account permissions get somehow affected by enabling the analytics federated auth flag.
 
To isolate the analytics federated auth from the unrelated Google API Client calls, we are replacing Google Cloud Service Account JSON token & Scopes with a simple API KEY with exclusive access to Google Web Search Index and Google Drive APIs. We have created this new key ad-hoc for this purpose.

The API key works from the production console. But we won't be able to know if it is compatible with the new Analytics Federated auth option until we enable the federated auth (again) in the production config.

### DIFF
--- a/config/initializers/google_api.rb
+++ b/config/initializers/google_api.rb
@@ -1,7 +1,6 @@
 require "google/apis/indexing_v3"
 
 GOOGLE_APIS_KEY = ENV.fetch("GOOGLE_APIS_KEY", "")
-GOOGLE_MAPS_API_KEY = ENV.fetch("GOOGLE_MAPS_API_KEY", "")
 GOOGLE_LOCATION_SEARCH_API_KEY = ENV.fetch("GOOGLE_LOCATION_SEARCH_API_KEY", "")
 
 if GOOGLE_APIS_KEY.empty?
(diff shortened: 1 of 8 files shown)


config/initializers/google_api.rb
```diff
@@ -1,20 +1,12 @@
 require "google/apis/indexing_v3"
 
-GOOGLE_API_JSON_KEY = ENV.fetch("GOOGLE_API_JSON_KEY", "")
-
-GOOGLE_MAPS_API_KEY = ENV.fetch("GOOGLE_MAPS_API_KEY", "")
```
scruti, line 5: This is not used in the project. So cleaning it up.

PR #7458, opened 2025-01-30
Reduce image creation on orgs

starswan:
## Changes in this PR:

Stop creating images for all organisations - just when they are needed. Appears to speed up tests by around 40% - my i5 desktop goes from 10 minutes to 7mins 21 secs with this change


scruti (approved):
CI Unit test runs from 6m to 4m.

You hit the nail on the head with this :clap:

PR #7464, opened 2025-01-31
Track analytics on google geocoding api hits

scruti:
## Trello card URL
https://trello.com/c/D974czyQ/1553-add-analytics-info-on-google-geocoding-api-hits

## Changes in this PR:

Sends a new custom event to DFE Analytics: `google_geocoding_api_hit`

This event and the contextual type, location, and result will help us track what queries hit the Google Geocoding API and how frequently they hit it. 

## Next steps:
Study possible patterns/issues regarding our caching approach.
EG: We're not caching queries that return `nil`. Are there many of those?

### DIFF
--- a/lib/geocoding.rb
+++ b/lib/geocoding.rb
@@ -16,7 +16,7 @@ class Geocoding
 
     Rails.cache.fetch([:geocoding, location], expires_in: CACHE_DURATION, skip_nil: true) do
       Geocoder.coordinates(location, lookup: :google, components: "country:gb").tap do |coords|
-         trigger_google_geocoding_api_hit_event(type: :coordinates, location:, result: coords)
+        trigger_google_geocoding_api_hit_event(type: :coordinates, location:, result: coords)
       end
     rescue Geocoder::OverQueryLimitError
       trigger_google_geocoding_api_hit_event(type: :coordinates, location:, result: "OVER_QUERY_LIMIT")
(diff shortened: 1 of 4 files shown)


lib/geocoding.rb
```diff
@@ -15,8 +15,11 @@ def coordinates
     return self.class.test_coordinates if Rails.application.config.geocoder_lookup == :test
 
     Rails.cache.fetch([:geocoding, location], expires_in: CACHE_DURATION, skip_nil: true) do
-      Geocoder.coordinates(location, lookup: :google, components: "country:gb")
+      coords = Geocoder.coordinates(location, lookup: :google, components: "country:gb")
+      trigger_google_geocoding_api_hit_event(type: :coordinates, location:, result: coords)
+      coords
```
starswan, line 20:
```suggestion
      Geocoder.coordinates(location, lookup: :google, components: "country:gb").tap do |coords|
         trigger_google_geocoding_api_hit_event(type: :coordinates, location:, result: coords)
      end
```

spec/lib/geocoding_spec.rb
```diff
@@ -27,19 +28,38 @@
         expect(Geocoder).not_to receive(:coordinates)
         subject.coordinates
       end
+
+      it "does not trigger a Google Geocoding API hit event", :dfe_analytics do
```
starswan, line 32: I'm wondering if the whole test shouldn't be :dfe_analytics enabled, so that we don't accidentally test for a negative when its switched off or typoed?

starswan (approved):
a couple of minor points

spec/lib/geocoding_spec.rb
scruti: Do you mean enabling analytics over all the tests in the file instead of on an individual basis? `RSpec.describe Geocoding, geocode: true, :dfe_analytics do`

PR #7469, opened 2025-02-03
Add working pattern preference to jobseeker profile

KyleMacPherson:
## Trello card URL

https://trello.com/c/Hth1hb35/1498-add-working-pattern-preference-to-personal-details-application

## Changes in this PR:

Add working pattern details text area to jobseeker profiles

## Next steps:


spec/system/jobseekers/jobseekers_can_add_job_preferences_to_their_profile_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    context "when adding job preferences" do
+      before do
+        visit jobseekers_profile_path
+        create(:job_preferences, completed_steps: [], builder_completed: false, jobseeker_profile: jobseeker.jobseeker_profile)
+      end
+
+      it "allows jobseekers to add job preferences" do
+        click_on "Add job preferences"
+
+        # submit empty form
```
scruti, line 22: Question: Weren't we trying to move away from doing all these error checks as system tests? They add-up quite a bit of time to the test runs and could be checked at lower-level testing.

app/views/jobseekers/profiles/job_preferences/_summary.html.slim
```diff
@@ -47,6 +47,14 @@ ruby:
...
         row.with_action(text: t("buttons.change"), href: jobseekers_job_preferences_step_path(step: step_name), visually_hidden_text: attribute.to_s.humanize)
       end
+
+      if step_name == :working_patterns && multistep.working_pattern_details.present?
+        summary_list.with_row do |row|
+          row.with_key text: "Working pattern details"
+          row.with_value text: multistep.working_pattern_details
+          row.with_action(text: t("buttons.change"), href: jobseekers_job_preferences_step_path(step: step_name), visually_hidden_text: "working pattern details")
+        end
+      end
```
scruti, line 57: :raising_hand_man: I would avoid adding this specific condition block if possible.
It parts away from the generic row loop to repeat the same with "hardcoded" values for a particular case.

Have you explored if we could extend the `attributes` hash to also include the `working_pattern_details`?

It looks like the main issue here is that the `working_pattern_details` doesn't map to its own step, but the `working_patterns` step.  Maybe we could add a `step:` key to each of the attributes elements. Something like:

```ruby
    working_patterns: {
      map_values: ->(value, step) { step.options[value] },
      join_with: ", ",
    },
    working_pattern_details: {
      step_name: `working_patterns`, # So maps to the correct step that defines this attribute.
      if: ->(form) { form.working_pattern_details.present? }
    },
```
Then some change on how the `step_name` is retrieved at the moment `step_name = multistep.class.delegated_attributes[attribute.to_s]` (not completely sure on what `delefate_attributes` magic does here) so it uses the attributes `step_name` if defined and theoretically that would be enough to get `working_pattern_details` row to work as everything else in the summary?

Also, if I read this right,  the current implementation output lists the working pattern details below the locations instead of immediately after the working patterns as per design.

scruti (changes requested):
Looks like it is missing some requirements on the max word limit. Also left a proposal to avoid tailored code departing from the current pattern.

spec/system/jobseekers/jobseekers_can_add_job_preferences_to_their_profile_spec.rb
KyleMacPherson: Ah, I didn't realise we had agreed on that. Have removed them! (We already have form tests for these, I just added this for completeness)

app/views/jobseekers/profiles/job_preferences/_summary.html.slim
KyleMacPherson: `Also, if I read this right, the current implementation output lists the working pattern details below the locations instead of immediately after the working patterns as per design.`

No, the details appeared below working patterns!

The existing logic seems to be written assuming each attribute has it's own step, and each value is an array, so I thought it would be more readable and obvious to just hardcode the row that departed from this pattern. That being said I'm happy to go with your preference and have changed it!

app/views/jobseekers/profiles/job_preferences/working_patterns.html.slim
```diff
@@ -9,5 +9,6 @@
   .govuk-grid-row
     .govuk-grid-column-two-thirds
       = f.govuk_collection_check_boxes :working_patterns, @step.options, :first, :last, legend: { size: "l", text: "Working patterns" }, caption: { text: "Job preferences", size: "l" }
+      = f.govuk_text_area :working_pattern_details, label: { text: t(".working_pattern_details.label"), size: "s" }, hint: { text: t(".working_pattern_details.hint") }
```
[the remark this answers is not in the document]
KyleMacPherson: Oops didn't notice this on the screenshot, thanks @scruti, good spot! 🙇

PR #7473, opened 2025-02-04
Refactor google api client initialization

scruti:
## Trello card URL
- https://trello.com/c/Axm4qiA1/1549-fix-issue-in-google-index-drive-apis-authentication-vs-analytics-wif-auth

## Changes in this PR:
**Short:**
Wraps global Google API requests authorization code into a Singleton Class with the hope of resolving authorization issues that may be caused by setting the authorization config globally.

**Long:**
Instead of having an initialiser setting a global authorization for all the Google API Client requests we extract the Api Client authorization into a singleton class.

That class will get the correct authorization and refresh the token when it expires.

The different Services and Jobs relying on this Google client will call this singleton instance to authorise their requests.

The intention of this is to stop setting global values at Rails initialization level, as it may have unintended consequences between Google Analytics internals mixing with the global authorization set for Google Drive and Search Index APIs.


config/initializers/google_api.rb
```diff
@@ -1,20 +1 @@
...
-end
-
-scope = ["https://www.googleapis.com/auth/indexing",
-         "https://www.googleapis.com/auth/drive"]
-
-key = StringIO.new(GOOGLE_API_JSON_KEY)
-authorizer = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: key,
-                                                                scope: scope)
-authorizer.fetch_access_token!
-Google::Apis::RequestOptions.default.authorization = authorizer
```
scruti, line 20: Main change: We were setting the authorization as "for all the Google APIs, use this global value!"

lib/google_indexing.rb
```diff
@@ -9,9 +9,11 @@ class GoogleIndexing
   attr_reader :service, :url
 
   def initialize(url)
-    return if api_key_empty?
+    @api_client = GoogleApiClient.instance
+    return if @api_client.missing_key?
 
     @service = API::IndexingService.new
+    @service.authorization = api_client.authorization
```
scruti, line 16: Main change: Now we set the authorization at the service scope level.

app/services/publishers/document_virus_check.rb
```diff
@@ -7,28 +8,37 @@ class Publishers::DocumentVirusCheck
   FILE_VIRUS_STATUS_CODE = 403 # 403 Permission denied when acknowledge_abuse: false
 
   def initialize(file)
+    @api_client = GoogleApiClient.instance
+    return unless api_client.authorization
+
+    drive_service.authorization = api_client.authorization
```
scruti, line 14: Main change: Now we set the authorization at the service scope level. Prior was a global setting.

lib/google_indexing.rb
```diff
@@ -9,9 +10,11 @@ class GoogleIndexing
   attr_reader :service, :url
 
   def initialize(url)
-    return if api_key_empty?
+    @api_client = GoogleApiClient.instance
+    return unless api_client.authorization
 
     @service = API::IndexingService.new
+    @service.authorization = api_client.authorization
     @url = url
```
scruti, line 18: Main change: Now we set the authorization at the service scope level. Prior was a global setting.

PR #7499, opened 2025-02-10
Convert current role to boolean (Part 1)

starswan:
## Changes in this PR:

Convert current_role in employments from string to boolean. Part 1 add new field and enable back-fill

### DIFF
--- a/app/helpers/job_applications_helper.rb
+++ b/app/helpers/job_applications_helper.rb
@@ -206,7 +206,7 @@ module JobApplicationsHelper
             reason_for_leaving: "Moving out of the area",
             subjects: Faker::Educator.subject,
             started_on: Faker::Date.in_date_period(year: 2016),
-            current_role: "no",
+            current_role: "",
             ended_on: Faker::Date.in_date_period(year: 2018),
           ),
           Employment.new(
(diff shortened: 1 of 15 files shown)


app/helpers/job_applications_helper.rb
```diff
@@ -200,7 +200,7 @@ def job_application_attributes # rubocop: disable Metrics/MethodLength, Metrics/
             reason_for_leaving: "Moving out of the area",
             subjects: Faker::Educator.subject,
             started_on: Faker::Date.in_date_period(year: 2016),
-            current_role: "no",
+            current_role: "",
```
KyleMacPherson, line 203: Seems fine, just curious, why are we changing from "no" to a blank string?

app/models/employment.rb
```diff
@@ -22,6 +36,7 @@ def duplicate
   end
 
   def current_role?
-    current_role == "yes"
+    current_role.present? && current_role != "no"
```
KyleMacPherson, line 39: [Minor] Probably something I am missing here but using current_role == "yes" here makes more sense to me as that's the logic we are using to set is_current_role on line 18.

KyleMacPherson (approved):

app/models/employment.rb
starswan: again, reverting

PR #7504, opened 2025-02-11
Export casual vacancies with VacancyType of temporary to DWP find a job

KyleMacPherson:
## Changes in this PR:

Prior to this change, we were not sending a value for the VacancyType column if the vacancy had a contract_type of "casual". This change means that we send populate VacancyType with 3 which maps to "temporary" on DWPs side. I am not sure if we are happy mapping "casual" -> "temporary" and am open to opinions.

See screenshot for DWP VacancyType mappings or see pages 20/21 [here](https://static.findajob.dwp.gov.uk/images/find_a_job_bulk_upload_spec_v3.0.pdf)

Screenshot 2025-02-11 at 15 27 12

## Next steps:

### DIFF
--- a/app/services/vacancies/export/dwp_find_a_job/published_and_updated_vacancies/parsed_vacancy.rb
+++ b/app/services/vacancies/export/dwp_find_a_job/published_and_updated_vacancies/parsed_vacancy.rb
@@ -17,7 +17,7 @@ module Vacancies::Export::DwpFindAJob::PublishedAndUpdatedVacancies
     CONTRACT_TYPE_MAPPING = {
       "permanent" => TYPE_PERMANENT_ID,
       "fixed_term" => TYPE_CONTRACT_ID,
-      "casual" => TYPE_TEMPORARY_ID
+      "casual" => TYPE_TEMPORARY_ID,
     }.freeze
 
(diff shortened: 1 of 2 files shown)


app/services/vacancies/export/dwp_find_a_job/published_and_updated_vacancies/parsed_vacancy.rb
```diff
@@ -67,6 +68,8 @@ def type_id
         TYPE_PERMANENT_ID
       when "fixed_term", "parental_leave_cover"
         TYPE_CONTRACT_ID
+      when "casual"
+        TYPE_TEMPORARY_ID
```
starswan, line 72: I quite like the pattern of looking this sort of thing up in a hash with fetch - thus resulting in a sentry alert if we add a value but forget to update the mapping. WDYT?

scruti: @starswan I was wondering how to keep track of future changes and if we could do some metaprogramming with the tests iterating over all the defined values in the Vacancy enums/constants... 
But I like the simplicity of the fetch approach :clap:

PR #7532, opened 2025-02-19
Move conflict validation to vacancy model

ddippolito:
## Trello card URL
https://trello.com/c/hnsRpqHg/1589-ats-api-prevent-vacancy-updates-resulting-in-duplicated-vacancies

## Changes in this PR:
Move conflict and duplicate validation for vacancies to model layer

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,10 +13,13 @@
 ActiveRecord::Schema[7.2].define(version: 2025_02_21_101041) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gist"
+  enable_extension "citext"
   enable_extension "fuzzystrmatch"
+  enable_extension "pg_trgm"
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
   enable_extension "postgis"
+  enable_extension "uuid-ossp"
 
   create_table "active_storage_attachments", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "name", null: false
@@ -328,8 +331,8 @@ ActiveRecord::Schema[7.2].define(version: 2025_02_21_101041) do
     t.datetime "updated_at", null: false
     t.date "account_closed_on"
     t.text "current_sign_in_ip_ciphertext"
-    t.text "last_sign_in_ip_ciphertext"
     t.string "govuk_one_login_id"
+    t.text "last_sign_in_ip_ciphertext"
     t.string "account_merge_confirmation_code"
     t.datetime "account_merge_confirmation_code_generated_at"
     t.index ["email"], name: "index_jobseekers_on_email", unique: true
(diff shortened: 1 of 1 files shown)


spec/models/vacancy_spec.rb
```diff
@@ -673,4 +673,71 @@
...
+
+      it "is invalid when a vacancy with same job_title, expires_at and organisations exists" do
+        create(:vacancy, :external, duplicate_attributes)
+
+        expect(vacancy).not_to be_valid
+        expect(vacancy.errors[:base])
+          .to include("A vacancy with the same job title, expiry date, and organisation already exists.")
+      end
+
+      it "stores the conflicting vacancy" do
```
KyleMacPherson, line 736: Again, I might be misunderstanding this but I thought we _didn't_ want to store them

app/models/vacancy.rb
```diff
@@ -103,6 +103,9 @@ class Vacancy < ApplicationRecord
 
   validates :slug, presence: true
   validate :enable_job_applications_cannot_be_changed_once_listed
+  validate :no_conflict_vacancy, if: -> { publisher_ats_api_client_id.present? && external_reference.present? }
```
KyleMacPherson, line 106: Disclaimer: I don't have much context on the API stuff.

Could we use the validates_uniqueness_of validation instead of the custom no_conflict_vacancy. So something like:

`validates :publisher_ats_api_client_id, uniqueness: { scope: :external_reference, message: "A vacancy with the provided ATS client ID and external reference already exists." }, if: -> { publisher_ats_api_client_id.present? && external_reference.present? }
`

I think we might also be able to enforce this one at the db level if we want, but maybe there is a reason not to do this?

`add_index :vacancies, [:publisher_ats_api_client_id, :external_reference], unique: true`

scruti: Good idea to add the unique index to also avoid the duplication at DB level.

On the suggestion to do a Rails uniqueness validation instead of a custom one:

If there is a conflict when creating/updating the vacancy from an API request, the API requires to identify if/what other vacancy conflicts with. This way can identify when generate the conflict status code and the lin to the existing vacancy.

So `find_conflicting_vacancy` public interface method (and internally `find_conflict_vacancy` and `find_duplicate_vacancy`) need to be defined anyway for the API Controller to query them.

Given that these methods are defined anyway, and also to help with following the same pattern as the `find_duplicate_vacancy` (that cannot be as easily moved to a rails out of box validation AFAI), the current implementation seems to provide "one way" to identify/validate the conflict rather than duplicating it into the rails uniqueness and then keep the method with the query anyways.

Not sure if there is a better way to approach it, but keeping the current approach for the model validation seems simpler? (on my eyes at least)

KyleMacPherson: Ah I see, makes sense. I thought there was probably there was something I was missing!

KyleMacPherson (approved):

db/schema.rb
```diff
@@ -780,7 +778,7 @@
   add_foreign_key "references", "job_applications"
   add_foreign_key "saved_jobs", "jobseekers"
   add_foreign_key "saved_jobs", "vacancies"
-  add_foreign_key "school_group_memberships", "organisations", column: "school_group_id"
+  add_foreign_key "school_group_memberships", "organisations", column: "school_group_id", validate: false
```
KyleMacPherson, line 781: [Minor] looks like some unrelated changes in this schema file.

PR #7533, opened 2025-02-19
Flexi work question

KyleMacPherson:
## Changes in this PR:

## Next steps:


db/migrate/20250218113307_add_flexi_working_details_provided_to_vacancies.rb
```diff
@@ -0,0 +1,7 @@
+class AddFlexiWorkingDetailsProvidedToVacancies < ActiveRecord::Migration[7.2]
+  # rubocop:disable Rails/ThreeStateBooleanColumn
```
KyleMacPherson, line 2: I have allowed 3 states here as I don't think we can really backfill `flexi_working_details_provided` for existing vacancies. I thought about creating a migration to set it to false if flexi_working was empty but seems like it kind of defeats the point of the ticket as those that have 'n/a' or similar here would have flexi_working_details_provided set to true, which I think is exactly what we are trying to avoid.

app/form_models/publishers/job_listing/about_the_role_form.rb
```diff
@@ -9,6 +9,8 @@ class Publishers::JobListing::AboutTheRoleForm < Publishers::JobListing::Vacancy
   validate :safeguarding_information_does_not_exceed_maximum_words, if: -> { safeguarding_information_provided == "true" }, unless: -> { vacancy.job_advert.present? || vacancy.about_school.present? }
   validates :further_details_provided, inclusion: { in: [true, false, "true", "false"] }, unless: -> { vacancy.job_advert.present? || vacancy.about_school.present? }
   validate :further_details_presence, if: -> { further_details_provided == "true" }, unless: -> { vacancy.job_advert.present? || vacancy.about_school.present? }
+  validates :flexi_working_details_provided, inclusion: { in: [true, false, "true", "false"] }
```
starswan, line 12: it would be more idiomatic to make this field an attribute of type boolean rather than a string. Then the checks aren't required, and the test becomes boolean rather than x == "true" (which as you allow 'true' is technically broken maybe?).

I don't think we need to persist the 'flexi working provided' value - check out TRN where IIRC I made the 'provided' value non-persisted, validates it based on true and loaded it based on data presence

app/views/publishers/vacancies/build/about_the_role.html.slim
```diff
@@ -51,14 +51,17 @@
...
-        rows: 5, required: true, aria: { hidden: false }),
-        value: form.flexi_working, field_name: "publishers_job_listing_about_the_role_form[flexi_working]", hint: t("jobs.flexi_working.hint"),
-        label: { text: t("jobs.flexi_working.publisher"), size: "s" })
+      = f.govuk_radio_buttons_fieldset :flexi_working_details_provided, legend: { size: "m", tag: nil }, class: ["flexi_working_details_provided"] do
+        = f.govuk_radio_button :flexi_working_details_provided, "true" do
+          = editor(form_input: f.govuk_text_area(:flexi_working,
+            rows: 5, required: true, aria: { hidden: false }),
+            value: form.flexi_working, field_name: "publishers_job_listing_about_the_role_form[flexi_working]", hint: t("jobs.flexi_working.hint"),
+            label: { text: t("jobs.flexi_working.publisher"), size: "s" })
+        = f.govuk_radio_button :flexi_working_details_provided, "false", link_errors: true
```
starswan, line 60: its idiomatic to link errors to the first value, not the last

app/form_models/publishers/job_listing/about_the_role_form.rb
```diff
@@ -38,9 +44,17 @@ def params_to_save
...
     }
   end
 
+  # Overriding load_form to include :flexi_working_details_provided since it's not part of `self.fields`. This is necessary because
+  # :flexi_working_details_provided is explicitly defined as a boolean attribute.
+  def self.load_form(model)
+    fields = self.fields << :flexi_working_details_provided
+    model.slice(*fields)
+  end
+
```
starswan, line 57: I have been using class << self syntax for class methods because I feel its more idiomatic than this style

app/form_models/publishers/job_listing/about_the_role_form.rb
```diff
@@ -38,9 +44,17 @@ def params_to_save
       safeguarding_information:,
       further_details_provided:,
       further_details:,
+      flexi_working_details_provided:,
     }
   end
 
+  # Overriding load_form to include :flexi_working_details_provided since it's not part of `self.fields`. This is necessary because
+  # :flexi_working_details_provided is explicitly defined as a boolean attribute.
```
starswan, line 52: is there any danger of an error due to this being missing from the list of fields? Do we need the fields method any more?

db/migrate/20250218113307_add_flexi_working_details_provided_to_vacancies.rb
starswan: maybe we should have a comment saying that this has to exist because of the modelling, as the vacancy model stores part-created vacancies (where this field and its ilk can be unknown values)

starswan (approved):

PR #7553, opened 2025-02-25
Remove back_link from personal details name page

KyleMacPherson:
## Changes in this PR:

We recently had a user report a bug. The user could not turn on their profile because they had not completed their personal details. The name page crashed when they tried to enter their name in the jobseeker profile.

### Why did this happen?

The user's jobseeker profile had some personal details filled in but the name section was blank. In completed_steps it had work and phone_number sections marked as complete but not the name section. The name is the first step in the personal details flow.

The "Step not completed" error was raised by the `previous_step` defined in lib/multistep/form.rb. This error seems to be raised to ensure that multistep forms cannot be filled in out of order. As name was not completed but later steps in the personal details flow were completed it raised this error.

The `previous_step` method was called because of the back link on the name page of the personal details form.

### How did this happen?

I think the user's jobseeker profile got into this state due to the way that we populate a new jobseeker profile for a user. If the user has a job application, when they first visit the jobseeker profile page, we populate their profile with information from their previous application regardless of whether that application has been submitted or not.

In this case the user had both a draft and a submitted job application, I think that the user's jobseeker profile got populated based on the draft application which was missing some information.

### How to fix this?

I have tried to fix this both for other jobseekers who may already have their data in a similar state and to ensure other jobseekers do not get their data into a similar state.

1. Removing the back link from the name page of the personal details flow. This back link is not needed as it's the first page of the flow and the user can use the cancel button, or the browsers back button to return to the profile summary page. Removing this back link stops the error being raised for users who find themselves in a similar position to the user who raised the issue.

2. Only populate new jobseeker profiles with data from job applications if they are not in a draft state. This hopefully should ensure that the data from the job application is completed and valid and will not create a jobseeker profile that is missing important data.

## Next steps:

### DIFF
--- a/app/models/concerns/profile_section.rb
+++ b/app/models/concerns/profile_section.rb
@@ -6,7 +6,7 @@ module ProfileSection
       find_or_initialize_by(**).tap do |record|
         if record.new_record?
           # do not populated from draft applications as they make be incomplete and invalid.
-          if (previously_submitted_application = jobseeker(record).job_applications.where.not(status: :draft).last)
+          if (previously_submitted_application = jobseeker(record).job_applications.after_submission.last)
             copy_attributes(record, previously_submitted_application)
 
             block&.call(record)
--- a/app/models/concerns/profile_section.rb
+++ b/app/models/concerns/profile_section.rb
@@ -5,8 +5,9 @@ module ProfileSection
     def prepare(**, &block)
       find_or_initialize_by(**).tap do |record|
         if record.new_record?
-          if (previous_application = jobseeker(record).job_applications.last)
-            copy_attributes(record, previous_application)
+          # do not populated from draft applications as they make be incomplete and invalid.
+          if (previously_submitted_application = jobseeker(record).job_applications.where.not(status: :draft).last)
+            copy_attributes(record, previously_submitted_application)
 
             block&.call(record)
             before_save_on_prepare(record)
@@ -24,9 +25,9 @@ module ProfileSection
       record.jobseeker_profile.jobseeker
     end
 
-    def copy_attributes(record, previous_application)
+    def copy_attributes(record, previously_submitted_application)
       attributes_to_copy.each do |attribute|
-        record.assign_attributes(attribute => previous_application.public_send(attribute))
+        record.assign_attributes(attribute => previously_submitted_application.public_send(attribute))
       end
     end
 
--- a/app/views/jobseekers/profiles/personal_details/name.html.slim
+++ b/app/views/jobseekers/profiles/personal_details/name.html.slim
@@ -2,7 +2,6 @@
 
 - content_for :breadcrumbs do
   nav.govuk-breadcrumbs aria-label="Breadcrumbs"
-  = govuk_back_link text: t("buttons.back"), href: back_url, html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
--- a/app/views/jobseekers/profiles/personal_details/name.html.slim
+++ b/app/views/jobseekers/profiles/personal_details/name.html.slim
@@ -2,6 +2,7 @@
 
 - content_for :breadcrumbs do
   nav.govuk-breadcrumbs aria-label="Breadcrumbs"
+  = govuk_back_link text: t("buttons.back"), href: escape_path, html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
 
 .govuk-grid-row
   .govuk-grid-column-two-thirds
--- a/spec/models/jobseeker_profile_spec.rb
+++ b/spec/models/jobseeker_profile_spec.rb
@@ -36,8 +36,8 @@ RSpec.describe JobseekerProfile, type: :model do
       end
 
       it "does not use the details from the draft application" do
-        expect(profile.employments.map(&:job_title).sort).to be_empty
-        expect(profile.qualifications.map(&:institution).sort).to be_empty
+        expect(profile.employments.map(&:job_title)).to be_empty
+        expect(profile.qualifications.map(&:institution)).to be_empty
         expect(profile.qualified_teacher_status_year).to be_nil
         expect(profile.qualified_teacher_status).to be_nil
         expect(profile.personal_details.first_name).to be_nil
--- a/spec/models/jobseeker_profile_spec.rb
+++ b/spec/models/jobseeker_profile_spec.rb
@@ -14,38 +14,43 @@ RSpec.describe JobseekerProfile, type: :model do
     end
 
     context "when the jobseeker has a previously submitted application" do
-      let!(:previous_application) { create(:job_application, :status_submitted, jobseeker:) }
+      let(:first_name) { "Bill" }
+      let(:last_name) { "Smith" }
+      let!(:previous_application) { create(:job_application, :status_submitted, jobseeker:, first_name:, last_name:) }
 
       it "uses the details from the previous application" do
         expect(profile.employments.map(&:job_title).sort).to eq(previous_application.employments.map(&:job_title).sort)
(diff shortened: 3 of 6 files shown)


spec/system/jobseekers/jobseekers_can_manage_a_profile_spec.rb
```diff
@@ -127,8 +127,8 @@
       end
     end
 
-    describe "personal details if the jobseeker has a blank previous job application" do
-      let!(:previous_application) { create(:job_application, :status_draft, jobseeker:, first_name: nil, last_name: nil, phone_number: "01234567890") }
+    describe "personal details if the jobseeker has a blank previously submitted job application" do
+      let!(:previous_application) { create(:job_application, :status_submitted, jobseeker:, first_name: nil, last_name: nil, phone_number: "01234567890") }
```
scruti, line 131: A submitted application without data? How is this achievable? It seems a non-reachable context.

app/views/jobseekers/profiles/personal_details/name.html.slim
```diff
@@ -2,7 +2,6 @@
 
 - content_for :breadcrumbs do
   nav.govuk-breadcrumbs aria-label="Breadcrumbs"
-  = govuk_back_link text: t("buttons.back"), href: back_url, html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
```
KyleMacPherson, line 5: I believe that the error was raised by `back_url`, a helper method which is defined [here](https://github.com/DFE-Digital/teaching-vacancies/blob/2450de3af7e6d4767cb1b804a1ef8ad278a5a0a3/lib/multistep/controller.rb#L74). This method in turn calls the `previous_step` method.

spec/models/jobseeker_profile_spec.rb
```diff
@@ -22,6 +22,26 @@
...
+    end
+
+    context "when the jobseeker has a previous draft application" do
+      before do
+        create(:job_application, :status_draft, jobseeker:)
+      end
+
+      it "does not use the details from the draft application" do
+        expect(profile.employments.map(&:job_title).sort).to be_empty
+        expect(profile.qualifications.map(&:institution).sort).to be_empty
```
starswan, line 40: the sorts look superflous if we're checking for empty?

app/models/concerns/profile_section.rb
```diff
@@ -5,8 +5,9 @@ module ProfileSection
     def prepare(**, &block)
       find_or_initialize_by(**).tap do |record|
         if record.new_record?
-          if (previous_application = jobseeker(record).job_applications.last)
-            copy_attributes(record, previous_application)
+          # do not populated from draft applications as they make be incomplete and invalid.
+          if (previously_submitted_application = jobseeker(record).job_applications.where.not(status: :draft).last)
```
starswan, line 9: I was going to say that we should use a scope here and found after_submission - except it has a bug (created by yours truly) which forgets to add 'interviewing' to the list of 'after submission' values. Think we could use this, and change the implementation to 'not draft'?

starswan: rails 7 has a 'invert_where' for just this sort of thing

spec/models/jobseeker_profile_spec.rb
```diff
@@ -22,6 +22,26 @@
         expect(profile.qualified_teacher_status_year).to eq(previous_application.qualified_teacher_status_year)
         expect(profile.qualified_teacher_status).to eq(previous_application.qualified_teacher_status)
         expect(profile.personal_details.first_name).to eq(previous_application.first_name)
+        expect(profile.personal_details.last_name).to eq(previous_application.last_name)
+      end
+
+      it "creates a job preferences record" do
+        expect(profile.job_preferences).to be_present
```
starswan, line 29: sorry I've just noticed this. Is there a stronger assertion we can make?

spec/models/personal_details_spec.rb
```diff
@@ -33,6 +33,23 @@
...
+      before do
+        create(:job_application, :status_draft, jobseeker:)
+      end
+
+      it "uses the details from the previous application" do
+        expect(personal_details.first_name).to be_nil
+        expect(personal_details.last_name).to be_nil
+        expect(personal_details.phone_number).to be_nil
+        expect(personal_details.right_to_work_in_uk).to be_nil
+      end
```
starswan, line 46: not sure that the test proves the assertion in the it?

spec/system/jobseekers/jobseekers_can_manage_a_profile_spec.rb
```diff
@@ -127,19 +127,29 @@
...
+      let!(:previous_application) { create(:job_application, :status_submitted, jobseeker:, first_name: "Alfred", last_name: "Accelsior", phone_number: "01234567890") }
 
       before do
         visit jobseekers_profile_path
       end
 
       it "prefills the form with the jobseeker's provided personal details" do
         expect(page).to have_content(previous_application.phone_number)
+        expect(page).to have_content(previous_application.first_name)
+        expect(page).to have_content(previous_application.last_name)
```
starswan, line 140: I'm not a fan of this pattern - I worry that the values might  be nil or blank (either now or in the future) at which point the assertion isn't useful. I would much prefer to assert actual values (or let variables)

spec/models/jobseeker_profile_spec.rb
```diff
@@ -22,6 +22,26 @@
...
+        create(:job_application, :status_draft, jobseeker:)
+      end
+
+      it "does not use the details from the draft application" do
+        expect(profile.employments.map(&:job_title)).to be_empty
+        expect(profile.qualifications.map(&:institution)).to be_empty
+        expect(profile.qualified_teacher_status_year).to be_nil
+        expect(profile.qualified_teacher_status).to be_nil
+        expect(profile.personal_details.first_name).to be_nil
+        expect(profile.personal_details.last_name).to be_nil
```
starswan, line 44: this is a lot of negative asserts - presumably e.g. the employment list is empty, so the map() call does nothing useful. Is there a way of making this a positive test about what it **does** do (which I don't know BTW, but would like to)

app/models/concerns/profile_section.rb
```diff
@@ -5,8 +5,9 @@ module ProfileSection
     def prepare(**, &block)
       find_or_initialize_by(**).tap do |record|
         if record.new_record?
-          if (previous_application = jobseeker(record).job_applications.last)
-            copy_attributes(record, previous_application)
+          # do not populated from draft applications as they make be incomplete and invalid.
+          if (previously_submitted_application = jobseeker(record).job_applications.after_submission.last)
```
scruti, line 9: `after_submission` didn't know we had this. Nice!

spec/system/jobseekers/jobseekers_can_manage_a_profile_spec.rb
scruti: Didn't think about it, but now that you mention it, it is a good point.

KyleMacPherson: Good point, have fixed this one!

spec/models/jobseeker_profile_spec.rb
KyleMacPherson: I added an assertion with a be_persisted matcher, not really sure what else to assert for an empty model

spec/models/jobseeker_profile_spec.rb
KyleMacPherson: Removed the map, good call! It creates empty models so I am not sure there is much more we can do

app/models/concerns/profile_section.rb
KyleMacPherson: me neither, @starswan pointed it out in an earlier comment!

scruti (approved):
LGTM!

PR #7558, opened 2025-02-27
Fix internal external scopes

starswan:
## Changes in this PR:

whilst looking at the validation bug, found this defect which will kick in when API clients go live

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -94,8 +94,8 @@ class Vacancy < ApplicationRecord
   scope :published_on_count, ->(date) { published.where(publish_on: date.all_day).count }
   scope :visa_sponsorship_available, -> { where(visa_sponsorship_available: true) }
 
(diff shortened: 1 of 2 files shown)


app/models/vacancy.rb
```diff
@@ -94,8 +94,8 @@ class Vacancy < ApplicationRecord
   scope :published_on_count, ->(date) { published.where(publish_on: date.all_day).count }
   scope :visa_sponsorship_available, -> { where(visa_sponsorship_available: true) }
 
-  scope :internal, -> { where(external_source: nil) }
-  scope :external, -> { where.not(external_source: nil) }
+  scope :internal, -> { where(external_source: nil, publisher_ats_api_client_id: nil) }
+  scope :external, -> { internal.invert_where }
```
scruti, line 98: Had no idea about `invert_where`. Nice!

starswan: just learnt it myself. Rails 7 needed

PR #7560, opened 2025-02-27
TRN form tweaks: Remove duplicated code, reorganise validations and reduce duplicated validation messages.

scruti:
## Trello card URL

- https://trello.com/c/2LLFZmsR/1632-remove-repeated-validation-errors-on-professional-status-form-teacher-reference-number-attributes

## Changes in this PR:

There are 3 main changes on the job application professional status form submission:

1. Code to assign the TRN attributes from the jobseeker profile was defined in 2 different places: The form itself and the controller.
    Only the controller code was reachable. 
    This PR removes the but from controller logic and moves it to the form, as it aims to keep as much of the particular form logic within the form itself (especially when the controller is shared across the whole job application flow) and it allows deeper unit testing at form object model level.

2. Refactor validations grouping them by shared conditions, and add extensive validation testing.
    This was mainly done to be able to identify and fix any redundancy (next point), but also made the form validations easier to glance through and more manageable. 

2. Remove duplications on professional status form error messages.
    The combination of attribute validations, presence, format, inclusion... caused situations where the same error message was displayed for 2 different fields, or the same field has the same error message listed twice.
    We have simplified the validations over TRN fields so they don't step on each other.

### Before

Screenshot_20250226_204224

Screenshot_20250226_222455

### After

image

image

#### Only if selecting "Yes" on "Do you have a teacher reference number (TRN)?" but not entering the number
image

### DIFF
--- a/app/form_models/jobseekers/job_application/professional_status_form.rb
+++ b/app/form_models/jobseekers/job_application/professional_status_form.rb
@@ -47,9 +47,21 @@ module Jobseekers
 
       # These validations are only applied when the professional status section is marked as completed.
       #
-      # Equivalent to if: -> { professional_status_section_completed == true }.
       # Nested validations using "if" calls would override the parent "with_options if:" call instead of combining the conditions.
       # By using "unless" we can keep using "if" for nested validations and the "if & unless" conditions will be combined.
+      #
+      # BAD: Output would be "validates ... if: -> { condition B }".
+      # with_options if: -> { condition A }
+      #   validates ... if: -> { condition B }
+      # end
+      #
+      # GOOD: Output would be "validates ... unless: -> { !condition A }, if: -> { condition B }".
+      #       That is the same as "validates ... if: -> { condition A && condition B }".
+      # with_options unless: -> { !condition A }
+      #   validates ... if: -> { condition B }
+      # end
+      #
+      # Equivalent to if: -> { professional_status_section_completed == true }.
       with_options unless: -> { professional_status_section_completed != true } do
         validates :qualified_teacher_status, inclusion: { in: %w[yes no on_track] }
         validates :statutory_induction_complete, inclusion: { in: %w[yes no] }
@@ -67,6 +79,10 @@ module Jobseekers
       # Teacher reference number:
       # Its presence is required when the "has_teacher_reference_number" is "yes".
       # Its format is validated only when the number is provided.
+      # Having a qalified teacher status 'yes' will enforce the "has_teacher_reference_number" to be 'yes' what will force
+      # the "teacher_reference_number" to be present and formatted correctly.
+      # Codewise "has_teacher_reference_number" seems to be a redundant field as it can be inferred from the presence
+      # of the "teacher_reference_number".
       validates :teacher_reference_number, presence: true, if: -> { has_teacher_reference_number == "yes" }
       validates_format_of :teacher_reference_number, with: /\A\d{7}\z/, allow_blank: true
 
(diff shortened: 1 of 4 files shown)


app/form_models/jobseekers/job_application/professional_status_form.rb
```diff
@@ -45,16 +45,30 @@ def initialize(attributes = {})
...
-      validates_format_of :teacher_reference_number, with: /\A\d{7}\z/, allow_blank: true, if: -> { qualified_teacher_status.in?(%w[no on_track]) }
-      validates :has_teacher_reference_number, inclusion: { in: %w[yes] }, if: -> { qualified_teacher_status == "yes" && professional_status_section_completed }
-      validates :has_teacher_reference_number, inclusion: { in: %w[yes no] }, if: -> { qualified_teacher_status.in?(%w[no on_track]) && professional_status_section_completed }
+      # These validations are only applied when the professional status section is marked as completed.
+      #
+      # Equivalent to if: -> { professional_status_section_completed == true }.
+      # Nested validations using "if" calls would override the parent "with_options if:" call instead of combining the conditions.
+      # By using "unless" we can keep using "if" for nested validations and the "if & unless" conditions will be combined.
+      with_options unless: -> { professional_status_section_completed != true } do
```
KyleMacPherson, line 53: So we can't use `if` here and also use `if` for nested validations?

spec/form_models/jobseekers/job_application/professional_status_form_spec.rb
```diff
@@ -1,17 +1,169 @@
...
+            instance_double(JobseekerProfile, teacher_reference_number: nil, has_teacher_reference_number: nil)
+          end
+
+          it "does not set the TRI attributes" do
+            expect(form).to have_attributes(teacher_reference_number: nil,
+                                            has_teacher_reference_number: nil)
+          end
+        end
+
+        context "when explicitly stating no teacher reference number is provided" do
```
KyleMacPherson, line 155: [Minor] This made me think that we were passing a jobseeker_profile but also explicitly setting has_teacher_reference_number: "no" in the attributes. Might just be me but I would understand the test more easily something like "when the jobseeker profile has_teacher_reference_number set to no"

app/form_models/jobseekers/job_application/professional_status_form.rb
```diff
@@ -45,16 +45,30 @@ def initialize(attributes = {})
...
+        end
+
+        with_options if: -> { qualified_teacher_status.in?(%w[no on_track]) } do
+          validates :has_teacher_reference_number, inclusion: { in: %w[yes no] }
+        end
+      end
+
+      # Teacher reference number:
+      # Its presence is required when the "has_teacher_reference_number" is "yes".
+      # Its format is validated only when the number is provided.
```
KyleMacPherson, line 69: [Minor] Might be good to comment something about when qualified_teacher_status == "yes" then these validations will always be run (once the jobseeker has passed the has_teacher_reference_number validation).

KyleMacPherson (approved):

app/form_models/jobseekers/job_application/professional_status_form.rb
scruti: Exactly, that is what I was trying to alert about with the comment:
```
with_options if: -> { condition A }
  validate .... if: -> { condition B }
end
```
The output for this would be `  validate .... if: -> { condition B }` as the nested `if` overrides the parent.
Instead
```
with_options unless: -> { condition A }
  validate .... if: -> { condition B }
end
```
The output for this is `validate .... if: -> { condition B }, unless:  -> { condition A }`

KyleMacPherson: Ah I did see that from the comment but it looked so weird that I wanted to double check I understood, this makes a lot of sense: `The output for this is validate .... if: -> { condition B }, unless:  -> { !condition A }.`

Thanks for the great explanation 🙇

PR #7577, opened 2025-03-04
Rescue InvalidGeometry error in location filtering for job alerts

KyleMacPherson:
## Trello card URL

https://trello.com/c/MAlVPSX3/1638-bug-investigate-fix-error-on-daily-subscription-runs

## Changes in this PR:

We have been experiencing some RGeo::Error::InvalidGeometry errors when processing job alerts. I raised [a PR](https://github.com/DFE-Digital/teaching-vacancies/pull/7573) to add the invalid polygons (which were causing these errors) to the INVALID_POLYGONS to stop these errors occuring.

This PR ensures we handle any RGeo::Error::InvalidGeometry errors which could thrown in the future by invalid polygons that we have not yet added to INVALID_POLYGONS. This change rescues these errors and falls back to using co-ordinates rather than the polygon if the polygon is raising errors.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -119,7 +119,7 @@ class Subscription < ApplicationRecord
           return vacancies.select { |v| v.organisations.map(&:geopoint).any? { |point| polygon.area.contains?(point) } }
         end
       rescue RGeo::Error::InvalidGeometry => e
-        Rails.logger.error("Invalid geometry encountered for location #{polygon.name}: #{e.message}")
+        Sentry.capture_exception(e)
       end
 
       radius_in_metres = convert_miles_to_metres(radius_in_miles)
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -120,7 +120,7 @@ class Subscription < ApplicationRecord
         end
       rescue RGeo::Error::InvalidGeometry => e
         Sentry.with_scope do |scope|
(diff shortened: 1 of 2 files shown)


app/models/subscription.rb
```diff
@@ -111,17 +111,23 @@ def vacancies_matching(scope)
...
-        radius_in_metres = convert_miles_to_metres radius_in_miles
-        coordinates = Geocoding.new(location).coordinates
-        search_point = RGeo::Geographic.spherical_factory(srid: 4326).point(coordinates.second, coordinates.first)
-        vacancies.select { |v| v.organisations.map(&:geopoint).any? { |point| search_point.distance(point) < radius_in_metres } }
+      begin
+        if polygon.present? && !polygon.name.in?(INVALID_POLYGONS) && polygon.area.invalid_reason.nil?
+          return vacancies.select { |v| v.organisations.map(&:geopoint).any? { |point| polygon.area.contains?(point) } }
+        end
+      rescue RGeo::Error::InvalidGeometry => e
+        Sentry.capture_exception(e)
```
scruti, line 122: Could we add context to the error? I think Sentry allows for this (from what I see in other manual events sent to Sentry):

```suggestion
        Sentry.with_scope do |scope|
          scope.set_context("Polygon", { id: polygon.id, name: polygon.name } )
          Sentry.capture_exception(e)
        end
```

scruti (approved):

PR #7598, opened 2025-03-10
Prevent invalid employment records from submission

starswan:
## Trello card URL

https://trello.com/c/avSRIbAT/1655-bug-old-employment-records-can-be-invalid-mainly-due-a-lack-of-leaving-reason-insist-that-they-are-validated-before-application

## Changes in this PR:

Create a workflow so that invalid employment records copied from an old job application or profile have to be fixed up before job application can be submitted

### After
InvalidEmployment

### DIFF
--- a/app/form_models/jobseekers/job_application/employment_history_form.rb
+++ b/app/form_models/jobseekers/job_application/employment_history_form.rb
@@ -12,29 +12,39 @@ module Jobseekers
         end
 
         def unstorable_fields
-          %i[unexplained_employment_gaps_present employment_history_section_completed]
+          %i[unexplained_employment_gaps_present employment_history_section_completed employments]
         end
 
         def load_form(model)
-          super.merge(unexplained_employment_gaps: model.unexplained_employment_gaps).merge(completed_attrs(model, :employment_history))
+          super.merge(employments: model.employments,
+                      unexplained_employment_gaps: model.unexplained_employment_gaps)
+               .merge(completed_attrs(model, :employment_history))
         end
       end
-      attr_accessor(:unexplained_employment_gaps)
+      attr_accessor(:unexplained_employment_gaps, :employments)
 
-      validate :employment_history_gaps_are_explained
+      validate :employment_history_gaps_are_explained, if: -> { unexplained_employment_gaps_present == "true" && employment_history_section_completed }
+
+      validate :employment_records_are_all_valid
+
+      completed_attribute(:employment_history)
+
+      private
 
       attribute :unexplained_employment_gaps_present, :boolean
 
       def employment_history_gaps_are_explained
-        return unless unexplained_employment_gaps_present && employment_history_section_completed
-
         unexplained_employment_gaps.each_value do |details|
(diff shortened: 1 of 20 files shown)


app/models/jobseeker_profile.rb
```diff
@@ -51,7 +51,7 @@ class JobseekerProfile < ApplicationRecord
 
   def self.copy_attributes(record, previous_application)
     record.assign_attributes(
-      employments: previous_application.employments.map(&:duplicate),
+      employments: previous_application.employments.map(&:clone),
```
jriga: out of curiosity, why use `clone` here? Also the qualifications uses `duplicate`

app/form_models/jobseekers/job_application/employment_history_form.rb
```diff
@@ -14,27 +14,37 @@ def storable_fields
...
+          super.merge(employments: model.employments,
+                      unexplained_employment_gaps: model.unexplained_employment_gaps)
+               .merge(completed_attrs(model, :employment_history))
         end
       end
-      attr_accessor(*FIELDS, :unexplained_employment_gaps)
+      attr_accessor(*FIELDS, :unexplained_employment_gaps, :employments)
 
-      validate :employment_history_gaps_are_explained
+      validate :employment_history_gaps_are_explained, if: -> { unexplained_employment_gaps_present == "true" && employment_history_section_completed }
```
jriga, line 28: I thought we had migrated field like this `unexplained_employment_gaps_present` to boolean?

spec/system/jobseekers/jobseekers_can_add_employments_to_their_job_application_spec.rb
```diff
@@ -40,35 +40,35 @@
...
-    click_on I18n.t("buttons.save_employment")
-
-    all(:link, "Add another job").first.click
-
-    fill_in_employment_history(job_title: "Oldest job", start_month: "09", start_year: "2015", end_month: "06", end_year: "2019")
-
-    click_on I18n.t("buttons.save_employment")
+  context "with an employmeent history" do
+    before do
+      build(:employment, reason_for_leaving: nil, job_application: job_application, job_title: "Oldest job",
```
jriga, line 45: this built instance does not seem to be used anywhere

starswan: It's saved with validate: false - is there a better way of creating an invalid item from a factory?

app/models/jobseeker_profile.rb
starswan: that should now be reverted - clone was a mistake. Turns out #dup is the correct way, but dup copies parent associations

app/form_models/jobseekers/job_application/employment_history_form.rb
starswan: I've just managed to remove it - it was making failing scenarios pass as well by being nil in quite a few places

jriga (approved):

PR #7600, opened 2025-03-10
Add start page

KyleMacPherson:
## Changes in this PR:


app/components/dashboard_component/dashboard_component.html.slim
```diff
@@ -10,7 +10,7 @@
       span.govuk-caption-l = @organisation.name
       h1.govuk-heading-l = t("jobs.dashboard.#{@selected_type}.with_count", count: @vacancies.count)
     .govuk-grid-column-one-quarter
-      = govuk_button_to t("buttons.create_job"), organisation_jobs_path, class: "float-right"
+      = govuk_button_to t("buttons.create_job"), organisation_jobs_start_path, method: :get, class: "float-right"
```
jriga, line 13: you could use a link with the class `govuk-botton` here as it is just to get to the start page

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -7,6 +7,8 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
 
   helper_method :vacancy_statistics_form
 
+  def start; end
```
jriga, line 10: I think that we should stick to `new` as it's more in line with rails controller convention and I when we reach that action we have not yet created a vacancy so it a good place for that kind a page to be.

app/components/dashboard_component/dashboard_component.html.slim
KyleMacPherson: Ah good idea, thank you!

app/controllers/publishers/vacancies_controller.rb
jriga: had a chat offline abouth this and we agreed to leave as is. With the current codebase structure either ways have there own drawbacks. This can be reviews with the future work about forms.

jriga (approved):

PR #7611, opened 2025-03-12
improve flaky test again

starswan:
This test stopped another production deploy. Hopefully this is a better improvement

### DIFF
--- a/spec/system/publishers/publishers_can_manage_job_applications_for_a_vacancy_spec.rb
+++ b/spec/system/publishers/publishers_can_manage_job_applications_for_a_vacancy_spec.rb
@@ -117,18 +117,11 @@ RSpec.describe "Publishers can manage job applications for a vacancy" do
       end
     end
 
-    describe "shortlisted application" do
-      let(:status) { "shortlisted" }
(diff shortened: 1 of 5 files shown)


spec/system/publishers/publishers_can_manage_job_applications_for_a_vacancy_spec.rb
```diff
@@ -75,7 +75,9 @@
         find(".govuk-checkboxes__item").click
       end
       click_on I18n.t("publishers.vacancies.job_applications.candidates.update_application_status")
-      find(".govuk-tag--purple").click
+      # wait for page load
+      find(".govuk-radios")
```
KyleMacPherson, line 79: [Minor] I think it might be better to use expect(page).to have_css(".govuk-radios"), I think the waiting behaviour is the same but might be a bit easier to understand at a glance than finding an element but not interacting with it. Also the assertion failing might make the issue with the test a bit more obvious.

KyleMacPherson (approved):

scruti (approved):

PR #7635, opened 2025-03-18
Amazon SES mailer

starswan:
## Trello card URL

https://trello.com/c/wZGZ28Yl/1641-spike-build-dummy-mailer-client-using-amazon-ses-instead-of-govuk-notify

## Changes in this PR:

This is the (currently unused) Amazon SES Mail client, used to send ATS comms


jriga (approved):

app/mailers/amazon_ses_mailer.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  self.delivery_method = :smtp unless Rails.env.test?
+
+  default from: "ats.teachingvacancies@service.education.gov.uk"
+
+  def send_email(to:, subject:)
+    @to = to
+    mail(to: to, subject: subject, delivery_method_options: {
+      user_name: Rails.application.credentials.dig(:smtp, :user_name),
+      password: Rails.application.credentials.dig(:smtp, :password),
```
scruti, line 21: :raising_hand: The service uses AWS Parameter Store to store these secrets ) we have an entry per env on ...`/app/secrets`
I remember having to add the Rails application credentials just because we had to store a massive private key for GovUK One Login that exceeded parameter store limits. But if possible... we should keep all the secrets together.

scruti (changes requested):
Added a comment on where the new secrets are being stored.

scruti (approved):
LGTM!

PR #7637, opened 2025-03-18
Add explicit waits on JS flaky tests

scruti:
Went through the JS tests that failed between yesterday and today and added some explicit waiting times to all the JS flagged tests on those files.

### DIFF
--- a/spec/system/publishers/job_applications_draft_spec.rb
+++ b/spec/system/publishers/job_applications_draft_spec.rb
@@ -27,11 +27,11 @@ RSpec.describe "Draft job applications for publishers" do
       end
 
       it "shows a status 'tag' of 'unread'" do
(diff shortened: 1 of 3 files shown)


spec/system/publishers/job_applications_draft_spec.rb
```diff
@@ -27,11 +27,11 @@
       end
 
       it "shows a status 'tag' of 'unread'" do
-        expect(page).to have_css(".govuk-tag", text: "unread")
+        expect(page).to have_css(".govuk-tag", text: "unread", wait: 5)
```
KyleMacPherson, line 30: Wow, I didn't know you could do this! 🧠

KyleMacPherson (approved):

PR #7644, opened 2025-03-24
[ATS API] Soft delete vacancies

jriga:
## Trello card URL
[1694](https://trello.com/c/gTuRa2Oy)

## Changes in this PR:

- Add `Vacancy#trash!` method
  Move the logic out of the controller down in the models
  This makes it simpler to keep a consistent behaviour across UI and API

- ATS API does not list trashed vacancies

- External reference constraints applies to trash vacancies

### DIFF
--- a/app/controllers/publishers/vacancies_controller.rb
+++ b/app/controllers/publishers/vacancies_controller.rb
@@ -39,9 +39,7 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
   end
 
   def destroy
-    vacancy.supporting_documents.purge_later
-    vacancy.update_attribute(:status, :trashed)
-    remove_google_index(vacancy)
+    vacancy.trash!
     redirect_to organisation_jobs_with_type_path, success: t(".success_html", job_title: vacancy.job_title)
   end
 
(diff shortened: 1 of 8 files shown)


jriga:
[ATS API: Move to soft deletion for DELETE endpoint](https://trello.com/c/gTuRa2Oy/1694-ats-api-move-to-soft-deletion-for-delete-endpoint)

app/models/vacancy.rb
```diff
@@ -302,6 +302,23 @@ def find_conflicting_vacancy
...
+    remove_google_index
+  end
+
+  def remove_google_index
+    return if DisableExpensiveJobs.enabled?
+
+    url = Rails.application.routes.url_helpers.job_url(self)
+    RemoveGoogleIndexQueueJob.perform_later(url)
+  end
+
```
scruti, line 321: Nice!

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -39,9 +39,7 @@ def review
   end
 
   def destroy
-    vacancy.supporting_documents.purge_later
-    vacancy.update_attribute(:status, :trashed)
-    remove_google_index(vacancy)
+    vacancy.trash!
```
scruti, line 42: :clap:

PR #7646, opened 2025-03-24
Bulk rejection process

starswan:
## Changes in this PR:


config/routes.rb
```diff
@@ -452,7 +466,21 @@
...
+        end
+      end
     end
+
+    # scope constraints: { type: /(rejection)/ } do
+    #   resources :email_templates, only: %i[new create edit update destroy], controller: "publishers/email_templates"
+    #   # get "jobs(/:type)", to: "publishers/vacancies#index", defaults: { type: :published }, as: :jobs_with_type
+    # end
+
+    resources :email_templates, only: %i[new create edit update destroy], controller: "publishers/email_templates"
```
scruti: Same as above. This seems duplicated?

db/schema.rb
```diff
@@ -10,14 +10,14 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_10_03_121224) do
+ActiveRecord::Schema[8.0].define(version: 2025_10_15_125032) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gist"
   enable_extension "citext"
   enable_extension "fuzzystrmatch"
+  enable_extension "pg_catalog.plpgsql"
```
scruti: This makes me a little bit nervous. Guess if the review app has been published, it is not causing issues with the Azure DB 🤞

PR #7650, opened 2025-03-25
Delete subscription when email invalid

jriga:
## Trello card URL
[1701](https://trello.com/c/7d2KfmKy)

## Changes in this PR:

captures the exception only for that scenario otherwise re raise the exception

### DIFF
--- a/app/mailers/jobseekers/alert_mailer.rb
+++ b/app/mailers/jobseekers/alert_mailer.rb
@@ -1,4 +1,7 @@
 class Jobseekers::AlertMailer < Jobseekers::BaseMailer
+  rescue_from Notifications::Client::BadRequestError, with: :handle_invalid_email_exception
+  INVALID_EMAIL_REGEXP = %r{ValidationError: email_address}
+
   after_action :jobseeker
 
   self.delivery_job = AlertMailerJob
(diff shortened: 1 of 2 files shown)


spec/mailers/jobseekers/alert_mailer_spec.rb
```diff
@@ -61,6 +61,35 @@
     subscription.create_alert_run
   end
 
+  describe "exception handling" do
+    # exception message found here
+    # https://docs.notifications.service.gov.uk/ruby.html#send-a-file-by-email-response
```
scruti, line 66: :heart:

scruti (approved):

app/mailers/jobseekers/alert_mailer.rb
```diff
@@ -30,6 +33,12 @@ def alert(subscription_id, vacancy_ids)
 
   attr_reader :subscription_id
 
+  def handle_invalid_email_exception(exception)
+    return subscription.destroy! if exception.message.match?(INVALID_EMAIL_REGEXP)
+
+    raise exception
```
starswan, line 39: I'm not sure we should be destroying subscriptions - we can just make them inactive IMHO?

PR #7665, opened 2025-03-31
Fix invalid URI error on Vacancy breadcrumbs

scruti:
## Sentry issue
- https://teaching-vacancies.sentry.io/issues/6270428510

## Changes in this PR:
Sometimes the request referrer contains an invalid uri. EG: '"'
This breaks the display of the vacancy for the visitor.

Fix it by rescuing it by setting an empty referrer as the default.


KyleMacPherson (approved):
LGTM!

PR #7679, opened 2025-04-03
Validate incoming API requests using swagger schemas

starswan:
## Trello card URL

https://trello.com/c/CGTN1GZm/1734-api-feedback-eteach-fix-errors-when-invalid-values-are-provided-for-arrayenum

## Changes in this PR:

Validate incoming API requests using swagger schemas


scruti (approved):
Very good appproach that brings a very consistent way to validate the schemas and actually helpful feedback on the error messages to the clients.

Thanks for bringing it in. 

I've added a few bits after testing the API in the review app end to end and noticing some missing bits (like the need to explicitly require the gem to run the validations on the deployed environments or forcing the schema validation to be consistent when there is no "vacancy" main keyword).

Did full end-to-end testing, and I'm merging it so our first pilot tester can use this improvement!

PR #7708, opened 2025-04-16
Jobs management pagination

starswan:
## Trello card URL

https://trello.com/c/WoE7tkoX/1745-add-pagination-onto-hiring-staff-jobs-management-pages

## Changes in this PR:

Jobs management pagination

### DIFF
--- a/app/controllers/publishers/vacancies_controller.rb
+++ b/app/controllers/publishers/vacancies_controller.rb
@@ -30,9 +30,11 @@ class Publishers::VacanciesController < Publishers::Vacancies::BaseController
 
     vacancies = Vacancy.in_organisation_ids(current_publisher.accessible_organisations(current_organisation).map(&:id))
 
-    @vacancies = vacancies.public_send(VACANCY_TYPES.fetch(@selected_type))
+    vacancies = vacancies.public_send(VACANCY_TYPES.fetch(@selected_type))
                           .order(@sort.by => @sort.order)
                           .where.not(job_title: nil)
+    @pagy, @vacancies = pagy(vacancies)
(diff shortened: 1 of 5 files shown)


app/controllers/publishers/vacancies_controller.rb
```diff
@@ -13,11 +13,28 @@ def show
     @vacancy = VacancyPresenter.new(vacancy)
   end
 
+  # rubocop:disable Metrics/AbcSize
   def index
     @selected_type = (params[:type] || :live).to_sym
```
jriga, line 18: we should probably add a validating method in the controller even though there are  constraint the routes.rb. It's straight away visible that we used it with `send` and `public_send`

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -13,11 +13,28 @@ def show
...
     @publisher_preference = PublisherPreference.find_or_create_by(publisher: current_publisher, organisation: current_organisation)
     @sort = Publishers::VacancySort.new(current_organisation, @selected_type).update(sort_by: params[:sort_by])
-  end
+
+    vacancies =
+      if @publisher_preference.organisations.any?
```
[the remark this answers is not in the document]
starswan: This code has moved, so its hard come up with a new structure. All I'm doing is adding pagination.

app/controllers/publishers/vacancies_controller.rb
starswan: same as above - this is just code motion. Brakeman couldn't see this code before for some reason, it was still there...

jriga (approved):

PR #7711, opened 2025-04-22
Remove plpgsql

KyleMacPherson:
## Changes in this PR:

This change is to fix [this deployment issue](https://github.com/DFE-Digital/teaching-vacancies/actions/runs/14590558093/job/40925626193) we are running into. It removes plpgsql  as it's removed as an extension option and is installed by default so we don't need to specify it.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/terraform/app/modules/paas/aks_database.tf
+++ b/terraform/app/modules/paas/aks_database.tf
@@ -12,7 +12,7 @@ module "postgres" {
   use_azure                      = var.deploy_azure_backing_services
   azure_enable_backup_storage    = var.azure_enable_backup_storage
   azure_enable_monitoring        = var.enable_monitoring
-  azure_extensions               = ["btree_gist", "pgcrypto", "fuzzystrmatch", "plpgsql", "pg_trgm", "postgis"]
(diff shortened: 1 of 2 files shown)


terraform/app/modules/paas/aks_database.tf
```diff
@@ -12,7 +12,7 @@ module "postgres" {
   use_azure                      = var.deploy_azure_backing_services
   azure_enable_backup_storage    = var.azure_enable_backup_storage
   azure_enable_monitoring        = var.enable_monitoring
-  azure_extensions               = ["btree_gist", "pgcrypto", "fuzzystrmatch", "plpgsql", "pg_trgm", "postgis"]
+  azure_extensions               = ["btree_gist", "pgcrypto", "fuzzystrmatch", "pg_trgm", "postgis"]
```
starswan, line 15: do we need to disable the plpgsql extension in schema.rb as well

PR #7731, opened 2025-04-29
create analytics events on copy and relist journeys

starswan:
## Trello card URL

https://trello.com/c/IzjqtNvU/1794-analytics-events-bug-send-additional-documents-data-to-analytics

## Changes in this PR:

Add analytics events to copy and relist journeys

### DIFF
--- a/app/controllers/publishers/vacancies/documents_controller.rb
+++ b/app/controllers/publishers/vacancies/documents_controller.rb
@@ -20,7 +20,7 @@ class Publishers::Vacancies::DocumentsController < Publishers::Vacancies::BaseCo
 
   def destroy
     document = vacancy.supporting_documents.find(params[:id])
-    document.purge
+    document.purge_later
     send_dfe_analytics_event(:supporting_document_deleted, document.filename, document.byte_size, document.content_type)
 
     redirect_to after_document_delete_path, flash: { success: t("jobs.file_delete_success_message", filename: document.filename) }


spec/requests/publishers/vacancies/copies_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  after { sign_out(publisher) }
+
+  describe "POST /copy" do
+    let(:vacancy) { create(:vacancy, :with_application_form, :with_supporting_documents, organisations: [organisation]) }
+
+    it "doesn't crash" do
+      post organisation_job_copy_path(vacancy.id)
+      expect(response).to have_http_status(:found)
```
KyleMacPherson, line 21: Maybe we could add some additional assertions here to try to get more value out of the test, perhaps something like:

`expect { post organisation_job_copy_path(vacancy.id) }.to change(Vacancy, :count).by(1)`

KyleMacPherson (approved):
Very nice! Added a small comment on the test but looks good to me!

app/controllers/publishers/vacancies/make_duplicate_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        CopyVacancy.new(vacancy).call.tap do |new_vacancy|
+          if new_vacancy.include_additional_documents
+            new_vacancy.supporting_documents.attachments.each do |document|
+              send_dfe_analytics_event(:supporting_document_created, new_vacancy.id, document.blob)
+            end
+          end
+
+          if new_vacancy.application_form.present?
+            send_dfe_analytics_event(:supporting_document_created, new_vacancy.id, new_vacancy.application_form.attachment.blob)
+          end
```
scruti, line 18: Given that this logic seems to be needed on every case we use `CopyVacancy` and `CopyVacancy` is a service object, wouldn't be appropriate to move it to inside the `CopyVacancy#call` ?

starswan: the challenge is that the send_dfe_analytics_event method is in the controller...

scruti: I was assuming that `send_dfe_analytics_event` could be easily moved. But looking into detail, now I see that it is using the `request`  && `response`.

My main concern is that `CopyVacancy#call` is effectively invalid/buggy if it is not called exactly like this each time:
```ruby
        CopyVacancy.new(vacancy).call.tap do |new_vacancy|
          if new_vacancy.include_additional_documents
            new_vacancy.supporting_documents.attachments.each do |document|
              send_dfe_analytics_event(:supporting_document_created, new_vacancy.id, document.blob)
            end
          end

          if new_vacancy.application_form.present?
            send_dfe_analytics_event(:supporting_document_created, new_vacancy.id, new_vacancy.application_form.attachment.blob)
          end
```

Maybe this is not the right event to trigger and we would have to define a particular `:supporting_document_copied` within `CopyVacancy`  not depending on the request/response? 

And even if the exact `:supporting_document_created`  event is needed with request/response/publisher info as it is now, given that we need it each time that `CopyVacancy#call` is called. I would rather inject the request data so the `CopyVacancy` internally commands the event triggering if relevant. 

Anyways, this is just for your consideration, not a blocking comment. I will be completely ok if you decide to proceed with this as it is :+1:

spec/requests/publishers/vacancies/copies_spec.rb
starswan: That was my intention, but didn't re-visit after Brandon checked it over...

app/controllers/publishers/vacancies/copy_controller.rb
```diff
@@ -1,7 +1,6 @@
-class Publishers::Vacancies::CopyController < Publishers::Vacancies::BaseController
+class Publishers::Vacancies::CopyController < Publishers::Vacancies::MakeDuplicateController
   def create
-    vacancy.status = :draft
```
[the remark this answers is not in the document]
starswan: Yes the call already added the draft status

scruti (approved):

PR #7733, opened 2025-04-29
Discard soft deletes phase 1

starswan:
## Trello card URL

https://trello.com/c/arZfP3fr/1803-draftvacancy-step-1-remove-the-soft-deletion-from-the-vacancy-status-list

## Changes in this PR:

Phase 1 of soft deletes - just introduce and use soft deletes, adding a rake task to back-fill whilst keeping current functionality - specifically not just using the 'discarded' status

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -193,6 +193,8 @@ GEM
       responders
       warden (~> 1.2.3)
     diff-lcs (1.6.2)
+    discard (1.4.0)
+      activerecord (>= 4.2, < 9.0)
     docile (1.4.1)
     dotenv (3.1.8)
     dotenv-rails (3.1.8)
@@ -890,6 +892,7 @@ DEPENDENCIES
   debug (>= 1.0.0)
   devise
   dfe-analytics!
+  discard (~> 1.4)
   dotenv-rails
   factory_bot_rails
   faker


app/models/vacancy.rb
```diff
@@ -294,13 +296,10 @@ def find_conflicting_vacancy
     find_conflict_vacancy || find_duplicate_vacancy
   end
 
+  # soft delete so that all the stats etc are kept even after the vacancy no longer exists
   def trash!
-    return if trashed?
```
scruti, line 298: Question here:
wouldn't be better to keep the guard clause here?
`return if discarded?` 

If not multiple calls over `trash!` would repeatedly triggering the document purge job and google index removal API call, and I imagine also updating the discarding timestamp?

lib/tasks/dicard_trashed_vacancies.rake
```diff
@@ -0,0 +1,6 @@
+namespace :vacancies do
+  desc "Discard trashed vacancies"
+  task discard_trashed: :environment do
+    Vacancy.trashed.or(Vacancy.removed_from_external_system).update_all(discarded_at: Time.current)
```
scruti, line 4: If we try to be able to identify when a vacancy was trashed or removed from external system, could we use their `updated_at` instead of `Time.current` to set the `discarded_at` timestamp?

It may be imperfect, but I guess would be pretty accurate for the majority of vacancies.

scruti: I see that the[ follow-up PR](https://github.com/DFE-Digital/teaching-vacancies/pull/7732/files) removes the `trashed` status (:raised_hands: ). 
Would it be appropriate to use this task to also backfill all the `trashed` to `published`? 
Or `draft` vacancies could also be trashed, and we can't distinguish if a current `trashed` should map to `published` or `draft`? :thinking:

starswan: Thanks - I had intended to update the status to published, but forgot when copying the example from the discard docs. Don't think draft vacancies can be trashed - it's only external vacancies that can be soft deleted (in the current code base anyway)

app/models/vacancy.rb
starswan: I wasn't sure whether the scenario was even possible - its only the unit test that keeps the code executing. Now we only load  'kept' vacancies, it seems unlikely that double-trashing could ever take place

lib/tasks/dicard_trashed_vacancies.rake
scruti: Great then. Thanks for looking into this!

PR #7736, opened 2025-04-30
[SPIKE] Job Application Tab nav

jriga:
-------

## Trello card URL
[1755](https://trello.com/c/fZDkmt60)

### Issue
When performing an action and the selected job application is empty
Then the job application tab navigation would always display the `All` tab with an error message
even if we were on a different tab.

### Desired outcome
We would like to remain on the same tab after performing an action.

### Proposed solution
Pass on new parameter `origin` that would indicate the which tab triggered the action.
with this new parameter the controller is now able to redirect to the source tab.

### Technical approach
- Add a new scope `JobApplication.group_by_status`, returns a hash where the key is:
     * native status, shortlisted int 
     * a composite status used on the tab, aka new
     * all the records in scope, all
  and the value is an ActiveRecord::Relation for the selected statuses
     
- Add method `current_tab` in job_application_controller that returns the url anchor necessary to have the tab displayed after redirection.

### DIFF
--- a/app/controllers/publishers/vacancies/job_applications_controller.rb
+++ b/app/controllers/publishers/vacancies/job_applications_controller.rb
@@ -2,7 +2,7 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
   include Jobseekers::QualificationFormConcerns
   include DatesHelper
 
-  helper_method :employments, :form, :job_applications, :qualification_form_param_key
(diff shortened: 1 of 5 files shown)


jriga:
[Spike: application tab navigation](https://trello.com/c/fZDkmt60/1755-spike-application-tab-navigation)

jriga:
we will be keeping this approach without the new `job_application.group_by_status` scope

PR #7761, opened 2025-05-09
[bugfix]: Copy records using #duplicate rather than #dup

starswan:
## Trello card URL

https://trello.com/c/uaWbekjF/1830-active-profiles-application-data-copy-bug

## Changes in this PR:

Employments, training, qualifications  and professional body memberships are all modelled with 2 optional associations rather than one polymorphic one. This can cause issues as the #dup method in rails doesn't copy child associations, but it does copy parent ones, which means when we copy from profile -> job_application (and back again) we need to clear out the parent associations, otherwise we risk dual-linking the records which can cause loss of data when e.g. the profile is updated from the job application, it causes the records in the job application to be destroyed.

This only manifested itself when copying from a profile into a job application (which is currently only done if the profile is active and there are no non-draft applications) but it does then cause all applications for that user to be problematic. 

We also forgot to copy the professional body memberships back into the profile, so this PR does that too.

### Data & Schema Changes

- [X] Adds/removes model validations

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/models/concerns/associated_record.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module AssociatedRecord
```
scruti, line 3: Suggestion not request. Just wondering if this could be more explicit, given the coupling with 2 particular models
```suggestion
module ApplicationAndProfileAssociatedRecord
```

lib/tasks/remove_double_linked_associations.rake
```diff
@@ -0,0 +1,21 @@
+namespace :db do
+  desc "Update working_patterns to replace 'job_share' with 'part_time' and set is_job_share to true"
```
scruti, line 2: :point_up: This description needs to be changed to match the new task :P

scruti (approved):
Great work identifying and resolving the issue LGTM!

PR #7778, opened 2025-05-15
Self disclosure form

jriga:
## Trello card URL
[1798](https://trello.com/c/brOdxbPw)
[1800](https://trello.com/c/rkgpuLUl)

## Changes in this PR:

### Data & Schema Changes

- [X] Adds/removes model validations

<details>

### Integration Impact

- [X] DfE Analytics platform

</details>


app/controllers/publishers/vacancies/references_and_declarations_controller.rb
```diff
@@ -14,34 +14,48 @@ class ReferencesAndDeclarationsController < JobApplications::BaseController
 
       steps(*FORMS.keys)
 
+      # :nocov:
```
starswan, line 17: :nocov: is only for legacy code. I guess we need some more tests?

app/controllers/jobseekers/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  }.freeze
+
+  steps(*FORMS.keys)
+
+  def show
+    render_wizard
+  end
+
+  def update
+    if form.update!
```
starswan, line 23: This pattern is out of sync with the rest of the codebase. if @form.valid? then update is more the style that is used.

app/controllers/jobseekers/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,95 @@
+class Jobseekers::JobApplications::SelfDisclosureController < Jobseekers::BaseController
+  include Wicked::Wizard
+
+  before_action :redirect_when_self_disclosure_not_pending, except: :completed
+  before_action :form, except: :completed
```
starswan, line 5: This feels a little counter-intuitive - the method uses ||= as though it is a normal method, but it doesn't need to. Maybe it should be called set_form, and then controller uses @form throughout as it can be depended on to be set

app/controllers/jobseekers/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  steps(*FORMS.keys)
+
+  def show
+    render_wizard
+  end
+
+  def update
+    if form.update!
+      if next_step == "wicked_finish"
```
starswan, line 24: There is a constant for this - Wicked::FINISH_STEP

app/controllers/publishers/vacancies/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,49 @@
+class Publishers::Vacancies::JobApplications::SelfDisclosureController < Publishers::Vacancies::JobApplications::BaseController
+  before_action :not_enabled
+
+  helper_method :job_application
```
starswan, line 4: I've been trying to avoid this pattern in favour of just setting controller variables and passing them to the view

app/form_models/jobseekers/job_applications/self_disclosure/personal_details_form.rb
```diff
@@ -0,0 +1,30 @@
+module Jobseekers::JobApplications::SelfDisclosure
+  class PersonalDetailsForm < BaseForm
+    # include DateAttributeAssignment
+    # attr_reader :date_of_birth
+    # def date_of_birth=(value)
+    #   @date_of_birth = date_from_multiparameter_hash(value)
+    # end
```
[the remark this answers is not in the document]
starswan: Think this is the way to go rather than parsing in the controller

app/models/self_disclosure.rb
```diff
@@ -0,0 +1,28 @@
...
+  has_encrypted :has_unspent_convictions, :has_spent_convictions, :is_barred, :has_been_referred,
+                :is_known_to_children_services, :has_been_dismissed, :has_been_disciplined,
+                :has_been_disciplined_by_regulatory_body, :agreed_for_processing,
+                :agreed_for_criminal_record, :agreed_for_organisation_update,
+                :agreed_for_information_sharing, type: :boolean
+
+  def self.find_or_create_by_and_prefill!(job_application)
+    find_or_create_by!(
+      self_disclosure_request_id: job_application.self_disclosure_request.id,
+    ).prefill(job_application)
```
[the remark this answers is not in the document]
starswan: Think we only want to pre-fill on create, not on find

app/models/self_disclosure_request.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def self.create_and_notify_all!(job_applications)
+    transaction do
+      job_applications.map do |job_application|
+        find_or_create_by!(job_application_id: job_application.id).tap do |request|
+          job_application.self_disclosure_request = request
+          SelfDisclosure.find_or_create_by_and_prefill!(job_application)
+          Jobseekers::JobApplicationMailer.declarations(job_application).deliver_later
+          request.sent!
+        end
```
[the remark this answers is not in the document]
starswan: again, don't think we want an update if we find the record

spec/page_objects/pages/jobseeker/job_applications/self_disclosure/date_field.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module PageObjects
+  module Pages
+    module Jobseeker
+      module JobApplications
+        module SelfDisclosure
+          class DateField < SitePrism::Section
```
starswan, line 8: as this isn't a page, does it belong in a different namespace to PageObjects::Pages?

app/controllers/jobseekers/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def form_params
+    params
+      .fetch(form_params_key, {})
+      .tap { flatten_enumerable_value(it) }
+      .tap { parse_dob(it) }
+      .permit(*form_class.new.attributes)
+  end
+
+  def flatten_enumerable_value(hsh)
```
starswan, line 60: I'm not sure why we need this? Doesn' feel clear to me what's going on here.

app/controllers/publishers/vacancies/job_applications/self_disclosure_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def show
+    @self_disclosure = SelfDisclosurePresenter.new(job_application)
+    if @self_disclosure.request.manual?
+      @batch = JobApplicationBatch.create!(vacancy_id: job_application.vacancy.id).tap do
```
starswan, line 9: This doesn't appear to be used in any of the views, just it will just leak into the database?

app/views/publishers/vacancies/job_applications/self_disclosure/show.html.slim
```diff
@@ -0,0 +1,21 @@
...
+  = govuk_back_link text: t("app.back_link"), href: pre_interview_checks_organisation_job_job_application_path(vacancy.id, job_application), html_attributes: { "aria-label" => "Back navigation", role: "navigation" }
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    span.govuk-caption-l = @self_disclosure.header_text
+    h1.govuk-heading-xl class="govuk-!-margin-bottom-5 govuk-!-margin-top-0"
+      = @self_disclosure.applicant_name
+
+    - if @self_disclosure.request.completed?
+      = render "jobseekers/job_applications/self_disclosure/review/completed"
```
starswan, line 13: I have been trying to always pass variables to partials rather than them accessing globals directly. I always thought it was a standard, but don't know where from?

spec/factories/self_disclosure_requests.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :self_disclosure_request do
+    job_application
+    status { 2 }
```
starswan, line 4: can this not be an enum value? I can't remember what '2' means...

PR #7788, opened 2025-05-22
Upload job applications

KyleMacPherson:
## Changes in this PR:

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -91,7 +91,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   def submit
     raise ActionController::RoutingError, "Cannot submit application for non-listed job" unless vacancy.listed?
     raise ActionController::RoutingError, "Cannot submit non-draft application" unless job_application.draft?
-    binding.pry
+
     if review_form.valid? && all_steps_valid?(job_application.type)
       update_jobseeker_profile!(job_application) if review_form.update_profile
       job_application.submit!
@@ -157,7 +157,6 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def all_steps_valid?(job_application_type = nil)
-    binding.pry
     # Check that all steps are valid, in case we have changed the validations since the step was completed.
     # NB: Only validates top-level step forms. Does not validate individual qualifications, employments, or references.
     if job_application_type == "UploadedJobApplication"
@@ -215,7 +214,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def raise_unless_vacancy_enable_job_applications
-    raise ActionController::RoutingError, "Cannot apply for this vacancy" unless vacancy.enable_job_applications? || vacancy.receive_applications == "uploaded_form"
+    raise ActionController::RoutingError, "Cannot apply for this vacancy" unless vacancy.enable_job_applications? || vacancy.has_uploaded_form?
   end
 
   def review_form
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -11,7 +11,6 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
 
   def new
     send_dfe_analytics_event
-
     if session[:newly_created_user]
       @newly_created_user = true
       session.delete(:newly_created_user)
@@ -26,8 +25,13 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def create
-    new_job_application = current_jobseeker.job_applications.create(vacancy:)
-    redirect_to jobseekers_job_application_apply_path(new_job_application)
+    new_job_application = if vacancy.has_uploaded_form?
+      current_jobseeker.uploaded_job_applications.create!(vacancy:)
+    else
+      current_jobseeker.native_job_applications.create!(vacancy:)
+    end
+
+    redirect_to jobseekers_job_application_apply_path(new_job_application.id)
   end
 
   def pre_submit
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -31,8 +31,9 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def pre_submit
-    @form = Jobseekers::JobApplication::PreSubmitForm.new(completed_steps: job_application.completed_steps, all_steps: step_process.steps.excluding(:review).map(&:to_s))
-    if @form.valid? && all_steps_valid?
+    all_steps = job_application.is_a?(UploadedJobApplication) ? UploadedJobApplication::ALL_STEPS : step_process.steps.excluding(:review).map(&:to_s)
+    @form = Jobseekers::JobApplication::PreSubmitForm.new(completed_steps: job_application.completed_steps, all_steps: all_steps)
+    if @form.valid? && all_steps_valid?(job_application.type)
       redirect_to jobseekers_job_application_review_path(@job_application)
     else
       render :apply
@@ -65,6 +66,20 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
     @form = Jobseekers::JobApplication::PreSubmitForm.new(completed_steps: job_application.completed_steps, all_steps: step_process.steps.excluding(:review).map(&:to_s))
   end
 
+  def apply
+    if job_application.is_a?(UploadedJobApplication)
+      @form = Jobseekers::JobApplication::PreSubmitForm.new(
+        completed_steps: job_application.completed_steps,
+        all_steps: ["personal_details", "upload_form"],
+      )
+    else
+      @form = Jobseekers::JobApplication::PreSubmitForm.new(
+        completed_steps: job_application.completed_steps,
+        all_steps: step_process.steps.excluding(:review).map(&:to_s),
+      )
+    end
+  end
+
   def quick_apply
     raise ActionController::RoutingError, "Cannot quick apply if there's no profile or non-draft applications" unless quick_apply?
 
@@ -76,8 +91,8 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   def submit
     raise ActionController::RoutingError, "Cannot submit application for non-listed job" unless vacancy.listed?
     raise ActionController::RoutingError, "Cannot submit non-draft application" unless job_application.draft?
-
-    if review_form.valid? && all_steps_valid?
+    binding.pry
+    if review_form.valid? && all_steps_valid?(job_application.type)
       update_jobseeker_profile!(job_application) if review_form.update_profile
       job_application.submit!
       redirect_to jobseekers_job_application_post_submit_path job_application
@@ -141,14 +156,23 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
     profile.replace_memberships!(job_application.professional_body_memberships)
   end
 
-  def all_steps_valid?
+  def all_steps_valid?(job_application_type = nil)
+    binding.pry
     # Check that all steps are valid, in case we have changed the validations since the step was completed.
     # NB: Only validates top-level step forms. Does not validate individual qualifications, employments, or references.
-    step_process.steps.excluding(:review).all? { |step| step_valid?(step) }
+    if job_application_type == "UploadedJobApplication"
+      UploadedJobApplication::ALL_STEPS.all? { |step| step_valid?(step, job_application_type) }
+    else
+      step_process.steps.excluding(:review).all? { |step| step_valid?(step) }
+    end
   end
 
-  def step_valid?(step)
-    form_class = "jobseekers/job_application/#{step}_form".camelize.constantize
+  def step_valid?(step, job_application_type = nil)
+    form_class = if job_application_type == "UploadedJobApplication"
+                   "jobseekers/uploaded_job_application/#{step}_form".camelize.constantize
+                 else
+                   "jobseekers/job_application/#{step}_form".camelize.constantize
+                 end
 
     attributes = form_class.load_form(job_application)
 
@@ -191,7 +215,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def raise_unless_vacancy_enable_job_applications
-    raise ActionController::RoutingError, "Cannot apply for this vacancy" unless vacancy.enable_job_applications?
+    raise ActionController::RoutingError, "Cannot apply for this vacancy" unless vacancy.enable_job_applications? || vacancy.receive_applications == "uploaded_form"
   end
 
   def review_form
@@ -214,8 +238,9 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def review_form_params
+    all_steps = job_application.is_a?(UploadedJobApplication) ? UploadedJobApplication::ALL_STEPS : step_process.steps.excluding(:review).map(&:to_s)
     params.require(:jobseekers_job_application_review_form).permit(:confirm_data_accurate, :confirm_data_usage, update_profile: [])
-          .merge(completed_steps: job_application.completed_steps, all_steps: step_process.steps.excluding(:review).map(&:to_s))
+          .merge(completed_steps: job_application.completed_steps, all_steps: all_steps)
   end
 
   def withdraw_form_params
--- a/app/controllers/jobseekers/job_applications_controller.rb
+++ b/app/controllers/jobseekers/job_applications_controller.rb
@@ -35,9 +35,8 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def pre_submit
-    all_steps = job_application.is_a?(UploadedJobApplication) ? UploadedJobApplication::ALL_STEPS : step_process.steps.excluding(:review).map(&:to_s)
     @form = Jobseekers::JobApplication::PreSubmitForm.new(completed_steps: job_application.completed_steps, all_steps: all_steps)
-    if @form.valid? && all_steps_valid?(job_application.type)
+    if @form.valid? && all_steps_valid?
       redirect_to jobseekers_job_application_review_path(@job_application)
     else
       render :apply
@@ -67,21 +66,10 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
   end
 
   def apply
-    @form = Jobseekers::JobApplication::PreSubmitForm.new(completed_steps: job_application.completed_steps, all_steps: step_process.steps.excluding(:review).map(&:to_s))
-  end
-
-  def apply
-    if job_application.is_a?(UploadedJobApplication)
-      @form = Jobseekers::JobApplication::PreSubmitForm.new(
-        completed_steps: job_application.completed_steps,
-        all_steps: ["personal_details", "upload_form"],
-      )
-    else
-      @form = Jobseekers::JobApplication::PreSubmitForm.new(
-        completed_steps: job_application.completed_steps,
-        all_steps: step_process.steps.excluding(:review).map(&:to_s),
-      )
-    end
+    @form = Jobseekers::JobApplication::PreSubmitForm.new(
+      completed_steps: job_application.completed_steps,
+      all_steps: all_steps,
+    )
   end
 
   def quick_apply
@@ -96,7 +84,7 @@ class Jobseekers::JobApplicationsController < Jobseekers::JobApplications::BaseC
     raise ActionController::RoutingError, "Cannot submit application for non-listed job" unless vacancy.listed?
     raise ActionController::RoutingError, "Cannot submit non-draft application" unless job_application.draft?
 
-    if review_form.valid? && all_steps_valid?(job_application.type)
+    if review_form.valid? && all_steps_valid?
       update_jobseeker_profile!(job_application) if review_form.update_profile
       job_application.submit!
       redirect_to jobseekers_job_application_post_submit_path job_application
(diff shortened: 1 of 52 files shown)


app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -144,19 +151,7 @@ def update_jobseeker_profile!(job_application)
...
-    form_class = "jobseekers/job_application/#{step}_form".camelize.constantize
-
-    attributes = form_class.load_form(job_application)
-
-    form = form_class.new(attributes)
-
-    form.valid?.tap do
-      job_application.errors.merge!(form.errors)
-    end
+    Jobseekers::JobApplications::JobApplicationHandler.new(job_application, step_process).all_steps_valid?
```
starswan, line 154: did you consider just having a different step_process class for the new application type?

app/form_models/publishers/job_listing/application_form_form.rb
```diff
@@ -2,54 +2,19 @@ class Publishers::JobListing::ApplicationFormForm < Publishers::JobListing::Vaca
...
-
-    "other"
-  end
-
-  def other_application_email
-    return params[:other_application_email] if params[:other_application_email]
-
-    @vacancy.application_email unless @vacancy.application_email == @current_publisher&.email
-  end
+  attr_accessor(:application_form, :application_form_staged_for_replacement)
```
starswan, line 5: minor, but its more idiomatic to use attributes that attr_accessors. Doesn't make a lot of difference here TBH

app/services/jobseekers/job_applications/quick_apply.rb
```diff
@@ -29,4 +34,8 @@ def previously_submitted_application?
   def jobseeker_profile
     @jobseeker_profile ||= jobseeker.jobseeker_profile
   end
+
+  def has_data_available_to_prefill_with?
+    previously_submitted_application? || jobseeker_profile
```
starswan, line 39:
```suggestion
    previously_submitted_application? || jobseeker_profile.present?
```

db/migrate/20250509154018_add_type_to_job_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddTypeToJobApplications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :job_applications, :type, :string
```
starswan, line 3: amazed it lets you do this - when I added the type field to Vacancy, it insisted that I create a PR first to ignore the type column (so that this migration doesn't crash running instances which don't have the 2 types (yet)

db/schema.rb
```diff
@@ -259,6 +259,7 @@
     t.boolean "has_close_relationships"
     t.boolean "has_right_to_work_in_uk"
     t.boolean "has_safeguarding_issue"
+    t.string "type"
```
starswan, line 262: this should eventually be non-nullable (Think Vacancy needs to have its default removed...)

lib/tasks/backfill_type_for_job_applications.rake
```diff
@@ -0,0 +1,6 @@
+namespace :job_applications do
+  desc "Backfill job application type where it's nil"
+  task backfill_type: :environment do
+    JobApplication.where(type: nil).update_all(type: "NativeJobApplication")
```
starswan, line 4: Vacancy defaulted its type column which feels slightly cleaner than needing to run this script

app/models/vacancy.rb
```diff
@@ -310,6 +310,22 @@ def trash!
     remove_google_index
   end
 
+  def new_application_path
+    if has_uploaded_form?
+      Rails.application.routes.url_helpers.jobseekers_job_job_application_path(id)
+    else
+      Rails.application.routes.url_helpers.new_jobseekers_job_job_application_path(id)
+    end
+  end
```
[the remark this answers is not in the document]
jriga: an alternative, could be that we keep the existing route and add a before_action filter that would redirect to the endpoint for the first step of the upload job application

app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -31,6 +31,20 @@ def download_pdf
...
+      return
+    end
+
+    send_data(
+      @job_application.application_form.download,
+      filename: @job_application.application_form.filename.to_s,
+      type: @job_application.application_form.content_type,
+      disposition: "inline",
+    )
+  end
```
jriga, line 46: we could potentially have just one `download`  endpoint for both type of job application.

app/models/vacancy.rb
```diff
@@ -300,6 +300,22 @@ def is_a_teaching_or_middle_leader_role?
...
+
+  def has_uploaded_form?
+    receive_applications == "uploaded_form"
+  end
```
jriga, line 313: `enum` auto defines helper methods, so you can `vacancy.uploaded_form?` instead of defining has_uploaded_form?

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -26,12 +25,17 @@ def new
   end
 
   def create
-    new_job_application = current_jobseeker.job_applications.create(vacancy:)
-    redirect_to jobseekers_job_application_apply_path(new_job_application)
+    new_job_application = if vacancy.has_uploaded_form?
+                            current_jobseeker.uploaded_job_applications.create!(vacancy:)
+                          else
+                            current_jobseeker.native_job_applications.create!(vacancy:)
```
jriga, line 31: we could just have `new_job_application = current_jobseeker.job_applications.create!(vacancy:)`  and add a `before_create`  callback on the base model that would set the correct `type`

jriga:
question about the target branch of the PR, should it be `references-and-declarations` instead of `main` ?

app/views/jobseekers/job_applications/_banner.html.slim
```diff
@@ -14,6 +14,8 @@
     = job_application_status_tag(job_application.deadline_passed? ? :deadline_passed : job_application.status)
 
   .govuk-button-group class="govuk-!-margin-bottom-3 govuk-!-margin-top-3 govuk-!-display-none-print"
+    - if job_application.is_a?(UploadedJobApplication) && vacancy.application_form
+      = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id)
```
starswan, line 18:
```suggestion
    - if vacancy.application_form.present?
      = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id)
```

app/views/jobseekers/job_applications/_job_application_review_sections.html.slim
```diff
@@ -1,17 +1,20 @@
...
-  - when "other_religion"
-    = render "jobseekers/job_applications/review/non_catholic_religious_information", r: r, job_application: job_application
-= render "jobseekers/job_applications/review/references", r: r, job_application: job_application
-- if jobseeker_signed_in?
-  = render "jobseekers/job_applications/review/equal_opportunities", r: r
-= render "jobseekers/job_applications/review/ask_for_support", r: r, job_application: job_application
-= render "jobseekers/job_applications/review/declarations", r: r, job_application: job_application
+- if job_application.is_a?(UploadedJobApplication)
+  = render "jobseekers/job_applications/uploaded_job_applications/review/personal_details", r: r, job_application: job_application
+- else
```
starswan, line 3: it would be nice to put a dynamic dispatch method on the 2 subclasses rather than check the type? WDYT?

app/views/jobseekers/job_applications/_job_application_review_sections.html.slim
```diff
@@ -1,17 +1,20 @@
...
-  - when "catholic"
-    = render "jobseekers/job_applications/review/catholic_religious_information", r: r, job_application: job_application
-  - when "other_religion"
-    = render "jobseekers/job_applications/review/non_catholic_religious_information", r: r, job_application: job_application
-= render "jobseekers/job_applications/review/references", r: r, job_application: job_application
-- if jobseeker_signed_in?
-  = render "jobseekers/job_applications/review/equal_opportunities", r: r
-= render "jobseekers/job_applications/review/ask_for_support", r: r, job_application: job_application
-= render "jobseekers/job_applications/review/declarations", r: r, job_application: job_application
+- if job_application.is_a?(UploadedJobApplication)
```
starswan, line 1:
```suggestion
- if vacancy.has_uploaded_form?
```

app/views/jobseekers/job_applications/_show.html.slim
```diff
@@ -3,20 +3,21 @@
...
-        - navigation.with_anchor text: t(".training_and_cpds.heading"), href: "#training_and_cpds"
-        - navigation.with_anchor text: t(".professional_body_memberships.heading"), href: "#professional_body_memberships"
-        - navigation.with_anchor text: t(".employment_history.heading"), href: "#employment_history"
-        - navigation.with_anchor text: t(".personal_statement.heading"), href: "#personal_statement"
-        - if vacancy.religion_type.present?
-          - navigation.with_anchor text: t(".religious_information.heading"), href: "#following_religion"
-        - navigation.with_anchor text: t(".referees.heading"), href: "#referees"
-        - navigation.with_anchor text: t(".ask_for_support.heading"), href: "#ask_for_support"
-        - navigation.with_anchor text: t(".declarations.heading"), href: "#declarations"
+    - unless job_application.is_a?(UploadedJobApplication)
```
starswan, line 6:
```suggestion
    - unless vacancy.has_uploaded_form?
```

app/views/jobseekers/job_applications/apply.html.slim
```diff
@@ -12,64 +12,76 @@
...
       = render "shared/error_messages", model: @job_application, presenter: ErrorSummaryPresenter.new(@job_application.errors)
 
+  - if job_application.is_a?(UploadedJobApplication)
+    p.govuk-body = t(".download_application_form")
+
   h2.govuk-heading-m = t(".heading")
 
   = govuk_task_list do |task_list|
-    - task_list.with_item(title: t("jobseekers.job_applications.build.personal_details.heading"), href: jobseekers_job_application_build_path(job_application, :personal_details),
+    - if job_application.is_a?(UploadedJobApplication)
```
starswan, line 21:
```suggestion
    - if vacancy.has_uploaded_form?
```

jriga (commented):
app/controllers/jobseekers/uploaded_job_applications/personal_details_controller.rb
and
app/controllers/jobseekers/uploaded_job_applications/upload_application_forms_controller.rb

are 2 steps for the uploadedJobApplication, so wonder if it would appropriate to use wicked::wizard here since we've settled on this way to handle this type of form flows

config/locales/jobseekers/job_application/personal_details_form.yml
```diff
@@ -100,11 +126,25 @@ en:
...
+        first_name: First name
+        last_name: Last name
+        phone_number: Phone number
+        personal_details_section_completed_options:
+          false: No, I'll come back to it later
+          true: Yes, I've completed this section
+      jobseekers_uploaded_job_application_upload_application_form_form:
+       upload_application_form_section_completed_options:
+          false: No, I'll come back to it later
+          true: Yes, I've completed this section
```
scruti, line 140: Question (I may be lost here). Are these translations used at all?
The view seems to be calling the `helpers.label.jobseekers_job_application_personal_details_form.first_name` labels rather than these ones.

And it works as they're the same wording. We could just refer to them or just use the others. Don't know the best practice on that.

db/schema.rb
KyleMacPherson: Am planning to address this one in a follow up PR once we have no null types left after running the rake task!

app/controllers/jobseekers/job_applications_controller.rb
KyleMacPherson: I ended up encapsulating this in vacancy, just because i find callbacks make things a bit more difficult to debug and understand for me personally!

app/views/jobseekers/job_applications/_banner.html.slim
KyleMacPherson: Nope, good call! 🙇

config/locales/jobseekers/job_application/personal_details_form.yml
KyleMacPherson: We use these in app/views/jobseekers/uploaded_job_applications/personal_details/edit.html.slim

app/controllers/jobseekers/job_applications/base_controller.rb
```diff
@@ -2,9 +2,7 @@ class Jobseekers::JobApplications::BaseController < Jobseekers::BaseController
   helper_method :job_application
 
   def step_process
-    Jobseekers::JobApplications::JobApplicationStepProcess.new(
-      job_application: job_application,
-    )
+    job_application.step_process
```
starswan, line 5: think this line can look at the vacancy and create the appropriate step_process for the job application. IMHO step_process is a controller concern, and we might want other flows which are nothing to do with job application type (which I'm still unsure of, as they are totally dependent on the vacancy type (hence this seems to violate DRY principles) - looking at the vacancy we can always obtain job application 'type' (hence step process)

app/controllers/jobseekers/uploaded_job_applications/personal_details_controller.rb
```diff
@@ -0,0 +1,43 @@
+class Jobseekers::UploadedJobApplications::PersonalDetailsController < Jobseekers::JobApplications::BaseController
+  before_action :set_job_application
+
+  helper_method :job_application, :vacancy
+
+  def edit
+    @form = Jobseekers::UploadedJobApplication::PersonalDetailsForm.new(@job_application.slice(:first_name, :last_name, :email_address, :phone_number, :has_right_to_work_in_uk))
```
starswan, line 7: This form should be in the same namespace as the controller, so we don't need this long name? In Jobseekers::UploadedJobApplications

app/controllers/jobseekers/uploaded_job_applications/personal_details_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  def set_job_application
+    @job_application = current_jobseeker.job_applications.draft.find(params[:uploaded_job_application_id])
+  end
+
+  def form_params
+    params.require(:jobseekers_uploaded_job_application_personal_details_form).permit(:first_name, :last_name, :phone_number, :email_address, :teacher_reference_number, :personal_details_section_completed, :has_right_to_work_in_uk)
+  end
+
+  def vacancy
+    @vacancy ||= job_application.vacancy
```
starswan, line 41: not a fan of this pattern - it's idiomatic to set the variable when required and use in the view, rather than relying on this method to be called and set @vacancy by magic. It's not an expensive call as it's a property of the job_application, so there's no real need to cache it

app/form_models/jobseekers/uploaded_job_application/upload_application_form_form.rb
```diff
@@ -0,0 +1,30 @@
+module Jobseekers
+  module UploadedJobApplication
+    class UploadApplicationFormForm < Jobseekers::JobApplication::BaseForm
+      include ActiveModel::Model
+      include ActiveModel::Attributes
+      include Jobseekers::JobApplication::CompletedFormAttribute
+
+      validates :application_form, form_file: Vacancy::DOCUMENT_VALIDATION_OPTIONS
+      validate :application_form_presence, if: -> { upload_application_form_section_completed }
```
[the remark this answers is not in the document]
KyleMacPherson: Oh yeah! 😂 🤦  Thanks, nice spot!

app/models/vacancy.rb
```diff
@@ -322,6 +321,23 @@ def is_a_teaching_or_middle_leader_role?
     job_roles.intersect?(%w[teacher head_of_year_or_phase head_of_department_or_curriculum sendco other_leadership])
   end
 
+  def new_application_path
+    if uploaded_form?
+      Rails.application.routes.url_helpers.jobseekers_job_job_application_path(id)
+    else
+      Rails.application.routes.url_helpers.new_jobseekers_job_job_application_path(id)
+    end
+  end
```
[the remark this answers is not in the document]
KyleMacPherson: Good call, have moved it

app/controllers/jobseekers/uploaded_job_applications/personal_details_controller.rb
KyleMacPherson: Ok, done! 🫡

app/controllers/jobseekers/job_applications/base_controller.rb
scruti: Agree on this point. Flow flow-related process would be better living outside of the model.

app/models/job_application.rb
```diff
@@ -118,6 +119,14 @@ def unexplained_employment_gaps
...
   end
 
+  def form_class_for(step)
+    "jobseekers/job_application/#{step}_form".camelize.constantize
+  end
+
+  def step_process
+    Jobseekers::JobApplications::JobApplicationStepProcess.new(job_application: self)
+  end
+
```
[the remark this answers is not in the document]
scruti: Agree with this bit. Better to keep flow logic outside the model!

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -26,12 +25,14 @@ def new
   end
 
   def create
-    new_job_application = current_jobseeker.job_applications.create(vacancy:)
-    redirect_to jobseekers_job_application_apply_path(new_job_application)
+    new_job_application = vacancy.build_job_application_for(current_jobseeker)
+    new_job_application.save!
```
scruti, line 29: If every time we call this we end up doing a `build` and then a `save!` immediately after (I think we always do that, right?)... couldn't we just define a `create!` method instead?
```suggestion
    new_job_application = vacancy.create_job_application_for!(current_jobseeker)
```

app/controllers/jobseekers/uploaded_job_applications/upload_application_forms_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Jobseekers::UploadedJobApplications::UploadApplicationFormsController < Jobseekers::JobApplications::BaseController
+  before_action :set_job_application
+
+  helper_method :job_application, :vacancy
```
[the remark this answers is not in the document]
KyleMacPherson: Ok, done! 🫡

app/controllers/jobseekers/job_applications_controller.rb
KyleMacPherson: Sure, done 🫡

app/services/jobseekers/job_applications/quick_apply.rb
```diff
@@ -19,7 +20,7 @@ def job_application
   attr_reader :jobseeker, :vacancy
 
   def new_job_application
-    @new_job_application ||= jobseeker.job_applications.create(vacancy: vacancy)
+    vacancy.create_job_application_for(jobseeker).tap(&:save!)
```
scruti, line 23: :scissors: 
```suggestion
    vacancy.create_job_application_for(jobseeker)
```

spec/components/dashboard_component_spec.rb
```diff
@@ -241,4 +241,67 @@
...
+    let(:vacancies) { [vacancy] }
+    let(:vacancy) do
+      create(:vacancy, :published, organisations: [organisation])
+    end
+
+    before do
+      create(:job_application, :status_submitted, vacancy: vacancy)
+      # rubocop:disable RSpec/SubjectStub
+      allow(subject).to receive(:include_job_applications?).and_return(include_applications)
+      # rubocop:enable RSpec/SubjectStub
```
scruti, line 256: You could set the context based on the `selected_type` (that is what dictates the inner behaviour of `#include_job_applications?`) and is already available to set as a `let` value in the subject initialisation.

```ruby
context "when the selected type is live" do
  let(:selected_type) { :live }

  it "renders the link to view applicants" do
  ...
end

context "when the selected type is draft" do
  let(:selected_type) { :draft }

  ...
end
```
This way, you don't have to define this "problematic" stub over the subject under test and also avoids stubbing a private method.

app/controllers/jobseekers/uploaded_job_applications/personal_details_controller.rb
KyleMacPherson: Yeah not so sure about that one, I couldn't get it to work anyway

spec/components/dashboard_component_spec.rb
KyleMacPherson: Good idea, thanks! Done!

spec/requests/publishers/job_applications_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe "Publishers::Vacancies::JobApplicationsController#download_application_form" do
+  let(:publisher) { create(:publisher) }
+  let(:organisation) { create(:school) }
+  let(:vacancy) { create(:vacancy, organisations: [organisation]) }
+  let(:job_application) { create(:uploaded_job_application, vacancy: vacancy) }
+
+  before do
+    sign_in publisher
```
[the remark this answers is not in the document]
KyleMacPherson: Done 🫡

spec/components/dashboard_component_spec.rb
```diff
@@ -241,4 +241,64 @@
...
+              href: Rails.application.routes.url_helpers.organisation_job_job_applications_path(vacancy.id),
+            )
+          end
+        end
+      end
+    end
+
+    context "when vacancy does not use either native or uploaded job application form" do
+      let(:uses_either_native_or_uploaded_job_application_form) { false }
+      let(:include_applications) { true }
```
scruti, line 289: :axe: 
```suggestion
```

spec/components/dashboard_component_spec.rb
```diff
@@ -241,4 +241,64 @@
...
+            I18n.t("jobs.manage.view_applicants", count: 1),
+            href: Rails.application.routes.url_helpers.organisation_job_job_applications_path(vacancy.id),
+          )
+        end
+      end
+
+      context "when the selected type is draft" do
+        let(:selected_type) { :draft }
+
+        context "and receive_applications != 'uploaded_form" do
```
scruti, line 275: :question: do we need this context at all? It doesn't seem to modify anything on the `vacancy`  and from what I see on the component code, the `draft` state never adds the link to view applicants. 

So this context seems not to be correct?

spec/components/dashboard_component_spec.rb
```diff
@@ -241,4 +241,64 @@
...
+
+    before do
+      create(:job_application, :status_submitted, vacancy: vacancy)
+      allow(vacancy).to receive(:allow_job_applications?).and_return(uses_either_native_or_uploaded_job_application_form)
+    end
+
+    context "when vacancy uses either native or uploaded job application form" do
+      let(:uses_either_native_or_uploaded_job_application_form) { true }
+
+      context "when the selected type is live" do
```
scruti, line 260: Suggestion: I guess also adding a context/test for `:expired` will increase the test coverage?

spec/components/dashboard_component_spec.rb
KyleMacPherson: oops yes good spot 👍

spec/components/dashboard_component_spec.rb
KyleMacPherson: Sure, done! 🫡

scruti (approved):
All good from my side/feedback. Thanks for going through all the feedback/changes interactions!

PR #7795, opened 2025-05-28
Draft vacancy model

starswan:
## Changes in this PR:

Add DraftVacancy model to database

### Data & Schema Changes

- [x] Adds/removes database fields

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -49,41 +49,41 @@ end
 
 # Vacancies at Bexleyheath school
 attrs = { organisations: [bexleyheath_school], phases: [bexleyheath_school.phase], publisher_organisation: bexleyheath_school, publisher: Publisher.all.sample }
-6.times { FactoryBot.create(:vacancy, :for_seed_data, :published, **attrs) }
-2.times { FactoryBot.create(:vacancy, :for_seed_data, :published, :no_tv_applications, **attrs) }
+6.times { FactoryBot.create(:vacancy, :for_seed_data, **attrs) }
+2.times { FactoryBot.create(:vacancy, :for_seed_data, :no_tv_applications, **attrs) }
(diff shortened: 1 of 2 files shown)


spec/factories/vacancies.rb
```diff
@@ -250,5 +228,26 @@
       school_offer { nil }
       flexi_working { nil }
     end
+
+    factory :draft_vacancy do
```
scruti, line 232: Question: Shouldn't this follow the same "class:..." as `factory :vacancy, class: "PublishedVacancy" do`

```suggestion
    factory :draft_vacancy, class: "DraftVacancy" do
```

app/models/published_vacancy.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  # soft delete so that all the stats etc are kept even after the vacancy no longer exists
+  def trash!
+    return if discarded?
+
+    supporting_documents.purge_later
+    discard!
+    remove_google_index
+  end
```
scruti, line 24: I believe we need this `trash!` call to be defined on the 3 models.

- In the `Vacancy` model, conditionally soft/hard deleting depending on the current Vacancy `status`.
- In this `PublishedVacancy` model, as it is.
- In the `DraftVacancy` mode,l doing a hard delete.

This way, when `trash!` is called while using `Vacancy` as we do at the moment, it will soft/hard delete depending on the status. 

And when we move to not use `Vacancy` but `DraftVacancy` and `PublishedVacancy` directly, each one will behave differently upon `trash!` call.

scruti (commented):
Left a comment regarding deletion behaviour, and a couple of minor non-blocking test comments. The changes look great!

scruti (approved):
Great one!

PR #7798, opened 2025-05-28
Pass extra `origin` params

jriga:
retains the id of the original tab

-------

## Trello card URL
[1755](https://trello.com/c/fZDkmt60)

## Changes in this PR:

Implements change suggested is #7736

### Solution
Pass on new parameter `origin` that would indicate the which tab
triggered the action.  with this new parameter the controller is now
able to redirect to the source tab.

### DIFF
--- a/app/controllers/publishers/vacancies/job_applications_controller.rb
+++ b/app/controllers/publishers/vacancies/job_applications_controller.rb
@@ -32,38 +32,41 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
   end
 
   def tag_single
-    prepare_to_tag([params.fetch(:id)])
+    prepare_to_tag([params.fetch(:id)], "all")
   end
 
   def tag
-    tag_params = params.require(:publishers_job_application_tag_form).permit(job_applications: [])
+    tag_params = params.require(:publishers_job_application_tag_form).permit(:origin, job_applications: [])
     if params["download_selected"] == "true"
       download_selected(tag_params)
     else
-      prepare_to_tag(tag_params.fetch(:job_applications).compact_blank)
+      origin = tag_params[:origin]
+      prepare_to_tag(tag_params.fetch(:job_applications).compact_blank, origin)
     end
   end
 
   def update_tag
-    update_tag_params = params.require(:publishers_job_application_status_form).permit(:status, job_applications: [])
+    update_tag_params = params.require(:publishers_job_application_status_form).permit(:origin, :status, job_applications: [])
 
     JobApplication.find(update_tag_params.fetch(:job_applications)).each do |job_application|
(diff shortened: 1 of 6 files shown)


app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -36,36 +36,41 @@ def tag_single
...
+    redirect_to organisation_job_job_applications_path(vacancy.id, anchor: update_tag_params[:origin])
   end
 
   def withdrawn; end
 
   private
 
-  def prepare_to_tag(job_applications)
+  # :nocov:
+  def prepare_to_tag(job_applications, origin = "all")
```
starswan, line 62: does this need a default? Don't we always provide this parameter?

app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -36,36 +36,41 @@ def tag_single
...
     @form = Publishers::JobApplication::TagForm.new(job_applications: job_applications)
     if @form.valid?
       @job_applications = vacancy.job_applications.where(id: @form.job_applications)
+      @origin = origin
       render "tag"
     else
-      render "index"
+      flash[origin&.to_sym] = @form.errors.full_messages
```
starswan, line 69: Not sure about the safe nav - it will set flash[nil] which looks wrong? Think it can just be removed

app/controllers/publishers/vacancies/job_applications_controller.rb
starswan, line 62:
```suggestion
  def prepare_to_tag(job_applications, origin)
```

app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -36,36 +36,41 @@ def tag_single
...
-    redirect_to organisation_job_job_applications_path(vacancy.id)
+    redirect_to organisation_job_job_applications_path(vacancy.id, anchor: update_tag_params[:origin])
   end
 
   def withdrawn; end
 
   private
 
-  def prepare_to_tag(job_applications)
+  # :nocov:
```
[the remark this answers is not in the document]
jriga: :recycle:

app/controllers/publishers/vacancies/job_applications_controller.rb
jriga: :recycle:

scruti (approved):
LGTM! (Guess will need to also confirm @starswan :heavy_check_mark:  on this)

starswan:
> LGTM! (Guess will need to also confirm @starswan ✔️ on this)

Converting the [] to fetch looks like a good idea, as the origin parameter should always be supplied

PR #7854, opened 2025-06-18
Fix self disclosure email

jriga:
Add missing self-disclosure link

-------


config/locales/mailers.yml
```diff
@@ -132,8 +132,9 @@ en:
       declarations:
         subject: Declarations
         heading: Self-disclosure form link
```
scruti, line 134: Not sure if this is used anymore?
```suggestion
```

PR #7858, opened 2025-06-18
use secret rather than ENV/ARG for RAILS_MASTER_KEY

starswan:
## Trello card URL

https://trello.com/c/WqgZtiOj/1923-do-not-pass-secrets-to-dockerfile-builds-in-arg-or-env-values

## Changes in this PR:

Use secrets for RAILS_MASTER_KEY

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -148,16 +148,18 @@ jobs:
 
       - name: Build and push docker image from builder target
         uses: docker/build-push-action@v6
+        env:
+          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
+          BUILDKIT_INLINE_CACHE: 1
         with:
           build-args: |
-            BUILDKIT_INLINE_CACHE=1
-            RAILS_MASTER_KEY=${{ secrets.RAILS_MASTER_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -148,16 +148,18 @@ jobs:
 
       - name: Build and push docker image from builder target
         uses: docker/build-push-action@v6
+        env:
+          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
+          BUILDKIT_INLINE_CACHE: 1
```
scruti, line 153: Was this `BUILDKIT_INLINE_CACHE` move needed? 
Our [docker docs](https://github.com/DFE-Digital/teaching-vacancies/blob/main/documentation/operations/infrastructure/docker.md#building-with-github-actions-workflow) (not written by me, so missing context here) explicitly states that this env var should be passed as a build-arg to the image.

starswan: Thanks - it was a bit hit and hope to start with. I've removed some other bits too

PR #7876, opened 2025-06-26
Job offered tab

jriga:
## Trello card URL

## Changes in this PR:

Add Job offered tab to publisher job application page

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/controllers/publishers/vacancies/job_applications_controller.rb
+++ b/app/controllers/publishers/vacancies/job_applications_controller.rb
@@ -1,9 +1,10 @@
+require "zip"
+
 class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::JobApplications::BaseController
   include Jobseekers::QualificationFormConcerns
   include DatesHelper
 
   before_action :set_job_application, only: %i[show download_pdf pre_interview_checks collect_references]
-
   before_action :set_job_applications, only: %i[index tag_single tag]
 
   def index
@@ -32,36 +33,36 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
   end
 
   def tag_single
-    prepare_to_tag([params.fetch(:id)], "all")
+    form_key = ActiveModel::Naming.param_key(Publishers::JobApplication::TagForm)
+    params[form_key] = {
+      origin: JobApplicationsHelper::JOB_APPLICATION_DISPLAYED_STATUSES.first,
+      job_applications: [params.fetch(:id)],
+    }
+    with_valid_tag_form { render "tag" }
   end
 
   def tag
-    tag_params = params.require(:publishers_job_application_tag_form).permit(:origin, job_applications: [])
-    if params["download_selected"] == "true"
-      download_selected(tag_params)
-    else
-      origin = tag_params[:origin]
-      prepare_to_tag(tag_params.fetch(:job_applications).compact_blank, origin)
+    with_valid_tag_form do |form|
+      case params["target"]
+      when "download" then download_selected(form.job_applications)
+      when "export"   then export_selected(form.job_applications)
+      when "emails"   then copy_emails_selected(form.job_applications)
+      when "declined" then render_declined_form(form.job_applications, form.origin)
+      else # when "update_status"
+        render "tag"
+      end
     end
   end
 
   def update_tag
-    update_tag_params = params.require(:publishers_job_application_status_form).permit(:origin, :status, job_applications: [])
-
-    applications = update_tag_params.fetch(:job_applications)
-    new_status = update_tag_params.fetch(:status).to_sym
-
-    if new_status == :interviewing
-      batch = JobApplicationBatch.create!(vacancy: vacancy)
-      JobApplication.find(applications).each do |ja|
-        batch.batchable_job_applications.create!(job_application: ja)
+    with_valid_tag_form(context: :update_tag) do |form|
+      case form.status
+      when "interviewing" then redirect_to_references_and_declarations(form.job_applications)
+      when "offered" then render_offered_form(form.job_applications, form.origin)
+      else
+        form.job_applications.find_each { it.update!(form.attributes) }
+        redirect_to organisation_job_job_applications_path(vacancy.id, anchor: form.origin)
       end
-      redirect_to organisation_job_job_application_batch_references_and_declaration_path(vacancy.id, batch.id, Wicked::FIRST_STEP)
-    else
-      JobApplication.find(applications).each do |job_application|
-        job_application.update!(status: new_status)
-      end
-      redirect_to organisation_job_job_applications_path(vacancy.id, anchor: update_tag_params[:origin])
     end
   end
 
@@ -78,6 +79,19 @@ class Publishers::Vacancies::JobApplicationsController < Publishers::Vacancies::
     @reference_requests = @job_application.referees.filter_map(&:reference_request)
   end
 
+  def update_all
+    form_params = params.require(:publishers_job_application_offer_decline_date_form)
+                          .permit(:origin, :offered_at, :declined_at, :status, job_applications: [])
+    form_params[:job_applications] = vacancy.job_applications.where(id: Array(form_params[:job_applications]).compact_blank)
+    @form = Publishers::JobApplication::OfferDeclineDateForm.new(form_params)
+    if @form.valid?
+      vacancy.job_applications.where(id: @form.job_applications).find_each { it.update!(@form.attributes) }
+      redirect_to organisation_job_job_applications_path(vacancy.id, anchor: @form.origin)
+    else
+      render(@form.status == "offered" ? "offered_date" : "declined_date")
+    end
+  end
+
   private
 
   def set_job_applications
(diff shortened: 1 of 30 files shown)


spec/views/publishers/vacancies/job_applications/index.html.slim_spec.rb
```diff
@@ -1,12 +1,12 @@
 require "rails_helper"
 
 RSpec.describe "publishers/vacancies/job_applications/index" do
-  let(:organisation) { build_stubbed(:school) }
-  let(:publisher) { build_stubbed(:publisher, organisations: [organisation]) }
+  let(:organisation) { create(:school) }
+  let(:publisher) { create(:publisher, organisations: [organisation]) }
```
starswan, line 5: There should be no need to convert these to create's - build_stubbed should always work for view specs, as they have id, created_at and relationships

app/models/job_application.rb
```diff
@@ -1,5 +1,5 @@
 class JobApplication < ApplicationRecord
-  before_save :update_status_timestamp, if: :will_save_change_to_status?
+  before_save :update_status_timestamp, if: %i[will_save_change_to_status? ignore_for_offered_and_declined?]
```
starswan, line 2: I think we want a timestamp for the 2 new statuses IMHO

app/form_models/publishers/job_application/tag_form.rb
```diff
@@ -1,8 +1,30 @@
...
   validates_length_of :job_applications, minimum: 1
+  validates_presence_of :status, on: :update_tag
+
+  def job_application_ids
+    return [] if job_applications.blank?
+
+    job_applications.pluck(:id)
+  end
+
+  def update_job_application_statuses
```
starswan, line 18: Think this needs a comment or a better name - I can't understand what this method is doing, or why it is here

app/form_models/publishers/job_application/offer_decline_date_form.rb
```diff
@@ -0,0 +1,33 @@
...
+  def job_application_ids
+    return [] if job_applications.blank?
+
+    job_applications.pluck(:id)
+  end
+
+  def attributes
+    hsh = super
+    ignore_field = status == "offered" ? "declined_at" : "offered_at"
+    hsh.delete(ignore_field)
```
starswan, line 26: personally I would use hash.except(key) to prevent altering something that I don't own

app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -32,36 +33,36 @@ def download_pdf
   end
 
   def tag_single
-    prepare_to_tag([params.fetch(:id)], "all")
+    form_key = ActiveModel::Naming.param_key(Publishers::JobApplication::TagForm)
+    params[form_key] = {
+      origin: JobApplicationsHelper::JOB_APPLICATION_DISPLAYED_STATUSES.first,
+      job_applications: [params.fetch(:id)],
+    }
```
starswan, line 40: again I find this awkward altering the params hash rather than passing parameters explicitly - I find it very hard to follow. Not sure I understand why we are passing JobApplicationsHelper::JOB_APPLICATION_DISPLAYED_STATUSES.first as the origin when we should always be getting it from the request...?

app/form_models/publishers/job_application/tag_form.rb
```diff
@@ -1,8 +1,30 @@
...
+  attribute :status, :string
+  attr_accessor :job_applications, :origin
 
   validates_length_of :job_applications, minimum: 1
+  validates_presence_of :status, on: :update_tag
+
+  def job_application_ids
+    return [] if job_applications.blank?
+
+    job_applications.pluck(:id)
```
starswan, line 15:
```suggestion
    job_applications.map(&:id)
```

app/helpers/job_applications_helper.rb
```diff
@@ -1,11 +1,14 @@
 module JobApplicationsHelper
+  JOB_APPLICATION_DISPLAYED_STATUSES = %i[submitted unsuccessful shortlisted interviewing offered].freeze
```
starswan, line 2: I think this constant suggests that the 'job offered' logic belongs possibly in a different partial - it seems to have altered the 'candidates' partial (and the controller) a bit too much for my liking

app/views/publishers/vacancies/job_applications/_declined_row.html.slim
```diff
@@ -0,0 +1,3 @@
+- row.with_cell(text: govuk_link_to(application.name, organisation_job_job_application_path(vacancy.id, application)))
+- row.with_cell(text: application.email_address)
+- row.with_cell(text: application.declined_at&.to_fs(:day_month_year))
```
starswan, line 3: if the row is 'declined' then declined_at must be set - see previous discussions about safe navigation operators

app/views/publishers/vacancies/job_applications/_offered_row.html.slim
```diff
@@ -0,0 +1,5 @@
+- row.with_cell(text: tag.div(f.govuk_check_box(:job_applications, application.id, link_errors: index.zero?, label: { hidden: true, text: "Select #{application.name}" }), class: "govuk-checkboxes--small"))
+- row.with_cell(text: govuk_link_to(application.name, organisation_job_job_application_path(vacancy.id, application)))
+- row.with_cell(text: application.email_address)
+- row.with_cell(text: application.offered_at&.to_fs(:day_month_year))
```
starswan, line 4: same here with offered_at

app/views/publishers/vacancies/job_applications/index.html.slim
```diff
@@ -5,38 +5,21 @@
 .govuk-grid-row
   .govuk-grid-column-full
     - if @vacancy.within_data_access_period?
-      - applications = @job_applications.group_by(&:status)
-      - new_ones = applications.fetch("submitted", []) + applications.fetch("reviewed", [])
-      - rejected = applications.fetch("unsuccessful", [])
-      - shortlisted = applications.fetch("shortlisted", [])
-      - interviewing = applications.fetch("interviewing", [])
```
starswan, line 12: I quite liked this logic being here rather than in the model. It should be simple enough - in fact I'm not sure why we didn't just use group_by here ;-)

app/views/publishers/vacancies/job_applications/index.html.slim
```diff
@@ -5,38 +5,21 @@
...
+          - count = (status == :offered ? categories[:offered].count + categories[:declined].count : categories[status].count)
+          - tabs.with_tab(label: t(".tab.#{status}.label", count:), id:)
+            - if categories[status].exists?
+              = render "candidates", form: @form, vacancy: @vacancy, candidates: categories[status], heading: t(".tab.#{status}.heading"), multi_select: status, origin: status
+            - else
+              = render EmptySectionComponent.new title: t(".tab.#{status}.none")
+            - if status == :offered && categories[:declined].exists?
+              = govuk_section_break(visible: true, size: "xl")
+              = render "candidates", form: @form, vacancy: @vacancy, candidates: categories[:declined], heading: t(".tab.declined.heading"), multi_select: :declined, origin: :declined
+
```
starswan, line 22: IMHO the previous version was easier to read.

app/controllers/publishers/vacancies/job_applications_controller.rb
```diff
@@ -78,6 +79,19 @@ def pre_interview_checks
...
+                          .permit(:origin, :offered_at, :declined_at, :status, job_applications: [])
+    form_params[:job_applications] = vacancy.job_applications.where(id: Array(form_params[:job_applications]).compact_blank)
+    @form = Publishers::JobApplication::OfferDeclineDateForm.new(form_params)
+    if @form.valid?
+      vacancy.job_applications.where(id: @form.job_applications).find_each { it.update!(@form.attributes) }
+      redirect_to organisation_job_job_applications_path(vacancy.id, anchor: @form.origin)
+    else
+      render(@form.status == "offered" ? "offered_date" : "declined_date")
+    end
+  end
```
starswan, line 93: almost wondering if this code should be somewhere else? 'update_all' doesn't feel like a good name - were updating offered/declined dates. Maybe it needs its own controlller?

app/form_models/publishers/job_application/tag_form.rb
```diff
@@ -1,8 +1,30 @@
...
   include ActiveModel::Model
   include ActiveModel::Validations
+  include ActiveModel::Attributes
 
-  attr_accessor :job_applications
+  attribute :status, :string
+  attr_accessor :job_applications, :origin
 
   validates_length_of :job_applications, minimum: 1
+  validates_presence_of :status, on: :update_tag
```
starswan, line 10: I don't understand why we would ever not validate the presence of status? We don't currently (which is a slight oversight as we always provide it) but only doing it 'sometimes' feels very odd

app/views/publishers/vacancies/job_applications/index.html.slim
```diff
@@ -5,38 +5,10 @@
...
-          - else
-            = render EmptySectionComponent.new title: t(".no_shortlisted")
-        - tabs.with_tab(label: "Interviewing (#{interviewing.size})", id: "interviewing")
-          - if interviewing.any?
-            = render "candidates", form: @form, vacancy: @vacancy, candidates: interviewing, heading: "Interviewing Applications", multi_select: "inter", origin: :interviewing
-          - else
-            = render EmptySectionComponent.new title: t(".no_interviewing")
+        - @form.tabs.each do |tab_name, count|
+          - tabs.with_tab(label: t(".tab.#{tab_name}.label", count:), id: tab_name)
+            = render "tab_panel", form: @form, vacancy: @vacancy, tab_name:
```
starswan, line 11: I'm not sure why this is an improvement? Each tab is potentially different, so this feels like a backward step. Adding the 'job offered' tab at the bottom of this feels like a better solution

spec/views/publishers/vacancies/job_applications/index.html.slim_spec.rb
jriga: the job_applications are retrieved with an active relation hence the change

PR #7877, opened 2025-06-26
Send emails for drafts to jobseeker account email

scruti:
Fixes both [this](https://teaching-vacancies.sentry.io/issues/6674552557/?environment=production&project=6212514&query=level%3Aerror&referrer=issue-stream&stream_index=1) and [this](https://teaching-vacancies.sentry.io/issues/6676724725/?environment=production&project=6212514&query=level%3Aerror&referrer=issue-stream&stream_index=0) Sentry errors.

We are getting errors on two mailers:
- The reminder emails about draft job applications. 
- The alerts to jobseekers with existing job application drafts for vacancies closed early by the publisher.

These errors are caused by the Job Application email address being missing on a subset of draft applications.

This happens when someone starts a job application draft, but the personal details sections haven't been imported from a prior application and haven't been filled out by the applicant.

There is a second issue using the personal details step email address: It may differ from the user account's email (EG: Users provide specific email addresses to receive job application communications from the school).

As a service, when communicating: "You may want to finish this draft before the vacancy closing date" or "You cannot longer appply to this vacancy", we are talking to the service user, not to the "applicant".

Switching to send this email to the jobseeker's email address, we achieve:
- Email will always be present, even when the job application hasn't filled the personal details section, avoiding errors/further filtering.
- The emails will be sent to the email associated to the user account, which seems more appropriate in this particular use case.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/mailers/jobseekers/vacancy_mailer.rb
+++ b/app/mailers/jobseekers/vacancy_mailer.rb
@@ -1,9 +1,10 @@
 class Jobseekers::VacancyMailer < Jobseekers::BaseMailer
   def draft_application_only(job_application)
     @job_application = job_application
+    @jobseeker = job_application.jobseeker
     @vacancy = job_application.vacancy
 
-    send_email(to: job_application.email_address,
+    send_email(to: @jobseeker.email,
                subject: I18n.t("jobseekers.vacancy_mailer.draft_application_only.subject", date: @vacancy.expires_at.to_date, job_title: job_application.vacancy.job_title))
   end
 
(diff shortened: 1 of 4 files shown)


app/mailers/jobseekers/vacancy_mailer.rb
```diff
@@ -1,9 +1,10 @@
 class Jobseekers::VacancyMailer < Jobseekers::BaseMailer
   def draft_application_only(job_application)
     @job_application = job_application
+    @jobseeker = job_application.jobseeker
     @vacancy = job_application.vacancy
 
-    send_email(to: job_application.email_address,
```
starswan, line 6: I think we need to do this for 'closed_early_ emails too - they seem to be the main source of the sentry error

starswan (approved):

PR #7895, opened 2025-07-04
Fix notifications spec

starswan:
## Changes in this PR:

The 'notifications' spec has been failing intermittently for a long time now. This is caused by the noticed library using 'insert_all' to create its notifications - this used the DB timestamp for created_at and updated_at (for efficiency) and hence they are all set to the same time during tests no matter how much time travelling we attempt.

This removed the 'created_at' delegation (which has been wrong for some time, it should have been in the 'noticed_methods' block) and replaces it with updating the notification records timestamp (in the test) to match the event (there is 1 notification for each recipient of an event)


scruti (approved):
Good catch!

PR #7898, opened 2025-07-04
Add `govuk-template--rebranded` to app layouts

peteryates:
The rebrand is currently half appplied because govuk-components was upgraded but the rebrand class wasn't added, so it's displaying a cyan dot on a black background instead of the blue one.

## Before 

image

## After

Screenshot From 2025-07-04 16-45-02


scruti:
Closing this as was superseeded by https://github.com/DFE-Digital/teaching-vacancies/pull/7968

Thanks @peteryates :bow:

PR #7942, opened 2025-07-25
Refactor publisher vacancies job application tag

jriga:
## Trello card URL

## Changes in this PR:

This is extracted from #7876 as a basis for changes coming to that part of the service.
No functional change is added at this stage.
Second PR #7943 contains the functional changes

* Added test coverage around job application status rendering both publisher and jobseeker
* Added test coverage around job application status transitions
* Fix little bug when no status is select on the tag page

Some spec are pending and commented out for now as the PR #7918 is yet to be merged


spec/form_models/publishers/job_application/tag_form_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      describe "validation" do
+        context "when validate status true" do
+          let(:validate_status) { true }
+
+          it { is_expected.to validate_length_of(:job_applications) }
+          it { is_expected.to validate_presence_of(:status) }
+          it { is_expected.to validate_inclusion_of(:status).in_array(%w[submitted unsuccessful reviewed shortlisted interviewing]) }
+        end
+
+        context "when validate status false" do
```
scruti, line 21: Would it make sense to test that the status presence/inclusion is not validated in this case?

app/form_models/publishers/job_application/tag_form.rb
```diff
@@ -1,8 +1,14 @@
...
 
-  attr_accessor :job_applications
+  VALID_STATUSES = %w[submitted unsuccessful reviewed shortlisted interviewing].freeze
+
+  attribute :status, :string
+  attr_accessor :job_applications, :origin, :validate_status
 
   validates_length_of :job_applications, minimum: 1
+  validates_presence_of :status, if: -> { validate_status }
+  validates_inclusion_of :status, in: VALID_STATUSES, if: -> { status.present? }
```
KyleMacPherson, line 13: If validate_status is false then would this allow us to set status to nil? Is that ok?

spec/form_models/publishers/job_application/tag_form_spec.rb
jriga: when the value of  ``validate_status``   evaluates to  truthy,  the validation occurs. 
I can rename the contexts to be more explicit.

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -0,0 +1,365 @@
...
+
+  describe "transition: no job application to draft" do
+    it "jobseeker can start a job application" do
+      run_with_jobseeker(jobseeker) do
+        #
+        # jobseeker starts an application
+        #
+        jobseeker_application_start_page.load(vacancy_id: vacancy.id)
+        jobseeker_application_start_page.btn_start_application.click
+        jobseeker.reload
```
scruti, line 45: Would be worth to assert on the `jobseeker.job_applications.draft.count` increasing by 1 on this click?

This or any other way to check that when visiting the Jobseeker Applications page there is a new draft there.

spec/form_models/publishers/job_application/tag_form_spec.rb
scruti: Yes got that bit, but as I see the both contexts, the tests are not covering that the presence & inclusion of the `status` is **only** validated when the `validate_status` is truth.

The `status` validation could also be happening when `validate_status: false` and the current tests would still pass. 

Does it make sense? Not sure I'm explaining myself very well.

app/form_models/publishers/job_application/tag_form.rb
jriga: yes that's fine as when we are on the tag page we don't know which status will be pick so it can be null.
when we perform the update_tag action the status has to be present and validated

app/views/publishers/vacancies/job_applications/tag.html.slim
```diff
@@ -11,20 +11,20 @@
...
+  - @form.job_applications.each do |ja|
     = f.hidden_field :job_applications, multiple: true, value: ja.id
 
-  = f.govuk_radio_buttons_fieldset :status do
-    = f.govuk_radio_button :status, :submitted
-    = f.govuk_radio_button :status, :unsuccessful
-    = f.govuk_radio_button :status, :reviewed
-    = f.govuk_radio_button :status, :shortlisted
+  = f.govuk_radio_buttons_fieldset :status, legend: nil do
+    - @form.class::VALID_STATUSES[..-2].each do |status|
```
KyleMacPherson, line 24: [Minor] I think `VALID_STATUSES.excluding("interviewing")` might be a bit easier to read/understand

spec/form_models/publishers/job_application/tag_form_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    RSpec.describe TagForm, type: :model do
+      subject(:tag_form) { described_class.new(job_applications:, status:, origin:, validate_status:) }
+
+      let(:job_applications) { create_list(:job_application, 2, :status_submitted) }
+      let(:status) { "shortlisted" }
+      let(:origin) { "submitted" }
+
+      describe "validation" do
+        context "when validate_status evalutes to truthy" do
+          let(:validate_status) { "false" }
```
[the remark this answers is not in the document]
scruti: Hey now the code proves what I was saying! 😅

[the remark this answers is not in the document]
scruti: I was shamefully removing my comment. Ok now this is even more confusing. 

I just wanted to assert that when the `validate_status` is `false`, the status validation doesn't apply 😿

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -0,0 +1,365 @@
...
+        expect(job_application.reload.status).to eq("reviewed")
+      end
+    end
+  end
+
+  describe "transition: reviewed to unsuccessful", :js do
+    let(:status) { "reviewed" }
+
+    it "jobseeker and publisher can view job application" do
+      run_with_jobseeker(jobseeker) do
```
KyleMacPherson, line 156: This block of code is run a 3 times (I think) in this spec, I think we could skip it here and in the next test so we only assert it once and maybe reduce the time it takes to run this spec a little bit

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -0,0 +1,365 @@
...
+        expect(display_status).to have_text("reviewed")
+
+        publisher_ats_applications_page.update_status(job_application) do |tag_page|
+          tag_page.select_and_submit("shortlisted")
+        end
+
+        expect(publisher_ats_applications_page.tab_panel.job_applications).to be_empty
+
+        #
+        # display not considering tab
```
KyleMacPherson, line 265: display shortlisted tab

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -0,0 +1,365 @@
...
+        expect(jobseeker_application_page.tag).to have_text("submitted") # TODO: should the jobseeker view reviewed instead of submitted here
+        expect(job_application.reload.status).to eq("reviewed")
+      end
+    end
+  end
+
+  describe "transition: reviewed to unsuccessful", :js do
+    let(:status) { "reviewed" }
+
+    it "jobseeker and publisher can view job application" do
```
KyleMacPherson, line 155: [Minor] We could change all of these test descriptions to describe what the publisher should see and what the jobseeker should see. For instance "publisher sees unsuccessful and jobseeker sees rejected tag" or something like that

scruti (approved):
LGTM!

spec/system/job_applications/status_transition_spec.rb
scruti, line 155:
```suggestion
    it "allows the publisher to reject a submitted job application and the jobseeker to see it as unsuccessful afterwards" do
```

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -0,0 +1,365 @@
...
+        expect(jobseeker_application_page).to be_displayed(id: job_application.id)
+        expect(jobseeker_application_page.tag).to have_text("unsuccessful")
+      end
+    end
+  end
+
+  describe "transition: reviewed to shortlisted", :js do
+    let(:status) { "reviewed" }
+
+    it "jobseeker and publisher can view job application" do
```
scruti, line 224:
```suggestion
    it "allows the publisher to shortlist a submitted job application and the jobseeker to see it as shortlisted afterwards" do
```

PR #7943, opened 2025-07-25
Job offered tab

jriga:
## Trello card URL

## Changes in this PR:
This is functional changes extracted from #7876.

* Remove publisher vacancies` tab all`
* Add `tab Job offered`

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


spec/system/job_applications/status_transition_spec.rb
```diff
@@ -194,50 +182,134 @@
...
+        expect(jobseeker_application_page.tag).to have_text("shortlisted")
+      end
 
-    it "allows the publisher to shortlist a submitted job application and the jobseeker to see it as shortlisted afterwards" do
       run_with_publisher(publisher) do
         publisher_ats_applications_page.load(vacancy_id: vacancy.id)
 
         #
-        # check tab all
+        # check tab new
```
scruti, line 217:
```suggestion
        # check shortlisted tab
```

spec/system/job_applications/status_transition_spec.rb
```diff
@@ -273,44 +346,35 @@
...
-        # without ATS references and self-disclosure collection through TV service
-        #
-        expect(publisher_ats_collect_references_page).to be_displayed(vacancy_id: vacancy.id)
-        publisher_ats_collect_references_page.answer_no
+        publisher_ats_applications_page.select_tab(:tab_interviewing)
+        expect(publisher_ats_applications_page.tab_panel.job_applications).to be_empty
 
         #
-        # display interviewing tab
+        # display offered tab
```
scruti, line 372:
```suggestion
        # display unsuccessful tab
```

spec/services/export_candidate_data_service_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    job_references
+    self_disclosure
+  end
+
+  describe ".call" do
+    it "creates new instance and calls export" do
+      expect(described_class).to receive(:new).with(job_applications).and_call_original
+      expect_any_instance_of(described_class).to receive(:export) # rubocop:disable RSpec/AnyInstance
+      described_class.call(job_applications)
+    end
```
scruti, line 46: Suggestion: Given that this requires the use of `expect_any_instance_of`. Couldn't we replace it with an explicit stub instead? It would still test that a new instance is created and that receives the `export` call.

```ruby
it "creates new instance and calls export" do
  service_instance = instance_double(described_class)
  expect(described_class).to receive(:new).with(job_applications).and_return(service_instance)
  expect(service_instance).to receive(:export)
  described_class.call(job_applications)
end
```

spec/factories/job_applications.rb
```diff
@@ -202,4 +202,27 @@
...
+  end
+
+  trait :status_declined do
+    transient do
+      submitted_at { 5.days.ago }
+      shortlisted_at { 4.days.ago }
+      interviewing_at { 2.days.ago }
+      offered_at { 1.day.ago }
+      declined_at { Time.zone.now }
+    end
```
scruti, line 224: ❓Question

Was there any need/benefit to define these attributes as `transient` ?. If I recall right, they map to existing DB fields.

app/models/job_application.rb
```diff
@@ -1,5 +1,5 @@
 class JobApplication < ApplicationRecord
-  before_save :update_status_timestamp, if: :will_save_change_to_status?
+  before_save :update_status_timestamp, if: %i[will_save_change_to_status? ignore_for_offered_and_declined?]
```
scruti, line 2: It took me a while to understand the purpose of `ignore_for_offered_and_declined?`.
So maybe we could add a comment here explaining that these two statuses' timestamps are set by manual input recording through forms instead of the automated timestamp. 
Thinking of someone stumbling into this bit of code with even less context than I (not having reviewed the PR introducing all this, but just landing on this file/behaviour).

scruti (approved):
LGTM! (left some minor comments for your consideration but no blockers)

Great job managing so many individual features into a single branch without going insane. 

Loved the addition of a new component for the tab panels!

PR #7960, opened 2025-08-01
Use referee presenter to render show page

jriga:
## Changes in this PR:

Following ticket [2027](https://trello.com/c/1vRjlE2I) where we generate the reference pdf
We can use the referee presenter to render the show html as well.

## Depends on 
#7943


scruti (approved):
Nice!

PR #7969, opened 2025-08-11
Statistics dashboard

starswan:
## Trello card URL

https://trello.com/c/q73sxE8U/1989-build-statistics-dashboard

## Changes in this PR:

The 'job listings' tab creates a 'publisher preference' if one doesn't exist (which is used by the query) - not sure what the correct behaviour should be 

### After
EqualOps

### DIFF
--- a/spec/requests/publishers/all_time_statistics_spec.rb
+++ b/spec/requests/publishers/all_time_statistics_spec.rb
@@ -18,22 +18,22 @@ RSpec.describe "Publisher all-time statistics" do
     before do
       create(:vacancy, publisher: publisher, organisations: [school], publish_on: 1.year.ago,
                        vacancy_analytics: build(:vacancy_analytics,
-                                                referrer_counts: { "Google" => 6, "Magic" => 8 }))
+                                                referrer_counts: { "google" => 6, "magic" => 8 }))
       create(:vacancy, publisher: publisher, organisations: [school],
                        vacancy_analytics: build(:vacancy_analytics,
-                                                referrer_counts: { "Google" => 6, "Magic" => 8 }))
+                                                referrer_counts: { "google" => 6, "magic" => 8 }))
       create(:vacancy, publisher: publisher, organisations: [school],
                        vacancy_analytics: build(:vacancy_analytics,
-                                                referrer_counts: { "Google" => 4, "Yahoo" => 12, "LinkedIn" => 13 }))
+                                                referrer_counts: { "google" => 4, "yahoo" => 12, "linkedin" => 13 }))
       create(:vacancy, publisher: publisher, organisations: [school],
                        vacancy_analytics: build(:vacancy_analytics,
-                                                referrer_counts: { "Google" => 14, "Indeed" => 5, "LinkedIn" => 15 }))
+                                                referrer_counts: { "google" => 14, "indeed" => 5, "linkedin" => 15 }))
     end
 
     it "includes old data in its calculations" do
       get(publishers_all_time_statistics_path(format: :csv))
 
-      expect(response.body.split("\n")).to eq(["Google,LinkedIn,Magic,Yahoo,Indeed", "30,28,16,12,5"])
+      expect(response.body.split("\n")).to eq(["Google,Linkedin,Magic,Yahoo,Indeed", "30,28,16,12,5"])
     end
   end
 


spec/system/publishers/publishers_can_view_statistics_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      within("#analytics") do
+        within(".govuk-summary-list__row:nth-child(1)") do
+          expect(page).to have_content("LinkedIn")
+          expect(page).to have_content("30")
+        end
+        within(".govuk-summary-list__row:nth-child(2)") do
+          expect(page).to have_content("Google")
+          expect(page).to have_content("24")
+        end
```
scruti, line 45: Shouldn't `Magic`, `Yahoo` and `Indeed` numbers also be asserted against here?

spec/system/publishers/publishers_can_view_statistics_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+    end
+  end
+
+  describe "equal opportunities page" do
+    before do
```
scruti, line 51: It seems to be a lot of test setup info in the `equal_opportunities_report` instances that doesn't seem to be used for any assertion. 

Don't know if it was forgotten or just want to keep the assertions lean. In the second case, I guess we could just not set up the data we don't assert against?

spec/requests/publishers/statistics_spec.rb
```diff
@@ -0,0 +1,98 @@
+require "rails_helper"
+
+RSpec.describe "Publisher statistics" do
```
scruti, line 3:
```suggestion
RSpec.describe "Publisher all time statistics" do
```

And maybe we could rename the file to `all_time_statistics_spec.rb` too?

spec/models/vacancy_spec.rb
```diff
@@ -332,6 +332,41 @@
     let(:expired_earlier_today) { create(:vacancy, expires_at: 5.hour.ago) }
     let(:expires_later_today) { create(:vacancy, expires_at: 1.hour.from_now) }
 
+    describe "#active_in_current_academic_year" do
```
scruti, line 335:
```suggestion
    describe ".active_in_current_academic_year" do
```

spec/models/vacancy_spec.rb
```diff
@@ -332,6 +332,41 @@
...
+        before do
+          travel_to Date.current.beginning_of_year.months_since(10)
+        end
+
+        it "finds current aug/sept" do
+          expect(Vacancy.active_in_current_academic_year.map(&:job_title)).to contain_exactly("expired_this_august", "expired_this_september")
+        end
+      end
+
+      context "when bwtween January and August" do
```
scruti, line 359:
```suggestion
      context "when between January and August" do
```

spec/models/vacancy_spec.rb
```diff
@@ -332,6 +332,41 @@
...
+        before do
+          travel_to Date.current.end_of_year.months_since(6)
```
scruti, line 361: Question:
Wouldn't it be clearer to travel to `Date.current.beginning_of_year.months_since(6)` and assert `contain_exactly("expired_last_august", "expired_last_september")`?

Is this not possible due to the `before` creating the vacancies before stubbing the current `Date` ?

app/models/vacancy.rb
```diff
@@ -132,6 +132,19 @@ class Vacancy < ApplicationRecord
...
   scope :search_by_full_text, ->(query) { vacancy_full_text_search_query(query) }
 
+  scope :active_in_current_academic_year, lambda {
+    august_31st = Date.current.end_of_year.months_ago(4)
+    academic_year_end = if Date.current >= august_31st
+                          august_31st + 1.year
+                        else
+                          august_31st
+                        end
+    academic_year_start = academic_year_end - 1.year
```
scruti, line 142: Isn't this off by one day? 
This will set the academic year start to `August 31st` and, according to the ticket, the start date should be `September 1st`.

app/controllers/publishers/statistics_controller.rb
```diff
@@ -0,0 +1,39 @@
+module Publishers
+  class StatisticsController < BaseController
```
scruti, line 2: Non-blocking suggestion:

I wonder if this logic would be better encapsulated as a Presenter for a collection of `PublishedVacancy` rather than as an abstract controller.

`PublisherVacancyStatisticsPresenter` ?

My reasoning behind this:
- This controller whole purpose seems to be sharing logic across the other 2 child controllers (what originally made me wonder if it could be shaped as a concern instead).
-  The majority of the logic here (besides `#vacancies` that would be the bit remaining in the controller) is dedicated to aggregating and massaging the `vacancies` collection data for display. What makes me lean towards a Presenter.
- Having its own Presenter class would allow for encapsulation of the private methods (EG: `#filter_reports_by` , `#merge_sum`) and defining and unit testing of the public interface `#equal_opportunities_data`, `#listing_data`.

app/views/layouts/_publisher_secondary_nav.html.slim
```diff
@@ -1,6 +1,6 @@
...
-    .govuk-width-container
-      = tabs html_attributes: { "aria-label": "publisher nav", id: "publisher-nav" } do |tabs|
-        - tabs.with_navigation_item text: "Job listings", link: organisation_jobs_with_type_path
-        - tabs.with_navigation_item text: "Candidate profiles", link: publishers_jobseeker_profiles_path
+.sub-navigation
+  .govuk-width-container
+    = tabs html_attributes: { "aria-label": "publisher nav", id: "publisher-nav" } do |tabs|
+      - tabs.with_navigation_item text: "Job listings", link: organisation_jobs_with_type_path
+      - tabs.with_navigation_item text: "Candidate profiles", link: publishers_jobseeker_profiles_path
+      - tabs.with_navigation_item text: "Statistics", link: publishers_current_year_statistics_path
```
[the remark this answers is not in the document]
starswan: There's a weird piece of code in job listings controller which I didn't replicate,which creates a default publisher preference record. I'm not really sure if it should be there or not. The review app could just create the preference in the seeds instead

spec/models/vacancy_spec.rb
scruti: > ruby convention is # for methods, not . ?

Instance methods yes, for class methods is `.`  AFAIK

app/controllers/publishers/statistics_controller.rb
starswan: Yes that worked really well. Uncovered a couple of missing formatting issues as well

PR #7971, opened 2025-08-11
Publishers ATS API: External vacancies validations fixes and refactors

scruti:
## Trello ticket
- https://trello.com/c/Ss1f1cWi/2071-allow-ats-api-clients-to-replublish-deleted-vacancies

## Changes in this PR:

Work based on the feedback from one of our clients when testing their integration with our API:

- We need to allow publishers to publish a Vacancy with the same `external_reference` as other vacancy/es that has been previously deleted by them.

Our current validation and DB unique index were impeding this. As was ensuring uniqueness even against deleted vacancies.

- Refactor: As a part of the work, we moved validations that only happen on external vacancies to the `ExternalVacancyValidator`.

### Data & Schema Changes

- [x] Adds/removes model validations

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings
- [X] Publisher ATS API (may require mapping updates or API versioning)

Could this change impact:
</details>


spec/services/publishers/ats_api/update_vacancy_service_spec.rb
```diff
@@ -221,7 +225,7 @@
   end
 
   context "when a vacancy with the same external reference exists" do
-    let!(:existing_vacancy) do
+    let(:existing_vacancy) do
```
starswan, line 228: personally I prefer the existing pattern - IMHO it's less prone to confusion during code changes. I actually fixed a bug recently where the test had a let w/o a bang and no 'phantom reference' so the test passed even though the code was incorrect because the side-effect didn't take place. I have used the 'create and find' pattern in the past which is maybe clearer?

app/validators/external_vacancy_validator.rb
```diff
@@ -8,12 +8,29 @@ def validate(record)
     )
 
     record.errors.add(:organisations, "No school(s) associated with vacancy") if record.organisations.empty?
+    validate_uniqueness_per_client(record)
+    validate_no_duplicate_vacancy(record)
     validate_job_title_length(record)
-    validate_expiry_date(record) if record.expires_at.present?
```
starswan, line 12: I prefer the original pattern rather than the short-circuit in the method - feels clearer that we only validate the expiry date if it exists, rather than the short-return (which I thought we agreed not to use on small methods) on the method itself

app/models/vacancy.rb
```diff
@@ -131,15 +131,6 @@ class Vacancy < ApplicationRecord
   extend VacancyFulTextSearchQueryModule
   scope :search_by_full_text, ->(query) { vacancy_full_text_search_query(query) }
 
-  # effectively we have two different types of vacancy - external ones and internal ones
```
starswan, line 134: It seems a shame to lose this comment and the idea - IMHO ExternalVacancy should be a separate type, as it has vastly different validation rules (and even different fields)

starswan (approved):
Left a few non-blocking comments

spec/services/publishers/ats_api/update_vacancy_service_spec.rb
scruti: I also favour the `let!` to be fair. I only did this because I needed to redefine the `existing_vacancy` in an inner context and `let!` wasn't allowing it.

Re-visiting this, I think this is just a sign that the `contexts` for deleted/non-deleted should be defined in parallel, not one inside another. 

Going to fix it 👷‍♂️

app/validators/external_vacancy_validator.rb
scruti: I treated it like any other method, where I would guard it against giving the presence of the `expires_at` for granted could cause 
```ruby
nil < Time.zone.today
NoMethodError: undefined method '<' for nil (NoMethodError)
```
But you're right, in a validation context where we want to apply a validation, the original form was clearer about when it runs.

PR #7996, opened 2025-08-19
Fix specs after state machine

jriga:
## Trello card URL
https://trello.com/c/usyCTDSW/2089-adjust-test-to-new-job-application-state-machine

## Changes in this PR:

Adjust specs for JobApplication status state machine

### DIFF
--- a/spec/factories/reference_requests.rb
+++ b/spec/factories/reference_requests.rb
@@ -1,0 +1,34 @@
+FactoryBot.define do
+  factory :reference_request do
+    referee
+    token { SecureRandom.uuid }
+    status { :created }
+
+    after(:stub) do |request|
+      request.email = request.referee&.email
+    end
+
+    after(:build) do |request|
+      request.email = request.referee&.email
+    end
+  end
+
+  trait :reference_requested do
+    status { :requested }
+
+    after(:create) do |request|
+      request.created!
+      request.requested!
+    end
+  end
+
+  trait :reference_received do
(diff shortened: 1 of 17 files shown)


spec/factories/reference_requests.rb
```diff
@@ -0,0 +1,34 @@
...
+      request.email = request.referee&.email
+    end
+  end
+
+  trait :reference_requested do
+    status { :requested }
+
+    after(:create) do |request|
+      request.created!
+      request.requested!
```
scruti, line 21: Question: Why are we doing this chain of status update writes in DB? These calls are so explicit that there must be a need for it, but I am completely missing it 😅 

I see the downside of the multiple DB writes that will happen each time one of these factories is created. But there must be an upside here! 🤔

jriga: This is a way to create the data used by the timeline component.
That trigger the model callback that papertrail adds to the model to create differents `versions` entries

starswan: IMHO the create should just be a create - there might be a trait which calls requested! (the created! call appears to be unnecessary, as that is the default state on line 5). Otherwise the tests result in different data from the code flows

starswan (approved):
1 tiny comment, would be good to fix

spec/factories/reference_requests.rb
jriga: :recycle:

PR #7997, opened 2025-08-19
Allow jobseekers to message publishers back

KyleMacPherson:
## Changes in this PR:

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/assets/stylesheets/actiontext.css
```diff
@@ -29,3 +29,9 @@
   padding: 0 !important;
   max-width: 100% !important;
 }
+
+/* Hide attachment button in Trix toolbar */
+trix-toolbar .trix-button--icon-attach {
```
starswan, line 34: Think we need a better comment - we're doing this for a reason, hopefully its clear that we're hiding the button ;-) I guess this is temporary as don't support uploads yet?

app/controllers/jobseekers/job_applications_controller.rb
```diff
@@ -110,6 +110,12 @@ def post_submit
 
   def show
     raise ActionController::RoutingError, "Cannot view draft application" if job_application.draft?
+
+    if params[:tab] == "messages"
+      conversation = job_application.conversations.first
+      @messages = conversation&.messages&.order(created_at: :desc) || []
```
starswan, line 116: could we not just display all messages here and so avoid this safe-nav operators? (which look scary IMHO)

app/views/jobseekers/job_applications/_show.html.slim
```diff
@@ -1,43 +1,50 @@
-= job_application_review(job_application, allow_edit: local_assigns[:allow_edit]) do |r|
-  - render "jobseekers/job_applications/job_application_review_sections", r: r, jobseeker: current_jobseeker
-  - r.with_above do
-    = yield
+= yield
```
starswan, line 1: this looks a very strange change - it seems to alter the with_above section for the current scenario. Was that the intention? Not sure about this stray 'yield'  statement on the first line. Maybe this code would be better inserted at a higher level rather than in this partial

app/views/publishers/vacancies/job_applications/_messages.html.slim
```diff
@@ -4,30 +4,22 @@
...
-          = f.govuk_error_summary
-
-          .govuk-form-group
-            = f.rich_text_area :content
-
-          = f.govuk_submit t(".send_message")
-          = govuk_link_to t(".cancel"), request.original_url, class: "govuk-!-margin-left-3"
-    - else
+    - unless params[:show_form] == "true"
+      / Inset text (only show when form is not visible)
```
starswan, line 15: This seems to undo a deliberate 'only check show_form once' idiom, so not sure why this has changed?

app/views/jobseekers/job_applications/_messages.html.slim
```diff
@@ -0,0 +1,22 @@
+.govuk-grid-row
+  .govuk-grid-column-full
+    - if @messages.any?
```
starswan, line 3: unsure about putting this logic here (don't allow new messages unless we have any). Think this should be in the model layer somehow, it's pretty obscure IMHO in this location

app/views/publishers/vacancies/job_applications/_messages.html.slim
```diff
@@ -4,30 +4,22 @@
       = govuk_button_link_to t(".print_page"), "#", class: "govuk-button--secondary", "data-controller": "utils", "data-action": "click->utils#print" unless params[:show_form] == "true"
       = govuk_button_link_to t(".send_message_to_candidate"), url_for(params.to_unsafe_h.merge(show_form: "true")), class: "govuk-button--secondary govuk-!-margin-left-3" unless params[:show_form] == "true"
 
-    - if params[:show_form] == "true"
-      div class="govuk-!-margin-top-6"
-        = govuk_details summary_text: t(".how_will_message_be_sent") do
-          = t(".message_sent_description")
```
starswan, line 10: did we mean to change the behaviour here? This boolean used to be required in order for the form to be shown

app/views/shared/_message_form.html.slim
```diff
@@ -0,0 +1,13 @@
+- if params[:show_form] == "true"
```
starswan, line 1: really unsure about this as a structure. We're rending a partial that may end up with nothing being rendered. Previous logic was better (just an if/else), this doesn't feel right (even more so as its a 'global' rather than a parameter)

spec/requests/publishers/vacancies/job_applications/messages_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  before do
+    sign_in(publisher, scope: :publisher)
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(Publishers::BaseController).to receive(:current_organisation).and_return(organisation)
+    # rubocop:enable RSpec/AnyInstance
+  end
+
+  after { sign_out publisher }
+
+  describe "POST /organisation/jobs/:vacancy_id/job_applications/:job_application_id/messages" do
```
starswan, line 18: I'm really unsure about creating these request specs. If we have coverage for functional specs, do we need these? Controllers should not do very much, so this feels like we're 'over-testing' .

spec/system/publishers/publishers_can_send_messages_to_applicants_spec.rb
```diff
@@ -67,4 +68,75 @@
...
+      click_button "Send message"
+
+      expect(page).to have_text("Message sent successfully")
+      expect(page).to have_text("Hello from publisher")
+      expect(page).to have_text(jobseeker_reply)
+      expect(page).to have_text(job_application.name.to_s)
+      expect(conversation.reload.messages.count).to eq(2)
+    end
+
+    it "shows jobseeker reply in publisher interface" do
```
starswan, line 121: This feels  like it should be a view spec rather than a functional one

app/controllers/publishers/vacancies/job_applications/messages_controller.rb
```diff
@@ -2,26 +2,37 @@ class Publishers::Vacancies::JobApplications::MessagesController < Publishers::V
...
+      render "publishers/vacancies/job_applications/show", status: :unprocessable_entity
     end
   end
 
   private
 
   def find_or_create_conversation
     @job_application.conversations.first ||
-      @job_application.conversations.build(title: Conversation.default_title_for(@job_application))
+      @job_application.conversations.create!(title: Conversation.default_title_for(@job_application))
```
starswan, line 32: this change creates a bug - it creates a conversation even if the message isn't valid.

app/assets/stylesheets/components/messages.scss
```diff
@@ -0,0 +1,12 @@
+@import 'govuk-frontend/dist/govuk/helpers/colour';
+
+// Message header styling for sender messages (blue) and recipient messages (grey)
+.message-header--blue {
+  background-color: govuk-colour('blue') !important;
+  color: govuk-colour('white') !important;
+}
+
+.message-header--grey {
```
starswan, line 9: IMHO it would be better to name this what it is (recipient_message) rather than message-header-grey

app/helpers/messages_helper.rb
```diff
@@ -0,0 +1,18 @@
+module MessagesHelper
+  def sender_is_current_user?(message, current_user_type)
+    (current_user_type == "jobseeker" && message.sender_type == "Jobseeker") ||
+      (current_user_type == "publisher" && message.sender_type == "Publisher")
+  end
```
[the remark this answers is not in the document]
KyleMacPherson: No because the organisation can have multiple publishers all of whom could be sending messages to the jobseeker on behalf of the organisation. I believe your proposed solution would show both other publishers in the organisation and the jobseeker's messages in the same colour which I don't think is the intended behaviour.

app/views/jobseekers/job_applications/_messages.html.slim
KyleMacPherson: Well this is decide whether to show a form, or to show content which tells the user that no messages exist yet so I think it's necessary in this case.

app/assets/stylesheets/components/messages.scss
KyleMacPherson: I don't think recipient_message is what this is. Every message is both a sender_message and a recipient_message as messages by their nature always have a sender and recipient. I think this name better describes what the style is.

spec/system/publishers/publishers_can_send_messages_to_applicants_spec.rb
KyleMacPherson: I think that's a fair comment but I also think for the sake of a one or two tests it's probably pragmatic to keep the related tests in a single file

app/views/shared/_message_form.html.slim
KyleMacPherson: Good point, done!

app/assets/stylesheets/components/messages.scss
```diff
@@ -0,0 +1,12 @@
+@import 'govuk-frontend/dist/govuk/helpers/colour';
+
+// Message header styling for sender messages (blue) and recipient messages (grey)
+.message-header--blue {
+  background-color: govuk-colour('blue') !important;
```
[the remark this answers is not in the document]
KyleMacPherson: My CSS skills are pretty non-existent so thanks, actually didn't really know the implications of this. 🙇  Corrected!

app/helpers/messages_helper.rb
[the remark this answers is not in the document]
starswan: don't think it does if you use current_user?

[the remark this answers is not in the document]
starswan: have a (think its nice) suggestion for you - if we introduce polymorphism into Message (PublisherMessage and JobSeekerMessage) then render collection: @messages will render '_publisher_message' and '_jobseeker_message' partials and a lot of this complexity will dissolve away...

app/views/jobseekers/job_applications/_messages.html.slim
starswan: why do we show a form based on whether they are current messages? When we have rejection messages won't that need changing? Hence I wondered if it should be a model concern rather than a view one

app/controllers/jobseekers/job_applications/messages_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Jobseekers::JobApplications::MessagesController < Jobseekers::JobApplications::BaseController
+  before_action :set_job_application
+  before_action :ensure_conversation_exists, only: [:create]
+  helper_method :vacancy
+
+  def create
+    message_form = Publishers::JobApplication::MessagesForm.new(message_form_params)
```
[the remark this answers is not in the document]
starswan: sadly there is another issue here - the form pattern doesn't work as action text is expecting certain ways of interacting with the model. Tests don't run for me locally, and IIRC this was the main reason for removing the forms in my PR.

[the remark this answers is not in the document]
starswan: I thought when I did my version that the form thing didn't work, and I am having issues running the tests locally with these changes.

app/views/jobseekers/job_applications/_show.html.slim
starswan: would it not be simpler just to include them in the template rather than yielding on line 1?

PR #8007, opened 2025-08-22
Retention Policy

jriga:
## Trello card URL
[2053](https://trello.com/c/5OUPpfRV)

## Changes in this PR:

* Add `Discard::Model` to self-disclosure and reference data models
* Set soft deletion threshold to 6 months
* Set hard deletion threshold to 5 years

* Group models that are using the same retention policy together

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/jobs/retention_policy_job.rb
```diff
@@ -0,0 +1,31 @@
...
+        y << JobReference.joins(referee: :job_application).merge(job_applications)
+        y << ReferenceRequest.joins(referee: :job_application).merge(job_applications)
+        # qualifying active job applications
+        y << job_applications
+      end
+
+      # draft job application when vacancy expired
+      y << JobApplication.joins(:vacancy).draft.where(updated_at: ...threshold).merge(Vacancy.expired)
+
+      y << Feedback.where(created_at: ...threshold) if hard_delete?
```
starswan, line 20: This looks weird - its the only reason the 'hard_delete' method is required, and IMHO it at least needs a comment. Ideally we should look to remove it so that our future selves don't go 'why is this anomaly here?'

app/jobs/retention_policy_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  def threshold
+    raise "define thresold period in subclass"
+  end
```
starswan, line 26: I'm a bit unsure about using inheritance for this - Josh Bloch says normally to use composition rather than inheritance. Actually Ruby make a potentially neat solution possible, as we have functions as parameters so this could all be in the subclass, passing parameters (including the &:discard or &:delete method). WDYT?

PR #8051, opened 2025-09-08
2111 ats requests statuses

jriga:
## Trello card URL
[2111](https://trello.com/c/n5ZXpgDd)

## Changes in this PR:
- copy change
- add common pre interviewing request statuses
- update self disclosure request code for new flow and statuses
- update reference request code for new flow and statuses
- add a rake task to update any existing pre interviewing requests

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


starswan (commented):
This looks very invasive - all we need is to allow the items to be marked as complete or declined - references already have a completed field that could be extended to cope with a third status (or have it as a tri-state nil/true/false.

app/helpers/job_references_helper.rb
```diff
@@ -1,19 +0,0 @@
...
-          "declined"
-        end
-      else
-        "pending"
-      end
-    else
-      "created"
-    end
-  end
-end
```
starswan, line 19: i don't understand why this helper has been removed - it clearly sows intent (status based on model) and has a test to go with it. Surely this should have been extended for the new cases?

app/models/concerns/interviewing_request.rb
```diff
@@ -0,0 +1,8 @@
+module InterviewingRequest
+  extend ActiveSupport::Concern
+
+  included do
+    # statuses start from 10 to avoid confusion with legacy statuses
+    enum :status, { created: 10, requested: 11, received: 12, completed: 13, declined: 14 }
+  end
```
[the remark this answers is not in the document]
starswan: IMHO this is too tiny to be a concern. I think we should do our best to avoid this sort of deep change as the code has now been released

app/form_models/publishers/vacancies/job_applications/mark_reference_as_received_form.rb
```diff
@@ -1,18 +0,0 @@
...
-        ATTRIBUTES = [:reference_satisfactory].freeze
-
-        ATTRIBUTES.each do |field|
-          attribute field, :boolean
-          validates field, inclusion: { in: [true, false], allow_nil: false }
-        end
-      end
-    end
-  end
-end
```
starswan, line 18: again I don't see the need to remove this class

app/controllers/referees/build_references_controller.rb
```diff
@@ -47,7 +47,10 @@ def update
       if @form.valid?
         @reference.update!(@form.params_to_save)
         # invalidate token after reference is complete
-        if @reference.complete?
+        if !@reference.can_give_reference?
+          @reference.mark_as_declined
```
starswan, line 51: IMHO this state transition is unnecessary - can_give_reference? already returns false if the reference is declined

app/views/publishers/vacancies/job_applications/reference_requests/show.html.slim
```diff
@@ -9,18 +9,28 @@
     h1.govuk-heading-l = @referee.job_application.name
 
     .govuk-body
-      - status = reference_request_status(@reference_request, @referee.job_reference)
+      - status = @reference_request.status
```
starswan, line 12: As we are now ignoring 'marked as complete', I am fairly certain this has a high probability of introducing a regression, especially as the test for the deleted method has been deleted

app/views/jobseekers/job_applications/self_disclosure/review/_completed.html.slim
```diff
@@ -23,3 +15,10 @@
           - if question.present?
             .govuk-body{class="govuk-!-font-weight-bold"} = question
           .govuk-body = answer
+
+- elsif @self_disclosure.request.completed?
+  .govuk-body = t(".manually_completed")
+
+- elsif @self_disclosure.request.received?
+  / received status when managed outside TVS
+  .govuk-body = t(".manually_received")
```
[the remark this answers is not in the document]
jriga: changed locale key to match request status

app/form_models/publishers/vacancies/job_applications/mark_reference_as_received_form.rb
jriga: this step has been remove in new flow

app/helpers/job_references_helper.rb
jriga: The reference request.status has the information about all the possible states making this helper unnecessary

app/models/reference_request.rb
```diff
@@ -1,26 +1,17 @@
...
   has_one :job_reference, dependent: :destroy
 
   validates :reference_id, uniqueness: true
-
-  # marked_as_complete is a seperate field as it can be done
-  # even when the request is in the 'created' state as the hiring
-  # staff can process references outside the service, but still mark
-  # the request as complete when they receive it.
-  enum :status, { created: 0, requested: 1, received: 2 }
-
```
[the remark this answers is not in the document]
jriga: `marked_as_complete` is provided by the status `completed`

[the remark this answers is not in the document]
starswan: IMHO these changes are too invasive, and there are not enough tests to demonstrate that these changes have not introduced a regression. I think a more subtle approach is required

app/helpers/job_references_helper.rb
starswan: I just don't agree with this re-write. This code (and test) were written because numerous defects were found during manual testing. Removing this feels like a retrograde step, especially as this is now live and working. I don't believe there was a need to alter the references implementation as part of this change.

PR #8052, opened 2025-09-08
Add logic for messaging permissions

KyleMacPherson:
## Trello card URL

https://trello.com/c/ozAifzRW/2119-comms-feature-state-machine

## Changes in this PR:

This PR hides messaging functionality depending on the status of the job application and the type of user that is trying to send the message (jobseeker or hiring staff). See trello ticket for full details.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/spec/models/job_application_spec.rb
+++ b/spec/models/job_application_spec.rb
@@ -320,7 +320,7 @@ RSpec.describe JobApplication do
     context "when status allows jobseeker to initiate messages" do
       %w[interviewing unsuccessful_interview offered declined].each do |allowed_status|
         context "when status is #{allowed_status}" do
-          let(:job_application) { create(:job_application, status: allowed_status) }
+          let(:job_application) { build_stubbed(:job_application, status: allowed_status) }
 
           it { is_expected.to be true }
         end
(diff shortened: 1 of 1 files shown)


spec/views/jobseekers/job_applications/_messages_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  let(:messages) { [] }
+
+  before do
+    assign(:job_application, job_application)
+    assign(:show_form, "false")
+    # rubocop:disable RSpec/VerifiedDoubles
+    model_name = double("ModelName", param_key: "publishers_job_application_messages_form")
+    errors = double("Errors", empty?: true, any?: false)
+    message_form = double("MessageForm", model_name: model_name, to_key: nil, persisted?: false, errors: errors, content: "")
+    # rubocop:enable RSpec/VerifiedDoubles
```
scruti, line 15: 🙋‍♂️ Since we're disabling a cop here, do we really need to do this? Couldn't we just use `instance_double` with a `Publishers::JobApplication::MessagesForm` instance?

Even if we need to stub its internal methods, would be nice if we could benefit of the `instance_double` check that ensures that the stubbed method exist on the real model.

But maybe there is a need?

spec/models/job_application_spec.rb
```diff
@@ -314,6 +314,110 @@
     end
   end
 
+  describe "#can_jobseeker_initiate_message?" do
+    subject { job_application.can_jobseeker_initiate_message? }
+
+    context "when status allows jobseeker to initiate messages" do
+      %w[interviewing unsuccessful_interview offered declined].each do |allowed_status|
+        context "when status is #{allowed_status}" do
+          let(:job_application) { create(:job_application, status: allowed_status) }
```
scruti, line 323: Suggestion: Use `build_stubbed` instead of `create` for these tests if possible, since we don't need the record in DB to test the method's logic and it is faster?

PR #8055, opened 2025-09-09
rename religious enums in JobApplication

starswan:
## Changes in this PR:

rename religious enums as they can be used more effectvely as method names

### DIFF
--- a/spec/system/publishers/publishers_can_select_a_job_application_for_interview_spec.rb
+++ b/spec/system/publishers/publishers_can_select_a_job_application_for_interview_spec.rb
@@ -123,7 +123,7 @@ RSpec.describe "Publishers can select a job application for interview", :perform
 
             let(:new_email) { Faker::Internet.email(domain: TEST_EMAIL_DOMAIN) }
 
-            scenario "with a good email" do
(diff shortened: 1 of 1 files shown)


spec/system/publishers/publishers_can_select_a_job_application_for_interview_spec.rb
```diff
@@ -3,341 +3,350 @@
...
             end
 
-            context "when marking reference as complete" do
-              before do
-                publisher_ats_pre_interview_checks_page.reference_links.first.click
-                click_on "Mark as received"
-              end
+            let(:new_email) { Faker::Internet.email(domain: TEST_EMAIL_DOMAIN) }
+
+            scenario "with a good email" do
```
KyleMacPherson, line 126: "with a good email" nested inside "when the referee email is incorrect" is quite confusing to me. I think it means when the referee email is incorrect but then we replace it with a valid email?

starswan: replaced with the word 'valid' instead - the other case was w/o an email at all

PR #8072, opened 2025-09-16
Messaging notifications and emails

KyleMacPherson:
## Trello card URL

https://trello.com/c/NSvyNh1m/2124-add-comms-to-jobseeker-notifications

## Changes in this PR:

This PR adds notifications/emails for messages, in bulk for publishers and for each individual message for jobseeker, and adds warning text to the messages page depending on the state of the job application.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/jobs/send_messages_received_yesterday_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  private
+
+  def publishers_with_messages_received_yesterday
+    Publisher.distinct
+             .joins(organisations: { vacancies: { job_applications: { conversations: :messages } } })
+             .where(messages: { type: "JobseekerMessage", read: false })
+             .where("DATE(messages.created_at) = ?", Date.yesterday)
```
scruti, line 19: I have run this query through an execution plan and highlighted a high cost on the messages query/filtering, as it triggers a sequential scan of the messages table:

```
Seq Scan on messages  (cost=0.00..21.38 rows=1 width=16)
  Filter: ((NOT read) AND ((type)::text = 'JobseekerMessage'::text) AND (date(created_at) = '2025-09-15'::date))
```

Image

Ran it through copilot and recommends 2 improvements:
1. Add a composite index on (type, read, created_at)
2.  Avoid wrapping `created_at` in a function (prevents use of an index on `created_at`.
The query must scan all rows to check the date after casting). Instead, use a range:
```
.where(messages: { type: "JobseekerMessage", read: false })
.where(messages: { created_at: Date.yesterday.beginning_of_day..Date.yesterday.end_of_day })
```

the above, would generate this SQL that would use the index:
```
WHERE messages.type = 'JobseekerMessage'
  AND messages.read = FALSE
  AND messages.created_at >= '2025-09-15 00:00:00'
  AND messages.created_at <= '2025-09-15 23:59:59.999999'
```

app/mailers/jobseekers/message_mailer.rb
```diff
@@ -0,0 +1,19 @@
+class Jobseekers::MessageMailer < Jobseekers::BaseMailer
+  def message_received(message)
+    @message = message
+    @jobseeker = message.conversation.job_application.jobseeker
+    @job_application = message.conversation.job_application
+    @vacancy = @job_application.vacancy
+    @organisation = @vacancy.organisation
+
+    @subject = if @job_application.status == "unsuccessful"
```
scruti, line 9: Non-blocking suggestion:
```suggestion
    @subject = if @job_application.unsuccessful?
```

app/models/jobseekers/message_received_notifier.rb
```diff
@@ -0,0 +1,58 @@
...
+      def timestamp
+        "#{day(created_at)} at #{format_time(created_at)}"
+      end
+
+      private
+
+      def unsuccessful_application?
+        job_application.status == "unsuccessful"
+      end
+
```
scruti, line 39: Non-blocking suggestion: as per above, we don't need this method.

app/jobs/send_messages_received_yesterday_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  private
+
+  def publishers_with_messages_received_yesterday
+    Publisher.distinct
+             .joins(organisations: { vacancies: { job_applications: { conversations: :messages } } })
+             .where(messages: { type: "JobseekerMessage", read: false })
+             .where(messages: { created_at: Date.yesterday.all_day })
```
scruti, line 19: 🙋‍♂️ 
As far as I understand, this is still doing a sequential scan over all the `messages` table, which can become a problem once the `messages` table grows.

I don't have the full picture of how the messages are being queried elsewhere, but if I assume this is the only bit of code doing the "search messages filtering by...", adding this partial index for unread messages would greatly improve this query (and the `count_messages_for_publisher` one) performance: 

```suggestion
add_index :messages, [:type, :created_at], where: "read = false", name: "index_messages_unread_on_type_created_at"
```

PR #8073, opened 2025-09-16
Fix document virus check tempfile download path

scruti:
## Trello card URL
- Related to: 
  - https://github.com/DFE-Digital/teaching-vacancies/pull/8069
  - https://github.com/DFE-Digital/teaching-vacancies/pull/8059
  -  
## Changes in this PR:

The document virus check process was using the project root as destination for the scanned file to be downloaded prior to be checked for safety and deleted.

This was innocuous until our infra team tried to change the permissions for our Docker images, and the file uploads started failing during the antivirus check.

Resolving it by downloading the files where they belong: The project "tmp" folder.
This folder has explicit permissions with the new Docker user permissions approach.

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/services/publishers/document_virus_check.rb
+++ b/app/services/publishers/document_virus_check.rb
@@ -21,7 +21,7 @@ class Publishers::DocumentVirusCheck
     drive_service.get_file(
       uploaded_file.id,
       acknowledge_abuse: false,
-      download_dest: uploaded_file.id.to_s,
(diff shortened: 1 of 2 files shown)


starswan (approved):
Seems fine, but would it be more idiomatic to use a ruby Tempfile (so the id just works, the file gets cleaned up by itself, and ends up in /tmp rather than Rails.root/tmp ?

PR #8089, opened 2025-09-22
change print button text

starswan:
## Trello card URL

https://trello.com/c/3iLFvFtm/2200-tweak-button-content-on-self-disclosure-and-references

## Changes in this PR:

Change print button text

### After
DownloadReference


spec/system/publishers/publishers_can_manage_self_disclosure_spec.rb
```diff
@@ -95,7 +95,6 @@
                       "#{publisher.given_name} #{publisher.family_name} - #{Date.current.to_fs.strip}"])
 
             expect(publisher_ats_self_disclosure_page.status.text).to eq("completed")
-            expect(publisher_ats_self_disclosure_page.button.text).to eq("Print self-disclosure")
```
scruti, line 98: Non-blocking: 
is this removed rather than replaced for some particular reason? Or was committed by mistake?

PR #8091, opened 2025-09-22
Message bug fixes

KyleMacPherson:
## Trello card URL

https://trello.com/c/kCARXpWf/2189-comms-bug-party-fixes

## Changes in this PR:

This PR fixes issues raised in the bug party including:

- changing the blue in the messages heading to be lighter for accessibility purposes
- moving the timestamp of message into the heading
- removing the conversation title
- adding publisher notifications when a jobseeker message is received
- changing the email subject to better reflect the intention of the email
- ordering candidate messages so that unread messages appear first
- showing the "Send message to hiring staff" button on the jobseeker side, rather than immediately rendering the message form
- allowing links in messages
- fixing links in notifications and on the candidate message page to link the user directly to the messages
- add PDF download of message history for hiring staff
- fix the cancel link

For more information see the board linked in the ticket.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/views/shared/_jobseeker_message.html.slim
```diff
@@ -3,6 +3,4 @@
     h3.govuk-summary-card__title
       = jobseeker_message_display_name(message, job_application)
   .govuk-summary-card__content
-    p.govuk-body class="govuk-!-font-weight-bold" = "Date: #{message.created_at.to_fs(:date_at_time)}"
-    p.govuk-body class="govuk-!-font-weight-bold" = message.conversation.title
-    p.govuk-body class="message-content" = simple_format(message.content)
+    p.govuk-body class="message-content" = message.content
```
scruti, line 6: Just a concern that may be overzealous: We need to be sure that the message input form is being sanitised when the user introduces it. Given that this change removed the `simple_format` sanitisacion.

spec/system/publishers/publishers_can_view_candidate_messages_spec.rb
```diff
@@ -160,4 +160,62 @@
       end
     end
   end
+
+  describe "message ordering" do
```
scruti, line 164: Given the cost of system tests. I wonder if we could assert the conversation order within the `viewing candidate messages` test above and save a specific setup and run dedicated to only test the ordering.

app/models/conversation.rb
```diff
@@ -8,20 +8,15 @@ class Conversation < ApplicationRecord
...
       .where(messages: { type: "JobseekerMessage", read: false })
       .distinct
   }
 
-  def self.default_title_for(job_application)
-    "Regarding application: #{job_application.vacancy.job_title}"
-  end
+  scope :ordered_by_unread_and_latest_message, lambda {
+    order(has_unread_messages: :desc, last_message_at: :desc)
+  }
```
scruti, line 19: As this is built now, it seems like conversations where the Publisher sent the latest message and the Jobseeker has still not read it will also be placed on top of the Publisher conversation list. Since the new `has_unread_messages` doesn't distinguish between `jobseeker` and `publisher`.

Is this a bug or a feature? 😅

app/models/message.rb
```diff
@@ -9,11 +9,20 @@ class Message < ApplicationRecord
...
   end
 
   def unread?
     !read?
   end
+
+  private
+
+  def update_conversation
+    conversation.update(last_message_at: created_at, has_unread_messages: true)
```
scruti, line 26: For data consistency, we probably want to ensure that if the conversation update fails for any reason, the message also doesn't get created. Adding a `!` into the conversation update would ensure the message doesn't get recorded unless the conversation gets updated:
 
```suggestion
    conversation.update!(last_message_at: created_at, has_unread_messages: true)
```

scruti (approved):
LGTM! Thanks for adressing all the suggestions!

PR #8115, opened 2025-09-30
Subscription alerts performance tweaks

scruti:
## Trello card URL
https://trello.com/c/2J2JST1M/2220-subscription-alert-runs-only-filter-by-location-if-there-are-any-vacancies-left-after-subscription-filters

## Changes in this PR:

[Only do subscription location filtering if needed](https://github.com/DFE-Digital/teaching-vacancies/commit/17ed3b5e4f6eb5593d7587e5bb69f855b266995f) 

The location handling was happening regardless of whether any vacancy was left after the non-location criteria have been applied.

This caused a lot of expensive computational calculations to happen for tens of thousands of subscriptions. When they should be just skipped, since the subscriptions would not have matched any vacancy by their search criteria, independently on the location of the vacancies.

Now we only run the location filtering/computations if the subscription has vacancy matches on the subscription criteria, and needs to calculate if the location criteria also match.

[Avoid reloading subscriptions from DB 1 by 1](https://github.com/DFE-Digital/teaching-vacancies/commit/6501328be0c8a4dfa4834a0718c5a76c44507fa9) 

The guarding for the subscription email being removed while the job is running results in over 100k individual DB reads.

This is caused by the `subscription.reload` call in the loop.

We can avoid this by moving that "fresh last-second from DB" check responsibility to the mailer itself. As the mailer runs asynchronously, and is already loading the subscription from DB to send the email to the subscriber.

### Before
Did a TDD approach to change this, confirming that, effectively, the location filtering calculations were being called even when not needed.
Screenshot From 2025-09-30 12-37-19

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -114,7 +114,14 @@ class Subscription < ApplicationRecord
     vacancies = scope.select do |vacancy|
       criteria.except(*JOB_ROLE_ALIASES, :location, :radius).all? { |criterion, value| FILTERS.fetch(criterion).call(vacancy, value) }
     end
-    self.class.handle_location(vacancies, criteria)
+    # Location handling is the most expensive operations, since they involve polygons or geocoding computations in DB.
+    # So do it last, and only if there are any vacancies left to filter after the other criteria have been applied over
+    # the in-memory vacancy set.
+    if vacancies.any?
+      self.class.handle_location(vacancies, criteria)
+    else
+      vacancies
+    end
   end
 
   extend DistanceHelper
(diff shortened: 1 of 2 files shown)


app/models/subscription.rb
```diff
@@ -154,16 +161,16 @@ def limit_by_location(vacancies, location, radius_in_miles)
     end
     # rubocop:enable Metrics/AbcSize
 
-    def handle_location(scope, criteria)
+    def handle_location(vacancies, criteria)
```
scruti, line 164: This is just a cognitive refactor. We're not dealing with a `scope` anymore, but an array of vacancies in memory.

KyleMacPherson (approved):

PR #8118, opened 2025-09-30
Remove vacancy legacy fields db

jriga:
## Trello card URL
https://trello.com/c/giXMS3L4/2190-forgotten-legacy-fields

## Changes in this PR:

Delete the vacancies table columns:
-     "safeguarding_information"
-     "safeguarding_information_provided" 

## Depends on PR
#8117 

Updates 
### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/vacancy.rb
+++ b/app/models/vacancy.rb
@@ -2,9 +2,6 @@ require "geocoding"
 
 # rubocop:disable Metrics/ClassLength
 class Vacancy < ApplicationRecord
-  # needed to remove legacy columns
-  self.ignored_columns += %w[safeguarding_information_provided safeguarding_information]
-
   extend FriendlyId
   extend ArrayEnum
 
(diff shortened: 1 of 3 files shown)


app/models/vacancy.rb
```diff
@@ -2,6 +2,9 @@
 
 # rubocop:disable Metrics/ClassLength
 class Vacancy < ApplicationRecord
+  # needed to remove legacy columns
+  self.ignored_columns += %w[safeguarding_information_provided safeguarding_information]
+
```
starswan, line 7: This belongs in the other PR, along with associated analytics changes. The only thing here should be the migration IMHO

db/schema.rb
```diff
@@ -821,8 +821,6 @@
     t.boolean "contact_number_provided"
     t.string "skills_and_experience"
     t.string "school_offer"
-    t.boolean "safeguarding_information_provided"
-    t.string "safeguarding_information"
```
starswan, line 825: you can remove the 2 ignored columns now we have removed them from the database

PR #8124, opened 2025-10-01
Allow MAT users to see messages for job applications at their schools

KyleMacPherson:
## Changes in this PR:

Prior to this change, only the members of the organisation for which the vacancy was published can see messages on the candidate messages page. This could cause an issue for instance because MAT staff would not be able to see messages on the candidate messages page which relate to job applications at one of their schools.

This PR fixes the issue, allowing MATs to see conversations on the candidate messages pages for jobs at all their schools.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


scruti (approved):
Left a couple of non-blocking suggestions. LGTM!

PR #8127, opened 2025-10-02
Reduce job alerts job memory usage

scruti:
## Trello card URL
- https://trello.com/c/agdJHLTG/2231-subscription-alerts-profile-the-job-run-locally-with-real-volume-of-alerts

## Changes in this PR:

Reduces the memory used by the job alert jobs by:
- Backfilling the legacy phases mapping into the subscription search criteria values instead of doing real-time mapping over the loop (what caused a lot of memory allocations). Now we will just "match" the phase without needing a mapping anymore. **Note: Found a missing mapping here. Added it**.
- Slightly tweak how we define and apply the filters, to reduce memory allocations inside the vacancies loop.
- Fewer intermediate objects and strings.

### Test for Daily Alerts with 2000 subscriptions over 390 new vacancies

| Metric | Original | After changes | Change | % |
| --- | --- | --- | --- | --- |
| Memory Used |	7,236,692 KB |	3,360,852 KB	 |-3,875,840 KB | -53% |
| Runtime |	~52 sec	| ~42 sec | -10 sec | -19% |

#### Also did a proper memory profiling of the job run before and after:

| Metric | Original | After changes | Change | % |
| --- | --- | --- | --- | --- |
| Allocated Memory (subscription.rb) | 911,620,848 bytes | 105,254,314 bytes | -806 MB | -90% |
| Allocated Objects (subscription.rb) | 15,628,903 objects | 2,098,999 objects | - 13.5 million objects | -87%|

##### These metrics only apply to the `subscriptions.rb` memory allocation.
There are also lots of allocations by the `rgeo` gem for geospatial calculations/checks by the location search. And those have not been affected. I plan to tackle those separately by reshaping how we do the subscription location filtering.

#### Before

All the phases (and how many times they appear in the search criteria) in our prod DB subscriptions:
```
{"nursery" => 20232,
 "primary" => 91073,
 "middle" => 28329,
 "secondary" => 69828,
 "16-19" => 7878,
 "sixth_form_or_college" => 21700,
 "through" => 13839,
 "middle_deemed_secondary" => 323,
 "sixteen_plus" => 410,
 "middle_deemed_primary" => 246,
 "all_through" => 177,
 "not_applicable" => 15,
 "" => 1}
```

#### After
Ran the backfill mapping tasks over a copy of prod DB:
Now all the values match our current `phases` possible values.
```
{"nursery" => 20232, 
 "primary" => 99627, 
 "secondary" => 78688, 
 "sixth_form_or_college" => 29988, 
 "through" => 14016}
```

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Subscription < ApplicationRecord
   enum :frequency, { daily: 0, weekly: 1 }
 
@@ -38,8 +40,16 @@ class Subscription < ApplicationRecord
   # during the deploy
   self.ignored_columns += %w[active]
 
+  # One/off code. Remove after running the rake task to normalise all subscriptions
+  LEGACY_PHASE_MAPPING = {
+    "middle_deemed_secondary" => %w[primary secondary].freeze,
+    "middle_deemed_primary" => %w[primary secondary].freeze,
+    "all_through" => %w[through].freeze,
+    "sixteen_plus" => %w[sixth_form_or_college].freeze,
+    "16-19" => %w[sixth_form_or_college].freeze,
+  }.freeze
+
   class << self
-    # map legacy phase filters onto current ones
     def phases_match?(vacancy, filter)
       phases = filter.map { |phase| phase.in?(%w[middle_deemed_secondary middle_deemed_primary]) ? "primary secondary" : phase }
           .map { |phase| phase == "all_through" ? "through" : phase }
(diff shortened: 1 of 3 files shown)


starswan (approved):
I guess the only challenge here is that the rake task needs to be run before the subscriptions kick in at 3pm, as we have wrecked the legacy ones in the code...

PR #8136, opened 2025-10-06
Blind recruitment

starswan:
## Trello card URL

https://trello.com/c/kqjNe7qm/2049-build-the-ability-to-anonymise-applications

## Changes in this PR:

### DIFF
--- a/app/models/job_application.rb
+++ b/app/models/job_application.rb
@@ -222,6 +222,10 @@ class JobApplication < ApplicationRecord
     vacancy.anonymise_applications? && status.in?(PRE_SHORTLIST_STATUSES)
   end
 
+  def address
+    [street_address, city, postcode, country].compact_blank
+  end
+
   private
 
   def update_conversation_searchable_content
--- a/app/models/job_application_pdf.rb
+++ b/app/models/job_application_pdf.rb
@@ -312,12 +312,7 @@ class JobApplicationPdf
   end
 
   def job_application_address
-    [
-      job_application.street_address,
-      job_application.city,
-      job_application.postcode,
-      job_application.country,
-    ].compact.join(", ")
+    job_application.address.join(", ")
   end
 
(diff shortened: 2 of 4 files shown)


starswan:
> the anonymisation should cover the job application PDF as well

Seems to already do it - what issue have you found?

jriga:
sorry I should have added more details:
- on the publisher job application show page, there is also a download pdf button that is missing the decoration call and we can see candidate details there

Also just noticed that the file name in the downloaded zip is in `Information not shown_Information not shown.pdf` we should maybe add the generated id there instead

app/models/job_application.rb
```diff
@@ -135,18 +135,8 @@ def active_status?
...
-      return Document["no_application_form.txt", "the candidate has no application for on record"] unless application_form.attached?
-
-      extension = File.extname(application_form.filename.to_s)
-      Document["application_form#{extension}", application_form.download]
-    else
-      pdf = JobApplicationPdfGenerator.new(self).generate
-      Document["application_form.pdf", pdf.render]
-    end
+  def has_pre_interview_checks?
+    status.in?(%w[interviewing unsuccessful_interview offered declined])
```
jriga: We've had a few bugs related to status management when we employ this imperative method definition. Whenever we update `STATUS_TRANSITION`  we risk introducing a bug.

We could instead define a constant that is populated with all the statuses after `shortlisted` from `STATUS_TRANSITION`

```suggestion
 pre_shortlisted, post_shortlisted = JobApplication::STATUS_TRANSITIONS.keys.split { it == "shortlisted" }
POST_SHORTLISTED = post_shortlisted.flat_map { JobApplication::STATUS_TRANSITIONS[it] }.uniq - ["withdrawn"]

status.in?(POST_SHORTLISTED)
```

jriga (approved):

PR #8582, opened 2026-03-07
Use :nocov: tags to make code coverage 100%

starswan:
Coverage data has been fluctuating recently which makes it hard to discover where coverage has been genuinely lost during a change. This makes the (line) coverage 100% as an experiment - at least this should be keepable at the maximum, with all 'uncovered' code marked with :nocov: tags


scruti (dismissed):
Thanks for this housekeeping. Nice to see so much unreferenced code being removed.

I left a couple of comments on rake testing and simplecove config files from things I suspect they have already resolved/removed in `main` for your consideration.

PR #8611, opened 2026-03-17
Refactor job application review component

richardpattinson:
## Trello card URL

https://trello.com/c/eNNwhCHm/2690-job-application-review-component

## Changes in this PR:

The job application review component is hard to understand and hard to change. Recent changes have just made it worse, and it is not clear how it could be improved. We should look at converting it into a plain old slim template to reduce the complexity

## Screenshots of UI changes:

No UI changes

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/helpers/components_helper.rb
```diff
@@ -6,11 +6,11 @@ module ComponentsHelper
     empty_section: "EmptySectionComponent",
     filters: "FiltersComponent",
     job_application_notes: "PublisherNotesOnJobApplicationComponent",
-    job_application_review: "JobApplicationReviewComponent",
     landing_page_link_group: "LandingPageLinkGroupComponent",
     map: "MapComponent",
     navigation_list: "NavigationListComponent",
     editor: "EditorComponent",
+    review_section: "JobApplicationReviewComponent::Section",
```
scruti: Since `JobApplicationReviewComponent` has been removed and what used to be a review component is now a review section component, could we transform `JobApplicationReviewComponent::Section` into `ReviewSectionComponent`? 

That way, we would keep the normalised structure of the components folder/namespacing.

scruti (commented):
Very nice refactor 👏 

The only thing for me is the structure/namespacing of the new component. I would rather align it as if it were a new "review section" component instead of keeping the old namespacing within a component that got removed.

scruti (approved):
LGTM!

PR #8643, opened 2026-03-24
Replace google drive with microsoft defender for vacancy and job application malware scanning

KyleMacPherson:
## Changes in this PR:

This PR removes the google drive antivirus flow for the hiring staff job publishing flow and the jobseeker job application flow. Instead we switch over to using microsoft defender for storage. The key difference is that the antivirus scanning is done async on azure. To account for this change, in the flows mentioned above we will allow users to continue through the wizard steps if the file has not been scanned and only block them from progressing if the file is flagged as malicious or the azure scan has errored. We will however stop them from actually publishing the job or submitting the application until the file has been scanned and tagged as safe. See below for more detailed information on each flow with screenshots.

**Hiring staff uploaded application form**
- allows hiring staff to upload application form and progress to review and publish page before file has been scanned or has been scanned and flagged as safe
- errors if uploaded application form has been scanned as unsafe
Screenshot 2026-04-14 at 16 03 52
- blocks publishing of vacancy until file has been scanned
Screenshot 2026-04-14 at 16 03 26
- blocks  publishing of vacancy if file has been scanned and either the scan errors or the file is flagged as unsafe and prompts user to delete the file and try again
Screenshot 2026-04-14 at 16 03 41

**Hiring staff additional documents**
- allows hiring staff to upload additional documents and progress to review and publish page before file has been scanned or has been scanned and flagged as safe
- errors if additional document has been scanned as unsafe
Screenshot 2026-04-14 at 15 07 05
- blocks publishing of vacancy until file has been scanned
Screenshot 2026-04-17 at 09 38 37
- blocks  publishing of vacancy if file has been scanned and either the scan errors or the file is flagged as unsafe and prompts user to delete the file and try again
Screenshot 2026-04-14 at 15 33 33

**jobseeker uploaded job application**
- allows user to upload a job application form and progress to the review and submit page before file has been scanned or has been scanned and flagged as safe
- errors if uploaded job application form has been scanned as unsafe
Screenshot 2026-04-14 at 14 46 27 
- blocks submission of job application until file has been scanned
- blocks submission of job application if file has been scanned and either the scan errors or the file is flagged as unsafe and prompts user to delete the file and try again
Screenshot 2026-04-14 at 14 46 17

**jobseeker catholic application form**
- allows user to upload a baptism certificate and progress to the review and submit page before file has been scanned or has been scanned and flagged as safe
- errors if uploaded baptism has been scanned as unsafe
Screenshot 2026-04-14 at 15 40 22
- blocks submission of job application until file has been scanned
- blocks submission of job application if file has been scanned and either the scan errors or the file is flagged as unsafe and prompts user to delete the file and try again
Screenshot 2026-04-14 at 15 40 38

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/form_models/base_form.rb
```diff
@@ -6,7 +6,7 @@ class BaseForm
     valid_file_types: %i[JPG JPEG PNG].freeze,
   }.freeze
 
-  VALID_DOCUMENT_TYPES = %i[PDF DOC DOCX].freeze
+  VALID_DOCUMENT_TYPES = %i[PDF DOC DOCX TXT].freeze
```
KyleMacPherson, line 9: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/models/vacancy.rb
```diff
@@ -78,19 +78,19 @@ class Vacancy < ApplicationRecord
   belongs_to :publisher_ats_api_client, optional: true
 
   DOCUMENT_FILE_SIZE_LIMIT = 20.megabytes
-  DOCUMENT_CONTENT_TYPES = %w[application/pdf application/msword application/vnd.openxmlformats-officedocument.wordprocessingml.document].freeze
+  DOCUMENT_CONTENT_TYPES = %w[application/pdf application/msword application/vnd.openxmlformats-officedocument.wordprocessingml.document text/plain].freeze
```
KyleMacPherson, line 81: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/models/vacancy.rb
```diff
@@ -78,19 +78,19 @@ class Vacancy < ApplicationRecord
...
 
   DOCUMENT_VALIDATION_OPTIONS = {
     file_type: :document,
     content_types_allowed: DOCUMENT_CONTENT_TYPES,
     file_size_limit: DOCUMENT_FILE_SIZE_LIMIT,
-    valid_file_types: %i[PDF DOC DOCX],
+    valid_file_types: %i[PDF DOC DOCX TXT],
```
KyleMacPherson, line 87: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/validators/form_file_validator.rb
```diff
@@ -39,7 +39,7 @@ def valid_file_size?(record, attribute, value)
   end
 
   def valid_file_type?(record, attribute, value)
-    content_type = MimeMagic.by_magic(value.tempfile)&.type
+    content_type = MimeMagic.by_magic(value.tempfile)&.type || value.content_type
```
KyleMacPherson, line 42: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/views/jobseekers/uploaded_job_applications/upload_application_forms/edit.html.slim
```diff
@@ -21,7 +24,7 @@
         = f.govuk_file_field :application_form,
           label: { text: "Upload application form", tag: "h1", size: "l" },
           hint: { text: t("helpers.hint.publishers_job_listing_application_form_form.application_form") },
-          accept: ".doc, .docx, .pdf",
+          accept: ".doc, .docx, .pdf, .txt",
```
KyleMacPherson, line 27: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/views/publishers/vacancies/build/application_form.html.slim
```diff
@@ -16,13 +16,16 @@ div data-controller="upload-documents" data-upload-documents-inactive-value=vaca
       = f.govuk_file_field :application_form,
         javascript: true,
         label: { size: "m" },
-          accept: ".doc, .docx, .pdf",
+          accept: ".doc, .docx, .pdf, .txt",
```
KyleMacPherson, line 19: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/views/publishers/vacancies/documents/new.html.slim
```diff
@@ -12,7 +12,7 @@
         javascript: true,
         label: { text: vacancy_form_page_heading(vacancy, step_process, back_path: back_path), tag: "h1", size: "l" },
         hint: { text: t("helpers.hint.publishers_job_listing_documents_form.documents") },
-        accept: ".doc, .docx, .pdf",
+        accept: ".doc, .docx, .pdf, .txt",
```
KyleMacPherson, line 15: This is just to allow us to test with eicar.txt to simulate a malicious file and will be removed before deploying.

app/controllers/jobseekers/uploaded_job_applications/upload_application_forms_controller.rb
```diff
@@ -8,6 +8,11 @@ def edit
   def update
     # Skip validation if no new file is submitted, but one is already attached
     if !form_params[:application_form] && @job_application.application_form.attached?
+      @form = Jobseekers::UploadedJobApplication::UploadApplicationFormForm.new(
+        form_params.to_h.merge(application_form: @job_application.application_form),
+      )
+      return render :edit if @form.invalid?
+
```
scruti, line 15: Not sure I understand if/why this change is needed at this stage. The original comment does not make sense now with the change of behaviour. So a comment update here seems it would help?

db/migrate/20260318095049_add_malware_scan_result_to_active_storage_blobs.rb
```diff
@@ -0,0 +1,8 @@
+class AddMalwareScanResultToActiveStorageBlobs < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :active_storage_blobs, :malware_scan_result, :integer, default: 0, null: false
+    add_index :active_storage_blobs, :malware_scan_result, algorithm: :concurrently
```
scruti, line 6: 🙋‍♂️ I don't think this index is needed or beneficial. 
We have only a few possible values (which makes the index less useful) and no bulk query operations over this new field (that I can see).

app/validators/form_file_validator.rb
```diff
@@ -10,7 +10,7 @@ def validate_each(record, attribute, value)
       [value].flatten.each do |file|
         valid_file_size?(record, attribute, file) &&
           valid_file_type?(record, attribute, file) &&
-          virus_free?(record, attribute, file)
+          (options[:skip_google_drive_virus_check] || virus_free?(record, attribute, file))
       end
```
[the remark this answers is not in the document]
KyleMacPherson: This is intended behaviour, some other uploads still rely on google drive for now.

app/form_models/publishers/job_listing/documents_confirmation_form.rb
```diff
@@ -2,10 +2,31 @@ class Publishers::JobListing::DocumentsConfirmationForm < BaseForm
...
+
+  private
+
+  def additional_documents_scan_safe
+    vacancy.supporting_documents.each do |doc|
+      blob = doc.blob
+      if blob.malware_scan_malicious? || blob.malware_scan_scan_error?
+        errors.add(:base, I18n.t("jobs.file_unsafe_error_message", filename: doc.filename))
+      end
+    end
```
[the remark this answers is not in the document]
KyleMacPherson: Addressed this one 👍

db/migrate/20260318095049_add_malware_scan_result_to_active_storage_blobs.rb
```diff
@@ -0,0 +1,8 @@
...
+  end
```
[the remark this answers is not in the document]
KyleMacPherson: Addressed this one 👍

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,22 @@
...
+    enum :malware_scan_result, {
+      pending: 0,
+      clean: 1,
+      malicious: 2,
+      scan_error: 3,
+    }, prefix: :malware_scan
+
+    after_create_commit do
+      FetchMalwareScanResultJob.set(wait: 2.seconds).perform_later(id)
+    end
```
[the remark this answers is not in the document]
KyleMacPherson: Addressed this one 👍

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,22 @@
...
+  ActiveStorage::Blob.class_eval do
+    enum :malware_scan_result, {
+      pending: 0,
+      clean: 1,
+      malicious: 2,
+      scan_error: 3,
+    }, prefix: :malware_scan
+
+    after_create_commit do
+      FetchMalwareScanResultJob.set(wait: 2.seconds).perform_later(id)
```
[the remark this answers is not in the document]
KyleMacPherson: Addressed this one 👍

app/controllers/publishers/vacancies/publish_controller.rb
```diff
@@ -1,10 +1,19 @@
...
       redirect_to organisation_job_path(vacancy.id), notice: t("messages.jobs.already_published")
+    elsif (not_safe_blobs = uploaded_files_not_safe).any?
+      messages = not_safe_blobs.map do |blob|
+        if blob.malware_scan_pending?
+          t("jobs.file_pending_scan_message", filename: blob.filename)
+        else
+          t("jobs.file_unsafe_error_message", filename: blob.filename)
+        end
+      end
+      redirect_to organisation_job_review_path(vacancy.id), alert: messages.join(". ")
```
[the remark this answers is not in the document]
KyleMacPherson: Addressed this one 👍

app/views/vacancies/listing/_job_details.html.slim
```diff
@@ -108,7 +108,7 @@ section#job-details class="govuk-!-margin-bottom-5"
     - elsif vacancy.receive_applications == "email"
       p.govuk-body = t("jobs.apply_via_email_html", email: application_email_link(vacancy))
       p.govuk-body = t("jobseekers.job_applications.no_cvs")
-      - if current_jobseeker.present?
+      - if current_jobseeker.present? && vacancy.application_form.blob.malware_scan_clean?
         = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id)
```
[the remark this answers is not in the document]
KyleMacPherson: I'm not too worried about this one as it should never actually be able to happen. An unsafe uploaded application form should never get to this point, here we are just not showing the download button to be extra safe.

app/controllers/publishers/vacancies/publish_controller.rb
```diff
@@ -20,5 +29,16 @@ def create
...
+  private
+
+  # Pending files are allowed to progress through the wizard steps but blocked here at publish time.
+  # This covers files still awaiting their antivirus scan result as well as malicious/errored ones.
+  def uploaded_files_not_safe
+    blobs = []
+    blobs << vacancy.application_form.blob if vacancy.application_form.attached?
+    blobs += vacancy.supporting_documents.map(&:blob) if vacancy.supporting_documents.attached?
+    blobs.reject(&:malware_scan_clean?)
+  end
```
scruti: Suggestion: I think this could be rather defined as `Vacancy#unsafe_blobs` method. So it could be unit tested in the model, and the controller would just ask the vacancy instead of check the vacancy internals & build the array

app/controllers/jobseekers/uploaded_job_applications/upload_application_forms_controller.rb
KyleMacPherson: Good point, sorry missed that one! Have updated it now, let me know if it makes sense please 🙏

db/migrate/20260318095049_add_malware_scan_result_to_active_storage_blobs.rb
KyleMacPherson: Removed!

app/controllers/publishers/vacancies/publish_controller.rb
KyleMacPherson: Good idea! I changed this now

app/form_models/jobseekers/job_application/catholic_form.rb
```diff
@@ -35,7 +35,10 @@ def load_form(model)
         validates :baptism_date, date: { on_or_before: :today }
       end
 
-      validates :baptism_certificate, form_file: Vacancy::DOCUMENT_VALIDATION_OPTIONS, presence: true, if: -> { section_completed && following_religion && religious_reference_type == "baptism_certificate" }
+      validates :baptism_certificate, form_file: Vacancy::DOCUMENT_VALIDATION_OPTIONS.merge(skip_google_drive_virus_check: true), presence: true, if: -> { section_completed && following_religion && religious_reference_type == "baptism_certificate" }
+      # Files awaiting an antivirus scan are allowed to progress through the wizard steps so jobseekers can complete other steps.
+      # Pending files are blocked at submit time in the review form.
+      validate :baptism_certificate_scan_safe, if: -> { section_completed && following_religion && religious_reference_type == "baptism_certificate" }
```
scruti: Suggestion: Since both validations are guarded under the same condition. Wrap them inside the condition to make it clearer.

```suggestion
      with_options if: -> { section_completed && following_religion && religious_reference_type == "baptism_certificate" } do
        validates :baptism_certificate, form_file: Vacancy::DOCUMENT_VALIDATION_OPTIONS.merge(skip_google_drive_virus_check: true), presence: true
        # Files awaiting an antivirus scan are allowed to progress through the wizard steps so jobseekers can complete other steps.
        # Pending files are blocked at submit time in the review form.
        validate :baptism_certificate_scan_safe
      end
```

app/form_models/jobseekers/job_application/catholic_form.rb
```diff
@@ -44,6 +47,15 @@ def baptism_date=(value)
...
       end
+
+      private
+
+      def baptism_certificate_scan_safe
+        return unless baptism_certificate.respond_to?(:blob)
+
+        blob = baptism_certificate.blob
+        errors.add(:baptism_certificate, :unsafe_file, filename: blob.filename) if blob.malware_scan_malicious? || blob.malware_scan_scan_error?
+      end
```
scruti, line 60: The new validations across the 3 different forms seem to do the same thing.  

I suspect we could directly include this as a new check within the `FormFileValidator` (with maybe an option to disable it for logos and others not using this), given that they're "paired" together and the malware check is a check over the form file. So could end by replacing the current `virus_free?` internal validation step within `FormFileValidator`.

Or create a separate `FileMalwareFreeValidator` if we want to end using https://github.com/igorkasyanchuk/active_storage_validations as the comment on https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/validators/form_file_validator.rb#L1-L3 suggests.

But given they both go together... maybe creating this check within the `FormFileValidator` seems simpler and more straightforward?

app/form_models/jobseekers/job_application/review_form.rb
```diff
@@ -1,11 +1,36 @@
...
+      def uploaded_file_scan_safe
+        return unless job_application
+
+        blob = if job_application.instance_of?(::UploadedJobApplication)
+                 job_application.application_form.blob if job_application.application_form.attached?
+               elsif job_application.baptism_certificate.attached?
+                 job_application.baptism_certificate.blob
+               end
+
+        return if blob.nil? || blob.malware_scan_clean?
```
scruti: This seems to be digging a bit too much into JobApplication internals. We could try to encapsulate these things into the `JobApplication`/variants models 🤔 

EG: 
- `job_application.uploaded_file` -> Returns the blob,
  - `UploadedJobApplication` defines the method as the return of the `application_form` if attached.
  - `NativeJobApplication` (I am assuming all non Uploaded are Native?) defines the same method to check the `baptism_certificate` blob if attached.
- `job_application.uploaded_file_clean?` -> As above, using different implementations in the JobApplication subclasses.
That way, the form would just need to query the `job_application` to see if is clean (maybe return true by default if no attachment?) and use the `uploaded_file` to generate the error if not.

In theory, we would achieve the same output with a more "OOP interface" approach, by just asking the JobApplication without knowing its internals.

app/services/fetch_malware_scan_result.rb
```diff
@@ -0,0 +1,75 @@
...
+    @blob_service ||=
+      AzureBlob::Client.new(
+        account_name: azure_storage_account_name,
+        access_key: azure_storage_access_key,
+        container: azure_storage_container,
+      )
+  end
+
+  def blob_tags
+    @blob_tags ||= blob_service.get_blob_tags(blob.key)
```
scruti: ❓ Is there any reason we need to explicitly re-produce the storage config and initialise a new Azure Blob client to retrieve the tags?

The config should be accessible directly from `blob.service` 

And I believe we can ever save the whole client setup and query the tags directly from the `blob`, which would save us quite a lot of boilerplating (if feasible!)🤞 

```suggestion
    @blob_tags ||= blob.service.client.get_blob_tags(blob.key)
```

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  after_create_commit do
+    if azure_virus_scan_supported?
+      FetchMalwareScanResultJob.set(wait: 2.seconds).perform_later(id)
+    elsif malware_scan_pending?
+      update_column(:malware_scan_result, self.class.malware_scan_results[:clean])
+    end
+  end
+end
```
scruti, line 28: This is a big change proposal and may not even work, but I think it is worth a look since it would simplify things if feasible:

When checking `ActiveStorage` interface while reviewing this PR, I bumped into the `ActiveStorage::Analyzer` ([link](https://api.rubyonrails.org/classes/ActiveStorage/Analyzer.html)) and [Analyzable concern link](https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blob/Analyzable.html).

It seems to be designed specifically to automatically run different analysis tools over the blob and add metadata into the blob `metadata` JSON column. And given that someone built a [gem adding a ClamAV ActiveStorage Analyzer](https://github.com/ackama/activestorage-clamav-analyzer), it seems building Analyzers for AV scans is a good pattern.

### What I like about it.
1. It automatically enqueues an[ ActiveStorage::AnalyzeJob](https://api.rubyonrails.org/classes/ActiveStorage/AnalyzeJob.html) for each Analyzer when the file is attached. Not having to define a custom job in our case.
2. It relies on the existing Blob `metadata` column. No need to extend the ActiveStorage::Blob default columns.
3. No need to do this extension over `ActiveSupport`. It is just appending our analyzer on config.
  - EG: `Rails.application.config.active_storage.analyzers.append DOCXAnalyzer`
4. Seems to be "The Rails way". And we should try to align with the Rails way (when possible!).

### What could be tricky/a problem
#### The asynchronous nature of Azure malware analysis.
This is when I wonder if you explored the Analyzer approach and found a blocker. Or it will be a blocker.

But I suspect that we could convert the current `ServiceObject` into an `Analyzer`, and raise an exception from it if the scan is still not available, which could turn into the AnalyzerJob failing and re-enqueuing following the Sidekiq on failure retries? 

EG (AI slop!):

```
# frozen_string_literal: true

class MalwareScanAnalyzer < ActiveStorage::Analyzer
  SCAN_RESULT_TAG_KEY = "Malware Scanning scan result"
  SCAN_RESULT_TAG_VALUE_CLEAN     = /no threats found/
  SCAN_RESULT_TAG_VALUE_ERROR     = /scan failed/
  SCAN_RESULT_TAG_VALUE_MALICIOUS = /malicious/

  def self.accept?(blob)
    # Completely made up!
    %w[azure_storage_documents azure_storage_images_and_logos].include?(blob.service_name)
  end

  def metadata
    tag = blob.service.client.get_blob_tags(blob.key)&.dig(SCAN_RESULT_TAG_KEY)&.downcase

    # Azure's scan is async — if the tag isn't present yet, raise so
    # ActiveStorage::AnalyzeJob is retried by Sidekiq with exponential backoff
    raise ScanPending if tag.nil?

    result = case tag
             when SCAN_RESULT_TAG_VALUE_CLEAN     then "clean"
             when SCAN_RESULT_TAG_VALUE_ERROR      then "scan_error"
             when SCAN_RESULT_TAG_VALUE_MALICIOUS  then "malicious"
             end

    { malware_scan_result: result }
  end

  class ScanPending < StandardError; end
end
```

Let me know your thoughts! 

A bit annoyed, I didn't know about this whole Analyzer bit and the metadata from the Blobs to propose it before reaching this point 🤦

KyleMacPherson: This does sound like a great solution and I have been working on this this afternoon. The major blocker I have come across is that I believe only one analyzer runs per blob unfortunately. So I think it's a great solution if we are happy to let images go unscanned (because we already use ImageAnalyzer) which seems reasonable but slightly risky to me. 

Otherwise we could patch or override the image analyzer to do both the image analysis stuff and the antivirus scan for images and use a different antivirus analyzer for other types of attachments but that doesn't seem like The Rails Way either.

Overall I think it's a pretty cool solution and I am up for pursuing it if you prefer it but I don't think it's quite as clean as it first looks.

scruti: Thanks for looking into this @KyleMacPherson. I completely missed the "first come, first (and only) served" approach from the Analyzers. I thought they could be combined :facepalm: 

Happy to go with the mixed approach you mention on the chat (EG: Relying on the metadata field instead of adding a migration, while keeping our AV job/service independent from the Rails analyzer approach).

app/services/fetch_malware_scan_result.rb
KyleMacPherson: Oh nice spot! Thats much better, have changed it!

app/form_models/jobseekers/job_application/catholic_form.rb
KyleMacPherson: I think we can address this one in the next PR when we remove the current virus_free? functionality if that's ok with you?

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def malware_scan_scan_error?
+    malware_scan_result == "scan_error"
+  end
+
+  def azure_virus_scan_supported?
+    # rubocop:disable Style/ClassEqualityComparison
+    service.class.name == "ActiveStorage::Service::AzureStorageService" &&
+      Rails.application.config.app_role.in?(%w[production staging qa review])
```
KyleMacPherson: enabling it in review so we can test in review apps manually, leaving a test failing on this for now and will remove before deploying

app/validators/form_file_validator.rb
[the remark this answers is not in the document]
KyleMacPherson: We only have this in place so we can use .txt files to manually test on review apps and will delete this change before deployment along with removing the ability for users to upload .txt files

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def malware_scan_scan_error?
+    malware_scan_result == "scan_error"
+  end
+
+  def azure_virus_scan_supported?
+    # rubocop:disable Style/ClassEqualityComparison
+    service.class.name == "ActiveStorage::Service::AzureBlobService" &&
+      Rails.application.config.app_role.in?(%w[production staging qa review])
```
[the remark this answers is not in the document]
KyleMacPherson: This mismatch is deliberately in place. We want to allow the review app to use azure scans for now to manually test this change. We don't want to deploy this without changing this so review apps do not use azure scans going forward. The mismatch is there so the test fails so we can't deploy it without making this change.

app/views/vacancies/listing/_job_details.html.slim
```diff
@@ -108,7 +108,7 @@ section#job-details class="govuk-!-margin-bottom-5"
     - elsif vacancy.receive_applications == "email"
       p.govuk-body = t("jobs.apply_via_email_html", email: application_email_link(vacancy))
       p.govuk-body = t("jobseekers.job_applications.no_cvs")
-      - if current_jobseeker.present?
+      - if current_jobseeker.present? && vacancy.application_form.blob.malware_scan_clean?
```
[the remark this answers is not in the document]
KyleMacPherson: Removing the vacancy.application_form.blob.malware_scan_clean? because we need to allow vacancy attachment blobs that haven't yet been backfilled to be displayed and downloaded

app/views/vacancies/listing/_job_details.html.slim
```diff
@@ -123,13 +123,15 @@ section#job-details class="govuk-!-margin-bottom-5"
...
+        - if vacancy.application_form.blob.malware_scan_clean?
+          = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id), class: "govuk-button--secondary"
       - elsif vacancy.application_link.present?
         p.govuk-body = t("jobseekers.job_applications.no_cvs")
         = apply_link(vacancy, class: "govuk-!-margin-bottom-5")
       - elsif vacancy.application_form.present?
         p.govuk-body = t("jobseekers.job_applications.no_cvs")
-        = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id)
+        - if vacancy.application_form.blob.malware_scan_clean?
+          = govuk_button_link_to t("buttons.download_application_form", size: number_to_human_size(vacancy.application_form.byte_size)), job_document_path(vacancy, vacancy.application_form.id)
```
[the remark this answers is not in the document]
KyleMacPherson: Removing the vacancy.application_form.blob.malware_scan_clean? because we need to allow vacancy attachment blobs that haven't yet been backfilled to be displayed and downloaded

spec/form_models/jobseekers/uploaded_job_application/upload_application_form_form_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Jobseekers::UploadedJobApplication::UploadApplicationFormForm, type: :model do
+  describe "application_form_scan_safe" do
```
scruti, line 6: What happens in this case when the blob scan is pending?

spec/form_models/publishers/job_listing/application_form_form_spec.rb
```diff
@@ -26,4 +26,30 @@
       end
     end
   end
+
+  describe "application_form_scan_safe" do
```
scruti, line 30: What happens in this case when the blob scan is pending?

spec/form_models/publishers/job_listing/documents_confirmation_form_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Publishers::JobListing::DocumentsConfirmationForm, type: :model do
+  let(:vacancy) { create(:vacancy, :with_supporting_documents) }
+  let(:form) { described_class.new({ upload_additional_document: "false" }, vacancy) }
+
+  describe "additional_documents_scan_safe" do
```
scruti, line 9: What happens in this case when some or the documents are still pending?

spec/services/publishers/document_virus_check_spec.rb
```diff
@@ -1,104 +0,0 @@
-require "rails_helper"
-require "google_api_client"
-
-RSpec.describe Publishers::DocumentVirusCheck do
```
scruti, line 4: Question: Why are we removing the spec but not the service code? Shouldn't we remove/keep both?

config/initializers/active_storage_malware_scanning.rb
```diff
@@ -0,0 +1,33 @@
...
+    pending: 0,
+    clean: 1,
+    malicious: 2,
+    scan_error: 3,
+  }, prefix: :malware_scan
+
+  # Override enum-generated method to also treat nil as pending (for legacy blobs pre-backfill)
+  def malware_scan_pending?
+    malware_scan_result.nil? || malware_scan_result == "pending"
+  end
```
scruti: If I understand correctly, the strategy here is going to be:
1 - Add the new column as `nil`
2 - Backfill all the nils as `clean`.
3- New attachments will be created as `nil`, marked as malicious/error/clean from Azure.
4 - `pending` enum doesn't seem to be used unless a further PR introduced the default later.

Based on that understanding (that may be wrong), wouldn't it be simpler to:
1 - Add the new column with `pending` as the default.
2 - As soon as the deploy goes through, run a task backfilling all the `pending` as `clean`.
3 - New attachments will be `pending` by default, no need for this method.

There is a time window between the deploy and running the task wher a file could be uploaded and marked as "clean" while containing a virus. But the risk of that seems very low and the code would be simpler?

Or I am missunderstanding this? 😅

scruti (commented):
Few comments/questions. Just trying to clarify I understood this before approving. Nothing big!

spec/form_models/jobseekers/uploaded_job_application/upload_application_form_form_spec.rb
KyleMacPherson: Pending scans do not raise an error because we don't want to block the user from continuing through the flow, pending scans are only blocked at the submitting stage.

spec/form_models/publishers/job_listing/application_form_form_spec.rb
KyleMacPherson: Pending scans do not raise an error because we don't want to block the user from continuing through the flow, pending scans are only blocked at the publishing stage.

spec/form_models/publishers/job_listing/documents_confirmation_form_spec.rb
KyleMacPherson: Pending scans do not raise an error because we don't want to block the user from continuing through the flow, pending scans are only blocked at the publishing stage.

spec/services/publishers/document_virus_check_spec.rb
KyleMacPherson: Good catch 🎣 I initially thought we could remove the document virus check code in this PR, but then realised we should keep it. Forgot to add back in the spec. Have added it back in now!

scruti (approved):
LGTM! Thanks for going through all the feedback/questions cycles on this 👏

PR #8666, opened 2026-03-27
Profiles: QTS/Professional Status updates

starswan:
## Trello card URL

https://trello.com/c/HnQ2ivZL/2672-profile-qualifications-journey
Edit: Think it is this one: https://trello.com/c/xHVsNEI4/2671-profile-professional-status-qts-journey (KM)

## Changes in this PR:

QTS Profiles page

### DIFF
--- a/config/locales/jobseekers/job_application/professional_status_form.yml
+++ b/config/locales/jobseekers/job_application/professional_status_form.yml
@@ -56,7 +56,7 @@ en:
         is_statutory_induction_complete: Have you completed your induction period?
         statutory_induction_complete_details: Provide more detail, for example, if you have started but not completed your induction, or you are %{link} (optional)
         statutory_induction_complete_details_hint_text: exempt from completing an induction
-        trn_link_text: find your teacher reference number (TRN)
+        trn_link_text: find your TRN
       jobseekers_job_application_professional_status_form:
         qualified_teacher_status_details_html: Please provide more detail (optional<span class='govuk-visually-hidden'> field</span>)</span>
         qualified_teacher_status_options:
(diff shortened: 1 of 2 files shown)


app/views/jobseekers/profiles/qualified_teacher_status/_qts_fields.html.slim
```diff
@@ -0,0 +1,23 @@
...
+  legend: { size: "l", tag: "h1" } do
+  = f.govuk_radio_button :qualified_teacher_status, "yes", link_errors: true
+    = f.govuk_text_field :qualified_teacher_status_year
+    = f.govuk_text_field :qts_age_range_and_subject
+
+  = f.govuk_radio_button :qualified_teacher_status, "no"
+    = f.govuk_text_area :qualified_teacher_status_details
+  = f.govuk_radio_button :qualified_teacher_status, "on_track"
+
+= f.govuk_radio_buttons_fieldset :has_teacher_reference_number, legend: { tag: "h2" },
```
KyleMacPherson: It looks like we are re-introducing has_teacher_reference_number which we got rid of a few months ago. I don't see this mentioned on the ticket, is this part of this work?

KyleMacPherson (changes requested):
The ticket seems to suggest that we just want to change some hint text, but this looks like it's adding in an extra, previously deleted field. I may well be missing context here but based on the ticket I don't think this looks quite right

app/views/jobseekers/profiles/qualified_teacher_status/_qts_fields.html.slim
scruti: Good catch @KyleMacPherson. I suspect the design on the Lucid is outdated from the start, since the "Current Profiles" initial screenshots contain the outdated design with the removed radio button, while in reality now looks like this:
image

@starswan you can double-check with Alice, but given how much of a pain was the radio with the `has_teacher_reference_number` field, and how relieved we were once removed, I wouldn't expect going back to the previous state of that question. So I'm pretty sure it is an outdated design.

KyleMacPherson: I asked in the TV Product channel and it looks like the screenshots used for the Lucid board are in fact just out of date so we don't want to add back in that question. Annoying, I know!

KyleMacPherson (approved):

PR #8681, opened 2026-03-31
Profiles: Work History

starswan:
## Trello card URL
https://trello.com/c/c4YrieFy/2723-profile-work-history-journey

## Changes in this PR:

Profile work history


KyleMacPherson (dismissed):
Looks good to me! One thing I noticed was on the ticket it says **Update fields “Start date” and “End date” to just 1 field “Duration”**. On the lucid board it says: _"can the devs calculate a duration so the jobseeker gets the same start/end date fields, but we should as '3 years and 6 months' to hiring staff?"_. I don't think we have addressed that in this change, but appreciate that requirement could well have changed and no longer be relevant!

Edit: Also think there may be changes to make to app/views/jobseekers/profiles/_summary.html.slim. I think we can remove `p class="govuk-!-margin-bottom-0" = employment.subjects`. I can't spot anything else at a glance but might be worth a quick check

starswan:
> Looks good to me! One thing I noticed was on the ticket it says **Update fields “Start date” and “End date” to just 1 field “Duration”**. On the lucid board it says: _"can the devs calculate a duration so the jobseeker gets the same start/end date fields, but we should as '3 years and 6 months' to hiring staff?"_. I don't think we have addressed that in this change, but appreciate that requirement could well have changed and no longer be relevant!
> 
> Edit: Also think there may be changes to make to app/views/jobseekers/profiles/_summary.html.slim. I think we can remove `p class="govuk-!-margin-bottom-0" = employment.subjects`. I can't spot anything else at a glance but might be worth a quick check

In the discussions (Chris and I) thought that it would be better to keep start/end dates for now as we already have them, and candidates will probably know them better than duration (we might want to keep them ordered by date anyway?) and the data migration was going to be tricky to time correctly

KyleMacPherson (approved):
Looks good to me!

PR #8705, opened 2026-04-09
Add view and delete uploaded reference functionality

KyleMacPherson:
## Changes in this PR:

In a previous PR i was adding functionality to view and delete previously uploaded files in preparation for the new anti-virus flow as we move from google drive to defender. I missed the uploaded reference in that PR.

I have had to slightly change the UI for this to allow deletion, so we no longer just show the file name on the pre-interview checks page for uploaded references but instead follow the same flow as other references where we click the referee name which links us to a page which shows the details and, in the case of uploaded references, allows us to delete the reference. Please see screenshots.

### Before
Screenshot 2026-04-13 at 09 19 37

### After
Screenshot 2026-04-13 at 09 15 28
Screenshot 2026-04-13 at 09 15 37
Screenshot 2026-04-13 at 09 15 46

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/views/publishers/vacancies/job_applications/reference_requests/show.html.slim
+++ b/app/views/publishers/vacancies/job_applications/reference_requests/show.html.slim
@@ -51,6 +51,12 @@
           - row.with_key(text: key)
           - row.with_value(text: value)
 
+    - if @reference_request.reference_form.attached?
+      h2.govuk-heading-m = t(".uploaded_reference.heading")
+      p.govuk-body
+        = govuk_link_to "#{@reference_request.reference_form.filename} (#{number_to_human_size(@reference_request.reference_form.byte_size)})", @reference_request.reference_form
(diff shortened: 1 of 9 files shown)


app/views/publishers/vacancies/job_applications/reference_requests/show.html.slim
```diff
@@ -50,6 +50,14 @@
         = summary_list.with_row do |row|
```
scruti, line 50: Being a little bit nitpicky here, if the attached/uploaded references have no values for the fields we may want to just not show the empty fields? 
So we may always show the `name` and condition the rest of "when no reference doc is attached"?

Regarding this
Image

scruti (approved):
LGTM! Left a non blocking suggestion on hiding some fields from the view when not applicable.

PR #8707, opened 2026-04-10
Fix browser dimensions

starswan:
## Changes in this PR:

Browser size is different between Selenium and Cuprite non-headless tests


scruti (commented):
Woopsie. My bad. 

Approved.

scruti (approved):

PR #8708, opened 2026-04-10
remove select tab from ATS tests - go straight to tab

starswan:
## Changes in this PR:

Several ATS tests go to the wrong tab and then select the correct one - this isn't very useful (or performant)


scruti (approved):
Nice catch! 👏

PR #8714, opened 2026-04-13
Fixup redis config with current

starswan:
## Changes in this PR:

Second try to fix redis config error - remove Redis.current as it relied on previous configuration


spec/services/vacancy_analytics_service_spec.rb
```diff
@@ -18,13 +18,13 @@
 
     it "increments the Redis counter for a normalized referrer" do
       expect {
-        described_class.track_visit(vacancy.id, referrer_url, hostname, {})
+        described_class.track_visit(mock_redis, vacancy.id, referrer_url, hostname, {})
       }.to change { Redis.current.get(redis_key).to_i }.by(1)
```
scruti: Shouldn't the references to `Redis.current` in this file be removed too?

starswan: thanks - I thought I had caught them all

scruti (approved):

PR #8729, opened 2026-04-17
reset_application_link for vacancy templates

starswan:
## Changes in this PR:

Override reset_application_link for templates as they don't have the application_link prperty


KyleMacPherson (approved):

spec/models/vacancy_template_spec.rb
```diff
@@ -47,4 +47,14 @@
                            other_extension_reason_details])
     end
   end
+
+  describe "resetting before save" do
```
KyleMacPherson: [minor] I don't exactly get how this description relates to the test, is there another way we could describe it? (non blocking!)

PR #8730, opened 2026-04-17
Avoid Publisher API V1 schema from automatically changing and add a drift guard

scruti:
## Changes in this PR:
https://trello.com/c/eBZpVot1

### Context
We've raised in meetings that we need to be careful about changing values in the vacancy model enumerables, as that could break our existing Publisher ATS API contract with existing clients.

Upon a closer look into our API schemas/definition, I've noticed that they're dynamically generated from the Vacancy model values, which is convenient for refreshing the docs without needing developer explicit action, but is dangerous and doesn't respect the immutability that a published API versioned schema should adhere to.

EG: if someone changed a Vacancy model constant (e.g. splitting `:secondary` into two phases), the rswag schema would silently regenerate with the new values, breaking existing API consumers who could no longer submit the old value. 

### We want to:
- Do not automatically change the published schema clients are built against when our Vacancy model changes.
- Make the developer aware when changes to the Vacancy model would break our existing API schema.
- Force the developer to resolve this, providing some guidance.

### We achieve it by
- Replacing all dynamic model references in `spec/swagger_helper.rb` (e.g. `Vacancy.phases.keys`) with static frozen `V1_*` constants, decoupling the published V1 API contract from future Vacancy model changes
- Adding `spec/swagger/v1/enum_consistency_spec.rb` — a failing guard spec that detects when model constants diverge from the frozen V1 values, with inline instructions for the three remediation scenarios (add value, remove/rename value, create V2)

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/spec/swagger/v1/enum_consistency_spec.rb
+++ b/spec/swagger/v1/enum_consistency_spec.rb
@@ -1,0 +1,55 @@
+require "rails_helper"
+require "swagger_helper"
+
+# This spec guards the V1 API contract against silent breakage caused by changes to Vacancy model constants.
+#
+# The ATS API V1 schema defines a fixed set of allowed enum values for fields like phases, key_stages, etc.
+# These are stored as frozen constants (V1_PHASES, V1_KEY_STAGES, ...) in swagger_helper.rb, deliberately
+# decoupled from the Vacancy model constants. This decoupling means a model change does NOT automatically
+# change the API contract — which is correct, because the API is a published, versioned interface.
+#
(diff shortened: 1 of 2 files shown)


KyleMacPherson (approved):
Good points. LGTM. Thanks for the comments in spec/swagger/v1/enum_consistency_spec.rb, think they will be very useful going foward!

PR #8744, opened 2026-04-22
remove and comment gems in gemfile

starswan:
## Trello card URL

https://trello.com/c/ABowv38i/2689-introduce-tool-in-ci-if-possible-to-check-for-unused-gems

## Changes in this PR:

Remove some gems, add comments about most others to describe their purpose

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,6 +93,7 @@ gem "paper_trail-globalid"
 # search keyword parsing into filters
 gem "parslet"
 gem "pg"
+#  vector searches in postgres used for vacancy search weightings
 gem "pg_search"
 # PDF table support
 gem "prawn-table"
@@ -102,8 +103,10 @@ gem "rack-cors"
 gem "rails", RAILS_VERSION # Explicitly declare rails so we can do a "bundle update rails" when needed.
 gem "rails_semantic_logger"
 gem "recaptcha"
+# Geographic point conversions
 gem "rgeo-geojson"
 gem "rgeo-proj4"
+# open API docs for external client integrations
 gem "rswag-api"
 gem "rswag-ui"
 # more limiting of Govuk Notify API requests
@@ -175,7 +178,6 @@ group :test do
   gem "fastimage"
   gem "mock_redis"
   gem "rack_session_access"
-  gem "redis-client"
   # maintained fork of rspec-retry
   gem "rspec-rebound"
   gem "selenium-webdriver"
--- a/Gemfile
+++ b/Gemfile
@@ -178,6 +178,8 @@ group :test do
   gem "fastimage"
   gem "mock_redis"
   gem "rack_session_access"
+  # needed to support mock_redis
+  gem "redis-client"
   # maintained fork of rspec-retry
   gem "rspec-rebound"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -21,88 +21,113 @@ gem "activerecord-import"
...
-gem "sanitize"
 gem "sentry-rails"
 gem "sentry-ruby"
 gem "sentry-sidekiq"
 # sidekiq 7 needs Redis 6.2.x which isn't available on Azure (yet)
 gem "sidekiq", "<7"
 gem "sidekiq-cron"
+# throttle sidekiq requests to avoid overwhelming the Govuk Notify API
 gem "sidekiq-limit_fetch"
-gem "skylight"
```
scruti, line 97: 🙋‍♂️ I think this one is needed for our [Skylight performance monitoring](https://www.skylight.io/app/applications/xsMWeSG9ned8/recent/6h/endpoints?query=agony%3A1..3)

Gemfile
```diff
@@ -21,88 +21,113 @@ gem "activerecord-import"
 gem "activerecord-postgis-adapter", ">= 10.0.1"
 gem "activerecord-session_store"
 gem "active_storage_validations"
-gem "addressable"
```
scruti, line 24: Since we explicitly use `Addressable::URI` calls within our project codebase. 
Shouldn't we keep this? 

They seem to keep working only as `adressable` is a dependency of other gems we use. But I believe this could cause removing `Capybara` or whatever, and suddenly unrelated  `Addressable::URI` calls breaking?

Gemfile
```diff
@@ -21,88 +21,113 @@ gem "activerecord-import"
...
+# slugs for vacancies
 gem "friendly_id"
+# guidance pages content
 gem "front_matter_parser"
+# postcode and location distance calculations
 gem "geocoder"
+# These 2 can be removed when we complete move to Azure attachment checking
 gem "google-apis-drive_v3"
 gem "google-apis-indexing_v3"
-gem "google-cloud-bigquery"
```
scruti, line 43: Shouldn't we keep this one? We are instantiating `Google::Cloud::Bigquery` directly within our codebase. It is a dependency from Analytics, but since we directly have a `config/initializers/big_query.rb` file, it seems our project directly depends on it.

Gemfile
```diff
@@ -21,88 +21,113 @@ gem "activerecord-import"
...
 gem "sentry-sidekiq"
 # sidekiq 7 needs Redis 6.2.x which isn't available on Azure (yet)
 gem "sidekiq", "<7"
 gem "sidekiq-cron"
+# throttle sidekiq requests to avoid overwhelming the Govuk Notify API
 gem "sidekiq-limit_fetch"
-gem "skylight"
 gem "slim-rails"
 gem "turbo-rails"
-gem "validate_url"
```
scruti, line 100: This also has a direct invocation from our codebase: https://github.com/DFE-Digital/teaching-vacancies/blob/main/app/form_models/publishers/job_listing/application_link_form.rb#L2

scruti (approved):

PR #8753, opened 2026-04-24
Replace slack notifications with teams notifications for workflows

rgwenter-mobi:
### Trello card

https://trello.com/c/5mophRLT/2659-add-teams-channel-gh-notifications

### Context

Remove slack notifications from workflows
Add any missing teams notifications

### Changes proposed in this pull request

Workflow updates

### Guidance to review

see successful workflows run by this PR

### DIFF
--- a/.github/workflows/accessibility_tests.yml
+++ b/.github/workflows/accessibility_tests.yml
@@ -51,12 +51,6 @@ jobs:
           role-duration-seconds: 3600
           role-skip-session-tagging: true
 
-      - name: Get secrets from AWS ParameterStore
-        if: failure()
-        uses: dkershner6/aws-ssm-getparameters-action@4fcb4872421f387a6c43058473acc1b22443fe13 # Pinned at v2.0.3
-        with:
-          parameterPairs: "/teaching-vacancies/github_action/infra/slack_webhook = SLACK_WEBHOOK"
-
       - name: Set environment variables from build output
         if: failure()
         run: |
@@ -64,16 +58,14 @@ jobs:
 
       - name: Notify twd_tv_dev channel on Pa11y test failures
         if: failure()
-        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661 # Pinned at v2.3.3
-        env:
-          SLACK_CHANNEL: twd_tv_dev
-          SLACK_USERNAME: CI Accessiblity Tests
-          SLACK_TITLE: Pa11y tests failures
-          SLACK_MESSAGE: |
+        uses: DFE-Digital/github-actions/send-to-teams-channel@master
+        with:
+          teams-webhook-url: ${{ secrets.TEAMS_WEBHOOK_URL }}
+          title: Pa11y tests failures
+          service: ${{ vars.TEAMS_MSG_SERVICE_NAME }}
+          message: |
             Pa11y accessibility tests failed against QA environment
             See: <${{ env.LINK_TO_RUN }}|Workflow run>
-          SLACK_WEBHOOK: ${{env.SLACK_WEBHOOK}}
-          SLACK_COLOR: failure
 
   lighthouse:
     name: Lighthouse tests
@@ -101,11 +93,6 @@ jobs:
           role-to-assume: Deployments
           role-duration-seconds: 3600
           role-skip-session-tagging: true
-
-      - name: Get secrets from AWS ParameterStore
-        uses: dkershner6/aws-ssm-getparameters-action@4fcb4872421f387a6c43058473acc1b22443fe13 # Pinned at v2.0.3
-        with:
-          parameterPairs: "/teaching-vacancies/github_action/infra/slack_webhook = SLACK_WEBHOOK"
 
       - name: Set lighthouse config values
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -299,13 +291,13 @@ jobs:
...
-        SLACK_TITLE: Deployment failure
-        SLACK_MESSAGE: 'Deployment of Docker image ${{ env.DOCKER_IMAGE_TAG }} to ${{ env.ENVIRONMENT }} - unsuccessful <!channel>'
-        SLACK_WEBHOOK: ${{env.SLACK_WEBHOOK}}
+      uses: DFE-Digital/github-actions/send-to-teams-channel@master
+      with:
+        teams-webhook-url: ${{ secrets.TEAMS_WEBHOOK_URL }}
+        title: Deployment failure
+        service: ${{ vars.TEAMS_MSG_SERVICE_NAME }}
+        message: |
+          'Deployment of Docker image ${{ env.DOCKER_IMAGE_TAG }} to ${{ env.ENVIRONMENT }} - unsuccessful <!channel>'
```
RMcVelia: I think the message fails if you put it in '' or ""
plus you can drop the <!channel> as that's slack only

.github/workflows/build_and_deploy.yml
```diff
@@ -223,18 +222,16 @@ jobs:
...
-          SLACK_TITLE: Build failure
-          SLACK_MESSAGE: |
+        uses: DFE-Digital/github-actions/send-to-teams-channel@master
+        with:
+          teams-webhook-url: ${{ secrets.TEAMS_WEBHOOK_URL }}
+          title: Build failure
+          service: ${{ vars.TEAMS_MSG_SERVICE_NAME }}
+          message: |
             Build failure on branch ${{env.GIT_BRANCH}} <!channel>
             See: <${{ env.LINK_TO_RUN }}|Workflow run>
```
RMcVelia: I'm not sure this will resolve as expected 
i.e. <${{ env.LINK_TO_RUN }}|Workflow run>
might need to test it

rgwenter-mobi: thanks, it doesn't 😞. 
My take on this is we're going to need to refactor the `send-to-teams-channel` action to handle custom links like this, we're using a [TextRun](https://learn.microsoft.com/en-us/adaptive-cards/schema-explorer/text-run) which doesn't support markdown and would need an `selectAction` block to do so.
I've changed the message so the link is included along with info on it but it is text, not a clickable link.

image

[sample message including link to branch](https://teams.microsoft.com/l/message/19:406d7e69161049328e4dd767b3d736fa@thread.tacv2/1777297137834?tenantId=fad277c9-c60a-4da1-b5f3-b3b8b34a82f9&groupId=5e035efe-5b2b-491b-9e3e-d832445e4ad1&parentMessageId=1777297137834&teamName=Schools%20Digital&channelName=TV%20Dev%20Notifications&createdTime=1777297137834)

RMcVelia: ah, that url is already in the 'workflow run' block just above, so we shouldn't need that in the message itself

rgwenter-mobi: you're right, I should have seen that 🤦 , have removed links to workflow run from the message part.
They also have links to the PR, Review app and Test results in the messages, adding as comments now...

.github/workflows/accessibility_tests.yml
```diff
@@ -137,15 +124,13 @@ jobs:
...
+          title: LightHouse tests results
+          service: ${{ vars.TEAMS_MSG_SERVICE_NAME }}
+          message: |
             LightHouse accessibility and performance tests ran against the  QA environment
-            <${{ env.INDEX_RESULT }}|Index page results>
-            <${{ env.JOBS_RESULT }}|Jobs page results>
-            <${{ env.SCHOOLS_RESULT }}|Schools page results>
-          SLACK_WEBHOOK: ${{ env.SLACK_WEBHOOK }}
-          SLACK_COLOR: good
+            Index page results - ${{ env.INDEX_RESULT }}
```
rgwenter-mobi, line 133: links included in message but as text only

.github/workflows/delete_review_app.yml
```diff
@@ -122,15 +117,12 @@ jobs:
...
+        teams-webhook-url: ${{ secrets.TEAMS_WEBHOOK_URL }}
+        title: Delete review app failure
+        service: ${{ vars.TEAMS_MSG_SERVICE_NAME }}
+        message: |
           Failed deletion of review app PR ${{env.PR_NUMBER}}
-          See: <${{ env.LINK_TO_RUN }}|Workflow run> - <${{ env.LINK_TO_PR }}|Pull request> - <${{ env.LINK_TO_APP }}|Review app>
-          <!channel>
-        SLACK_WEBHOOK: ${{env.SLACK_WEBHOOK}}
-        SLACK_COLOR: ${{ job.status }}
+          Pull Requst: ${{ env.LINK_TO_PR }}
```
rgwenter-mobi, line 127: links included in message but as text only

RMcVelia (approved):

PR #8778, opened 2026-04-30
Remove ignoring columns from Profile Employments

scruti:
## Trello card URL
Resolving an issue raised by a Performance Analyst.

## Changes in this PR:

Having two models using the same DB table and one of the models ignoring a subset of the table attributes results in DfE Analytics entities missing the keys in some of the employment rows, but not in others. What is effectively an inconsistent schema for the employments table in BigQ.

The PA alerted us about the issue and how BigQ as it is configured in Analytics cannot deal with this, breaking the analytics pipelines.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/profile_employment.rb
+++ b/app/models/profile_employment.rb
@@ -4,4 +4,17 @@ class ProfileEmployment < EmploymentRecord
   has_encrypted :organisation, :job_title
 
   self.table_name = "employments"
+
+  # This model was originally ignoring the following columns from the employments table:
+  # - main_duties
+  # - subjects
+  # -reason_for_leaving
+  # These columns have no use in the profile employment model.
+  #
+  # Unfortunately, this results in the DfE Analytics entity updates not including those keys, causing an inconsistency in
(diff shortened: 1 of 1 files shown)


app/models/profile_employment.rb
```diff
@@ -4,7 +4,4 @@ class ProfileEmployment < EmploymentRecord
   has_encrypted :organisation, :job_title
 
   self.table_name = "employments"
-
-  # these fields are not used in the profile version
-  self.ignored_columns += %i[main_duties subjects reason_for_leaving]
```
starswan, line 9: I would be tempted to add this back as a comment somehow , explaining that we would like to model it like this but BigQ prevents us from doing so

starswan (approved):
Suggestion about a comment

app/models/profile_employment.rb
scruti: Will add the comment. I also suspect this will be a non-issue once we move the DB Analytics feeding to Airbyte.

PR #8795, opened 2026-05-07
Antivirus final steps

KyleMacPherson:
## Trello card URL

https://trello.com/c/KHiroyZt/2805-use-microsoft-defender-for-storage-instead-of-google-drive-for-antivirus-scanning-for-other-uploads

## Changes in this PR:

Switch from using google drive to using microsoft defender for storage antivirus. This PR removes the google drive antivirus flow code and uses microsoft defender for cloud for the final 3 uploads (organisation photo, organisation logo, and uploaded reference).

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/validators/form_file_validator.rb
```diff
@@ -1,16 +1,12 @@
-# TODO: - this whole class could be replaced with https://github.com/igorkasyanchuk/active_storage_validations
-# Possibly this could happen when we move over to Azure storage, which will do the virus check for us rather than the
-# upload/download from google cloud as we currently have it.
```
starswan, line 3: should we keep this todo in here but slightly modified? As the virus check has now been moved, presumably we could move to that gem at a later date?

app/helpers/application_helper.rb
```diff
@@ -44,4 +44,10 @@ def consented_to_extra_cookies?
   def kcsie_link
     "https://www.gov.uk/government/publications/keeping-children-safe-in-education--2"
   end
+
+  def malware_scan_clean?(attachment)
+    return false unless attachment.attached?
+
+    attachment.blob.malware_scan_clean?
```
starswan, line 51: I'm not really sure about the location of this method. 'helpers' are 'view helpers' in rails, and this feels like object logic. Is it possible to mix it into ActiveStorage::Blob or somewhere similar?

starswan (approved):
1 small non-blocking comment

app/validators/form_file_validator.rb
scruti: Created a tech debt ticket for it: https://trello.com/c/Tq8Kmgs3/1738-replace-formfilevalidations

PR #8814, opened 2026-05-11
Religious character part 1 - add mappings from GIAS

starswan:
## Changes in this PR:

Add religious mappings from GIAS before using and relying on them to be populated

### DIFF
--- a/app/services/gias/import_schools_and_local_authorities.rb
+++ b/app/services/gias/import_schools_and_local_authorities.rb
@@ -13,6 +13,10 @@ class Gias::ImportSchoolsAndLocalAuthorities
   class ImportFailure < StandardError
   end
 
+  # New GIAS team now have some data dictionary information available at
+  # https://dfedigital.atlassian.net/wiki/spaces/GTP/pages/6155337742/Data+Dictionary
+  # but it's not quite complete/definitive
+  # (e.g establishment status says open/closed/proposed when reality is 'open proposed to close' and 'proposed to open')
   def call
     log_benchmark("Importing schools and local authorities") do
       import_errors = Gias::Data.new(SCHOOLS_AND_LOCAL_AUTHORITIES_CSV).each_slice(BATCH_SIZE).flat_map do |group|
@@ -118,6 +122,7 @@ class Gias::ImportSchoolsAndLocalAuthorities
       town: row["Town"],
       phase: row["PhaseOfEducation (code)"].to_i,
       url: Addressable::URI.heuristic_parse(row["SchoolWebsite"]).to_s,
+      religious_character: row.fetch("ReligiousCharacter (name)").presence || "None",
       gias_data: row.to_h,
     }.merge(school_location_data(row)).transform_values(&:presence)
   end
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -161,6 +161,7 @@
   - current_sign_in_ip
   :organisations:
   - gias_data_hash
+  - religious_character
   - slug
   - searchable_content
   :vacancy_analytics:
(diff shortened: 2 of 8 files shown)


config/analytics_blocklist.yml
```diff
@@ -161,6 +161,11 @@
   - current_sign_in_ip
   :organisations:
   - gias_data_hash
+  - religious_character
+  - number_of_pupils
+  - trust_school_flag_code
+  - trusts_code
+  - school_capacity
```
scruti: 🙋‍♂️ I suspect the only reason `gias_data_hash` was blocked from being sent to analytics was to avoid duplication of this computed field vs the raw `gias_data` JSON being sent already to Analytics.

As we are now removing both `gias_data` and `gias_data_hash`, I would assume these new fields should be included in the `analytics.yml` rather than blocked.

(And this is going to be one we need to catch up on with our Perf. Analyst!)

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -118,6 +122,11 @@ def school_data(row) # rubocop:disable Metrics/MethodLength
       town: row["Town"],
       phase: row["PhaseOfEducation (code)"].to_i,
       url: Addressable::URI.heuristic_parse(row["SchoolWebsite"]).to_s,
+      religious_character: row.fetch("ReligiousCharacter (name)").presence || "None",
+      number_of_pupils: row.fetch("NumberOfPupils"),
+      school_capacity: row.fetch("SchoolCapacity"),
+      trust_school_flag_code: row.fetch("TrustSchoolFlag (code)"),
+      trusts_code: row.fetch("Trusts (code)"),
       gias_data: row.to_h,
     }.merge(school_location_data(row)).transform_values(&:presence)
```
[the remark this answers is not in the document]
starswan: This comment is just plain wrong - I *want* the crash to happen early if the CSV schema changes. And it's interesting that the LLM thinks that a column can be missing, because Ruby's CSV library makes that impossible (IIRC)

config/analytics_blocklist.yml
starswan: Yes I spoke to John today, and I probably went on at length far too much about our interface with analytics, and GIAS data in general...

app/services/gias/import_schools_and_local_authorities.rb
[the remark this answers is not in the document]
KyleMacPherson: I don't have a problem with fetch raising the KeyError and am fine with this change.

If I was going to nitpick, I wonder if the inconsistency could potentially be a bit confusing in the future. Maybe a future dev would wonder why we treated these fields differently? We could add a method to validate that we received all required headers maybe. Don't really think it's a big deal/necessary though.

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -118,6 +122,11 @@ def school_data(row) # rubocop:disable Metrics/MethodLength
       town: row["Town"],
       phase: row["PhaseOfEducation (code)"].to_i,
       url: Addressable::URI.heuristic_parse(row["SchoolWebsite"]).to_s,
+      religious_character: row.fetch("ReligiousCharacter (name)").presence || "None",
+      number_of_pupils: row.fetch("NumberOfPupils"),
+      school_capacity: row.fetch("SchoolCapacity"),
+      trust_school_flag_code: row.fetch("TrustSchoolFlag (code)"),
+      trusts_code: row.fetch("Trusts (code)"),
```
KyleMacPherson: Not 100% sure on what these fields will be used for but I think the source of truth for school membership to trusts/school groups at the moment is imported in app/services/gias/import_trusts.rb (TRUST_MEMBERSHIPS_CSV). These values also don't seem to be related to the religious mapping, do we need them?

config/analytics_blocklist.yml
KyleMacPherson: I think these fields be removed from the blocklist now
  - number_of_pupils
  - trust_school_flag_code
  - trusts_code
  - school_capacity

KyleMacPherson (approved):

PR #8827, opened 2026-05-13
Trn ab test

starswan:
## Trello card URL

https://trello.com/c/vlnrtehQ/2807-schools-and-colleges-ab-test-ask-for-trns-at-the-moment-the-jobseeker-clicks-apply

## Changes in this PR:

AB Test for apply journey


scruti:
Note: Release blocked until we finish migration from Sidekiq to SolidQueue and remove Sidekiq. More details on Trello ticket comments.

scruti:
It seems that `Split` depends on `Redis`, and this is not configurable.

We plan to remove Redis in the coming months and migrate our cache to SolidCache, so it doesn't seem that this tool is feasible for our service. Have you considered a DB backed alternative like field_test? https://github.com/ankane/field_test ?

scruti (changes requested):
I don't think an A/B tool that has a hard dependency on Redis is going to be viable for us, since we're moving away from Redis in the coming months.

scruti (commented):
Seems Copilot raised 2 a couple things that need to be actioned or discarded if not relevant. 

One about a possible error when redirecting to an external link. 
And second about the test form being behind a login and not available for non-registered jobseekers trying to apply for the vacancy.

Guardfile
```diff
@@ -97,7 +97,7 @@ guard :rspec, rspec_options do
   end
 end
 
-guard :rubocop, cli: ["-A", "--no-parallel", "--no-server"], all_on_start: false do
+guard :rubocop, cli: ["-A", "--no-parallel"], all_on_start: false do
```
scruti: Needs a rebase, I guess? (This was added in a separate PR, if I recall right?

scruti (approved):
LGTM!

PR #8833, opened 2026-05-15
New gias import

starswan:
GIAS imports using discards rather than complex queries - pre cursor to colleges page

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -48,7 +48,6 @@ class Organisation < ApplicationRecord
 
   has_many :vacancy_templates, dependent: :destroy
 
-  scope :not_closed, -> { where.not(establishment_status: CLOSED_ESTABLISHMENT_STATUSES) }
   scope :schools, -> { where(type: "School") }
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
@@ -68,7 +67,7 @@ class Organisation < ApplicationRecord
 
   scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES).or(where(detailed_school_type: nil)) }
 
-  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts).registered_for_service }
+  scope :visible_to_jobseekers, -> { schools.kept.not_out_of_scope.or(Organisation.trusts).registered_for_service }
 
   scope :only_faith_schools, -> { where.not(religious_character: NON_FAITH_RELIGIOUS_CHARACTER_TYPES) }
 
@@ -129,7 +128,7 @@ class Organisation < ApplicationRecord
 
     if local_authority_code && local_authorities_extra_schools
       school_urns = local_authorities_extra_schools[local_authority_code]
-      School.where(urn: school_urns)
+      School.kept.where(urn: school_urns)
     else
       School.none
     end
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -51,7 +51,6 @@ class Organisation < ApplicationRecord
 
   has_many :vacancy_templates, dependent: :destroy
 
-  scope :not_closed, -> { where.not(establishment_status: CLOSED_ESTABLISHMENT_STATUSES) }
   scope :schools, -> { where(type: "School") }
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -51,6 +51,7 @@ class Organisation < ApplicationRecord
 
   has_many :vacancy_templates, dependent: :destroy
 
+  scope :not_closed, -> { where.not(establishment_status: CLOSED_ESTABLISHMENT_STATUSES) }
   scope :schools, -> { where(type: "School") }
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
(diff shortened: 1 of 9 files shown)


app/models/organisation.rb
```diff
@@ -51,7 +60,7 @@ class Organisation < ApplicationRecord
 
   scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES).or(where(detailed_school_type: nil)) }
 
-  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts).registered_for_service }
+  scope :visible_to_jobseekers, -> { schools.kept.not_out_of_scope.or(Organisation.trusts).registered_for_service }
```
KyleMacPherson: We are going to need some kind of backfill task for this I think. Otherwise between the time this is deployed and the time that the next GIAS import is run all schools will be visible to jobseekers.

app/models/school.rb
```diff
@@ -2,21 +2,35 @@ class School < Organisation
...
+                        INDEPENDENT_SCHOOL_TYPE,
+                        "Special schools",
+                        "Universities",
+                        ACADEMY_TYPE,
+                        FREE_SCHOOL_TYPE,
+                        "Welsh schools",
+                        "Other types",
+                        COLLEGE_SCHOOL_TYPE,
+                        "Online provider"].freeze
+  EXCLUDED_SCHOOL_TYPES = ["Universities", "Welsh schools", "Online providers"].freeze
```
KyleMacPherson: Should this be `Online provider`? (singular). Have we tested that this change succesfully recreates the behaviour of our gias imports?

  ```suggestion
  EXCLUDED_SCHOOL_TYPES = ["Universities", "Welsh schools", "Online provider"].freeze
  ```

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -1,141 +1,150 @@
...
-      import_errors = Gias::Data.new(SCHOOLS_AND_LOCAL_AUTHORITIES_CSV).each_slice(BATCH_SIZE).flat_map do |group|
-        group.each do |row|
-          local_authorities.add(group_data(row))
-          schools.push(school_data(row))
-          memberships.push(membership_data(row))
+  class << self
+    def call
+      log_benchmark("Importing schools and local authorities") do
+        import_errors = Gias::Data.new(SCHOOLS_AND_LOCAL_AUTHORITIES_CSV).each_slice(BATCH_SIZE).flat_map do |group|
+          import_group  group
```
KyleMacPherson: This part was hard for me to understand before digging into what this code actually did. I think this is because I was surprised that the return value of a method called import_group is the failed instances (I know import_batch previously did the same!). I feel like naming it something like import_batch_and_return_failures would make it a easier to understand without refactoring the code too much.

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -1,141 +1,150 @@
...
+        end
+        memberships.push(membership_data(row))
+      end
+      import_batch(local_authorities, schools, memberships).tap do
+        School.where(urn: discarded.map { |school_row| school_row.fetch(:urn) }).discard_all
+      end
+    end
 
-  attr_reader :local_authorities, :schools, :memberships
+    def import_batch(local_authorities, schools, memberships)
```
KyleMacPherson, line 74: maybe `import_batch_and_return_failed_instances`?

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -1,141 +1,150 @@
...
-    SchoolGroup.import(
-      local_authorities.to_a,
-      on_duplicate_key_update: {
-        conflict_target: [:local_authority_code],
-        columns: local_authorities.first.keys,
-      },
-    )
-  end
+    def import_schools(schools)
+      imported_schools = schools.map { |s| s.merge(discarded_at: nil) }
```
KyleMacPherson: Is there a risk that if the job crashes before we run `discard_all` later, every school in this batch will be left as `kept`? If so, would it be better to set `discarded_at` on a row-by-row basis rather than setting everything as kept to start with, then bulk discarding excluded/closed schools afterwards?

KyleMacPherson (commented):
This looks like a good improvement. I have left a few comments but my major concern is that we don't know if this is successfully recreating the behaviour of the old import (as far as i know). How do we know it keeps the same schools visible to jobseekers as on main?

app/models/organisation.rb
starswan: We need to split this into 2 again like we did for the religious character. Good catch.

app/models/school.rb
starswan: Its actually online providers

app/services/gias/import_schools_and_local_authorities.rb
starswan: importing isn't likely to crash (if validations fail, that's where the failed instances come from) - but I have now wrapped the batch in a transaction (in the other PR) to make doubly-sure)

app/models/school.rb
KyleMacPherson: So we have `Online provider` in `VALID_SCHOOL_TYPES` and `EXCLUDED_DETAILED_SCHOOL_TYPES` and `Online providers` in `EXCLUDED_SCHOOL_TYPES`

starswan: yep - there are 2 distinct fields. school_type is 'online providers' - all of those have detailed_school_type of 'online provider'

app/services/gias/import_schools_and_local_authorities.rb
starswan: I thought it seemed reasonable that the import returned the failures - thats what ActiveRecord#import does

app/models/school.rb
starswan: should we tighten this even more to make it clear? Every 'type' has a fixed set of 'subtypes', so we could validate them easily

PR #8841, opened 2026-05-19
Flakey dwp spec

starswan:
## Changes in this PR:

Fix up flakey DWP spec

### DIFF
--- a/spec/services/vacancies/export/dwp_find_a_job/closed_early_spec.rb
+++ b/spec/services/vacancies/export/dwp_find_a_job/closed_early_spec.rb
@@ -4,31 +4,8 @@ RSpec.describe Vacancies::Export::DwpFindAJob::ClosedEarly do
   describe "#call" do
     subject { described_class.new("2024-05-01") }
 
-    let(:vacancy_expired_old) { create(:vacancy, publish_on: 4.days.ago, expires_at: 2.days.ago) }
-    let(:vacancy_manually_expired) do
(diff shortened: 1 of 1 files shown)


spec/services/vacancies/export/dwp_find_a_job/closed_early_spec.rb
```diff
@@ -41,6 +41,9 @@
       vacancy_manually_expired2
       vacancy_naturally_expired
 
+      allow(Vacancies::Export::DwpFindAJob::ClosedEarlyVacancies::Query)
```
KyleMacPherson, line 17: So are we happy to not test the query logic in this one? (I think that's ok because it's covered in closed_early_vacancies/query_spec.r) If so we can probably simplify the set up as we don't really need all those vacancies here right?

KyleMacPherson (approved):

spec/services/vacancies/export/dwp_find_a_job/closed_early_spec.rb
starswan: Yes that was my thought. We already test the query, so we're really just testing the XML output which is sometimes in a different order

KyleMacPherson: That makes sense! We could get rid of `vacancy_expired_old` and `vacancy_naturally_expired` then I think

PR #8842, opened 2026-05-19
new GIAS import strategy phase 1

starswan:
## Changes in this PR:

Part 1 of new GIAS import strategy to support colleges in an easier to understand manner.

Have tested that old and new strategies produce the same answer using live data set locally


KyleMacPherson (approved):
Nice!

PR #8853, opened 2026-05-26
fix issue with incoming DSILogin records failing 

starswan:
## Trello card URL

https://trello.com/c/67sAt6O5/2873-bug-updateusersindb-crashing-and-possibly-preventing-future-updates

## Changes in this PR:

Some publisher records have trailing spaces in their emails, or their emails have missing MX records. Avoid strict validation for publishers coming in from sync with DSI as these errors can't be fixed easily

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -14,7 +14,8 @@ class Publisher < ApplicationRecord
 
   has_encrypted :family_name, :given_name
 
-  validates :email, email_address: true, if: -> { email_changed? } # Allows data created prior to validation to still be valid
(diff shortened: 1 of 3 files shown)


app/models/publisher.rb
```diff
@@ -14,7 +14,8 @@ class Publisher < ApplicationRecord
 
   has_encrypted :family_name, :given_name
 
-  validates :email, email_address: true, if: -> { email_changed? } # Allows data created prior to validation to still be valid
+  # don't use strict MX validation for publisher email - they come from DSI Login so can't really be fixed
```
scruti, line 17: Thanks for the comment

scruti (approved):
LGTM

PR #8854, opened 2026-05-26
Fix legacy integrations duplicating vacancies

scruti:
## Trello card URL

Reported from a school getting hundreds of duplicated adverts: https://trello.com/c/SKD4W2Si/2881-bug-hundreds-of-duplicated-instances-for-a-school-vacancy

## Changes in this PR:

When the legacy integrations source/feed has leading or trailing spaces in the external reference value (EG: external_reference: " CODE_01") it resulted with the same vacancy being re-created as a new vacancy in each import attempt (once per hour!).

This happened because the external_reference gets normalised by Rails ActiveRecord when creating the new vacancy (EG: "CODE_01"), and further pulls calling "find_or_create_by" does not match "CODE_01" with " CODE_01", resulting in a new record instead of matching the existing one.

Resolution: We are stripping possible spaces from the external reference when matching against existing data in our service.

### The offending value in the source:
image

### The consequence:
image

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/services/vacancies/import/sources/fusion.rb
+++ b/app/services/vacancies/import/sources/fusion.rb
@@ -23,7 +23,7 @@ class Vacancies::Import::Sources::Fusion
       # An external vacancy is by definition always published
       v = PublishedVacancy.find_or_initialize_by(
         external_source: SOURCE_NAME,
-        external_reference: result["reference"],
+        external_reference: result["reference"]&.strip,
       )
 
       # Consider publish_on date to be the first time we saw this vacancy come through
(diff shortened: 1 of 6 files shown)


app/services/vacancies/import/sources/fusion.rb
```diff
@@ -23,7 +23,7 @@ def each
       # An external vacancy is by definition always published
       v = PublishedVacancy.find_or_initialize_by(
         external_source: SOURCE_NAME,
-        external_reference: result["reference"],
+        external_reference: result["reference"]&.strip,
```
starswan, line 26: external_reference is surely a mandatory field? So we don't need the 'safe_nav' - its more like a 'fetch'. Shouldn't we fail if they send XML which doesn't match the schema?

scruti: This would allow us to rely on `ExternalVacancyValidator` to validate the `external_reference` presence, resulting in a new record in `FailedImportedVacancy` instead of an uncontrolled exception blowing up the whole feed import.

starswan (approved):

PR #8860, opened 2026-05-27
Solid queue Part III

starswan:
## Trello card URL

https://trello.com/c/WgknYrGz/2870-spike-could-we-move-from-sidekiq-to-solidqueue-dependency-of-trn-ab-test

## Changes in this PR:

Migrate to solid queue


spec/configuration/scheduled_jobs_spec.rb
```diff
@@ -8,7 +8,7 @@
     ApplicationJob.descendants.map(&:name)
   end
   let(:gem_jobs) { %w[DfE::Analytics::Jobs::EntityTableCheckJob] }
-  let(:scheduled_jobs) { YAML.load_file("./config/schedule.yml").map { |_, v| v["class"] }.uniq }
+  let(:scheduled_jobs) { YAML.load_file("./config/recurring.yml").map { |_, v| v["class"] }.uniq.compact }
```
scruti, line 11: Should we rename this spec file?

.devcontainer/docker-compose.yml
```diff
@@ -46,6 +44,18 @@ services:
...
+    restart: unless-stopped
+    volumes:
+      - postgres_data_dbq:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_DB: postgres
+      POSTGRES_PASSWORD: postgres
+    ports:
+      - "5443:5432"
+
```
scruti: Thanks for including the devcontainers 👏

config/analytics_blocklist.yml
```diff
@@ -22,28 +22,141 @@
...
+  - queue_name
+  - priority
+  - concurrency_key
+  - expires_at
+  - created_at
   :active_storage_variant_records:
   - id
   - blob_id
   - variation_digest
-  :action_text_rich_texts:
```
scruti, line 29: Is there any reason for all these pre-existing value changes? Doesn't seem to follow an alphabetical order or anything. I'm struggling to follow up on this shuffling (besides the new solid_queue values)

config/queue.yml
```diff
@@ -0,0 +1,26 @@
+default: &default
+  dispatchers:
+    - polling_interval: 1
+      batch_size: 500
+  workers:
+    - queues: "*"
```
scruti: If I understood the SolidQueue Readme correctly, I suspect `"*"` would cause this worker to read from any queue (including `low`, `notify`, `dbintensive`, `jobalerts`). So maybe we should use `default` as per this example? https://github.com/rails/solid_queue#performance-considerations

PR #8867, opened 2026-05-29
Work history word counts

KyleMacPherson:
## Trello card URL
https://trello.com/c/d07sBlzu/2840-schools-review-guidance-on-how-brief-the-work-history-section-should-be

## Changes in this PR:

This change adds a word counter to work history sections, and validates that the user cannot enter more than the word count.

### After
Screenshot 2026-06-01 at 15 11 15
Screenshot 2026-06-01 at 15 10 58
Screenshot 2026-06-01 at 15 10 47

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/models/employment_record.rb
```diff
@@ -3,6 +3,9 @@
 class EmploymentRecord < ApplicationRecord
   self.abstract_class = true
 
+  MAIN_DUTIES_MAX_WORDS = 150
+  REASON_FOR_LEAVING_MAX_WORDS = 50
```
scruti: It looks like these 2 constants are only relevant to `Employment` model, as `ProfileEmployment` ignores those.

Shouldn't they be defined in `Employment` rather than in this abstract class?

config/locales/jobseekers/employment_form.yml
```diff
@@ -36,8 +40,12 @@ en:
               present: End date cannot be entered for current role
             job_title:
               blank: Enter your job title
+            main_duties_words:
+              too_long: Main duties must be %{count} words or less
             organisation:
               blank: Enter a school or other organisation
+            reason_for_leaving_words:
+              too_long: Reason for leaving role must be %{count} words or less
```
scruti: ❓ Question:
 
Are these used anywhere?
From my understanding, `main_duties` and `reason_for_leaving` are not used by the `ProfileEmployment`.

app/models/employment.rb
```diff
@@ -5,6 +5,8 @@ class Employment < EmploymentRecord
 
   # KSIE dictates that we need a reason_for_leaving even for current role
   validates :reason_for_leaving, :main_duties, presence: true, if: -> { job? }
+  validates :main_duties_words, length: { maximum: MAIN_DUTIES_MAX_WORDS }, if: -> { main_duties.present? }
+  validates :reason_for_leaving_words, length: { maximum: REASON_FOR_LEAVING_MAX_WORDS }, if: -> { reason_for_leaving.present? }
```
scruti: ❓ Question:
Since these validations/errors are not added to `main_duties` and `reason_for_leaving`, wouldn't this cause the error messages upon form submission not to link to the respective fields they should link to?

scruti (commented):
Left a couple comments/questions.

Also, could the PR description include before/after screenshots with the interface changes? 

Got a bit confused with some of the Trello ticket comments regarding what to expect to see in the changes. So having a screenshot with "this is the output" would help to review!
Thanks!

app/models/employment_record.rb
KyleMacPherson: Yep, good spot. I previously had all of this on the employment_record as I didn't realise that we now have this split between employment_records, employments and profile_employments and was a bit confused about the difference between each. Forgot to move this over.

config/locales/jobseekers/employment_form.yml
KyleMacPherson: Another one I forgot to get rid of!

app/models/employment.rb
KyleMacPherson: Good catch!

KyleMacPherson:
> Left a couple comments/questions.
> 
> Also, could the PR description include before/after screenshots with the interface changes?
> 
> Got a bit confused with some of the Trello ticket comments regarding what to expect to see in the changes. So having a screenshot with "this is the output" would help to review! Thanks!

Agreed, very confusing ticket. I had to read the comments to understand what to do! Have updated it with some screenshots now, but basically we are adding word count and validation on the word count to 2 fields on the employments page.

scruti (approved):
LGTM! Thanks for the prompt action on the feedback!

PR #8868, opened 2026-05-29
Refactor malware scan logic into model concern

scruti:
## Trello card URL
https://trello.com/c/OaFUjk43/2894-refactor-of-malware-scanning-codebase

## Changes in this PR:

This PR attempts to organise the existing blob/uploads malware scanning code into a more OOP/Rails friendly approach.

It shouldn't include behavioural changes, only 2 main code grouping changes:

### 1 - All the blob malware-related code gets grouped into a Rails model concern included in the Blob class.

This keeps the same class extension as was already present but:
a) removes the service object and includes that interface in the Blob itself.
b) encapsulates all the malware scanning related code into a model concern, so a single file contains all the related logic.

### 2 - The handling of unsafe attachments is moved away from the Job

The job responsibility is left to trigger/retry the malware scan fetching.

The malware scan from the blob attempts to trigger the appropriate model to handle an unsafe scan result if the model has specific handling for it.

This way, the logic regarding how each unsafe attachment is handled lives within the model that defines the attachment, as this is business logic belonging to that particular model/entity.

### Before
image

### Step 1: Removing service object and grouping everything as a Concern included in Blob model

image

### Step 2: Remove unsafe handling logic from the Job and delegate it to each model

image

## Testing that we keep expected behaviour

image

----
image

----
image

----
image

----
image

### Jobseeker AV testing
image

----

image

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/jobs/fetch_malware_scan_result_job.rb
```diff
@@ -6,54 +6,11 @@ class FetchMalwareScanResultJob < ApplicationJob
...
-      Publishers::ReferenceDocumentMalwareScanMailer.reference_removed(job_application).deliver_later
-    end
-  end
-
-  def notifier_for(attachment_name)
-    case attachment_name
-    when "photo" then Publishers::OrganisationPhotoMalwareScanNotifier
-    when "logo"  then Publishers::OrganisationLogoMalwareScanNotifier
-    end
-  end
```
scruti, line 58: Moved all the handling to the respective models. Following a kind of "This model implements the  `handle_unsafe_attachment` interface" approach.

The idea is to avoid the job from becoming a God object and leave its responsibility to be only the calling & re-attempting until successful.

app/jobs/fetch_malware_scan_result_job.rb
```diff
@@ -6,54 +6,11 @@ class FetchMalwareScanResultJob < ApplicationJob
 
   def perform(blob_id)
     blob = ActiveStorage::Blob.find(blob_id)
-    FetchMalwareScanResult.new(blob: blob).call
-
-    blob.reload
+    blob.process_malware_scan_result!
```
scruti, line 9: Found that the service object `FetchMalwareScanResult.new(blob: blob)` has the OOP equivalent of `blob.fetch_malware_scan_result` (especially when no other dependencies are around). 
So used the opportunity, as I was injecting a concern with MalwareScan behaviour into the Blob class, to also convert this Service Object into one of the methods there.

KyleMacPherson (approved):
Nice, looks great to me!

PR #8869, opened 2026-06-01
2799 fe jobseekers home page accordions

richardpattinson:
## Trello card URL

https://trello.com/c/dY4CNXVc/2799-fe-jobseekers-home-page-accordions

## Changes in this PR:

Implement two new accordions split by school jobs and fe jobs. Deprecate the old landing_page_link_component.rb and replace with a vacancy_counter to make the queries faster. Update content + styling. 

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


app/helpers/landing_pages_helper.rb
```diff
@@ -36,4 +36,251 @@ def linked_ect_status(vacancy)
 
     landing_page_link_or_text({ ect_statuses: [vacancy.ect_status] }, vacancy.ect_status.humanize)
   end
+
+  def landing_page_teaching_roles_list
+    { "teacher-jobs" => "teacher", "head-of-year-or-phase-jobs" => "head_of_year_or_phase", "head-of-department-or-curriculum-jobs" => "head_of_department_or_curriculum", "assistant-headteacher-jobs" => "assistant_headteacher", "deputy-headteacher-jobs" => "deputy_headteacher", "headteacher-jobs" => "headteacher", "sendco-jobs" => "sendco", "other-leadership-roles-jobs" => "other_leadership" }
+  end
```
starswan: IMHO all these methods should just be frozen constants

app/controllers/home_controller.rb
```diff
@@ -1,6 +1,16 @@
...
+    @school_scope = PublishedVacancy.live.in_organisation_ids(Organisation.in_scope_schools.pluck(:id))
+    @role_counts = VacancyCounter.role_counts(scope: @school_scope)
+    @phase_counts = VacancyCounter.phase_counts(scope: @school_scope)
+    @working_pattern_counts = VacancyCounter.working_pattern_counts(scope: @school_scope)
+    @subjects_counts = VacancyCounter.subject_counts(scope: @school_scope)
+    @fe_scope = PublishedVacancy.live.in_organisation_ids(Organisation.colleges.pluck(:id))
+    @fe_role_counts = VacancyCounter.role_counts(scope: @fe_scope)
+    @fe_phase_counts = VacancyCounter.phase_counts(scope: @fe_scope)
+    @fe_working_pattern_counts = VacancyCounter.working_pattern_counts(scope: @fe_scope)
+    @fe_subjects_counts = VacancyCounter.subject_counts(scope: @fe_scope)
```
starswan: I think it would be nice to group these together logically (fe vs school) and then pass the group to the relevant partial as required. A nice use of the Data object maybe? Hopefully that turns the scopes into method level entities rather than view-scope ones

config/landing_pages.yml
```diff
@@ -206,6 +288,348 @@ shared:
     subjects:
       - Spanish
 
+  ### FE Subjects
```
starswan: It seems a shame to duplicate the 'Colleges' type on all of these queries (in a massively complicated config file). It might be better to create a new config file where the organisation type is specified once and the subjects are nested within it somehow?

spec/system/jobseekers/jobseekers_can_view_school_jobs_accordion_spec.rb
```diff
@@ -0,0 +1,85 @@
+require "rails_helper"
+
+RSpec.describe "Jobseeker views the 'School jobs by role, phase and subject' accordion" do
```
starswan: do we still need all these system specs? Could they at least be converted into view specs?

app/services/vacancy_counter.rb
```diff
@@ -0,0 +1,40 @@
...
+    "ICT and Computer Science": %i[ICT Computing],
+    "Politics, Humanities and Social Sciences": [:Politics, :Humanities, :"Social sciences"],
+    "Psychology, Sociology and RE": [:Psychology, :Philosophy, :Sociology, :"Religious education"],
+    Science: %i[Biology Chemistry Physics],
+  }.freeze
+
+  class << self
+    def role_counts(scope:)
+      scope.map(&:job_roles).flatten.group_by(&:itself).symbolize_keys.transform_values(&:count)
+    end
```
starswan, line 21:
```suggestion
    def role_counts(scope:)
        scope.group(:job_roles)
                  .count
                  .transform_keys { |keylist| keylist.map { |k| Vacancy::JOB_ROLES.invert.fetch(k).to_sym } }
                  .map { |keys, value| keys.map { |k| { k => value } } }
                  .flatten
                  .reduce(&:merge)
    end
```

This looks scary but it works - maybe you could find a more elegant solution? It does it with group_by in the DB followed by the transformation to bring it into the original format

config/landing_pages.yml
starswan: IIRC these relationships are already coded in one of the helper constants, so we should be able to use that rather than duplicating the relationships here (DRY)

app/services/vacancy_counter.rb
richardpattinson: I've written an updated query that I think works well - have a look and let me know what you think

app/helpers/landing_pages_helper.rb
```diff
@@ -36,4 +36,290 @@ def linked_ect_status(vacancy)
...
+      woodworking-joinery-and-carpentry-fe-jobs
+    ],
+  ].freeze
+
+  def landing_page_fe_subjects_list
+    FE_SUBJECTS_LIST
+  end
+
+  def landing_page_fe_subjects_columns
+    FE_SUBJECTS_COLUMNS
```
starswan: didn't know this was possible. Quite a clever idea!

app/services/vacancy_counter.rb
starswan: much nicer!

app/services/vacancy_counter.rb
```diff
@@ -0,0 +1,52 @@
...
+           .group("r.role")
+           .count
+           .transform_keys { |k| Vacancy::JOB_ROLES.invert[k].to_sym }
+    end
+
+    def phase_counts(scope:)
+      scope.joins("CROSS JOIN LATERAL unnest(phases) AS p(phase)")
+           .group("p.phase")
+           .count
+           .transform_keys { |k| Vacancy::PHASES.invert[k] }
```
starswan: better to use fetch rather than [] here

app/views/home/_landing_page_links.html.slim
```diff
@@ -1,141 +1,202 @@
...
+          h3.govuk-heading-s = t(".support_roles")
+          ul.govuk-list
+            - landing_page_fe_support_roles_list.each do |landing_page, role|
+              - count = fe_counts.role_counts.fetch(role.to_sym, 0)
+              li = govuk_link_to t("landing_pages.#{landing_page}.name", count: count), landing_page_path(landing_page), class: "govuk-link--no-visited-state"
+        .govuk-grid-column-one-third
+          h3.govuk-heading-s = t(".fe_phases")
+          ul.govuk-list
+            - landing_page_fe_phases_list.each do |landing_page, phase|
+              - count = fe_counts.phase_counts.fetch(phase.to_sym, 0)
```
starswan: could we not remove the duplication between fe/schools by passing fe_counts and school_counts respectively?

app/views/home/_landing_page_links.html.slim
```diff
@@ -1,141 +1,202 @@
...
 
-      .govuk-grid-column-one-third
-        = landing_page_link_group(use_locations: true) do |group|
-          - group.with_title_landing_page "london"
-          - group.with_landing_page "bexley"
-          - group.with_landing_page "havering"
-          - group.with_landing_page "barnet"
-          - group.with_landing_page "enfield"
-          - group.with_landing_page "bromley"
+    - accordion.with_section(heading_text: t(".jobs_by_location")) do
```
starswan, line 21: This partial seems too long. Maybe we can split it into its 3 parts? (fe job counts, school job counts and regional counts)

app/helpers/landing_pages_helper.rb
```diff
@@ -36,4 +36,290 @@ def linked_ect_status(vacancy)
...
+        child_landing_page = CHILD_SUBJECTS_LIST[child_subject]
+        next unless child_landing_page
+
+        [child_landing_page, counts_by_subject.fetch(child_subject, 0)]
+      }.to_h
+      [landing_page, [job_count, child_subjects_counts]]
+    end
+  end
+
+  FE_SUBJECTS_LIST = {
```
starswan: Still think that this is a duplicate of the mapping between the FE subject landing pages and the subject that it refers to - so we don't need the YML file? Maybe it needs a better name and/or home (this module is geting a bit big) to help manage the complexity

app/views/home/_landing_page_links.html.slim
```diff
@@ -1,141 +1,8 @@
...
-        = group.with_landing_page "sixth-form-or-college-jobs"
-        = group.with_landing_page "through-school-jobs"
-      = landing_page_link_group title: t(".working_patterns"), classes: "govuk-grid-column-one-third" do |group|
-        = group.with_landing_page "full-time-school-jobs"
-        = group.with_landing_page "part-time-school-jobs"
-        = group.with_landing_page "school-job-shares"
-
+      = render "home/school_counts_section", school_counts: school_counts
+    - accordion.with_section(heading_text: t(".jobs_by_fe_type_and_role")) do
+      = render "home/fe_counts_section", fe_counts: fe_counts
```
starswan: minor point - these small partials just add to the complexity. In-lining them here would be better IMHO, so that the (lack of) duplication is clear to the readerf

spec/jobs/import_from_vacancy_source_job_spec.rb
```diff
@@ -165,7 +165,7 @@ def each(...)
           "starts_on" => (Date.today + 1.year).strftime("%Y-%m-%d"),
           "stats_updated_at" => nil,
           "subjects" => [],
-          "uk_geolocation" => "POINT (469405.28816918284 178423.54064973217)",
+          "uk_geolocation" => FailedImportedVacancy.first.vacancy["uk_geolocation"],
```
starswan, line 168: stet? Don't think this change is relevant?

starswan (approved):

PR #8870, opened 2026-06-01
Show all in scope schools even without registered publishers

KyleMacPherson:
## Trello card URL

https://trello.com/c/D1SPHXwe/2871-bug-schools-and-colleges-page-by-association-doesnt-list-all-schools-in-scope

## Changes in this PR:

This PR changes our visible_to_jobseekers scope to show all schools that are open and in scope even if they have not registered for our service.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -68,7 +68,7 @@ class Organisation < ApplicationRecord
 
   scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES).or(where(detailed_school_type: nil)) }
 
-  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts.registered_for_service) }
+  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts) }
 
   scope :only_faith_schools, -> { where.not(religious_character: NON_FAITH_RELIGIOUS_CHARACTER_TYPES) }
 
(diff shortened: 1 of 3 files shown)


app/models/organisation.rb
```diff
@@ -68,7 +68,7 @@ class Organisation < ApplicationRecord
 
   scope :not_out_of_scope, -> { where.not(detailed_school_type: Organisation::OUT_OF_SCOPE_DETAILED_SCHOOL_TYPES).or(where(detailed_school_type: nil)) }
 
-  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts).registered_for_service }
+  scope :visible_to_jobseekers, -> { schools.not_closed.not_out_of_scope.or(Organisation.trusts.registered_for_service) }
```
starswan: Do we still want to filter trusts like this? (I think no if the goal is to allow a job-seeker to be able to see all schools that meet our eligibility criteria) - also I think there is a failure mode where user can't block their org because they are not registered, but once they register they can see the user even though the user would like to have biocked them?

KyleMacPherson: That's a great point! I'm a bit worried we might start showing out of scope or closed trusts though (if that is a thing?) because we don't seem to have any rules around what trusts are in scope. I wonder if we only allow relevant trusts to register for the service and that's how it works?

KyleMacPherson: Ok after doing a bit of reading it seems MATs always include state schools by definition so I think we are fine to show all of them. Have updated the code to do that!

starswan (approved):

PR #8891, opened 2026-06-08
Bundler 4 0.13

starswan:
## Changes in this PR:

Upgrade bundler version, and isolate it to prevent automatic updates breaking our Dockerfiles


.github/dependabot.yml
```diff
@@ -29,6 +29,11 @@ updates:
       - dependency-name: "view_component"
         update-types: ["version-update:semver-major"]
 
+      # This needs to be done by hand as dependabot doesn't spot our Dockerfiles with the bundler
+      # version coded in them
+      - dependency-name: "bundler"
+        update-types: ["version-update:semver-major"]
```
scruti: This seems to only ignore updates to bundler major versions. But minor and patch upgrades would still be brought up. 
If we want to completely remove dependabot upgrades for Bundler, we would need to add those too!
```suggestion
      # This needs to be done by hand as dependabot doesn't spot our Dockerfiles with the bundler
      # version coded in them
      - dependency-name: "bundler"
        update-types: ["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
```

.github/dependabot.yml
```diff
@@ -73,6 +78,7 @@ updates:
           - "activerecord"
           - "activestorage"
           - "activesupport"
+          - "bundler"
```
scruti: If I understand this right, this line would only exclude `bundler` from the `general-gems` grouping into a single PR, not exclude the `bundler` minor & patch upgrades from being opened into separate PRs by dependabot.
```suggestion
```

scruti (approved):

PR #8896, opened 2026-06-08
fix fusion format

starswan:
## Trello card URL

https://trello.com/c/QB5xpUTt/2914-fix-legacy-fusion-ats-integration-feed

## Changes in this PR:

Fix the broken fusion feed


spec/services/vacancies/import/sources/fusion_spec.rb
```diff
@@ -119,7 +119,7 @@
       context "when working_patterns includes `flexible`" do
         let(:response_body) do
           hash = JSON.parse(super())
-          hash["result"].first["workingPatterns"] = "full_time,flexible"
+          hash[0]["workingPatterns"] = "full_time,flexible"
```
scruti: I believe now this is `array` rather than `hash` but not blocking!

scruti (approved):
🚤 !!!

PR #8903, opened 2026-06-10
remove VacancyFormSequence#dependent_steps

starswan:
## Trello card URL

https://trello.com/c/nM5HBavW/2916-vacancyformsequence-can-we-get-rid-of-dependentsteps-method

## Changes in this PR:

Remove dependent_steps method and redundant base class for VacancyFormSequence.

dependent_steps was trying to re-work previous steps for a vacancy based on changing other answers on a published vacancy, but all examples now prevent those changes (e.g. application form type cannot be changed once published) so the entire method is now redundant


scruti (approved):
LGTM. The 🤖  seems to have caught a couple of little bits in the tests.

PR #8910, opened 2026-06-11
Do not show closed trusts in visible_to_jobseekers scope

KyleMacPherson:
## Trello card URL

https://trello.com/c/YiLOUNyw/2924-remove-closed-trusts-from-schools-page

## Changes in this PR:

Prior to this change we were showing closed trusts, this change ensures we only show currently open trusts (according to GIAS data).

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -52,6 +52,7 @@ class Organisation < ApplicationRecord
   scope :schools, -> { where(type: "School") }
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
+  scope :trusts_not_closed, -> { trusts.where.not("gias_data ->> 'Group Status' IN (?)", CLOSED_ESTABLISHMENT_STATUSES) }
   scope :local_authorities, -> { school_groups.where.not(local_authority_code: nil) }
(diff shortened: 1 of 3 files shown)


app/models/organisation.rb
```diff
@@ -52,6 +52,7 @@ class Organisation < ApplicationRecord
   scope :schools, -> { where(type: "School") }
   scope :school_groups, -> { where(type: "SchoolGroup") }
   scope :trusts, -> { school_groups.where.not(uid: nil) }
+  scope :trusts_not_closed, -> { trusts.where.not("gias_data ->> 'Group Status' IN (?)", CLOSED_ESTABLISHMENT_STATUSES) }
```
starswan, line 55: I was hoping we could just map the Group Status to our existing establishment_status field in Gias::ImportTrusts#trust_data (much simpler IMHO). I ran a quick spike and got just 'Open' and 'Closed' values - so we could use our existing not_closed scope

KyleMacPherson: Ah I see. I think we can deploy this if you're happy with it then I can raise another PR to make those changes to the import?

starswan (approved):

PR #8914, opened 2026-06-12
don't create vacancies for less active team members

starswan:
## Changes in this PR:

Allow less active team members (outside the main developer/tester group) to still have access to test systems, but don't fill their inboxes with emails about expires vacancies


scruti (approved):
Sorry didn't see this for a while. Good idea.

scruti:
@starswan I believe we can release this one?

PR #8949, opened 2026-06-22
2941 bug catholic job application review page with baptism cert option assumes presence of attachment

richardpattinson:
## Trello card URL

https://trello.com/c/uiZnKXI9/2941-bug-catholic-job-application-review-page-with-baptism-cert-option-assumes-presence-of-attachment

### Changes in this PR:

Fixes a bug on the Catholic job application review page where selecting the "baptism certificate" option (without attaching a file) caused an error. The partial was calling baptism_certificate? — which returns true when the option is selected — before checking whether a file was actually attached, then immediately trying to access blob attributes on a non-existent attachment.

The fix adds a .attached? guard to the condition, and also switches to the malware_scan_clean? helper rather than calling directly on the blob.

No — this is a straightforward defensive guard clause.

Screenshots of UI changes:

No visual change; the fix prevents a crash on the review page when a jobseeker selects the baptism certificate option but does not upload a file.

### Before

Error on review page when baptism_certificate? is true but no file is attached.

### After

Review page renders correctly; the baptism certificate row is skipped when no file has been attached.

Checklists:

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/controllers/jobseekers/job_applications/baptism_certificates_controller.rb
+++ b/app/controllers/jobseekers/job_applications/baptism_certificates_controller.rb
@@ -4,7 +4,7 @@ class Jobseekers::JobApplications::BaptismCertificatesController < Jobseekers::J
   before_action :set_job_application
 
   def destroy
-    job_application.baptism_certificate.purge_later
+    job_application.remove_baptism_certificate
     redirect_to jobseekers_job_application_build_path(job_application, :catholic)
   end
 
--- a/spec/system/jobseekers/jobseekers_can_complete_a_religious_job_application_spec.rb
+++ b/spec/system/jobseekers/jobseekers_can_complete_a_religious_job_application_spec.rb
@@ -171,6 +171,32 @@ RSpec.describe "Jobseekers can complete a religious job application" do
               expect(page).to have_current_path(jobseekers_job_application_build_path(job_application, :catholic))
               expect(job_application.reload.baptism_certificate.attached?).to be false
             end
+
+            scenario "deleting the baptism certificate resets the reference type and marks the form incomplete" do
+              complete_from_references_page
+              within("#catholic") { click_on "Change" }
+              click_on I18n.t("buttons.delete")
+
+              expect(page).to have_current_path(jobseekers_job_application_build_path(job_application, :catholic))
+
+              # expect no radio button checked for religious reference type
+              expect(page).to have_no_checked_field(I18n.t("helpers.label.jobseekers_job_application_catholic_form.religious_reference_type_options.baptism_certificate"))
+
+              # expect the section to be marked as incomplete
+              expect(page).to have_checked_field("No, I'll come back to it later")
+            end
+
+            scenario "the review page does not crash after the jobseeker deletes the baptism certificate and chooses a different reference type" do
+              complete_from_references_page
+              within("#catholic") { click_on "Change" }
+              click_on I18n.t("buttons.delete")
+              find("label[for='jobseekers-job-application-catholic-form-religious-reference-type-no-religious-referee-field']").click
+              choose "No, I'll come back to it later"
+              click_on I18n.t("buttons.save")
+
+              expect(page).to have_current_path(jobseekers_job_application_review_path(job_application))
+              expect(page).to have_no_content("blank_baptism_cert.pdf")
+            end
           end
         end
 
--- a/app/models/job_application.rb
+++ b/app/models/job_application.rb
@@ -244,6 +244,15 @@ class JobApplication < ApplicationRecord
     [street_address, city, postcode, country].compact_blank
   end
 
+  def remove_baptism_certificate
+    baptism_certificate.purge_later
+    update!(
+      religious_reference_type: nil,
+      completed_steps: completed_steps - %w[catholic],
+      in_progress_steps: (in_progress_steps + %w[catholic]).uniq,
+    )
+  end
+
   private
 
   def update_conversation_searchable_content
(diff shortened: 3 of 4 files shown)


app/views/jobseekers/job_applications/review/_catholic_religious_information.html.slim
```diff
@@ -43,10 +43,10 @@
...
           - row.with_key text: t("helpers.label.jobseekers_job_application_catholic_form.religious_referee_phone")
           - row.with_value text: job_application.religious_referee_phone
-      - elsif job_application.baptism_certificate?
+      - elsif job_application.baptism_certificate? && job_application.baptism_certificate.attached?
         - summary_list.with_row do |row|
           - row.with_key text: t("jobseekers.job_applications.review.religious_information.baptism_certificate")
-          - row.with_value text: (job_application.baptism_certificate.blob.malware_scan_clean? ? govuk_link_to("#{job_application.baptism_certificate.filename}  (#{number_to_human_size(job_application.baptism_certificate.byte_size)})", job_application.baptism_certificate, download: "true") : job_application.baptism_certificate.filename.to_s)
+          - row.with_value text: (malware_scan_clean?(job_application.baptism_certificate) ? govuk_link_to("#{job_application.baptism_certificate.filename}  (#{number_to_human_size(job_application.baptism_certificate.byte_size)})", job_application.baptism_certificate, download: "true") : job_application.baptism_certificate.filename.to_s)
```
starswan: This change is not required IMHO - the malware_scan_clean? view helper needs a different name (something like attached_and_clean?) to reflect its actual purpose (that's its implementation too) - this has already checked attached? so there is no need to check it again

app/views/jobseekers/job_applications/review/_catholic_religious_information.html.slim
```diff
@@ -43,10 +43,10 @@
         - summary_list.with_row do |row|
           - row.with_key text: t("helpers.label.jobseekers_job_application_catholic_form.religious_referee_phone")
           - row.with_value text: job_application.religious_referee_phone
-      - elsif job_application.baptism_certificate?
+      - elsif job_application.baptism_certificate? && job_application.baptism_certificate.attached?
```
starswan: The original code is fine - we are checking the state which should be synonymous with the attachment

app/controllers/jobseekers/job_applications/baptism_certificates_controller.rb
```diff
@@ -5,6 +5,12 @@ class Jobseekers::JobApplications::BaptismCertificatesController < Jobseekers::J
 
   def destroy
     job_application.baptism_certificate.purge_later
+    job_application.update!(
+      religious_reference_type: nil,
+      completed_steps: job_application.completed_steps.reject { |s| s == "catholic" },
+      in_progress_steps: (job_application.in_progress_steps + ["catholic"]).uniq,
+      imported_steps: job_application.imported_steps.reject { |s| s == "catholic" },
+    )
```
starswan: As per previous conversation, this (a) needs a massive comment to explain (b) should be an operation on the job_application (which should reduce the size of the comment) (c) should have a test explaining the behaviour (that the radio button is now un-selected, so that application is invalid, and that the 'completed' radio is now set to 'No' - also IMHO this shows that 'in_progress' doesn't really make sense, and so could/should be removed in a seperate PR

app/controllers/jobseekers/job_applications/baptism_certificates_controller.rb
```diff
@@ -4,6 +4,11 @@ class Jobseekers::JobApplications::BaptismCertificatesController < Jobseekers::J
   before_action :set_job_application
 
   def destroy
+    # Only the attachment is purged here - religious_reference_type and the catholic step's
+    # completed/in_progress state are intentionally left untouched. CatholicForm's baptism_certificate
+    # validation (unconditional on section_completed) will refuse to save the catholic step again until
+    # the jobseeker either re-attaches a file or picks a different religious_reference_type, so the
+    # resulting mismatch (type == "baptism_certificate" with no attachment) can never be persisted.
```
starswan: sadly this statement is false - we need to call into the model to 'do the right thing' here

app/form_models/jobseekers/job_application/catholic_form.rb
```diff
@@ -35,7 +35,12 @@ def load_form(model)
         validates :baptism_date, date: { on_or_before: :today }
       end
 
-      with_options if: -> { section_completed && following_religion && religious_reference_type == "baptism_certificate" } do
+      # Deliberately NOT gated on section_completed (unlike the baptism_date group above): a jobseeker
+      # can delete an already-attached certificate via BaptismCertificatesController#destroy without
+      # touching religious_reference_type. If presence were only checked on complete, they could save the
+      # step as "incomplete" and leave religious_reference_type == "baptism_certificate" with no file
+      # attached, which crashes the review page when it tries to read the (nonexistent) attachment's blob.
```
starswan: we don't want to do this. We *can* keep the options in sync (which were being done before)

spec/system/jobseekers/jobseekers_can_complete_a_religious_job_application_spec.rb
```diff
@@ -171,6 +171,29 @@
               expect(page).to have_current_path(jobseekers_job_application_build_path(job_application, :catholic))
               expect(job_application.reload.baptism_certificate.attached?).to be false
             end
+
+            scenario "the jobseeker cannot leave the step with baptism certificate selected but no file attached" do
+              complete_from_references_page
+              within("#catholic") { click_on "Change" }
+              click_on I18n.t("buttons.delete")
+              choose "No, I'll come back to it later"
```
starswan: This test should reflect the actual required behaviour (as originally coded) - so (a) radio buttons no longer selected and (b) 'completed' radio button now set false.

starswan (approved):

PR #8959, opened 2026-06-24
FE: rework location text to say 'college' on vacancy view screen 

starswan:
## Trello card URL

https://trello.com/c/BefvJ0Ir/2955-update-the-location-component

## Changes in this PR:

Update location text

### DIFF
--- a/spec/views/vacancies/show.html.slim_spec.rb
+++ b/spec/views/vacancies/show.html.slim_spec.rb
@@ -9,8 +9,6 @@ RSpec.describe "vacancies/show" do
     render
   end
 
-  # after { sign_out jobseeker if jobseeker.present? }
-
   describe "job posting metadata" do
     let(:jobseeker) { nil }
     let(:json_ld) { JSON.parse(rendered.html.css("script.jobref").inner_text, symbolize_names: true) }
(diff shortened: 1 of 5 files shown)


spec/views/vacancies/show.html.slim_spec.rb
```diff
@@ -2,15 +2,14 @@
...
-    end
+    create(:location_polygon, name: "hertfordshire")
+
+    allow(view).to receive_messages(current_jobseeker: jobseeker)
     assign :vacancy, vacancy.decorate
     render
   end
 
-  after { sign_out jobseeker if jobseeker.present? }
+  # after { sign_out jobseeker if jobseeker.present? }
```
KyleMacPherson: Should we keep this?

KyleMacPherson (approved):
Approved! There is a commented out line in a test we need to do something with I think `after { sign_out jobseeker if jobseeker.present? }`

spec/views/vacancies/show.html.slim_spec.rb
starswan: no there was never any need to sign in / out in the view spec

PR #8982, opened 2026-06-29
Import private beta only

starswan:
## Trello card URL

https://trello.com/c/Cv2kY37f/2919-restrict-access-to-the-fe-live-environment-to-the-pilot-group

## Changes in this PR:

Import private beta colleges only
 
#### FE College without being listed in the private beta group
image

#### FE College being listed in the private beta group
image


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -111,11 +111,11 @@ def id_token
   end
 
   def organisation_from_request
+    org_data = auth_hash.dig("extra", "raw_info", "organisation")
     # https://github.com/DFE-Digital/login.dfe.public-api#how-do-ids-map-to-categories-and-types
     case (cat_id = auth_hash.dig("extra", "raw_info", "organisation", "category", "id"))
```
scruti: I assume the intention behind extracting `org_data` was to reuse it across the method calls?

```suggestion
    case (cat_id = org_data.dig("category", "id"))
```

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -111,11 +111,11 @@ def id_token
...
     when Publishers::DfeSignIn::OrgIdMappings::CATEGORIES[:single_establishment]
-      ensure_establishment_type_supported!
-      School.find_by!(urn: auth_hash.dig("extra", "raw_info", "organisation", "urn"))
+      find_school org_data.fetch("urn")
     when Publishers::DfeSignIn::OrgIdMappings::CATEGORIES[:local_authority]
       SchoolGroup.find_by!(local_authority_code: auth_hash.dig("extra", "raw_info", "organisation", "establishmentNumber"))
```
scruti:
```suggestion
      SchoolGroup.find_by!(local_authority_code: org_data.fetch("establishmentNumber"))
```

app/controllers/omniauth_callbacks_controller.rb
scruti: Same here with `org_data.fetch("uid")`

scruti, line 141: Another case for `org_data`

spec/system/publishers/authentication/oauth_spec.rb
```diff
@@ -241,6 +241,28 @@
...
     end
 
+    context "with a college not part of private beta" do
+      let(:school) { create(:college, :discarded) }
+
+      before do
+        stub_publisher_authentication_step(
+          email: dsi_email_address,
+          school_urn: school.urn.to_s,
+          establishment_type: { id: "49", name: "Online provider" },
```
scruti: ☝️ I think this is a different test case. It is testing a non FE College establishment type (same case ase above one), not a FE College not part of the list.

config/data/colleges.csv
```diff
@@ -1,211 +1,211 @@
...
+10007636,N,WORKING MEN'S COLLEGE CORPORATION,Statutory FE Sector - Designated Institution,
+10094574,N,WRITTLE COLLEGE LIMITED,Statutory FE Sector - Designated Institution,HEI subsidiary
+10000055,N,ABINGDON AND WITNEY COLLEGE,Statutory FE Sector - General Further Education College,
+10004927,N,ACTIVATE LEARNING,Statutory FE Sector - General Further Education College,
+10057981,N,ADA NATIONAL COLLEGE FOR DIGITAL SKILLS,Statutory FE Sector - General Further Education College,
+10000415,N,ASKHAM BRYAN COLLEGE,Statutory FE Sector - General Further Education College,
+10000528,N,BARKING AND DAGENHAM COLLEGE,Statutory FE Sector - General Further Education College,
+10000533,N,BARNET & SOUTHGATE COLLEGE,Statutory FE Sector - General Further Education College,
+10000536,N,BARNSLEY COLLEGE,Statutory FE Sector - General Further Education College,
+10000560,Y,BASINGSTOKE COLLEGE OF TECHNOLOGY,Statutory FE Sector - General Further Education College,
```
scruti, line 22: 🙋 If we want to release the code changes today without giving access to any FE college, we need to set all these values to `N` and later on change the private beta ones to `Y` to give them access when going live. Right?

scruti (changes requested):
I think the only bit left is Setting all the colleges to `N` before merging this into `fe-colleges` and `fe-colleges` into `main`. Right?

app/controllers/omniauth_callbacks_controller.rb
scruti: This is still calling auth_hash dig directly. But not a blocker.

scruti (approved):
Agreed to set every college to `N` in a later PR

app/controllers/omniauth_callbacks_controller.rb
starswan: that one isn't covered by a test

PR #8997, opened 2026-07-01
remove change apply type rows when step process doesn't allow it

starswan:
## Trello card URL

https://trello.com/c/EkUvDjvi/2989-bug-fest-hiring-staff-job-listing-screen

## Changes in this PR:

Don't display change rows when journey hasn't included the steps we're going back to

### After
CollegeAfter

### DIFF
--- a/spec/views/publishers/vacancies/review.html.slim_spec.rb
+++ b/spec/views/publishers/vacancies/review.html.slim_spec.rb
@@ -2,21 +2,47 @@ require "rails_helper"
 
 RSpec.describe "publishers/vacancies/review" do
   let(:school) { build_stubbed(:school) }
-  let(:vacancy) { build_stubbed(:vacancy, publish_on: publish_date) }
   let(:vacancy_presenter) { vacancy.decorate }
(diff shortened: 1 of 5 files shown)


spec/views/publishers/vacancies/review.html.slim_spec.rb
```diff
@@ -2,21 +2,47 @@
 
 RSpec.describe "publishers/vacancies/review" do
   let(:school) { build_stubbed(:school) }
-  let(:vacancy) { build_stubbed(:vacancy, publish_on: publish_date) }
   let(:vacancy_presenter) { vacancy.decorate }
   let(:step_process) { Publishers::Vacancies::VacancyStepProcess.new(:review, vacancy: vacancy, organisation: school) }
```
scruti, line 6: I suspect that, to reflect reality, the `organisation: school` should be turn into a generic `organisation`, to be able to flip between for a school and for a college step processes.

Seems the logic in our case is based in `vacancy.for_an_fe_college?` method, but I would expect the org parameter for the step process to also be a college in that case?

scruti (approved):
Approved with a little non-blocking test suggestion.

PR #9008, opened 2026-07-03
disable uploads

starswan:
## Trello card URL

https://trello.com/c/6rTpgNdp/2917-granular-proper-disabling-of-trix-attachments

## Changes in this PR:

Disable trix uploads

### After
SendMessageUpload

image

### DIFF
--- a/app/views/shared/_message_form.html.slim
+++ b/app/views/shared/_message_form.html.slim
@@ -5,7 +5,7 @@ div class="govuk-!-margin-top-6"
   = form_for message_object, url: form_url, method: form_method do |f|
     = f.govuk_error_summary
 
-    .govuk-form-group.message-form
+    .govuk-form-group.message-form data-controller="disable-uploads"
       = f.rich_text_area :content, aria: { label: "Message" }
       noscript
         = f.govuk_text_area :content, label: { text: "Message" }, rows: 5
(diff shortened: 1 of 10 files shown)


KyleMacPherson:
If we are disabling attachments for all trix editors I wonder if this is the best approach? I _think_ we could follow the pattern we use for initTrixURLNormalizer and disable attachments globally and then we wouldn't need to remember to disable them for each new trix editor we introduce.

If we did want to follow the pattern outlined in the linked article maybe we should scope it to the element rather than the document but I think the only reason to do that would be if we wanted to disable it for some questions one a page but not others.

KyleMacPherson (approved):
After discussion in our weekly dev chat I understand we actually do not want to disable attachments on all trix editors (we want to keep attachments on the messages for instance)!

Approved!

app/views/shared/_message_form.html.slim
```diff
@@ -5,7 +5,7 @@ div class="govuk-!-margin-top-6"
   = form_for message_object, url: form_url, method: form_method do |f|
     = f.govuk_error_summary
 
-    .govuk-form-group.message-form
+    .govuk-form-group.message-form data-controller="disable-uploads"
```
scruti: 🙋‍♂️ We have re-enabled the attachments in the messages. So this should not include the upload-disabling controller.

scruti (changes requested):
We need to allow file uploads in the messages editor. 

All the rest seems good! (Tested it in the Review App and it works 👏 )

scruti (approved):
Tested button presence and copy & paste in the Messages (meant to accept attachments) and School Description (meant not to accept attachments).

Both behaviours work as expected 👏

PR #9009, opened 2026-07-03
Remove submodel forms

starswan:
## Trello card URL

https://trello.com/c/tM9Ar79k/3015-remove-redundant-form-pattern-from-sub-models-following-the-pattern-used-by-employment-controller

## Changes in this PR:

When we have sub-models in our code-base, it is probably better to use Rails idioms and set validations on the models themselves rather than creating duplicate 'form' objects that have a 1-1 correspondence with the model. This has already been done with the 'Employment' model, so this is just following up on the pattern for consistency.

It also removes 26 lines of code (12863 vs 12889)


db/migrate/20260708151811_convert_sub_model_nullable_columns.rb
```diff
@@ -0,0 +1,28 @@
+class ConvertSubModelNullableColumns < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
```
starswan, line 4: All 3 of these have been checked on production, and there are no records which violate these new contraints

scruti: Nice, thanks for checking these things. We always forget this 😅

scruti (approved):
Great changes 👏

PR #9017, opened 2026-07-08
Display vacancy pre-spike

starswan:
## Trello card URL

https://trello.com/c/3AIlLRdg/3023-vacancy-show-template-tidy-up

## Changes in this PR:

Moved variable access from partials into the main template, passing parameters to partials as per rails idioms. Turns out that preview re-uses the 'show' partial but with all apart from 'vacancy' not set - this now shows that these variables are explicitly set to nil


KyleMacPherson (approved):
Very nice 👏 LGTM!

PR #9035, opened 2026-07-10
Add overrides to ensure ccorrect location names + titalisation

richardpattinson:
## Trello card URL

https://trello.com/c/R00ZieDO/3031-text-updates-main-page-schools-and-colleges-jobs-by-location

## Changes in this PR:

Add method to strip the upcase from certain link words (like upon and The). Add an exception for Yorkshire and the Humber.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/models/location_landing_page.rb
+++ b/app/models/location_landing_page.rb
@@ -20,7 +20,9 @@ class LocationLandingPage < LandingPage
   end
 
   def name
-    (MAPPED_LOCATIONS[location.tr("-", " ")] || location).titleize.gsub(/\bAnd\b/, "and")
+    key = location.tr("-", " ")
+
+    (MAPPED_LOCATIONS[key] || location).titleize.gsub(/\b(And|Of|The|Upon)\b/, &:downcase)
   end
 
   private


app/models/location_landing_page.rb
```diff
@@ -1,6 +1,14 @@
 class LocationLandingPage < LandingPage
   attr_reader :location
 
+  # Overrides for landing pages whose underlying (search) location name would otherwise produce a
```
starswan: I'm really unsure about both the approach and the whole thing. I think this should be held off until we can review locations properly

scruti: Agree on this.

app/models/location_landing_page.rb
```diff
@@ -20,7 +28,9 @@ def initialize(location)
   end
 
   def name
-    (MAPPED_LOCATIONS[location.tr("-", " ")] || location).titleize.gsub(/\bAnd\b/, "and")
+    key = location.tr("-", " ")
+
+    DISPLAY_NAME_OVERRIDES[key] || (MAPPED_LOCATIONS[key] || location).titleize.gsub(/\b(And|Of|The|Upon)\b/, &:downcase)
```
scruti: Happy to release the bit about doing substitutions on the titleize.

scruti (approved):

PR #9045, opened 2026-07-13
Use solid queue adapter in SolidQueueJob

starswan:
## Changes in this PR:

Set wrong explicit adapter in SolidQueueJob


scruti (approved):
Wops

PR #9056, opened 2026-07-15
Upgrade sass module system

KyleMacPherson:
## Trello card URL

https://trello.com/c/5KUMh1tS/2915-frontend-upgrade-deprecated-sass-import-rules

## Changes in this PR:

This change migrates the Sass codebase from using the deprecated @import rule to the new @use/@forward rules. I tried to just run sass-migrator to convert all @import to @use/@forward but that didn't quite work, I had to make the following changes:

- Fixed sass-migrator using wrong namespace (helpers) for govuk-frontend replaced with @use 'govuk-frontend/dist/govuk/base' as *
- Added missing @use 'govuk-frontend/dist/govuk/index'. sass-migrator failed to convert the original @import, causing all govuk component styling to break.
- Replaced all @extend calls (@extend doesn't work with this new module system).
  - Created mixins in _mixins.scss for reusable patterns (textarea, list, body, heading styles)
  - Removed @extend in posts/index.scss and manually added in styles/mixins (couldn't just add classes to the views because we use kramdown to turn markdown -> html)
  - Replaced @extend in _global.scss with mixin calls, and added govuk-body directly to the small number of bare p tags we had in views
  - Removed _error-summary.scss — the custom classes it defined (govuk-info-summary, govuk-success-summary) were unused
  - Removed --incomplete and --header custom modifiers from _inset-text.scss. Can no longer use extend and didn't want to hardcode in styles so instead added the relevant govuk classes directly in the HTML (we could control the HTML here unlike the Kramdown situation)

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


scruti (approved):
LGTM. Copilot left a couple of comments you may want to look into.

PR #9060, opened 2026-07-16
Rework polygon import to enable faster polygon searches

starswan:
## Trello card URL

https://trello.com/c/N15hxa7F/3052-investigate-speed-of-location-searches-the-subscriptions-spec-takes-2-minutes-even-when-setup-once-and-torn-down-at-the-end

## Changes in this PR:

Current code doesn't validate polygons on import - turns out that several of them are invalid, and calling 'make_valid' on them doesn't always make them valid. This makes all polygons valid, iterating until they become correct. This reduces the accuracy of some polygons, but allows the removal of a very expensive 'valid' check during location searching, which speeds up the test showing this behaviour from 1 min 20 to 4 seconds, and presumably has a similar effect on production searches


config/initializers/location_data.rb
```diff
@@ -45,3 +45,15 @@
 end
 
 REDIRECTED_LOCATION_LANDING_PAGES = YAML.load_file("config/data/redirected_location_landing_pages.yml").freeze
+
+# Only use Faraday for ONS data - open Response class and hard-code cache value as ONS returns max-age=30 (seconds)
+# even though data file is over 3 years old
```
scruti, line 50: This is the bit they're changing to 1 hour, right? (Guess better than 30seconds!)

starswan: I've removed this - with my new strategy of importing composites and their dependencies first, things work a lot quicker - it was taking 35 minutes, now its done in less than 2

app/models/location_polygon.rb
```diff
@@ -33,11 +33,15 @@ def self.mapped_name(location)
 
   def self.find_valid_for_location(location)
     polygon = with_name(location)
-    if polygon.present? && polygon.area.invalid_reason.nil?
-      polygon
-    end
+    polygon.presence
```
scruti: It doesn't seem that this method adds anything (besides the presence call) over the `with_name` method. Couldn't just replace it with calls to `LocationPolygon.with_name(location)` ?

app/models/vacancy.rb
```diff
@@ -120,17 +120,13 @@ class Vacancy < ApplicationRecord
   # Not called from the code but frequently used for filtering during manual debugging sessions
   scope :external, -> { where.not(external_source: nil).or(where.not(publisher_ats_api_client_id: nil)) }
 
-  # we need these 3 tiny modules to provide 'scoping glue' between the model and the queries
-  # so that if we can use PublishedVacancy and DraftVacancy safely
-  extend VacancyFilterQueryModule
+  scope :search_by_filter, ->(filters) { VacancyFilterQuery.new(all).call(filters) }
```
scruti: Nice refactors  👏

app/services/ons_data_import/create_composites.rb
```diff
@@ -1,24 +1,41 @@
...
+          set_uk_area_data(location_polygon, quoted_constituents, new_tolerance)
+          location_polygon.touch
+          location_polygon.reload
+          # simplecov:disable
+          if location_polygon.area_data_valid?
+            Rails.logger.info("Created composite polygon for '#{name}' tolerance #{new_tolerance}")
+            break
+          end
+          # simplecov:enable
+        end
```
scruti, line 26: Ohhhh nice!

It took me a bit to understand this bit. 

I suggest adding a comment on top of the `(0..).each` block describing that it will incrementally increase the tolerance until the location_polygon is valid.

Also... couldn't this become an infinite loop if, by any chance, the `location_polygon.area_data_valid?` is never `true`?

Maybe we should set a limit on iterations to avoid this?

scruti (commented):
Great improvements. Love the new approach.

Left a couple suggestions and I guess it needs a rebase since we have introduced `Faraday` into `main` while this poor PR was waiting forever for a review (sorry!).

Will happily relook at it once it gets fresh and ready :)

scruti (approved):
🚀 👏

PR #9064, opened 2026-07-17
Rename only_faith_schools scope

KyleMacPherson:
## Changes in this PR:

This change renames scope only_faith_schools to faith_schools. I think this scope name prior to this change was a bit surprising as it doesn't follow our scope naming conventions (i.e we use in_scope_schools rather than only_in_scope_schools etc). I thought it may have been used to determine if a MAT included only_faith_schools or something like that but that does not seem to be the intention.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


scruti (approved):
Nice tidy up.

PR #9078, opened 2026-07-21
Redo subscription query

starswan:
## Trello card URL

https://trello.com/c/N15hxa7F/3052-investigate-speed-of-location-searches-the-subscriptions-spec-takes-2-minutes-even-when-setup-once-and-torn-down-at-the-end

## Changes in this PR:

Whilst investigating performance of location searches, I discovered that we had re-implemented about half of the VacancySearch in the job alert search. Re-wrote the subscription search in terms of the original search where possible, removing 58 lines of code (13173 vs 13231)

Some tests are marked as pending - subscriptions with empty filters are supposed to return no hits rather than all hits

### DIFF
--- a/app/models/location_polygon.rb
+++ b/app/models/location_polygon.rb
@@ -33,11 +33,7 @@ class LocationPolygon < ApplicationRecord
 
   def self.find_valid_for_location(location)
     polygon = with_name(location)
-    if polygon.present? && polygon.area.invalid_reason.nil?
-      polygon
-    end
-  rescue RGeo::Error::InvalidGeometry
-    nil
+    polygon.presence
   end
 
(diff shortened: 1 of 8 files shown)


app/models/location_polygon.rb
```diff
@@ -33,11 +33,7 @@ def self.mapped_name(location)
 
   def self.find_valid_for_location(location)
     polygon = with_name(location)
-    if polygon.present? && polygon.area.invalid_reason.nil?
-      polygon
-    end
-  rescue RGeo::Error::InvalidGeometry
-    nil
+    polygon.presence
```
scruti: It doesn't seem that this method adds anything (besides the presence call) over the `with_name` method. Couldn't just replace it with calls to `LocationPolygon.with_name(location)` ?

PR #9080, opened 2026-07-21
Changed action-block to override_action-none

jsomerset:
## Trello card URL
https://trello.com/c/chHAU4LZ/2846-add-rate-limiting-for-teaching-vacancies

## Changes in this PR:
Bugfix to update sql-protection rule. Despite the Terraform plan showing no issues, the aws_wafv2_web_acl_rule resource is showing as invalid during the apply stage. This change will attempt to resolve the issue by replacing the action block with override_action, which appears to be used instead when using AWS managed rule groups.

## To review:
Confirm that Terraform documentation uses the same structure when deploying AWS managed rule groups, as covered here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl_rule#managed-rule-group-statement:~:text=resource%20%22aws_wafv2_web_acl_rule%22%20%22aws_managed_rules,%3D%20true%0A%20%20%7D%0A%7D

Compare screenshots from AWS console when creating rate limit rules vs AWS managed rule groups - Note that action is only included in rate-limit and override is only included in managed rules:
image
image

### DIFF
--- a/app/models/school_group.rb
+++ b/app/models/school_group.rb
@@ -23,7 +23,7 @@ class SchoolGroup < Organisation
   # Allows School Groups/MATs to use religious application forms for head office vacancies.
   # School Groups don’t store religious character information, so we treat a group with at least one faith school as a faith organisation.
   def faith_school?
-    schools.only_faith_schools.exists?
+    schools.faith_schools.exists?
   end
 
   def all_organisations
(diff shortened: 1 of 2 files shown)


steves-mobilise (approved):
Looks good to me.

starswan (approved):

PR #9098, opened 2026-07-27
Simplecov deprecations

starswan:
## Changes in this PR:

After upgrading to Simplecov 1.0, fix all the deprecation notices that were generated as a result

### DIFF
--- a/.undercover
+++ b/.undercover
@@ -1,1 +1,1 @@
--c origin/further-education
+-c origin/main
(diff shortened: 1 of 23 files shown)


.undercover
```diff
@@ -1 +1 @@
--c origin/further-education
```
scruti, line 1: Oopsie!

scruti (approved):
LGTM!

PR #9109, opened 2026-07-29
Fe qts required (QTS required for FE phase 2 - Hiring staff)

starswan:
## Trello card URL

https://trello.com/c/rwVcEDGP/3072-hiring-staff-for-fe-additional-step-does-the-applicant-need-qts-before-ect-question

## Changes in this PR:

Add college specific field for QTS required

### Data & Schema Changes

<details>

### Integration Impact

- [X] DfE Analytics platform
- [X] Legacy imports mappings
- [X] Publisher ATS API (may require mapping updates or API versioning)

Could this change impact:
</details>

### DIFF
--- a/app/views/publishers/vacancies/vacancy_review_sections/_about_the_role.html.slim
+++ b/app/views/publishers/vacancies/vacancy_review_sections/_about_the_role.html.slim
@@ -3,15 +3,26 @@ h2 class="govuk-heading-m govuk-!-margin-bottom-4"
 
 = govuk_summary_list(html_attributes: { id: "about_the_role" }) do |summary_list|
 
-  - if vacancy.job_roles.include?("teacher") && !vacancy.ect_status.nil?
-    - summary_list.with_row(html_attributes: { id: "ect_status" }) do |row|
-      - row.with_key
-        = t("jobs.ect_status")
-      - row.with_value
-        = vacancy.readable_ect_status
-      - row.with_action text: t("buttons.change"),
-                  href: organisation_job_build_path(vacancy.id, :about_the_role, "back_to_#{action_name}": "true"),
-                  visually_hidden_text: t("jobs.ect_status")
+  - if vacancy.job_roles.include?("teacher")
+    - if vacancy.for_an_fe_college?
+      - summary_list.with_row(html_attributes: { id: "needs_qts_status" }) do |row|
+        - row.with_key
+          = t("helpers.legend.publishers_job_listing_about_the_role_form.fe_role_qts_required")
+        - row.with_value
+          = vacancy.fe_role_qts_required
+        - row.with_action text: t("buttons.change"),
+          href: organisation_job_build_path(vacancy.id, :about_the_role, "back_to_#{action_name}": "true"),
+          visually_hidden_text: t("helpers.legend.publishers_job_listing_about_the_role_form.fe_role_qts_required")
+
+    - unless vacancy.ect_status.nil?
+      - summary_list.with_row(html_attributes: { id: "ect_status" }) do |row|
+        - row.with_key
+          = t("jobs.ect_status")
+        - row.with_value
+          = vacancy.readable_ect_status
+        - row.with_action text: t("buttons.change"),
+                    href: organisation_job_build_path(vacancy.id, :about_the_role, "back_to_#{action_name}": "true"),
+                    visually_hidden_text: t("jobs.ect_status")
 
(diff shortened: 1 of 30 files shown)


richardpattinson (dismissed):
Looks great to me

app/controllers/publishers/vacancies_controller.rb
```diff
@@ -51,7 +51,9 @@ def save_and_finish_later
 
   def create
     # anonymise_applications defaults to false for past applications, so we have to explicitly set nil here
+    # needs_qts_status is the same - it will be filled in by the about_the_role screen
```
scruti: If I am understanding this right. The DB schema defaults are being used to "backfill" the existing values, but we always override it to an initial `nil` value for future vacancies, so the user has to make the `nil` to `true/false` choice?

In that case, wouldn't it be better not to set a default in the schema and to do backfill tasks if/when needed, rather than having to override the default for all new data?

And for the anonymise_applications one, I believe we could remove the default and the overrides?

app/views/publishers/vacancies/vacancy_review_sections/_about_the_role.html.slim
```diff
@@ -3,15 +3,26 @@ h2 class="govuk-heading-m govuk-!-margin-bottom-4"
...
-      - row.with_action text: t("buttons.change"),
-                  href: organisation_job_build_path(vacancy.id, :about_the_role, "back_to_#{action_name}": "true"),
-                  visually_hidden_text: t("jobs.ect_status")
+  - if vacancy.job_roles.include?("teacher")
+    - if vacancy.for_an_fe_college?
+      - summary_list.with_row(html_attributes: { id: "needs_qts_status" }) do |row|
+        - row.with_key
+          = t("helpers.legend.publishers_job_listing_about_the_role_form.needs_qts_status")
+        - row.with_value
+          = vacancy.needs_qts_status? ? "No" : "Yes"
```
scruti: 🙋‍♂️ 
I think this is inverted. Maybe should resort to the translations rather than hardcoding it?

### Selected option:
Image

----

### Review page:

Image

db/migrate/20260810152440_add_needs_qts_status.rb
```diff
@@ -0,0 +1,9 @@
+class AddNeedsQtsStatus < ActiveRecord::Migration[8.0]
+  # rubocop:disable Rails/ThreeStateBooleanColumn
+  def change
+    # default all existing vacancies to needing QTS status
+    add_column :vacancies, :needs_qts_status, :boolean, default: true
```
scruti: Question: Was this defaulting requested?

I see two possible issues here:
1. We're setting a default value for every vacancy in the system. But there is no backfill task to match the logic that we apply later (`true` for teaching roles, `false` for non-teaching roles, answer `true/false` for FE teaching roles). So the default seems inconsistent between historical vacancies and new ones.
2. If this is only relevant to a subset of vacancies. We may want to distinguish who explicitly filled what by not defaulting it on form params and just add a form validation requesting it when relevant?

Not defaulting it would also allow us not to have to override the default with every initialisation.

scruti (changes requested):
It looks like the review page answer is inverted.

And I suspect we want to re-think the whole defaulting approach vs backfill and answer enforcement only when needed.

db/migrate/20260810152440_add_needs_qts_status.rb
starswan: The challenge is needing the default for existing vacancies. The pattern we normally use is to introduce the field first, and then change it later in a different PR. I guess I forgot that as this got changed in the middle. Will introduce first in DB, then we can be correct afterwards. The API needs to get it right too...

spec/validators/json_swagger_validator_spec.rb
```diff
@@ -10,7 +10,7 @@
                                .except(:enable_job_applications, :contact_number_provided,
                                        :contact_number, :fixed_term_contract_duration, :further_details_provided,
                                        :hired_status, :include_additional_documents,
-                                       :listed_elsewhere, :ect_status, :school_visits, :needs_qts_status,
+                                       :listed_elsewhere, :ect_status, :school_visits, :fe_role_qts_required,
```
scruti: Just waiting until [this gets done in the first PR](https://github.com/DFE-Digital/teaching-vacancies/pull/9159/changes#r3796266471) and this diff disappears 🥷

scruti (approved):
LGTM!

PR #9112, opened 2026-07-30
Fix pagy error

richardpattinson:
## Trello card URL

https://trello.com/c/Z1U0J8Va/3101-investigate-pagy-error

## Changes in this PR:

Fixes a `Pagy::VariableError` (~630 events / 90 days, `VacanciesController#index`) that was crashing with a 500 whenever a request arrived with an out-of-range or non-numeric `page` param (e.g. `?page=0`, `?page=-1`)

Pagy's `overflow: :last_page` config (in `config/initializers/pagy.rb`) only handles pages that are too *high* — it clamps back to the last page. It doesn't cover pages below the minimum, which raises `Pagy::VariableError` instead and was going unhandled anywhere in the app.

- Override `pagy_get_page` in `ApplicationController` to clamp the page param to a minimum of 1 before it ever reaches Pagy, rather than rescuing the exception after the fact. This fixes the bug for every controller that uses `pagy` (vacancies, organisations, colleges, notifications, etc.), not just `VacanciesController`.
- Add request spec coverage for `page=0` and `page=-1` returning `200` (served as page 1) instead of raising.
- Also includes an unrelated one-line contentving a local merge conflict: adds the missing`saved_jobs.index.deadline_passed` locale key (`"Job closed"`), which `app/views/jobseekers/saved_jobs/index.html.sa`job_application_status_tag(:deadline_passed)` but had no translation for.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,6 +40,12 @@ class ApplicationController < ActionController::Base
 
   private
 
+  # Clamp out-of-range/non-numeric page params (e.g. ?page=0 or ?page=-1) to page 1
+  # instead of letting Pagy raise Pagy::VariableError.
+  def pagy_get_page(vars)
+    [super.to_i, 1].max
+  end
+
   def show_cookies_banner?
     cookies["consented-to-additional-cookies-v3"].blank?
   end
--- a/spec/requests/vacancies_spec.rb
+++ b/spec/requests/vacancies_spec.rb
@@ -6,5 +6,15 @@ RSpec.describe "Vacancies" do
       get jobs_path
       expect(response.headers["X-Robots-Tag"]).to eq("noarchive")
     end
+
+    it "clamps to page 1 instead of raising when page is 0" do
+      get jobs_path(page: 0)
+      expect(response).to have_http_status(:ok)
+    end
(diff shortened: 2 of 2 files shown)


scruti:
Hi @richardpattinson. 
Overriding a gem's internal method brings risks, like (as it is actually happening with Pagy new major version) the method internals changing or the method being removed in the gem
Sometimes is needed, but it is a last-resort thing to do rather than something we should try out of convenience. 

Pagy seems to have a full rewrite and upgrade that requires a manual migration. This may be a good chance (if you want) to attempt takling the upgrade and see how the new version deals with the `0` and `-1` page params?

And even if it keeps blowing up, maybe a global controller `page` param sanitisation on our end would be less invasive than overwriting the gem method. 

I've just added a ticket for the upgrade. Feel free to pick it up! https://trello.com/c/xtsdq43M/1755-upgrade-pagy-gem-to-v43

EDIT: Ok so upgrading the gem is not an option. Happy wit hthe current implementation then.

scruti (approved):

PR #9141, opened 2026-08-06
Add new support roles

richardpattinson:
## Trello card URL

https://trello.com/c/BHm9tbIW/3102-support-roles-to-be-added-for-fe-hiring-staff-to-fill-vacancies-in-non-teaching-roles

## Changes in this PR:

Adds 5 new roles to the support job-role group, positioned between "Pastoral, health and welfare" and "Other support role":

- Leadership and management
- Business support, administration and corporate services
- Student support, learning and progress
- Marketing and communications
- Apprenticeships and commercial services

Model (app/models/vacancy.rb): new entries added to Vacancy::JOB_ROLES (fresh IDs 17–21) and inserted in order into Vacancy::SUPPORT_JOB_ROLES — this array drives display order everywhere roles are listed (publisher job forms, jobseeker profile/search, subscriptions, dashboard filters).

Labels: added to config/locales/forms.yml (job listing form options) and config/locales/jobseekers/profile/job_preferences.yml (jobseeker profile summary).

Public API contract (spec/swagger_helper.rb): added to V1_JOB_ROLES, per that file's documented process for additive/non-breaking enum changes.

SEO landing pages (main + FE site): new page config in config/landing_pages.yml/config/fe_landing_pages.yml, translated title/heading/meta description in config/locales/landing_pages.yml, and slug mappings in app/helpers/landing_page_lists_helper.rb. Also added the 5 roles to the existing support-jobs aggregate landing page.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/helpers/landing_page_lists_helper.rb
+++ b/app/helpers/landing_page_lists_helper.rb
@@ -62,6 +62,11 @@ module LandingPageListsHelper
     "fe-catering-cleaning-site-management-jobs" => "catering_cleaning_and_site_management",
     "fe-it-support-jobs" => "it_support",
     "fe-pastoral-health-welfare-jobs" => "pastoral_health_and_welfare",
+    "fe-leadership-management-jobs" => "leadership_and_management",
+    "fe-business-support-administration-corporate-services-jobs" => "business_support_administration_and_corporate_services",
+    "fe-student-support-learning-progress-jobs" => "student_support_learning_and_progress",
+    "fe-marketing-communications-jobs" => "marketing_and_communications",
+    "fe-apprenticeships-commercial-services-jobs" => "apprenticeships_and_commercial_services",
     "fe-other-support-roles-jobs" => "other_support",
   }.freeze
 
--- a/app/helpers/vacancies_helper.rb
+++ b/app/helpers/vacancies_helper.rb
@@ -25,8 +25,12 @@ module VacanciesHelper
     end
   end
 
-  def fe_college_job_roles_options
+  def fe_college_teaching_job_roles_options
     [["teacher", I18n.t("helpers.label.publishers_job_listing_job_role_form.teaching_job_role_options.teacher")]]
+  end
+
+  def fe_college_support_job_roles_options
+    Vacancy::FE_SUPPORT_JOB_ROLES.map { |option| [option, I18n.t("helpers.label.publishers_job_listing_job_role_form.fe_support_job_role_options.#{option}")] }
   end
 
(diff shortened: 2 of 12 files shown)


scruti (changes requested):
As I understand and I got told when asking during recent meetings. These new job roles should be ony made available for FE Publishers.

Non FE Publishers should see the existing list without the additional roles.

And the Publisher ATS API shouldn't allow for FE specific roles either.

spec/swagger_helper.rb
```diff
@@ -20,6 +20,11 @@
   catering_cleaning_and_site_management
   it_support
   pastoral_health_and_welfare
+  leadership_and_management
+  business_support_administration_and_corporate_services
+  student_support_learning_and_progress
+  marketing_and_communications
+  apprenticeships_and_commercial_services
```
scruti: 🙋‍♂️ I don't think we can extend the API

FE colleges cannot publish through our API at the moment. 
And School/MATs shouldn't be able to publish vacancies through the API using the new FE-exclusive Job Roles. 

So, the API list of accepted roles should be kept as it is at the moment.

I think a viable solution would be:

- To kept the API documentation as it is: Modify the `enum_consistency_spec` test to exclude the `FE_SUPPORT_JOB_ROLES`. 
  - Something like: `"job_roles" => { v1: V1_JOB_ROLES, model: -> { Vacancy.job_roles.keys - Vacancy::FE_SUPPORT_JOB_ROLES } },` . 
  - Confirm that the new roles don't show up in the API docs (`/ats-api-docs` path in the review app).
- To ensure those roles couldn't be used on external integrations even if undocumented: Add a custom validation on `ExternalVacancyValidator` (shared by legacy integrations and the API) to ensure that the accepted incoming job roles are only in the pre-existing list and none of the `Vacancy::FE_SUPPORT_JOB_ROLES`. 
  - The error code and message should be the same as the default Rails error for when we include `foobar` in the list of roles and `foobar` doesn't belong to `Vacancy.job_roles` list.

scruti (changes requested):
The API side needs some massaging to:

- Do not extend the API docs with invalid (at the moment) FE roles.
- Do not allow FE roles to be injected into the API/Legacy integrations even when undocumented.

Left a rough proposal of action in a comment.

scruti (approved):
LGTM! (No idea about the review app error, but looks unrelated to your changes).

PR #9155, opened 2026-08-11
don't track vacancy view when no referrer

starswan:
## Trello card URL

https://trello.com/c/Oyqb7A53/3147-trackvacancyviewjob-could-we-not-trigger-it-for-bots-spiders

## Changes in this PR:

Don't track vacancy views without referrers

### DIFF
--- a/app/controllers/vacancies_controller.rb
+++ b/app/controllers/vacancies_controller.rb
@@ -26,7 +26,9 @@ class VacanciesController < ApplicationController
     vacancy = PublishedVacancy.kept.listed.friendly.find(params[:id])
     # Only track views which have a referrer - this should reduce bots while in practice not affecting user tracking
     # Referrer is an 'optional, not to be trusted' header, but every browser sends it correctly at the time of writing (Aug 2026)
-    TrackVacancyViewJob.perform_later(vacancy_id: vacancy.id, referrer_url: request.referer, hostname: request.host, params: request.query_parameters) if request.referer.present?
+    if request.referer.present? || params.key?(:utm_campaign)
+      TrackVacancyViewJob.perform_later(vacancy_id: vacancy.id, referrer_url: request.referer, hostname: request.host, params: request.query_parameters)
+    end
 
     @saved_job = vacancy.saved_jobs.find_by(jobseeker: current_jobseeker)
     @job_application = vacancy.job_applications.find_by(jobseeker: current_jobseeker)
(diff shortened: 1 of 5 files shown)


spec/views/vacancies/show.html.slim_spec.rb
```diff
@@ -9,6 +9,54 @@
...
+    end
+
+    context "with a website vacancy" do
+      let(:vacancy) { build_stubbed(:vacancy, :apply_via_website, organisations: [school]) }
+
+      scenario "jobseeker does not see a tag on jobs that don't allow to apply through Teaching Vacancies" do
+        expect(rendered).to have_no_css("strong.govuk-tag--green", text: I18n.t("vacancies.listing.enable_job_applications_tag"))
+      end
+
+      scenario "a jobseeker can click on the application link" do
```
scruti: I personally find it confusing to mix `it` and `scenario` within the spec file. Would it be ok to move this to `it` format? 😅 

EDIT: Oh I see what happened here, we moved the tests from a system test into a view test. First, thanks, good call. Second... yes better to set them as `it` :D

scruti (dismissed):
Excellent work moving tests away from system to request/view when appropriate. 

Left a comment as the copy paste of the tests into the view specs was lacking the change of test format from `scenario` to `it`

app/controllers/vacancies_controller.rb
```diff
@@ -24,7 +24,10 @@ def show
     end
 
     vacancy = PublishedVacancy.kept.listed.friendly.find(params[:id])
-    TrackVacancyViewJob.perform_later(vacancy_id: vacancy.id, referrer_url: request.referer, hostname: request.host, params: request.query_parameters)
+    # Only track views which have a referrer - this should reduce bots while in practice not affecting user tracking
+    # Referrer is an 'optional, not to be trusted' header, but every browser sends it correctly at the time of writing (Aug 2026)
+    TrackVacancyViewJob.perform_later(vacancy_id: vacancy.id, referrer_url: request.referer, hostname: request.host, params: request.query_parameters) if request.referer.present?
```
KyleMacPherson: Won't this mean that we don't track views coming from job alert emails which probably makes up a reasonable chunk of our vacancy views?

scruti: Oops. Didn't think about the alert links at all. Good shout.

KyleMacPherson (changes requested):
Just adding "Request changes" so we don't accidentally merge before thinking about the tracking from job alert emails

scruti (approved):
Nice!

KyleMacPherson (approved):

KyleMacPherson:
Looks good to me!

PR #9158, opened 2026-08-11
Upgrade rails to 8.1

richardpattinson:
## Trello card URL

https://trello.com/c/JNubAvNC/3175-upgrade-to-rails-v8

## Changes in this PR:

Upgraded from Rails 8.0.5.1 → 8.1.3.1. Changes span 4 files:

Gemfile / Gemfile.lock — bumped RAILS_VERSION ceiling to < 8.2 and ran bundle update rails plus dependents (activerecord-postgis-adapter 11.0.0→11.1.1, rgeo-activerecord 8.0.0→8.1.0, rack, erb, reline, mail, etc.). No gem downgrades or removals.

config/application.rb — three changes, two of which fix real breaking changes surfaced by the upgrade:
- config.load_defaults 8.0 → 8.1
- raise_on_open_redirects = false → action_on_open_redirect = :log (the old config key is deprecated in 8.1 and raises on boot)
- Added explicit config.active_storage.variant_processor = :mini_magick — Rails has defaulted this to :vips since 7.1, but it was only ever lazily required. Rails 8.1 now eagerly requires the processor gem at boot, and since this app only ever installed mini_magick (never ruby-vips), the app could no longer boot at all without this fix.

app/controllers/publishers/candidate_messages_controller.rb — Conversation.for_organisations merges in a scope carrying .distinct, which Rails 8.1 newly refuses to combine with .update_all (deprecation → hard error in 8.2). Added .distinct(false) before the update_all call in toggle_archive.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/controllers/publishers/candidate_messages_controller.rb
+++ b/app/controllers/publishers/candidate_messages_controller.rb
@@ -22,6 +22,8 @@ class Publishers::CandidateMessagesController < Publishers::BaseController
 
     Conversation.for_organisations(organisation_ids)
                 .where(id: conversation_ids)
+                # distinct(false) is used here to avoid a PG::InvalidColumnReference error when the query is executed. This is because the update_all method does not support DISTINCT queries, and the conversation_ids array may contain duplicate values. By using distinct(false), we ensure that the query does not include a DISTINCT clause, which allows the update_all method to execute without errors.
+                .distinct(false)
                 .update_all(archived: params[:archive_action] == "archive")
 
     if params[:archive_action] == "archive"
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 ruby "4.0.2"
 
-RAILS_VERSION = "< 8.1".freeze
+RAILS_VERSION = "< 8.2".freeze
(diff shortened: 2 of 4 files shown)


app/controllers/publishers/candidate_messages_controller.rb
```diff
@@ -22,6 +22,7 @@ def toggle_archive
 
     Conversation.for_organisations(organisation_ids)
                 .where(id: conversation_ids)
+                .distinct(false)
```
scruti, line 26: @richardpattinson  is this related to the upgrade? Curious about what may be causing this change.

scruti (approved):

app/controllers/publishers/candidate_messages_controller.rb
scruti: Clarified over chat. Comes from this deprecation: 
`Deprecate usage of WITH, WITH RECURSIVE and DISTINCT with update_all.`
https://guides.rubyonrails.org/8_1_release_notes.html#active-record-deprecations

PR #9159, opened 2026-08-11
Introduce fe_role_qts_required field (QTS required for FE - phase 1)

starswan:
## Trello card URL

https://trello.com/c/rwVcEDGP/3072-hiring-staff-for-fe-additional-step-does-the-applicant-need-qts-before-ect-question

## Changes in this PR:

Introduce needs_qts_status field with back-fill via default value

<details>

### Integration Impact

- [X] DfE Analytics platform
- [X] Legacy imports mappings

Could this change impact:
- [X] Existing subscription alerts (will legacy subscription search filters break?)
- [X] Legacy vacancy copying (will copied vacancies fail new validations?)
</details>

### DIFF
--- a/db/migrate/20260810152440_add_needs_qts_status.rb
+++ b/db/migrate/20260810152440_add_needs_qts_status.rb
@@ -1,0 +1,8 @@
+class AddNeedsQtsStatus < ActiveRecord::Migration[8.0]
+  # rubocop:disable Rails/ThreeStateBooleanColumn
+  def change
+    add_column :vacancies, :fe_role_qts_required, :boolean
+    add_column :vacancy_templates, :fe_role_qts_required, :boolean
+  end
+  # rubocop:enable Rails/ThreeStateBooleanColumn
+end
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -467,6 +467,7 @@ shared:
     - job_address_town
     - job_address_county
     - job_address_postcode
+    - fe_role_qts_required
   vacancies:
     - id
     - job_title
@@ -540,6 +541,7 @@ shared:
     - job_address_town
     - job_address_county
     - job_address_postcode
+    - fe_role_qts_required
   failed_imported_vacancies:
     - id
     - import_errors
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_23_204840) do
+ActiveRecord::Schema[8.1].define(version: 2026_08_10_152440) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gist"
   enable_extension "citext"
@@ -974,6 +974,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_23_204840) do
     t.string "external_advert_url"
     t.string "external_reference"
     t.string "external_source"
+    t.boolean "fe_role_qts_required"
     t.string "fixed_term_contract_duration"
     t.string "flexi_working"
     t.boolean "flexi_working_details_provided"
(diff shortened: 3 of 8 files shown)


db/migrate/20260810152440_add_needs_qts_status.rb
```diff
@@ -0,0 +1,9 @@
+class AddNeedsQtsStatus < ActiveRecord::Migration[8.0]
+  # rubocop:disable Rails/ThreeStateBooleanColumn
+  def change
+    # default all existing vacancies to needing QTS status
+    add_column :vacancies, :needs_qts_status, :boolean, default: true
+    add_column :vacancy_templates, :needs_qts_status, :boolean, default: true
```
KyleMacPherson: What's the thinking behind defaulting this to true? My first thought is that it might be more accurate to set up a backfill task that sets needs_qts_status to false if the vacancy is for a support role and true if it's for a teaching role. What do you think?

starswan: hadn't thought of that - was trying to work out how to introduce this with little pain, but I think you're right - the next Pr requires job_role == "teacher"

scruti (approved):

lib/tasks/backfill_needs_qts_status.rake
```diff
@@ -0,0 +1,8 @@
+desc "Backfill needs QTS status"
+task backfill_needs_qts_status: :environment do
+  PublishedVacancy.live.find_each do |vacancy|
+    needs_qts_status = vacancy.job_roles.include?("teacher")
```
KyleMacPherson: So do we want to have needs_qts_status as false for headteachers, head of departments and similar?

KyleMacPherson (approved):
Looks good to me! The schema diff is pretty huge but as far as I could see the fields which were not relevant to this PR where just changing order right?

scruti (approved):
re-approving with latest backfill changes

spec/validators/json_swagger_validator_spec.rb
```diff
@@ -10,7 +10,7 @@
                                .except(:enable_job_applications, :contact_number_provided,
                                        :contact_number, :fixed_term_contract_duration, :further_details_provided,
                                        :hired_status, :include_additional_documents,
-                                       :listed_elsewhere, :ect_status, :school_visits,
+                                       :listed_elsewhere, :ect_status, :school_visits, :needs_qts_status,
```
scruti:
```suggestion
                                       :listed_elsewhere, :ect_status, :school_visits, :fe_role_qts_required,
```

scruti (changes requested):
Just noticed a disparity on the ATS API json swagger validator spec attribute name: https://github.com/DFE-Digital/teaching-vacancies/pull/9109/changes#diff-ef6fd1171f8856abca75f8833bc975bbcd45c73679d736004ff36c9ac86e6a64R13

starswan:
> Just noticed a disparity on the ATS API json swagger validator spec attribute name: https://github.com/DFE-Digital/teaching-vacancies/pull/9109/changes#diff-ef6fd1171f8856abca75f8833bc975bbcd45c73679d736004ff36c9ac86e6a64R13

Think this is correct - we're removing attributes from the vacancy that aren't in the API. Its not a great test, I agree

spec/validators/json_swagger_validator_spec.rb
scruti: @starswan `needs_qts_status` is not an existing Vacancy attribute, and this PR is no longer introducing `needs_qts_status` but `fe_role_qts_required`.

So I believe this was renamed and should be replaced for `fe_role_qts_required` as per my code suggestion above? Your second PR is renaming it (I guess as it got fixed on the follow up branch) but it seems should be directly added here with the right name? https://github.com/DFE-Digital/teaching-vacancies/pull/9109/changes#diff-ef6fd1171f8856abca75f8833bc975bbcd45c73679d736004ff36c9ac86e6a64R13

starswan: sorry got completely confused. Have done it now

scruti (approved):
All good!

PR #9170, opened 2026-08-13
2846 monitoring resources iac

jsomerset:
## Trello card URL
https://trello.com/c/2rXHKl6g/2847-tv-availability-monitoring

## Changes in this PR:

This PR is to retrospectively import the monitoring resources deployed in Azure in to IAC. These include:
Application insights
Availability tests
Metric alerts
Action group + members

## To review
Review the code changes, check the test workflow outputs linked below, and and/or run the following command to check the plan output:
make production terraform-plan CONFIRM_PRODUCTION=YES

There will be some in-place updates on the monitoring resources for some scopes and resource IDs, however, this is expected due to some minor formatting changes. The terraform import blocks require a slightly different formatting than what is already in the ARM JSON (e.g. "microsoft.insights" needs to be capitalised to "Microsoft.Insights" for the import to work), so the updates reflect this but ultimately do not change the config or functionality of the resources.

There are also some changes to the geo locations of the availability tests. These are also expected as the Azure API returns the population name rather than the display name, however, these are the same locations. This can be confirmed here - https://learn.microsoft.com/en-gb/azure/azure-monitor/app/availability#location-population-tags:~:text=Australia%E2%80%AFEast,ru%2Dmsa%2Dedge

Running the plan locally will also show the recipients of the action group being removed, this is expected as we should not be storing personal information like names, email addresses and phone numbers in the code. Instead, this information will be stored as secrets in GitHub, and passed in when running Terraform in the workflow, which should result in the action group being imported without any additional changes. This can be confirmed in the following workflow run - https://github.com/DFE-Digital/teaching-vacancies/actions/runs/32049504954

When reviewing the workflow run above, please check the output of both the QA and production environments. The reason for this is that there is logic in the terraform and workflow yaml that will only import the monitoring resources (and include the variables as secrets) for the production environment.

Please also check that the changes to the workflow files (.github/actions/deploy/action.yml and .github/workflows/build_and_deploy.yml) match what is in the temporary test workflows (.github/actions/deploy_2847/action.yml and .github/workflows/build_and_deploy_2847.yml), in terms of additional parameters and secrets being passed in as environment varibles.

### DIFF
--- a/.github/actions/deploy_2847/action.yml
+++ b/.github/actions/deploy_2847/action.yml
@@ -5,12 +5,6 @@ inputs:
   environment:
     description: Environment
     required: true
-  tag:
-    description: Docker tag
-    required: true
-  pr_id:
-    description: Pull Request
-    required: false
   aws-access-key-id:
     description: AWS-ID
     required: true
@@ -32,7 +26,6 @@ inputs:
   action_group_sms_receivers:
     description: JSON string containing variable values for action group sms receivers
     required: false
-
 
 runs:
   using: composite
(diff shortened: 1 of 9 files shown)


hopep-mobilise:
You'll need to fix the Terraform layout
terraform fmt -recursive -check
terraform fmt -recursive
terraform fmt -recursive -check

hopep-mobilise (approved):
All checks and tests carried out and all appears to be correct.
The test and terraform will need to be fixed to merge

PR #9171, opened 2026-08-13
Strip spaces from submitted vacancy application urls

scruti:
## Changes in this PR:

When publishers submit an application link form with a URL with leading/trailing spaces, the value will be normalised (by stripping the spaces) and the URL accepted (if valid) rather than showing a validation error.

This has been confusing some users who seem to think the URL format was the issue.

Also extending tests to ensure tracking params will be accepted by the URL validator.

### Before
image
image

### After

No error, goes to the next wizard step.
image

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


KyleMacPherson (approved):
Nice!

PR #9172, opened 2026-08-13
Add new fe job alert landing page

KyleMacPherson:
## Trello card URL

https://trello.com/b/1QyLnb9W/teaching-vacancies-sprint-board

## Changes in this PR:

This ticket adds a landing page for fe job alerts and corrects an issue with the ect job alerts landing page which was causing the ect filter to not be ticked.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


config/routes.rb
```diff
@@ -376,6 +376,7 @@
   end
 
   get "sign-up-for-ECT-job-alerts", to: "subscriptions#new", as: "ect_job_alerts", defaults: { ect_job_alert: true, search_criteria: { teaching_job_roles: ["ect_suitable"] } }
```
scruti: >  and corrects an issue with the ect job alerts landing page which was causing the ect filter to not be ticked.

If I recall right, the issue is that `search_criteria: teaching_hob_roles` is the wrong key. I don't think this is fixed? 🤔 

Maybe wasn't comitted/pushed?

KyleMacPherson: Ah yeah I actually fixed this in a commmit but forgot to push it. Good shout, thanks Marc!

scruti (approved):
LGTM! Thanks for fixing the bug!

PR #9181, opened 2026-08-18
Convert subscription presenter into a decorator

starswan:
## Changes in this PR:

Convert subscription presenter into decorator


scruti (approved):
Nice!

PR #9193, opened 2026-08-19
Validate time formatting on interview time form

KyleMacPherson:
## Trello card URL
https://trello.com/c/6MQl5PFl/3192-bug-ats-flow-hs-add-interview-time

## Changes in this PR:

Prior to this change if a user entered an incorrect time format it would not show the user an error and would instead come up with a weird result. For instance Fisal entered 2.30pm and got 6pm as the result. This PR fixes this issue, raising an error if the time is not formatted as suggested in the hint text.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/config/initializers/active_model_type_time_or_string.rb
+++ b/config/initializers/active_model_type_time_or_string.rb
@@ -10,12 +10,12 @@ module ActiveModel
         end
       end
 
-      private
-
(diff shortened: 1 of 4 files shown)


scruti (approved):
LGTM. I like the Copilot's suggestion about stripping whitespaces.

PR #9199, opened 2026-08-21
Reduce complexity in VacancyFilterQuery

starswan:
## Changes in this PR:

Reduce complexity in VacancyFilterQuery to help implement new filter functionalty

### DIFF
--- a/app/queries/vacancy_filter_query.rb
+++ b/app/queries/vacancy_filter_query.rb
@@ -63,8 +63,6 @@ class VacancyFilterQuery
         built_scope.merge(Organisation.faith_schools).distinct
       elsif school_types.include?("special_school")
         built_scope.merge(Organisation.where(detailed_school_type: Organisation::SPECIAL_SCHOOL_TYPES)).distinct
-      else
-        built_scope
       end
     end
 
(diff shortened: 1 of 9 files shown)


app/queries/vacancy_filter_query.rb
```diff
@@ -1,136 +1,128 @@
...
+        Vacancy.with_any_of_phases(phases)
+      end
     end
-  end
 
-  def apply_job_roles(keys, built_scope, filters)
-    filtered_roles_as_strings = keys.flat_map { |key|
-      job_roles(filters[key])
-    }.compact
+    def job_roles_filters(keys, filters)
```
scruti: Non blocking suggestion: 
Since this method is used once and the call is `job_roles_filters(%i[job_roles teaching_job_roles support_job_roles], filters)`
with the `keys` argument being static. Could we simplify this method by transforming it into a `job_roles_filters(filters)` and keeping the `keys` list as a constant/or static list inside the method?

(unless needed otherwise).

scruti (approved):
Really nice refactor 👏 

Left a non blocking suggestion.

PR #9208, opened 2026-08-26
2855 - Debugging review app reconcile

steves-mobilise:
## Trello card URL

[Link](https://trello.com/c/ZWznDc8Y/2855-add-review-app-cleanup-workflow)

## Changes in this PR:

A simple change to add a debugging delete workflow that just outputs the event type that calls the workflow, along with PR number of review app it would have deleted, without any deletions taking place.
Also updates the schedule cron to every 30 minutes so that we can see the details during the day.

This is due to the fact that the deletion process is continually skipped for this service, even though the review apps are found.

## Guidance to review

There is no true way to test this as we need to determine how it is being called when scheduled (hence the change to the schedule cron).
However, I have proved that no review apps will be deleted during this time by running it against the **test branch** and setting `dry_run` to `false` here:

[Test branch: dry_run==false](https://github.com/DFE-Digital/teaching-vacancies/actions/runs/32943639119/job/98099645552)

I then ran it against **main** with `dry_run` set to `true` here:

[main branch: dry_run==true](https://github.com/DFE-Digital/teaching-vacancies/actions/runs/32944133593)

This shows that no review apps were deleted during the testing process.

## Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Applied DevOps label
- [x] Tested by running locally

### DIFF
--- a/.github/workflows/debug_delete_review_app.yml
+++ b/.github/workflows/debug_delete_review_app.yml
@@ -1,0 +1,147 @@
+name: Delete review app
+
+on:
+  pull_request:
+    types: [closed, unlabeled]
+    paths-ignore:
+      - "bigquery/**"
+      - "documentation/**"
+      - "terraform/common/**"
+      - "**.md"
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Pull Request number to delete (EG: 1234 for review-pr-1234)"
+        required: true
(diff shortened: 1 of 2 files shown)


matt-mobilise (approved):
Checked the output Steve, all looking good.

PR #9216, opened 2026-08-27
Add use my current location on homepage

KyleMacPherson:
## Trello card URL

https://trello.com/c/xJF4hBXF/3108-spike-search-based-on-users-location-from-the-browser

## Changes in this PR:

This PR adds a "Use my current location" link on the homepage/shool search page/colleges page when the user views it on mobile. We actually already have similar links on the jobs search page and the job alerts page.

It also autopopulates the location field on these pages when a user is using mobile/tablet and they have given us permission to use their location.

### Data & Schema Changes

<details>

### Integration Impact

- [ ] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>

### DIFF
--- a/app/assets/javascript/js_components/locationFinder/locationFinder.js
+++ b/app/assets/javascript/js_components/locationFinder/locationFinder.js
@@ -26,6 +26,14 @@ const LocationFinder = class extends Controller {
     this.input.addEventListener('focus', () => {
       this.removeErrorMessage();
     });
+
+    if (this.input.value === '' && navigator.permissions) {
+      navigator.permissions.query({ name: 'geolocation' }).then((result) => {
+        if (result.state === 'granted') {
+          this.findLocation();
+        }
+      });
+    }
   }
 
   findLocation() {
--- a/app/assets/javascript/js_components/locationFinder/locationFinder.js
+++ b/app/assets/javascript/js_components/locationFinder/locationFinder.js
@@ -27,12 +27,8 @@ const LocationFinder = class extends Controller {
       this.removeErrorMessage();
     });
 
-    if (this.input.value === '' && navigator.permissions) {
-      navigator.permissions.query({ name: 'geolocation' }).then((result) => {
-        if (result.state === 'granted') {
-          this.findLocation();
-        }
-      });
+    if (this.input.value === '' && navigator.geolocation) {
+      this.findLocation();
     }
   }
 
@@ -46,6 +42,9 @@ const LocationFinder = class extends Controller {
         this.onFailure();
         logger.log(`${LOGGING_MESSAGE}: ${error.message}`);
       });
+    }, () => {
+      this.onFailure();
+      logger.log(LOGGING_MESSAGE);
     });
   }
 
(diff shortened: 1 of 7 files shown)


app/views/home/_search.html.slim
```diff
@@ -5,6 +5,7 @@
     .search__form
       = render "vacancies/search/keyword", f: f
       = render "vacancies/search/location", f: f, label_text: t(".location_label"), show_hint: false, desktop: true
+      = render "vacancies/search/current_location", target: "location-field"
```
scruti: 🙋‍♂️  This one is getting triggered on page load, without having clicked the link.

spec/system/jobseekers/jobseekers_can_search_from_home_page_spec.rb
```diff
@@ -5,6 +5,10 @@
     visit root_path
   end
 
+  scenario "shows the use my location button" do
+    expect(page).to have_css(".location-finder")
+  end
```
scruti, line 10: Not sure this assertion is worth adding a feature spec. Wouldn't we achieve the same with a view spec?

app/assets/javascript/js_components/locationFinder/locationFinder.js
```diff
@@ -26,6 +26,10 @@ const LocationFinder = class extends Controller {
     this.input.addEventListener('focus', () => {
       this.removeErrorMessage();
     });
+
+    if (this.input.value === '' && navigator.geolocation && this.element.offsetParent !== null) {
+      this.findLocation();
+    }
   }
```
scruti, line 33: 🙋  This autofill triggers a browser location permission prompt automatically when visiting the page. 

I believe this is something we don't want? It is a UX annoyance, and closing the dialogue a couple of times, for example, causes browsers like Chrome to automatically block for good the location dialogue for the site for good, breaking any future clicks on the location button.

Could we avoid that doing it only if the user has already given location permissions after clicking the link for the first time?

EG (AI fix suggestion):

```js
connect() {
  this.input = document.getElementById(this.inputValue);

  this.input.addEventListener('focus', () => {
    this.removeErrorMessage();
  });

  this.autoFindLocationIfPermitted();
}

autoFindLocationIfPermitted() {
  if (this.input.value !== '' || !navigator.geolocation || this.element.offsetParent === null) {
    return;
  }

  if (!navigator.permissions || !navigator.permissions.query) {
    return;
  }

  navigator.permissions.query({ name: 'geolocation' }).then((status) => {
    if (status.state === 'granted') {
      this.findLocation();
    }
  }).catch(() => {});
}
```

KyleMacPherson: Ah great point. I didn't realise this as I had already granted the permissions when testing, thanks Marc. Great spot 🙇

app/assets/javascript/js_components/locationFinder/locationFinder.test.js
```diff
@@ -94,3 +94,70 @@ describe('when coordinates are not available from browser geolocation API', () =
     expect(document.getElementById('location-finder__error')).toBe(null);
   });
 });
+
+describe('automatically finding the location', () => {
```
scruti, line 98: Nice tests!

scruti (approved):
Nice! Now it works flawlessly (on my rought testing).

I think the PR needs to be rebased to catch up with the accessibility tests check.

PR #9233, opened 2026-09-02
3146 dfe sign in sync users jobs

richardpattinson:
## Trello card URL

https://trello.com/c/7cBdX8IM/3146-dfe-sign-in-sync-users-jobs-rewrite-jobs-with-better-progress-and-connection-error-handling

## Changes in this PR:

Reworks the two DfE Sign-in sync jobs (`UpdateDSIUsersInDbJob` and `ExportDSIUsersToBigQueryJob`) to fix the reliability and speed problems raised in the ticket.

**Problem:** each job walked every DSI API page serially in one long-running job (30+ minutes), so one failing page failed the entire run — and the BigQuery table was deleted before the new data was confirmed retrievable, so a partial failure could leave it empty until the next successful retry.

**What changes:**

- **Per-page jobs instead of one long job.** A `DSIExportRun` now fans out one job per page (`FetchDSIUsersPageJob` / `FetchDSIUsersExportPageJob` / `FetchDSIApproversExportPageJob`), capped with `limits_concurrency` to avoid hammering DSI with dozens of simultaneous requests. Retries happen per page instead of restarting the whole run.
- **Users and approvers exports split apart.** `ExportDSIUsersToBigQueryJob` used to export both in one job; approvers now run independently as `ExportDSIApproversToBigQueryJob`.
- **Table swap only happens after a confirmed-good run.** `FinalizeDSIUsersExportJob` / `FinalizeDSIApproversExportJob` delete-and-replace the BigQuery table only once every page in a run has fetched successfully — no more window where the table sits empty mid-retry.
- **Stuck runs get surfaced, not swallowed.** The fan-out design fails safe: a permanently-failing page just leaves its run stuck rather than raising or touching the table. `MonitorDSIExportRunsJob` alerts Sentry if a run hasn't completed within 3 hours, and prunes run records after 3 days.
- **Deliberately not using Rails 8.1's ActiveJob Continuation.** Since work is now fanned out into independently-retryable per-page jobs rather than one serial job, a deploy restart only loses in-flight pages, not the whole run — which covers the same need without adding that dependency.

### Is there anything specific you want feedback on?

- `limits_concurrency to: 5` on the fetch jobs and the 3-hour stale-run threshold in `MonitorDSIExportRunsJob` are both best guesses — flag if you'd tune either differently.
- Whether `DSIExportRun`/`DSIExportRunPage` feel like the right amount of tracking machinery for this, versus something simpler.
- The call to skip ActiveJob Continuation (see above) — happy to reconsider if you'd rather see it used regardless.

### Before

N/A — background job changes only, no UI.

### After

N/A — background job changes only, no UI.

### Data & Schema Changes

- [x] Adds/removes model validations
- [x] Adds/removes database fields

<details>
<summary>If any of the above options has changed then the author must check/resolve all of t

### Integration Impact

- [x] DfE Analytics platform
- [ ] Legacy imports mappings

Could this change impact:
</details>


scruti:
Had some comments on priv. Back to active development.

app/services/publishers/dfe_sign_in/big_query_export/users.rb
```diff
@@ -3,8 +3,12 @@ class Users < Base
     TABLE_NAME = "dsi_users".freeze
 
     def call
+      # Fetch every page before touching the table: `dsi_users` is a lazy enumerator, so
+      # forcing it here means a DSI failure on any page aborts before the table is deleted,
+      # rather than leaving it empty with no replacement data.
+      pages = dsi_users.to_a
```
scruti, line 9: 🙋 This gets all the DSI user JSON pages into memory in one go, defeating the whole pagination purpose, and I guess consuming a lot of the worker memory.

I would rather compromise and only delete the table after successfully fetching the first page. Then insert the table data for the first page and iterate from the second page.

app/services/publishers/dfe_sign_in/big_query_export/approvers.rb
```diff
@@ -3,8 +3,12 @@ class Approvers < Base
     TABLE_NAME = "dsi_approvers".freeze
 
     def call
+      # Fetch every page before touching the table: `dsi_approvers` is a lazy enumerator, so
+      # forcing it here means a DSI failure on any page aborts before the table is deleted,
+      # rather than leaving it empty with no replacement data.
+      pages = dsi_approvers.to_a
```
scruti, line 9: Same as https://github.com/DFE-Digital/teaching-vacancies/pull/9233/changes#r3988456133

config/recurring.yml
```diff
@@ -61,6 +61,11 @@ production:
     class: 'ExportDSIUsersToBigQueryJob'
     queue: low
 
+  export_approvers:
+    schedule: '05 01 * * *' # "At 01:05."
```
scruti, line 65: Not sure if there was an intention not to overlap here, but 1:00 vs 1:05 will result in the export_users and export_approvers jobs overlapping anyway.

I think that running both jobs in parallel should be fine, so we can launch both at 1:00.

lib/dfe_sign_in/api.rb
```diff
@@ -9,11 +9,39 @@ module API
...
     def dsi_approvers
-      PaginatedUsers.new(APPROVERS_ENDPOINT, APPROVERS_PAGE_SIZE).results
+      approvers_pagination.results
+    end
+
+    def dsi_approvers_page_count
+      approvers_pagination.number_of_pages
+    end
+
+    def dsi_approvers_page(page)
```
scruti, line 33: This is not used anywhere at the moment but I guess if you go back to pagination calls in `Publishers::DfeSignIn::BigQueryExport::Approvers` will be needed again?

spec/services/publishers/dfe_sign_in/big_query_export/users_spec.rb
```diff
@@ -97,6 +97,12 @@
       it "raises a runtime error" do
         expect { subject.call }.to raise_error(RuntimeError)
       end
+
+      it "does not touch the existing table, so it is not left empty" do
```
scruti, line 101: This is within "when the user table does not exist in the dataset" but should be within "when the user table exists in the dataset"

spec/services/publishers/dfe_sign_in/big_query_export/approvers_spec.rb
```diff
@@ -98,6 +98,12 @@
       it "raises a runtime error" do
         expect { subject.call }.to raise_error(RuntimeError)
       end
+
+      it "does not touch the existing table, so it is not left empty" do
```
scruti, line 102: This is within "when the approver table does not exist in the dataset" but should be within "when the approver table exists in the dataset" (BTW the context needs renaming as it incorrectly references "users").

app/jobs/export_dsi_approvers_to_big_query_job.rb
```diff
@@ -0,0 +1,10 @@
+# This is loaded correctly by Zeitwerk due to a custom inflection in config/inflections.rb
```
scruti, line 1:
```suggestion
# This is loaded correctly by Zeitwerk due to a custom inflection in config/initializers/inflections.rb
```

app/jobs/export_dsi_users_to_big_query_job.rb
scruti, line 1:
```suggestion
# This is loaded correctly by Zeitwerk due to a custom inflection in config/initializers/inflections.rb
```

scruti (approved):
LGTM!

PR #9242, opened 2026-09-03
Spike rework vacancy display

starswan:
## Trello card URL

https://trello.com/c/3AIlLRdg/3235-vacancy-show-template-tidy-up

## Changes in this PR:

Spike improved simple vacancy display


spec/presenters/vacancy_decorator_spec.rb
```diff
@@ -345,4 +345,75 @@ def normalize_html(html)
...
+
+        it "returns Fixed term (duration)" do
+          expect(decorated.readable_contract_information).to eq "Fixed term - 6 months"
+        end
+      end
+
+      context "when is_parental_leave_cover is true" do
+        let(:is_parental_leave_cover) { true }
+
+        it "returns Fixed term (duration)" do
```
scruti, line 413:
```suggestion
        it "returns Fixed term (duration) and the leave cover suffix" do
```

app/views/publishers/vacancies/show.html.slim
```diff
@@ -37,4 +37,10 @@
 - else
   = render "review_banner", vacancy: @vacancy, current_organisation: @current_organisation
 
-  = render "sections", vacancy: @vacancy, current_organisation: @current_organisation, step_process: @step_process
```
scruti, line 40: 🙋 I think the `sections` partial is no longer used and could be removed?

scruti (changes requested):
I believe the `sections` partial has been orphaned and should be removed now?

Nice refactor. It is much better now.

scruti (approved):
LGTM!

PR #9254, opened 2026-09-09
Delete unknown GIAS records

starswan:
## Trello card URL

https://trello.com/c/m2ITiQg3/3236-delete-orgs-removed-from-gias

## Changes in this PR:

soft-delete GIAS records which have no longer in GIAS data set

### DIFF
--- a/app/services/gias/import_schools_and_local_authorities.rb
+++ b/app/services/gias/import_schools_and_local_authorities.rb
@@ -26,7 +26,7 @@ class Gias::ImportSchoolsAndLocalAuthorities
         raise ImportFailure, import_errors.map(&:errors) if import_errors.any?
 
         # This is run every day, so discard old records which are clearly no longer in GIAS.
-        School.where(updated_at: ..1.week.ago).discard_all
+        School.where(updated_at: ..1.month.ago).discard_all
       end
     end
 
--- a/app/services/gias/import_schools_and_local_authorities.rb
+++ b/app/services/gias/import_schools_and_local_authorities.rb
@@ -23,9 +23,10 @@ class Gias::ImportSchoolsAndLocalAuthorities
         import_errors = Gias::Data.new(SCHOOLS_AND_LOCAL_AUTHORITIES_CSV).each_slice(BATCH_SIZE).flat_map do |group|
           import_group uk_colleges, group
         end
+        raise ImportFailure, import_errors.map(&:errors) if import_errors.any?
+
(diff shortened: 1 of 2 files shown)


app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -23,6 +23,8 @@ def call
         import_errors = Gias::Data.new(SCHOOLS_AND_LOCAL_AUTHORITIES_CSV).each_slice(BATCH_SIZE).flat_map do |group|
           import_group uk_colleges, group
         end
+        # This is run every day, so discard old records which are clearly no longer in GIAS.
+        School.where(updated_at: ..1.week.ago).discard_all
```
KyleMacPherson: Should we have this after the importer raises errors so we only discard them if the import has run successfully but not included the schools?

I vaguely recall that sometimes our imports fail for a while through no fault of our own but I can't really remember any specifics. Do you remember if it sometimes longer than a week?

app/services/gias/import_schools_and_local_authorities.rb
```diff
@@ -71,11 +73,16 @@ def import_group(uk_colleges, group) # rubocop:disable Metrics/AbcSize
...
+    # rubocop:disable Performance/Sum
     def import_batch(local_authorities, schools, memberships)
-      import_local_authorities(local_authorities).failed_instances +
-        import_schools(schools).failed_instances +
-        import_memberships(local_authorities, schools, memberships).failed_instances
+      [import_local_authorities(local_authorities),
+       import_schools(schools),
+       import_memberships(local_authorities, schools, memberships)]
+        .map(&:failed_instances)
+        .reduce(:+)
```
KyleMacPherson: Nice, I think that's way clearer! I think we could also use flat_map here

app/services/gias/import_schools_and_local_authorities.rb
scruti: Very good proposal. I was nervous about this bit, as it would allow deleting all the schools if the job fails for a week, and I was about to propose combining it with only calling these if other schools were updated recently.
But what you propose is way more solid.

scruti: Also, are we 100% sure all the orgs in the DB are updated/touched daily? (To triple-check)

starswan: Yes, I was a bit surprised that it touched things even when they don't change

scruti (approved):

PR #9257, opened 2026-09-09
Rework vacancy stats collection

starswan:
## Trello card URL

https://trello.com/c/3uwmeMXo/3255-redis-removal-step-1-remove-redis-from-vacancy-view-tracking

## Changes in this PR:

Rework vacancy stats collection. The SQL executed is:

"SELECT \"vacancy_analytics\".* FROM \"vacancy_analytics\" WHERE \"vacancy_analytics\".\"vacancy_id\" = '004823f5-1e14-4cea-aa63-14c2e446cc76' ORDER BY \"vacancy_analytics\".\"id\" ASC LIMIT 1 FOR UPDATE"

and then either 
""INSERT INTO \"vacancy_analytics\" (\"vacancy_id\", \"referrer_counts\", \"created_at\", \"updated_at\") VALUES ('004823f5-1e14-4cea-aa63-14c2e446cc76', '{\"facebook\":1}', '2026-09-10 07:51:45.177084', '2026-09-10 07:51:45.177084') RETURNING \"id\"

or 

"UPDATE \"vacancy_analytics\" SET \"referrer_counts\" = '{\"facebook\":2}', \"updated_at\" = '2026-09-10 07:51:58.399337' WHERE \"vacancy_analytics\".\"id\" = 'bb88d0c9-44fd-40e6-a2b1-bdd1469d4802'"

all done by ActiveRecord (as existing)


scruti:
Related documentation needs updating. Or, given that we are removing the buffering, maybe we can just remove this document? https://github.com/DFE-Digital/teaching-vacancies/blob/main/documentation/service/technical/tracking-vacancy-views.md

config/recurring.yml
```diff
@@ -201,11 +201,6 @@ production:
     class: 'ImportFromVacancySourcesJob'
     queue: default
 
-  aggregate_vacancy_referrer_stats:
```
scruti, line 204: Should we remove this in a follow-up PR?
My concern with removing it here is that as soon as the new code gets deployed, we will lose any counters that exist in Redis. Having a follow-up PR removing the aggregation would allow Redis counts to be flushed into DB between releases.

app/services/vacancy_analytics_service.rb
```diff
@@ -1,58 +1,19 @@
...
+      return if vacancy_id.blank?
 
-  def self.update_stats_in_database(vacancy_updates)
-    vacancy_updates.each do |vacancy_id, new_referrer_counts|
-      # Skip if the associated vacancy does not exist. This may happen if the vacancy was deleted after the visit was tracked.
-      next unless PublishedVacancy.exists?(id: vacancy_id)
+      referrer = normalize_referrer(referrer_url, hostname, params)
 
       VacancyAnalytics.transaction do
         analytics = VacancyAnalytics.where(vacancy_id: vacancy_id).lock(true).first_or_initialize
```
scruti, line 9: Heads up, I'm suggesting DB performance optimisations, as this will be hit a lot and we're not buffering it anymore. So the fewer SQL operations, the better:

The `"SELECT "vacancy_analytics".* FROM "vacancy_analytics" WHERE "vacancy_analytics"."vacancy_id" = '004823f5-1e14-4cea-aa63-14c2e446cc76' ORDER BY "vacancy_analytics"."id" ASC LIMIT 1 FOR UPDATE"` can only lock when the record already exist. So it seems to still allow race conditions in the `insert`.

Also according to the 🤖 analysis on the output:

```
BEGIN                        ──┐
SELECT … FOR UPDATE          ──┤ ← lock acquired
   (Ruby: read hash, add 1)    │
SELECT 1  (uniqueness check)   │   lock still held
UPDATE                         │   lock still held
COMMIT                       ──┘ ← lock released
```
> The row lock is held across three more network round trips plus the Ruby in between. Every other request for that same vacancy queues behind all of it

vs
```
INSERT … ON CONFLICT DO UPDATE   ← lock acquired and released inside this one statement
```
A single SQL statement and would also allow us to remove the `VacancyAnalytics.transaction`.

The 🤖  recommended approach:

> upsert takes an on_duplicate: clause, so this stays within AR apart from the jsonb_set fragment:

```
class VacancyAnalyticsService
  class << self
    def track_visit(vacancy_id:, referrer_url:, hostname:, params:)
      return if vacancy_id.blank?

      referrer = normalize_referrer(referrer_url, hostname, params)

      VacancyAnalytics.upsert(
        { vacancy_id: vacancy_id, referrer_counts: { referrer => 1 } },
        unique_by: :vacancy_id,
        on_duplicate: increment_referrer_count(referrer),
      )
    end

    private

    # jsonb has no native "increment a key", so the SET clause has to be SQL.
    # Bound, never interpolated: `referrer` derives from the Referer header.
    def increment_referrer_count(referrer)
      Arel.sql(
        VacancyAnalytics.sanitize_sql_array([<<~SQL.squish, referrer, referrer]),
          referrer_counts = jsonb_set(
            vacancy_analytics.referrer_counts,
            ARRAY[?::text],
            to_jsonb(COALESCE((vacancy_analytics.referrer_counts ->> ?)::bigint, 0) + 1)
          ),
          updated_at = now()
        SQL
      )
    end
  end
end
```

> Why sanitize_sql_array rather than string interpolation: normalize_referrer can return (host_split - tld_split).last, derived from the attacker-controlled Referer header. PublicSuffix would likely reject anything exotic and fall through to "invalid", but that's an accident of validation, not a guarantee — don't build an injection surface on it.

spec/services/vacancy_analytics_service_spec.rb
```diff
@@ -54,68 +44,4 @@
       expect(described_class.normalize_referrer(referrer_url, hostname, { utm_medium: "email" })).to eq("jobalert")
     end
   end
-
-  describe ".aggregate_and_save_stats" do
```
scruti, line 58: We should add some of these tests to the `.track_visit` method, since the logic has been moved there rather than removed. EG: Incrementing an existing counter value.

app/services/vacancy_analytics_service.rb
```diff
@@ -1,58 +1,19 @@
...
+    def track_visit(vacancy_id:, referrer_url:, hostname:, params:)
+      return if vacancy_id.blank?
 
-  def self.update_stats_in_database(vacancy_updates)
-    vacancy_updates.each do |vacancy_id, new_referrer_counts|
-      # Skip if the associated vacancy does not exist. This may happen if the vacancy was deleted after the visit was tracked.
-      next unless PublishedVacancy.exists?(id: vacancy_id)
+      referrer = normalize_referrer(referrer_url, hostname, params)
 
       VacancyAnalytics.transaction do
```
scruti, line 8: ☝️ We do not need the `transaction` block anymore, since the upsert is a single SQL query.

PR #9262, opened 2026-09-10
Remove noise from components and their usage

starswan:
## Trello card URL

https://trello.com/c/oRWtJ9Tc/3251-spike-review-our-own-viewcomponents

## Changes in this PR:

Remove noise around components which have unused options


app/components/landing_page_link_group_component.rb
```diff
@@ -1,11 +1,9 @@
 class LandingPageLinkGroupComponent < ApplicationComponent
   include FailSafe
 
-  def initialize(title: nil, subgroup: false, use_locations: false, list_class: "", classes: [], html_attributes: {})
-    super(classes: classes, html_attributes: html_attributes)
+  def initialize(use_locations: false, subgroup: false)
```
scruti, line 4: ❓ Question! 

Does anything passes/uses `subgroup` anymore? Since the view removed the `elsif @subgroup` title display distinction, I believe subgroups would be render with the same format as the parent group and could be an issue?

So not sure if removing the logic from the view but keeping it in the component is actually good? Or we should opt to keep it/remove it from both places?

Or am I misreading it?

scruti:
Just one question for clarification before approving. Looks good!

app/components/landing_page_link_group_component.rb
starswan: This is a bit of a challenge - I'm a bit scared of it as it has no tests and doesn't render at all in test mode. But I think I understand it well enough now

scruti (approved):
LGTM!

PR #9263, opened 2026-09-10
Rename unusual tvs component usage

starswan:
## Trello card URL

https://trello.com/c/oRWtJ9Tc/3251-spike-review-our-own-viewcomponents

## Changes in this PR:

Rename component usage to tvs_ to show that we are using a non-standard component, when probably a Govuk one should have been sufficient


scruti (approved):
Much better 👏

PR #9279, opened 2026-09-16
Replaced action group import with data block

jsomerset:
## Trello card URL
https://trello.com/c/pFSQOgp6/3061-new-import-approach-for-tv

## Changes in this PR:
This PR is a follow up to a previous emergency PR that removed resources from the state due to visibility of sensitive data in the workflow logs. The previous PR removed the imported action group resource from the state file as it contained contact information, as well as the monitor alert resources that referenced it as an endpoint.

These changes will re-import the alert rules, however, the action group will be referenced using a data block instead of being imported as a managed resource. The reason for this is that when the action group is a managed resource, the email and SMS receivers are stored as part of the resource config, meaning they are referenced in plans and apply logs. When using a data block, none of this information is stored in terraform or processed in the workflows, and is instead configured manually in the portal while allowing all other monitoring resources to remain in IAC.

## To Review
Run the following commands to review the plan outputs for each environment:
```
make production terraform-plan CONFIRM_PRODUCTION=YES
```
This should show the monitor alert rules as resources being imported without any modifications - Note that one highlighted change is that the plan shows the webhook_properties field being removed from the action block of the metric alert resources. This field is an empty map so is interpreted as null by Terraform and should have no effect on the resources themselves).
```
make qa terraform-plan 
```
```
make staging terraform-plan 
```
These should produce no changes to the monitoring resource and are just to confirm that the code only applies to prod, where the monitoring resources are deployed.


hopep-mobilise (approved):
Plans run in qa and production and imports were as expected. Accepted that the data block will not display the user details.

PR #9280, opened 2026-09-16
Remove new application path

starswan:
## Trello card URL

https://trello.com/c/J3WXueLj/3279-remove-bannerbuttoncomponent

## Changes in this PR:

Whilst investigating the BannerButtonComponent, I realised that this page was subject to confusion.
A helper with a name ending in '_path' was being used, which could easily be mistaken for a route. 
Removing the helper allowed the code to be greatly simplified as 'new_application_path' only decided the application path for an uploaded vacancy, but often was being called in the context of an uploaded vacancy (or in the opposite context) - and one route is a get and the other a post, so the code was still having to switch on uploaded_vacancy? so that it could be a POST rather than a GET. 

I am considering a follow-up PR to make the 'apply for vacancy' route the same regardless of the vacancy type, which will simplify the code even more, and will also simplify the TRN experiment code


app/views/jobseekers/saved_jobs/index.html.slim
```diff
@@ -30,12 +30,13 @@ h1.govuk-heading-l = t(".page_title")
...
-              - card.with_action_item link: govuk_link_to(safe_join([t(".view"), tag.span(" for #{saved_job.vacancy.job_title}", class: "govuk-visually-hidden")]), jobseekers_job_application_path(JobApplication.find_by(vacancy_id: saved_job.vacancy.id)))
-            - elsif saved_job.vacancy.enable_job_applications? && PublishedVacancy.live.exists?(saved_job.vacancy.id) && current_jobseeker.job_applications.draft.find_by(vacancy_id: saved_job.vacancy.id).present?
-              - card.with_action_item link: govuk_link_to(safe_join([t(".continue"), tag.span(" for #{saved_job.vacancy.job_title}", class: "govuk-visually-hidden")]), jobseekers_job_application_review_path(JobApplication.find_by(vacancy_id: saved_job.vacancy.id)))
-            - elsif saved_job.vacancy.enable_job_applications? && PublishedVacancy.live.exists?(saved_job.vacancy.id)
-              - card.with_action_item link: govuk_link_to(safe_join([t(".apply"), tag.span(" for #{saved_job.vacancy.job_title}", class: "govuk-visually-hidden")]), new_application_path(saved_job.vacancy))
+            - if saved_job.vacancy.enable_job_applications? && PublishedVacancy.live.exists?(saved_job.vacancy.id)
```
scruti, line 33: Nice refactor

scruti (approved):
Nice refactor 👏

PR #9285, opened 2026-09-18
Vacancy display tidy

starswan:
## Trello card URL

https://trello.com/c/tvcXdl8D/3222-vacancy-modelling-move-from-vacancy-discarding-attributes-value-to-failing-validation

## Changes in this PR:

Make view code more consistent, add comments around Resettable module.

Move some resettable triggers into attribute overrides so that they apply in memory rather than just to persisted objects to reduice opportunities for data with invalid shapes being introduced during tests


app/models/concerns/resettable.rb
```diff
@@ -1,3 +1,21 @@
...
+# then the 'details' field doesn't make sense.
+# This was originally done in a before_save hook, but that it not ideal, as it is still
+# possible to create an invalid object in memory, so some rules have been moved
+# to attribute assignment overloads.
+# These have not all been done because the array_enums do not support this
+# (working_patterns, job_roles, key_stages)
+# (array_enum is hand-rolled in this project)
+# and the vacancy_template doesn't have all the fields e.g. application_link
+#
+# Refactroing of this is also hindered by it being called directly by the legacy
```
scruti, line 12:
```suggestion
# Refactoring of this is also hindered by it being called directly by the legacy
```

app/models/concerns/resettable.rb
```diff
@@ -1,3 +1,21 @@
+# This module is used to keep the vacancy/vacancy_template shape 'correct'
```
scruti, line 1: Excellent comment, much clearer now.

app/models/concerns/resettable.rb
```diff
@@ -1,3 +1,21 @@
...
+# These have not all been done because the array_enums do not support this
+# (working_patterns, job_roles, key_stages)
+# (array_enum is hand-rolled in this project)
+# and the vacancy_template doesn't have all the fields e.g. application_link
+#
+# Refactroing of this is also hindered by it being called directly by the legacy
+# vacancy import feature - so more progress could possibly be made once that feature has been removed.
+#
+# Ideally these 'shapes' would be validated, but it's quite difficult to do
+# as e.g. hidden content behind de-selected radio buttons is still transmitted, so somnething simple
```
scruti, line 16:
```suggestion
# as e.g. hidden content behind de-selected radio buttons is still transmitted, so something simple
```

app/presenters/vacancy_decorator.rb
```diff
@@ -77,6 +77,10 @@ def fe_role_qts_required
     I18n.t("helpers.label.publishers_job_listing_about_the_role_form.fe_role_qts_required_options.#{model.fe_role_qts_required}")
   end
 
+  delegate :key_stages, :subjects
+
+  delegate :receive_applications, :application_link, :further_details, :fixed_term_contract_duration, :contact_number, :application_email, :actual_salary
+
```
scruti, line 83: Question: Is there any reason to separate these delegators from the main list and place them here at the bottom?

scruti (approved):
Nice

app/presenters/vacancy_decorator.rb
starswan: no reason - just some fallout from the exploration. Have moved them up to the top again with the others

PR #9289, opened 2026-09-18
Vacancy validations (spike)

starswan:
## Trello card URL

https://trello.com/c/tvcXdl8D/3222-vacancy-modelling-move-from-vacancy-discarding-attributes-value-to-failing-validation

## Changes in this PR:

Spike to test vacancy validations as an idea - add a validation, and change form params_to_save to match the validation that the form is already doing anyway


scruti (approved):
LGTM! I like this approach as each form gets more responsible on its params combination to propragate into DB.

